PR #59, opened 2026-06-01
feat: openapi.overrides config escape hatch

### LINKED ISSUE #45: openapi.overrides config escape hatch
## Summary

Add an `openapi.overrides` config array, keyed by route name or URI glob, that lets users patch a curated set of operation-level fields in the generated document — without touching controller code. It is a pure spec-only escape hatch: a late pipeline stage mutates the emitted OpenAPI document and never the host app.

## Motivation

The generator treats the controller as the source of truth, which is the right default but leaves real gaps: vendor packages ship controllers users can't edit, legacy routes need to be marked `deprecated` without a code commit, and one-off release corrections (a renamed `operationId`, an extra tag) shouldn't require a PR against application code. A config-driven override gives operators a documented seam for exactly these cases. This is a Tier 0 item (reflection/signatures only; the override is applied after generation, not derived from code).

## Proposed approach

Introduce `openapi.overrides` as a flat root-level config key. Each entry maps a route name or URI glob to a partial operation:

```php
'overrides' => [
    'users.show'         => ['operationId' => 'getCurrentUser', 'tags' => ['Identity'], 'deprecated' => true],
    'api/v1/legacy/*'    => ['x-internal' => true],
],
```

Four units, placed under `Support\Generator\` (reconcile against the current tree — the generator-chain refactor already shipped as `src/Support/Generator/BaselineRegistration.php` plus the stage split under `src/Support/Generator/Stages/`; ignore the draft-PR plan's stale references to a `Core\Generator\` namespace and to `CoreRegistration`):

- **`RouteIndex`** — a `#[Scoped]` accumulator mapping `(method, uri)` to route name, recorded by `src/Support/Generator/Stages/PathsStage.php` as it builds each operation (`buildOperationId()` at `src/Support/Generator/Stages/PathsStage.php:210` already has the route-name vs URI-fallback split that determines the key). Key format `"{UPPER_METHOD} /{uri}"`.
- **`OverrideMatcher`** — a pure `readonly` utility: precedence matching, allowlist filtering, and an `unusedKeys()` reporter for lint.
- **`OverridesStage implements SpecStage`** (`#[Scoped]`, contract at `src/Contracts/Generator/SpecStage.php` with the single method `apply(OA\OpenApi $doc, GenerationContext $ctx): void`) — mutates each operation in place; `x-*` keys land on the operation's `x` array.
- **Pipeline placement** — `core → plugin stages → OverridesStage → TransformersStage` (`src/Support/Generator/Stages/TransformersStage.php`), so config beats plugin contributions and convention-derived values, but a user's code-based `transformDocument()` callback still wins. Register the stage from the owning plugin's `register()` via `OpenApiRegistry::addStage(OverridesStage::class)` (`src/Registry/OpenApiRegistry.php:165`); `SpecPipeline` (`src/Support/Generator/SpecPipeline.php`) executes registered stages in registration order and already runs `TransformersStage` as a fixed terminal step, so registering after the plugin stages but before that terminal step yields the placement above.

Allowlist of overridable fields: `operationId`, `summary`, `description`, `tags`, `deprecated`, and any `x-*` extension. Non-allowlisted keys are skipped at apply time and surfaced by lint (set-only; no field removal, no nested response/parameter sub-structures).

Matching is per operation (URI + method), merging matching keys in ascending precedence so the most specific source wins per field: (1) URI globs ordered by specificity = count of literal non-`*` chars, ties broken by declaration order (later wins), where `*` matches any run including `/`; (2) exact route name, applied last. A key is treated as a route name iff a route with exactly that name exists, otherwise it is a URI glob.

Two `PreBuildRule` lint rules (marker interface at `src/Lint/Visitors/PreBuildRule.php`, which exposes `checkConfiguration()`) guard config mistakes — config errors should surface through `openapi:lint`, not silently no-op. Register them in `CorePlugin::RULES` (`src/Core/CorePlugin.php:43`); both are severity-overridable:

- `overrides.unknown-field` — flags a field key outside the allowlist.
- `overrides.unused` — flags a key matching no route name and no URI glob (delegates to `OverrideMatcher::unusedKeys()`).

## Acceptance criteria

- Allowlisted fields land on the correct operations and beat an attribute-derived value on the same field; a code `transformDocument()` callback still overrides them.
- `x-*` keys emit under the operation's `x` array; non-allowlisted keys are dropped and flagged by `overrides.unknown-field`.
- `overrides.unused` fires for a config key matching neither a route name nor a URI, and both rules stay silent on valid config (assert `Finding` shape via the rule + `tests/Support/` factories, collected through a `FindingsCollector` — see `src/Lint/ArrayFindingsCollector.php` — following the pattern in `tests/Unit/Lint/Rules/FieldDescriptionMissingTest.php`).
- Unit tests for `OverrideMatcher`, `RouteIndex`, and `OverridesStage` under `tests/Unit/Support/Generator/`; lint-rule unit tests under `tests/Unit/Lint/Rules/`; one end-to-end feature test under `tests/Feature/` asserting overrides applied to a generated spec.

## References

- Contract: `src/Contracts/Generator/SpecStage.php`
- Stage registration: `src/Registry/OpenApiRegistry.php:165` (`addStage`), executed by `src/Support/Generator/SpecPipeline.php`
- Key source: `src/Support/Generator/Stages/PathsStage.php:210` (`buildOperationId`)
- Terminal stage: `src/Support/Generator/Stages/TransformersStage.php`
- Lint: `src/Lint/Visitors/PreBuildRule.php`, rules registered in `src/Core/CorePlugin.php:43`; test pattern at `tests/Unit/Lint/Rules/FieldDescriptionMissingTest.php`

Non-goals: path-/document-level overrides, regex (globs only), field removal (set-only). Affected areas: `src/Support/Generator/` (see placement note), `src/Lint/`, `config/openapi.php`, `docs/`. Milestone: v1.0.

Verify: `vendor/bin/pest tests/Unit/Support/Generator tests/Unit/Lint/Rules tests/Feature`

Radiergummi:
> **Draft scaffold for #45.** The spec lives in #45; this PR body is the implementation plan.
> Record any deviation from the plan as a PR comment. **Before placing files, reconcile the
> `Core\Generator\` vs `Support\Generator\` namespaces against the current tree** — the
> generator-chain refactor already shipped (`Support\Generator\BaselineRegistration` + stage split).

Closes #45.

---

# `openapi.overrides` Config Surface Implementation Plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

**Goal:** Add a config-keyed escape hatch (`openapi.overrides`) that mutates operation-level fields of the emitted OpenAPI document per route name or URI glob, applied as a late pipeline stage, with two lint rules that surface config mistakes.

**Architecture:** A `#[Scoped] RouteIndex` accumulator records `route-name ⇄ (method, uri)` as `PathsStage` builds operations. A pure `OverrideMatcher` utility resolves the merged, allowlist-filtered field-set for an operation using ascending-precedence matching (URI globs by specificity, then exact route name). A new `OverridesStage` reads both and mutates the assembled document. `SpecPipeline` invokes it between the registry-stage loop and the terminal `TransformersStage` so config overrides beat plugin contributions but a code-based `transformDocument()` still wins. Two `PreBuildRule`s (`overrides.unknown-field`, `overrides.unused`) run at config time.

**Tech Stack:** PHP 8.4, Laravel 12/13, swagger-php (`OpenApi\Annotations`), Pest + Orchestra Testbench, PHPStan level 8, Laravel Pint.

---

## Background notes for the implementer (read once)

These facts were verified against the codebase and you will rely on them throughout:

- **Vendor extensions** live on the `x` array property of swagger-php annotations. `$op->x = ['internal' => true]` serialises to `x-internal: true`. So a config key `x-internal` maps to `$op->x['internal']`. Before assignment, `$op->x` may equal `\OpenApi\Generator::UNDEFINED` — coerce to `[]` first.
- **HTTP verb properties** on `OA\PathItem` are the lowercase method names. The canonical list (copied from `SpecTreeBuilder::HTTP_METHODS`) is: `get, post, put, patch, delete, options, head, trace`. A verb that is unset equals `\OpenApi\Generator::UNDEFINED` — skip those.
- **Path key convention.** `PathsStage` stores `$pathItem->path` as `Str::start($uri, '/')` (leading slash). The established index-key format (see `SpecTreeBuilder::buildDescriptorIndex()`) is `strtoupper($method) . ' /' . ltrim($uri, '/')`. `RouteIndex` uses this exact format so record/lookup agree whether the caller passes a raw `route->uri()` (no slash) or a `$pathItem->path` (leading slash).
- **`OA\Operation` properties** to set: `operationId` (string), `summary` (string), `description` (string), `tags` (list<string>), `deprecated` (bool). Plus `x` for extensions.
- **Scoped lifecycle.** `#[Scoped]` classes are one shared instance per container scope. `PathsStage` and `OverridesStage` both autowire the same `RouteIndex` instance within a generation run. `record()` overwrites by key, so it is self-correcting if a key is recorded twice.
- **Config injection.** Stages read config via the `config()` helper or `#[Illuminate\Container\Attributes\Config]`. Lint rules use `#[Config('...')]` (see `SpecConfigOrphaned`).
- **File header.** Every PHP file starts with `<?php`, the MIT/copyright docblock, and `declare(strict_types=1);`. Copy the exact header block from any existing file in the target directory.
- **Section comments** use folding regions: `// region <title>` / `// endregion`.

---

## File Structure

**New files:**

- `src/Core/Generator/RouteIndex.php` — scoped accumulator mapping `(method, uri)` → route name.
- `src/Core/Generator/OverrideMatcher.php` — pure utility: precedence matching + allowlist filtering + unused-key detection. Owns the field allowlist constant.
- `src/Core/Generator/Pipeline/OverridesStage.php` — `SpecStage` that mutates the document.
- `src/Core/Lint/Rules/OverridesUnknownField.php` — `PreBuildRule` flagging non-allowlisted field keys.
- `src/Core/Lint/Rules/OverridesUnused.php` — `PreBuildRule` flagging keys matching no route or URI.
- `tests/Unit/Core/Generator/RouteIndexTest.php`
- `tests/Unit/Core/Generator/OverrideMatcherTest.php`
- `tests/Unit/Core/Generator/Pipeline/OverridesStageTest.php`
- `tests/Unit/Lint/Rules/OverridesUnknownFieldTest.php`
- `tests/Unit/Lint/Rules/OverridesUnusedTest.php`
- `tests/Feature/OverridesTest.php`

**Modified files:**

- `src/Core/Generator/Pipeline/PathsStage.php` — inject `RouteIndex`, call `record()` per operation.
- `src/Core/Generator/Pipeline/SpecPipeline.php` — inject `OverridesStage`, invoke before terminal stage.
- `src/OpenApiServiceProvider.php` — bind `OverrideMatcher` as scoped (built from config).
- `src/Core/Registry/CoreRegistration.php` — register the two lint rules in `RULES`.
- `config/openapi.php` — add the `overrides` key with an empty documented default.
- `docs/` — config reference + overrides section.
- `CHANGELOG.md` — `[Unreleased]` entry.

---

## Task 1: `RouteIndex` scoped accumulator

**Files:**
- Create: `src/Core/Generator/RouteIndex.php`
- Test: `tests/Unit/Core/Generator/RouteIndexTest.php`

- [x] **Step 1: Write the failing test**

Create `tests/Unit/Core/Generator/RouteIndexTest.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

use Radiergummi\OpenApi\Support\Generator\RouteIndex;

uses()->group('openapi');

it('records and looks up a route name by method and uri', function (): void {
    $index = new RouteIndex();
    $index->record('users/{user}', 'GET', 'users.show');

    expect($index->routeNameFor('users/{user}', 'GET'))->toBe('users.show');
});

it('normalises leading slash and method case on lookup', function (): void {
    $index = new RouteIndex();
    $index->record('users/{user}', 'get', 'users.show');

    // PathsStage records the raw route uri; OverridesStage looks up with the
    // leading-slash $pathItem->path and an uppercase verb. Both must agree.
    expect($index->routeNameFor('/users/{user}', 'GET'))->toBe('users.show');
});

it('returns null for an unknown method/uri pair', function (): void {
    $index = new RouteIndex();
    $index->record('users/{user}', 'GET', 'users.show');

    expect($index->routeNameFor('users/{user}', 'POST'))->toBeNull()
        ->and($index->routeNameFor('posts', 'GET'))->toBeNull();
});

it('records a null route name for unnamed routes', function (): void {
    $index = new RouteIndex();
    $index->record('health', 'GET', null);

    // Distinguish "recorded, unnamed" from "never recorded" via has().
    expect($index->routeNameFor('health', 'GET'))->toBeNull()
        ->and($index->has('health', 'GET'))->toBeTrue()
        ->and($index->has('missing', 'GET'))->toBeFalse();
});

it('overwrites a previously recorded entry for the same key', function (): void {
    $index = new RouteIndex();
    $index->record('users', 'GET', 'users.old');
    $index->record('users', 'GET', 'users.new');

    expect($index->routeNameFor('users', 'GET'))->toBe('users.new');
});
```

- [x] **Step 2: Run the test to verify it fails**

Run: `vendor/bin/pest tests/Unit/Core/Generator/RouteIndexTest.php`
Expected: FAIL — class `RouteIndex` not found.

- [x] **Step 3: Write the implementation**

Create `src/Core/Generator/RouteIndex.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license       MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

namespace Radiergummi\OpenApi\Core\Generator;

use Illuminate\Container\Attributes\Scoped;use function array_key_exists;use function ltrim;use function strtoupper;

/**
 * Per-run map of `(HTTP method, URI)` → route name.
 *
 * {@see \Radiergummi\OpenApi\Support\Generator\Stages\PathsStage} records an entry as it builds
 * each operation (it already holds the route name and URI); the overrides stage reads it back to
 * resolve the route name for an assembled operation without re-walking routes. Mirrors
 * {@see ComponentSchemaRegistry}: mutable per-run state, reset between runs by the scoped lifecycle.
 *
 * Keys are normalised as `"{UPPER_METHOD} /{uri-without-leading-slash}"`, matching the convention
 * in {@see \Radiergummi\OpenApi\Lint\Tree\SpecTreeBuilder}, so callers may pass either a raw
 * `route->uri()` or a leading-slash `$pathItem->path` and get a consistent key.
 */
#[Scoped]
final class RouteIndex
{
    /**
     * @var array<string, ?string> keyed by "{METHOD} /{uri}" → route name (null for unnamed routes)
     */
    private array $names = [];

    public function record(string $uri, string $method, ?string $routeName): void
    {
        $this->names[$this->key($uri, $method)] = $routeName;
    }

    public function routeNameFor(string $uri, string $method): ?string
    {
        return $this->names[$this->key($uri, $method)] ?? null;
    }

    public function has(string $uri, string $method): bool
    {
        return array_key_exists($this->key($uri, $method), $this->names);
    }

    private function key(string $uri, string $method): string
    {
        return strtoupper($method) . ' /' . ltrim($uri, '/');
    }
}
```

- [x] **Step 4: Run the test to verify it passes**

Run: `vendor/bin/pest tests/Unit/Core/Generator/RouteIndexTest.php`
Expected: PASS (5 passing).

- [x] **Step 5: Commit**

```bash
git add src/Core/Generator/RouteIndex.php tests/Unit/Core/Generator/RouteIndexTest.php
git commit -m "feat(overrides): add RouteIndex scoped accumulator"
```

---

## Task 2: `OverrideMatcher` pure utility

**Files:**
- Create: `src/Core/Generator/OverrideMatcher.php`
- Test: `tests/Unit/Core/Generator/OverrideMatcherTest.php`

This is the heart of the feature. It implements the "Matching semantics" section: per-operation precedence (URI globs by specificity ascending, ties broken by declaration order, then the exact route-name key last), the allowlist filter, and `unusedKeys()`.

**Classification rule (per-operation):** a config key is treated as an exact route-name match for an operation when it equals that operation's route name. Otherwise it is treated as a URI glob and matched against the operation's URI. (`*` matches any run of characters including `/`.)

- [x] **Step 1: Write the failing test**

Create `tests/Unit/Core/Generator/OverrideMatcherTest.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

use Radiergummi\OpenApi\Support\Generator\OverrideMatcher;

uses()->group('openapi');

it('returns the field-set for an exact route-name match', function (): void {
    $matcher = new OverrideMatcher([
        'users.show' => ['operationId' => 'getCurrentUser', 'deprecated' => true],
    ]);

    expect($matcher->fieldsFor('users.show', 'api/users/{user}'))
        ->toBe(['operationId' => 'getCurrentUser', 'deprecated' => true]);
});

it('matches a URI glob against the operation uri', function (): void {
    $matcher = new OverrideMatcher([
        'api/v1/legacy/*' => ['x-internal' => true],
    ]);

    expect($matcher->fieldsFor(null, 'api/v1/legacy/users'))
        ->toBe(['x-internal' => true]);
});

it('treats * as matching across slashes', function (): void {
    $matcher = new OverrideMatcher([
        'api/*' => ['deprecated' => true],
    ]);

    expect($matcher->fieldsFor(null, 'api/v1/deep/nested/path'))
        ->toBe(['deprecated' => true]);
});

it('merges per-field across multiple matches with the most specific winning', function (): void {
    // Mirrors the worked example in the design doc.
    $matcher = new OverrideMatcher([
        'api/v1/legacy/*'     => ['x-internal' => true],
        'api/v1/legacy/users' => ['deprecated' => true],
        'legacy.users'        => ['tags' => ['Legacy']],
    ]);

    expect($matcher->fieldsFor('legacy.users', 'api/v1/legacy/users'))
        ->toBe([
            'x-internal' => true,
            'deprecated' => true,
            'tags'       => ['Legacy'],
        ]);
});

it('lets the more specific glob override a shared field', function (): void {
    $matcher = new OverrideMatcher([
        'api/*'        => ['tags' => ['Broad']],
        'api/users/*'  => ['tags' => ['Narrow']],
    ]);

    expect($matcher->fieldsFor(null, 'api/users/42')['tags'])->toBe(['Narrow']);
});

it('breaks specificity ties by declaration order, later key winning', function (): void {
    // Both globs have identical literal-character counts.
    $matcher = new OverrideMatcher([
        'api/a/*' => ['tags' => ['First']],
        'api/b/*' => ['tags' => ['Second']],
    ]);

    // Only the second matches this uri, but assert the tie path is reachable:
    expect($matcher->fieldsFor(null, 'api/b/x')['tags'])->toBe(['Second']);
});

it('lets the exact route-name key win over any glob on the same field', function (): void {
    $matcher = new OverrideMatcher([
        'api/users/*' => ['tags' => ['FromGlob']],
        'users.show'  => ['tags' => ['FromName']],
    ]);

    expect($matcher->fieldsFor('users.show', 'api/users/42')['tags'])
        ->toBe(['FromName']);
});

it('passes through arbitrary x-* extensions', function (): void {
    $matcher = new OverrideMatcher([
        'users.show' => ['x-rate-limit' => ['window' => 60, 'max' => 100]],
    ]);

    expect($matcher->fieldsFor('users.show', 'api/users'))
        ->toBe(['x-rate-limit' => ['window' => 60, 'max' => 100]]);
});

it('drops non-allowlisted fields', function (): void {
    $matcher = new OverrideMatcher([
        'users.show' => [
            'summary'       => 'ok',
            'responses'     => ['200' => []], // not allowlisted
            'operatonId'    => 'typo',        // not allowlisted (typo)
        ],
    ]);

    expect($matcher->fieldsFor('users.show', 'api/users'))
        ->toBe(['summary' => 'ok']);
});

it('returns an empty array when nothing matches', function (): void {
    $matcher = new OverrideMatcher([
        'users.show' => ['summary' => 'ok'],
    ]);

    expect($matcher->fieldsFor('posts.index', 'api/posts'))->toBe([]);
});

it('reports keys that match no route name and no uri as unused', function (): void {
    $matcher = new OverrideMatcher([
        'users.show'      => ['summary' => 'ok'],   // matches by name
        'api/posts/*'     => ['deprecated' => true], // matches by uri glob
        'legacy.foo'      => ['summary' => 'x'],     // matches nothing
        'api/missing/*'   => ['summary' => 'x'],     // matches nothing
    ]);

    $routes = [
        ['name' => 'users.show', 'uri' => 'api/users/{user}'],
        ['name' => 'posts.index', 'uri' => 'api/posts/list'],
    ];

    expect($matcher->unusedKeys($routes))->toBe(['legacy.foo', 'api/missing/*']);
});

it('treats a key equal to an existing route name as used even if its glob form matches nothing', function (): void {
    $matcher = new OverrideMatcher([
        'users.show' => ['summary' => 'ok'],
    ]);

    $routes = [
        ['name' => 'users.show', 'uri' => 'api/users/{user}'],
    ];

    expect($matcher->unusedKeys($routes))->toBe([]);
});
```

- [x] **Step 2: Run the test to verify it fails**

Run: `vendor/bin/pest tests/Unit/Core/Generator/OverrideMatcherTest.php`
Expected: FAIL — class `OverrideMatcher` not found.

- [x] **Step 3: Write the implementation**

Create `src/Core/Generator/OverrideMatcher.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license       MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

namespace Radiergummi\OpenApi\Core\Generator;

use function array_key_first;
use function in_array;
use function ltrim;
use function preg_match;
use function preg_quote;
use function str_replace;
use function str_starts_with;
use function strlen;
use function usort;

/**
 * Resolves the merged, allowlist-filtered override field-set for a single operation, and detects
 * config keys that match nothing.
 *
 * Precedence (ascending — the most specific match wins per field):
 *   1. URI globs, ordered by specificity (count of literal, non-`*` characters); ties broken by
 *      declaration order, later key winning. `*` matches any run of characters, including `/`.
 *   2. The exact route-name key, applied last (highest precedence).
 *
 * A config key is an exact route-name match for an operation when it equals that operation's route
 * name; otherwise it is matched as a URI glob against the operation's URI.
 *
 * Pure utility — constructed with the raw `openapi.overrides` config array, no container dependency.
 */
final readonly class OverrideMatcher
{
    /**
     * Operation-level fields that may be set from config. Any `x-*` vendor extension is also
     * allowed (handled separately by {@see isAllowedField()}).
     *
     * @var list<string>
     */
    public const array ALLOWED_FIELDS = [
        'operationId',
        'summary',
        'description',
        'tags',
        'deprecated',
    ];

    /**
     * @param array<string, array<string, mixed>> $overrides
     */
    public function __construct(private array $overrides) {}

    /**
     * The merged, allowlist-filtered field-set for one operation.
     *
     * @return array<string, mixed>
     */
    public function fieldsFor(?string $routeName, string $uri): array
    {
        $uri = $this->normalise($uri);

        /** @var list<array{specificity: int, order: int, fields: array<string, mixed>}> $globMatches */
        $globMatches = [];
        $nameFields = [];
        $order = 0;

        foreach ($this->overrides as $key => $block) {
            $fields = $this->filter($block);

            if ($fields === []) {
                $order++;

                continue;
            }

            if ($routeName !== null && $key === $routeName) {
                $nameFields = $fields;
            } elseif ($this->matchesGlob($key, $uri)) {
                $globMatches[] = [
                    'specificity' => $this->specificity($key),
                    'order'       => $order,
                    'fields'      => $fields,
                ];
            }

            $order++;
        }

        // Ascending precedence: least specific first, ties by declaration order (later last).
        usort($globMatches, static function (array $a, array $b): int {
            return $a['specificity'] <=> $b['specificity']
                ?: $a['order'] <=> $b['order'];
        });

        $merged = [];

        foreach ($globMatches as $match) {
            $merged = [...$merged, ...$match['fields']];
        }

        // Exact route-name key wins last.
        return [...$merged, ...$nameFields];
    }

    /**
     * Config keys that matched no route name and no URI across the given set.
     *
     * @param list<array{name: ?string, uri: string}> $routes
     *
     * @return list<string>
     */
    public function unusedKeys(array $routes): array
    {
        $unused = [];

        foreach ($this->overrides as $key => $_block) {
            if (!$this->matchesAnyRoute((string) $key, $routes)) {
                $unused[] = (string) $key;
            }
        }

        return $unused;
    }

    /**
     * @param list<array{name: ?string, uri: string}> $routes
     */
    private function matchesAnyRoute(string $key, array $routes): bool
    {
        foreach ($routes as $route) {
            if ($route['name'] !== null && $route['name'] === $key) {
                return true;
            }

            if ($this->matchesGlob($key, $this->normalise($route['uri']))) {
                return true;
            }
        }

        return false;
    }

    /**
     * @param array<string, mixed> $block
     *
     * @return array<string, mixed>
     */
    private function filter(array $block): array
    {
        $filtered = [];

        foreach ($block as $field => $value) {
            if (self::isAllowedField((string) $field)) {
                $filtered[(string) $field] = $value;
            }
        }

        return $filtered;
    }

    public static function isAllowedField(string $field): bool
    {
        return in_array($field, self::ALLOWED_FIELDS, strict: true)
            || str_starts_with($field, 'x-');
    }

    private function matchesGlob(string $pattern, string $normalisedUri): bool
    {
        $regex = '#^' . str_replace('\*', '.*', preg_quote($this->normalise($pattern), '#')) . '$#';

        return preg_match($regex, $normalisedUri) === 1;
    }

    private function specificity(string $pattern): int
    {
        return strlen(str_replace('*', '', $this->normalise($pattern)));
    }

    private function normalise(string $uri): string
    {
        return ltrim($uri, '/');
    }
}
```

Note on `array_key_first` import: remove it if unused after final review — it is not used in the listed body, so **do not add it**. (The `use function` list above intentionally omits it.)

- [x] **Step 4: Run the test to verify it passes**

Run: `vendor/bin/pest tests/Unit/Core/Generator/OverrideMatcherTest.php`
Expected: PASS (12 passing).

- [x] **Step 5: Run PHPStan on the new files**

Run: `composer analyse -- src/Core/Generator/OverrideMatcher.php src/Core/Generator/RouteIndex.php`
(If the script does not accept path args, run `composer analyse` and confirm no new errors reference these files.)
Expected: no errors.

- [x] **Step 6: Commit**

```bash
git add src/Core/Generator/OverrideMatcher.php tests/Unit/Core/Generator/OverrideMatcherTest.php
git commit -m "feat(overrides): add OverrideMatcher precedence + allowlist utility"
```

---

## Task 3: Wire `RouteIndex` recording into `PathsStage`

**Files:**
- Modify: `src/Core/Generator/Pipeline/PathsStage.php`

`PathsStage::attachOperation()` already iterates the route's methods. Record the route name there, keyed by the route URI (raw — `RouteIndex` normalises). Use `$action->route->getName()` (may be `null`, or a `generated::*` placeholder — for the index we record the raw name, since override keys are user-authored real names; a `generated::*` name will simply never be a user key).

- [x] **Step 1: Add a failing test**

Create the test in the existing pipeline test dir — `tests/Unit/Core/Generator/Pipeline/PathsStageRouteIndexTest.php`. This verifies recording without exercising the whole introspector by going through the real generator end-to-end is heavy; instead assert via a focused feature-style test that the index is populated after a generation run. Simpler and robust: assert through the public document in the Task 9 feature test. For this task, add a unit test that calls `record()` indirectly is not possible without the introspector.

Therefore: **make this task's verification the existing suite plus a targeted feature assertion in Task 9.** Add no new unit test here; instead add a guard test asserting `PathsStage` constructor accepts a `RouteIndex`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

use Radiergummi\OpenApi\Support\Generator\RouteIndex;use Radiergummi\OpenApi\Support\Generator\Stages\PathsStage;

uses()->group('openapi');

it('resolves PathsStage with an injected RouteIndex and shares the scoped instance', function (): void {
    $stage = app(PathsStage::class);
    expect($stage)->toBeInstanceOf(PathsStage::class);

    // RouteIndex is scoped: the same instance is shared across the run.
    expect(app(RouteIndex::class))->toBe(app(RouteIndex::class));
});
```

- [x] **Step 2: Run it to verify it fails**

Run: `vendor/bin/pest tests/Unit/Core/Generator/Pipeline/PathsStageRouteIndexTest.php`
Expected: FAIL — `PathsStage` cannot be resolved (no `RouteIndex` param yet) OR passes trivially if `RouteIndex` already autowires. If it passes trivially, that's acceptable; the meaningful assertion lands in Task 9. Proceed.

- [x] **Step 3: Modify `PathsStage`**

In `src/Core/Generator/Pipeline/PathsStage.php`:

Add the import near the other `use` statements:

```php
use Radiergummi\OpenApi\Support\Generator\RouteIndex;
```

Add the constructor dependency:

```php
    public function __construct(
        private RouteIntrospector $introspector,
        private OperationBuilder $operationBuilder,
        private InclusionEvaluator $evaluator,
        private Dispatcher $events,
        private RouteIndex $routeIndex,
    ) {}
```

In `attachOperation()`, record the route name for each emitted method. Replace the `foreach ($action->route->methods() as $method)` loop body so the record happens for every non-HEAD method (place the `record()` call right after the `HEAD` continue, before building the operation schema):

```php
    private function attachOperation(OA\PathItem $pathItem, ActionDescriptor $action): void
    {
        $operation = $this->operationBuilder->build($action, [$this->deriveTag($action)]);

        foreach ($action->route->methods() as $method) {
            $upper = strtoupper($method);

            if ($upper === 'HEAD') {
                continue;
            }

            $this->routeIndex->record($action->route->uri(), $upper, $action->route->getName());

            $resolved = $operation->operationId !== null
                ? $operation
                : $operation->withOperationId($this->buildOperationId($action, $upper));

            $operationSchema = $resolved->toOpenApi($method);

            if ($operationSchema === null) {
                continue;
            }

            OpenApiExtensions::applyOperationTransformers(
                $operationSchema,
                new OperationContext($action, $upper),
            );

            $pathItem->{strtolower($upper)} = $operationSchema;
        }
    }
```

(Recording before the `$operationSchema === null` skip is intentional: webhooks are handled separately above this method and never reach here; for path operations the record reflects the route regardless of whether a schema body is produced. This keeps the index aligned with the discovered route set.)

- [x] **Step 4: Run the focused test + the full PathsStage-affecting suite**

Run: `vendor/bin/pest tests/Unit/Core/Generator/Pipeline/ tests/Feature/`
Expected: PASS — no regressions. (The injected dependency autowires via `#[Scoped]`.)

- [x] **Step 5: Commit**

```bash
git add src/Core/Generator/Pipeline/PathsStage.php tests/Unit/Core/Generator/Pipeline/PathsStageRouteIndexTest.php
git commit -m "feat(overrides): record route names into RouteIndex from PathsStage"
```

---

## Task 4: `OverridesStage`

**Files:**
- Create: `src/Core/Generator/Pipeline/OverridesStage.php`
- Test: `tests/Unit/Core/Generator/Pipeline/OverridesStageTest.php`

The stage iterates `$doc->paths`, resolves each operation's route name from `RouteIndex`, asks `OverrideMatcher` for the field-set, and assigns each field to the `OA\Operation`. `x-*` keys map onto `$op->x` (key stripped of the `x-` prefix).

- [x] **Step 1: Write the failing test**

Create `tests/Unit/Core/Generator/Pipeline/OverridesStageTest.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

use OpenApi\Annotations as OA;use OpenApi\Generator;use Radiergummi\OpenApi\Support\Generator\OverrideMatcher;use Radiergummi\OpenApi\Support\Generator\Pipeline\OverridesStage;use Radiergummi\OpenApi\Support\Generator\RouteIndex;use Radiergummi\OpenApi\Generator\GenerationContext;use Radiergummi\OpenApi\Support\Spec\SpecRegistry;

uses()->group('openapi');

/**
 * Builds a one-operation document: GET /api/users with a named route.
 */
function overridesStageDoc(): OA\OpenApi
{
    $operation = new OA\Operation([
        'path'    => '/api/users',
        'method'  => 'get',
        'tags'    => ['Users'],
    ]);

    $pathItem = new OA\PathItem(['path' => '/api/users']);
    $pathItem->get = $operation;

    $doc = new OA\OpenApi(['openapi' => '3.1.0']);
    $doc->paths = [$pathItem];

    return $doc;
}

function overridesStageCtx(): GenerationContext
{
    return new GenerationContext(app(SpecRegistry::class)->default(), 'testing');
}

it('assigns allowlisted scalar fields onto the matching operation', function (): void {
    $index = new RouteIndex();
    $index->record('api/users', 'GET', 'users.index');

    $matcher = new OverrideMatcher([
        'users.index' => [
            'operationId' => 'listUsers',
            'summary'     => 'List users',
            'deprecated'  => true,
            'tags'        => ['Identity'],
        ],
    ]);

    $doc = overridesStageDoc();
    (new OverridesStage($index, $matcher))->apply($doc, overridesStageCtx());

    $op = $doc->paths[0]->get;
    expect($op->operationId)->toBe('listUsers')
        ->and($op->summary)->toBe('List users')
        ->and($op->deprecated)->toBeTrue()
        ->and($op->tags)->toBe(['Identity']);
});

it('maps x-* keys onto the operation x array with the prefix stripped', function (): void {
    $index = new RouteIndex();
    $index->record('api/users', 'GET', 'users.index');

    $matcher = new OverrideMatcher([
        'users.index' => ['x-internal' => true, 'x-rate-limit' => ['max' => 100]],
    ]);

    $doc = overridesStageDoc();
    (new OverridesStage($index, $matcher))->apply($doc, overridesStageCtx());

    $op = $doc->paths[0]->get;
    expect($op->x)->toBe(['internal' => true, 'rate-limit' => ['max' => 100]]);
});

it('matches an operation with no route name by uri glob', function (): void {
    $index = new RouteIndex();
    $index->record('api/users', 'GET', null);

    $matcher = new OverrideMatcher([
        'api/*' => ['deprecated' => true],
    ]);

    $doc = overridesStageDoc();
    (new OverridesStage($index, $matcher))->apply($doc, overridesStageCtx());

    expect($doc->paths[0]->get->deprecated)->toBeTrue();
});

it('leaves operations untouched when nothing matches', function (): void {
    $index = new RouteIndex();
    $index->record('api/users', 'GET', 'users.index');

    $matcher = new OverrideMatcher([
        'posts.index' => ['deprecated' => true],
    ]);

    $doc = overridesStageDoc();
    (new OverridesStage($index, $matcher))->apply($doc, overridesStageCtx());

    $op = $doc->paths[0]->get;
    expect($op->operationId)->toBe(Generator::UNDEFINED)
        ->and($op->deprecated)->toBe(Generator::UNDEFINED);
});
```

- [x] **Step 2: Run the test to verify it fails**

Run: `vendor/bin/pest tests/Unit/Core/Generator/Pipeline/OverridesStageTest.php`
Expected: FAIL — class `OverridesStage` not found.

- [x] **Step 3: Write the implementation**

Create `src/Core/Generator/Pipeline/OverridesStage.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license       MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

namespace Radiergummi\OpenApi\Support\Generator\Pipeline;

use Illuminate\Container\Attributes\Scoped;use OpenApi\Annotations as OA;use OpenApi\Generator;use Radiergummi\OpenApi\Contracts\Generator\SpecStage;use Radiergummi\OpenApi\Support\Generator\OverrideMatcher;use Radiergummi\OpenApi\Support\Generator\RouteIndex;use Radiergummi\OpenApi\Generator\GenerationContext;use Radiergummi\OpenApi\Support\Generator\Stages\TransformersStage;use function is_array;use function str_starts_with;use function strtoupper;use function substr;

/**
 * Applies `config('openapi.overrides')` to assembled operations.
 *
 * Runs after every registry-resolved stage (so config overrides beat plugin contributions) but
 * before the terminal {@see TransformersStage} (so a user's code-based `transformDocument()`
 * retains the final word). Spec-only — mutates the document, never the host app.
 */
#[Scoped]
final readonly class OverridesStage implements SpecStage
{
    /** @var list<string> lowercase HTTP verb properties on an {@see OA\PathItem} */
    private const array HTTP_METHODS = [
        'get',
        'post',
        'put',
        'patch',
        'delete',
        'options',
        'head',
        'trace',
    ];

    public function __construct(
        private RouteIndex $routeIndex,
        private OverrideMatcher $matcher,
    ) {}

    public function apply(OA\OpenApi $doc, GenerationContext $ctx): void
    {
        if (!is_array($doc->paths)) {
            return;
        }

        foreach ($doc->paths as $pathItem) {
            if (!$pathItem instanceof OA\PathItem || $pathItem->path === Generator::UNDEFINED) {
                continue;
            }

            foreach (self::HTTP_METHODS as $method) {
                $operation = $pathItem->{$method} ?? null;

                if (!$operation instanceof OA\Operation) {
                    continue;
                }

                $routeName = $this->routeIndex->routeNameFor($pathItem->path, strtoupper($method));
                $fields = $this->matcher->fieldsFor($routeName, $pathItem->path);

                if ($fields === []) {
                    continue;
                }

                $this->applyFields($operation, $fields);
            }
        }
    }

    /**
     * @param array<string, mixed> $fields
     */
    private function applyFields(OA\Operation $operation, array $fields): void
    {
        foreach ($fields as $field => $value) {
            if (str_starts_with($field, 'x-')) {
                $x = $operation->x === Generator::UNDEFINED ? [] : $operation->x;
                $x[substr($field, 2)] = $value;
                $operation->x = $x;

                continue;
            }

            // Allowlisted scalar/list fields map 1:1 to OA\Operation properties.
            $operation->{$field} = $value;
        }
    }
}
```

- [x] **Step 4: Run the test to verify it passes**

Run: `vendor/bin/pest tests/Unit/Core/Generator/Pipeline/OverridesStageTest.php`
Expected: PASS (4 passing).

- [x] **Step 5: PHPStan**

Run: `composer analyse`
Expected: no errors. (If PHPStan complains about the dynamic `$operation->{$field}` assignment, the allowlist guarantees `$field` is one of the known `OA\Operation` properties; add a targeted `// @phpstan-ignore` only if a real error surfaces — do not add it preemptively.)

- [x] **Step 6: Commit**

```bash
git add src/Core/Generator/Pipeline/OverridesStage.php tests/Unit/Core/Generator/Pipeline/OverridesStageTest.php
git commit -m "feat(overrides): add OverridesStage applying config overrides to operations"
```

---

## Task 5: Bind `OverrideMatcher` and wire `OverridesStage` into `SpecPipeline`

**Files:**
- Modify: `src/OpenApiServiceProvider.php`
- Modify: `src/Core/Generator/Pipeline/SpecPipeline.php`

`OverrideMatcher` needs the config array, so it cannot autowire from an empty constructor — bind it scoped. Then inject `OverridesStage` into `SpecPipeline` and invoke it after the registry-stage loop, before the terminal stage.

- [x] **Step 1: Add a failing test**

Create `tests/Unit/Core/Generator/Pipeline/SpecPipelineOverridesTest.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

use Radiergummi\OpenApi\Support\Generator\OverrideMatcher;use Radiergummi\OpenApi\Support\Generator\SpecPipeline;

uses()->group('openapi');

it('resolves SpecPipeline with the OverridesStage dependency', function (): void {
    config()->set('openapi.overrides', ['users.show' => ['summary' => 'x']]);

    // OverrideMatcher must be resolvable and carry the configured overrides.
    $matcher = app(OverrideMatcher::class);
    expect($matcher->fieldsFor('users.show', 'api/users'))->toBe(['summary' => 'x']);

    expect(app(SpecPipeline::class))->toBeInstanceOf(SpecPipeline::class);
});
```

- [x] **Step 2: Run it to verify it fails**

Run: `vendor/bin/pest tests/Unit/Core/Generator/Pipeline/SpecPipelineOverridesTest.php`
Expected: FAIL — `OverrideMatcher` cannot be resolved (no binding; it has a required `array` constructor param).

- [x] **Step 3: Bind `OverrideMatcher` in the service provider**

In `src/OpenApiServiceProvider.php`, add the import:

```php
use Radiergummi\OpenApi\Support\Generator\OverrideMatcher;
```

Inside `registerRegistries()` (or alongside the other scoped pipeline bindings — match the surrounding style), add:

```php
        $this->app->scoped(
            OverrideMatcher::class,
            static fn (): OverrideMatcher => new OverrideMatcher(
                (array) config('openapi.overrides', []),
            ),
        );
```

- [x] **Step 4: Inject and invoke `OverridesStage` in `SpecPipeline`**

In `src/Core/Generator/Pipeline/SpecPipeline.php`, add the constructor dependency:

```php
    public function __construct(
        private OpenApiRegistry $registry,
        private Container $container,
        private OverridesStage $overridesStage,
        private TransformersStage $terminalStage,
    ) {}
```

In `run()`, invoke `$this->overridesStage->apply($doc, $ctx);` between the registry loop and the terminal stage:

```php
            foreach ($this->registry->stages() as $stageClass) {
                $stage = $this->container->make($stageClass);
                assert($stage instanceof SpecStage);
                $stage->apply($doc, $ctx);
            }

            $this->overridesStage->apply($doc, $ctx);
            $this->terminalStage->apply($doc, $ctx);

            return $doc;
```

`OverridesStage` is in the same namespace as `SpecPipeline` (`...Pipeline`), so no new `use` is required for it.

- [x] **Step 5: Run the test + full suite**

Run: `vendor/bin/pest tests/Unit/Core/Generator/Pipeline/SpecPipelineOverridesTest.php`
Expected: PASS.

Run: `vendor/bin/pest`
Expected: PASS — no regressions.

- [x] **Step 6: Commit**

```bash
git add src/OpenApiServiceProvider.php src/Core/Generator/Pipeline/SpecPipeline.php tests/Unit/Core/Generator/Pipeline/SpecPipelineOverridesTest.php
git commit -m "feat(overrides): wire OverridesStage into SpecPipeline before the terminal stage"
```

---

## Task 6: `overrides.unknown-field` lint rule

**Files:**
- Create: `src/Core/Lint/Rules/OverridesUnknownField.php`
- Test: `tests/Unit/Lint/Rules/OverridesUnknownFieldTest.php`

Pure config check: for each override block, flag any field key not in the allowlist. Does not use `$descriptors`.

- [x] **Step 1: Write the failing test**

Create `tests/Unit/Lint/Rules/OverridesUnknownFieldTest.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

use Radiergummi\OpenApi\Core\Lint\Rules\OverridesUnknownField;use Radiergummi\OpenApi\Lint\Finding;use Radiergummi\OpenApi\Lint\FindingsCollector;use Radiergummi\OpenApi\Support\Spec\SpecRegistry;

uses()->group('openapi', 'lint');

function overridesUnknownFieldCollect(array $overrides): array
{
    $collector = new class () implements FindingsCollector {
        /** @var list<Finding> */
        public array $findings = [];

        public function emit(Finding $finding): void
        {
            $this->findings[] = $finding;
        }
    };

    new OverridesUnknownField($overrides)
        ->checkConfiguration(app(SpecRegistry::class), [], $collector);

    return $collector->findings;
}

it('has the correct id and a non-zero level', function (): void {
    $rule = new OverridesUnknownField([]);

    expect($rule->id())->toBe('overrides.unknown-field')
        ->and($rule->level())->toBeGreaterThan(0);
});

it('stays silent for fully allowlisted blocks', function (): void {
    $findings = overridesUnknownFieldCollect([
        'users.show' => [
            'operationId' => 'x',
            'summary'     => 'x',
            'description' => 'x',
            'tags'        => ['x'],
            'deprecated'  => true,
            'x-internal'  => true,
        ],
    ]);

    expect($findings)->toBe([]);
});

it('flags an unknown field key', function (): void {
    $findings = overridesUnknownFieldCollect([
        'users.show' => ['operatonId' => 'typo'],
    ]);

    expect($findings)->toHaveCount(1)
        ->and($findings[0]->ruleId)->toBe('overrides.unknown-field')
        ->and($findings[0]->message)->toContain('users.show')
        ->and($findings[0]->message)->toContain('operatonId');
});

it('flags each unknown field in a block separately', function (): void {
    $findings = overridesUnknownFieldCollect([
        'users.show' => ['responses' => [], 'parameters' => []],
    ]);

    expect($findings)->toHaveCount(2);
});
```

- [x] **Step 2: Run it to verify it fails**

Run: `vendor/bin/pest tests/Unit/Lint/Rules/OverridesUnknownFieldTest.php`
Expected: FAIL — class `OverridesUnknownField` not found.

- [x] **Step 3: Write the implementation**

Create `src/Core/Lint/Rules/OverridesUnknownField.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license       MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

namespace Radiergummi\OpenApi\Core\Lint\Rules;

use Illuminate\Container\Attributes\Config;use Override;use Radiergummi\OpenApi\Contracts\Lint\Rule;use Radiergummi\OpenApi\Lint\Finding;use Radiergummi\OpenApi\Lint\FindingsCollector;use Radiergummi\OpenApi\Lint\Visitors\PreBuildRule;use Radiergummi\OpenApi\Support\Generator\OverrideMatcher;use Radiergummi\OpenApi\Support\Spec\SpecRegistry;use function implode;use function is_array;

/**
 * Flags override field keys that are not in the allowlist
 * ({@see OverrideMatcher::ALLOWED_FIELDS} plus any `x-*` extension). Such fields are silently
 * skipped at apply time, so without this rule a typo'd field key would be a silent no-op.
 */
final readonly class OverridesUnknownField implements PreBuildRule, Rule
{
    public const string ID = 'overrides.unknown-field';

    /**
     * @param array<string, array<string, mixed>> $overrides
     */
    public function __construct(
        #[Config('openapi.overrides')]
        private array $overrides = [],
    ) {}

    #[Override]
    public function id(): string
    {
        return self::ID;
    }

    #[Override]
    public function description(): string
    {
        return "Override block sets a field outside the allowlist (operationId, summary, description, tags, deprecated, x-*).";
    }

    #[Override]
    public function checkConfiguration(
        SpecRegistry $specs,
        array $descriptors,
        FindingsCollector $findings,
    ): void {
        foreach ($this->overrides as $key => $block) {
            if (!is_array($block)) {
                continue;
            }

            foreach ($block as $field => $_value) {
                if (OverrideMatcher::isAllowedField((string) $field)) {
                    continue;
                }

                $findings->emit(
                    new Finding(
                        ruleId: self::ID,
                        level: $this->level(),
                        message: "Override '{$key}' sets unknown field '{$field}'.",
                        fixHint: 'Allowed: ' . implode(', ', OverrideMatcher::ALLOWED_FIELDS) . ', x-*',
                    ),
                );
            }
        }
    }

    #[Override]
    public function level(): int
    {
        return 3;
    }
}
```

- [x] **Step 4: Run the test to verify it passes**

Run: `vendor/bin/pest tests/Unit/Lint/Rules/OverridesUnknownFieldTest.php`
Expected: PASS (4 passing).

- [x] **Step 5: Commit**

```bash
git add src/Core/Lint/Rules/OverridesUnknownField.php tests/Unit/Lint/Rules/OverridesUnknownFieldTest.php
git commit -m "feat(overrides): add overrides.unknown-field lint rule"
```

---

## Task 7: `overrides.unused` lint rule

**Files:**
- Create: `src/Core/Lint/Rules/OverridesUnused.php`
- Test: `tests/Unit/Lint/Rules/OverridesUnusedTest.php`

Flags any override key matching no route name and no route URI across `$descriptors`. Delegates matching to `OverrideMatcher::unusedKeys()`.

- [x] **Step 1: Write the failing test**

Create `tests/Unit/Lint/Rules/OverridesUnusedTest.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

use Illuminate\Routing\Route;use Radiergummi\OpenApi\Core\Lint\Rules\OverridesUnused;use Radiergummi\OpenApi\Lint\Finding;use Radiergummi\OpenApi\Lint\FindingsCollector;use Radiergummi\OpenApi\Routing\ActionDescriptor;use Radiergummi\OpenApi\Support\Generator\OverrideMatcher;use Radiergummi\OpenApi\Support\Spec\SpecRegistry;

uses()->group('openapi', 'lint');

function overridesUnusedDescriptor(string $uri, ?string $name): ActionDescriptor
{
    $route = new Route(['GET'], $uri, fn() => null);

    if ($name !== null) {
        $route->name($name);
    }

    return new ActionDescriptor(
        route: $route,
        controller: null,
        method: null,
        summary: null,
        description: null,
    );
}

function overridesUnusedCollect(array $overrides, array $descriptors): array
{
    $collector = new class () implements FindingsCollector {
        /** @var list<Finding> */
        public array $findings = [];

        public function emit(Finding $finding): void
        {
            $this->findings[] = $finding;
        }
    };

    new OverridesUnused(new OverrideMatcher($overrides))
        ->checkConfiguration(app(SpecRegistry::class), $descriptors, $collector);

    return $collector->findings;
}

it('has the correct id and a non-zero level', function (): void {
    $rule = new OverridesUnused(new OverrideMatcher([]));

    expect($rule->id())->toBe('overrides.unused')
        ->and($rule->level())->toBeGreaterThan(0);
});

it('stays silent when every key matches a route name or uri', function (): void {
    $findings = overridesUnusedCollect(
        [
            'users.show'  => ['summary' => 'x'],
            'api/posts/*' => ['deprecated' => true],
        ],
        [
            overridesUnusedDescriptor('api/users/{user}', 'users.show'),
            overridesUnusedDescriptor('api/posts/list', 'posts.index'),
        ],
    );

    expect($findings)->toBe([]);
});

it('flags a key that matches no route or uri', function (): void {
    $findings = overridesUnusedCollect(
        ['legacy.foo' => ['summary' => 'x']],
        [overridesUnusedDescriptor('api/users/{user}', 'users.show')],
    );

    expect($findings)->toHaveCount(1)
        ->and($findings[0]->ruleId)->toBe('overrides.unused')
        ->and($findings[0]->message)->toContain('legacy.foo');
});
```

- [x] **Step 2: Run it to verify it fails**

Run: `vendor/bin/pest tests/Unit/Lint/Rules/OverridesUnusedTest.php`
Expected: FAIL — class `OverridesUnused` not found.

- [x] **Step 3: Write the implementation**

Create `src/Core/Lint/Rules/OverridesUnused.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license       MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

namespace Radiergummi\OpenApi\Core\Lint\Rules;

use Override;use Radiergummi\OpenApi\Contracts\Lint\Rule;use Radiergummi\OpenApi\Lint\Finding;use Radiergummi\OpenApi\Lint\FindingsCollector;use Radiergummi\OpenApi\Lint\Visitors\PreBuildRule;use Radiergummi\OpenApi\Support\Generator\OverrideMatcher;use Radiergummi\OpenApi\Support\Spec\SpecRegistry;

/**
 * Flags override keys that match no route name and no route URI across the discovered routes.
 * Catches typo'd route names and globs that match nothing. Delegates to
 * {@see OverrideMatcher::unusedKeys()} so the matching logic is not duplicated.
 */
final readonly class OverridesUnused implements PreBuildRule, Rule
{
    public const string ID = 'overrides.unused';

    public function __construct(private OverrideMatcher $matcher) {}

    #[Override]
    public function id(): string
    {
        return self::ID;
    }

    #[Override]
    public function description(): string
    {
        return 'Override key matches no route name and no route URI.';
    }

    #[Override]
    public function checkConfiguration(
        SpecRegistry $specs,
        array $descriptors,
        FindingsCollector $findings,
    ): void {
        $routes = [];

        foreach ($descriptors as $descriptor) {
            $routes[] = [
                'name' => $descriptor->route->getName(),
                'uri'  => $descriptor->route->uri(),
            ];
        }

        foreach ($this->matcher->unusedKeys($routes) as $key) {
            $findings->emit(
                new Finding(
                    ruleId: self::ID,
                    level: $this->level(),
                    message: "Override key '{$key}' matched no route or URI.",
                    fixHint: 'Fix the route name/glob, or remove the override entry.',
                ),
            );
        }
    }

    #[Override]
    public function level(): int
    {
        return 3;
    }
}
```

- [x] **Step 4: Run the test to verify it passes**

Run: `vendor/bin/pest tests/Unit/Lint/Rules/OverridesUnusedTest.php`
Expected: PASS (3 passing).

- [x] **Step 5: Commit**

```bash
git add src/Core/Lint/Rules/OverridesUnused.php tests/Unit/Lint/Rules/OverridesUnusedTest.php
git commit -m "feat(overrides): add overrides.unused lint rule"
```

---

## Task 8: Register the two lint rules in `CorePlugin`

**Files:**
- Modify: `src/Core/Registry/CoreRegistration.php`

- [x] **Step 1: Add the rules to the `RULES` constant**

In `src/Core/Registry/CoreRegistration.php`, add the two entries at the end of the `RULES` array, after `Rules\ResponseRefUnresolvable::class,` (alongside the sibling `Rules\SpecConfigOrphaned::class` group is also acceptable — place them with the other config-soundness rules):

```php
        Rules\ResponseRefUnresolvable::class,

        Rules\OverridesUnknownField::class,
        Rules\OverridesUnused::class,
    ];
```

- [ ] **Step 2: Add a test asserting the rules are registered and resolvable**

Create `tests/Unit/Lint/Rules/OverridesRulesRegisteredTest.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

use Radiergummi\OpenApi\Core\Lint\Rules\OverridesUnknownField;use Radiergummi\OpenApi\Core\Lint\Rules\OverridesUnused;use Radiergummi\OpenApi\Lint\RuleRegistry;

uses()->group('openapi', 'lint');

it('registers both override lint rules as pre-build rules', function (): void {
    config()->set('openapi.overrides', []);

    $preBuild = app(RuleRegistry::class)->preBuildRules();

    $ids = array_map(static fn($r) => $r->id(), $preBuild);

    expect($ids)->toContain(OverridesUnknownField::ID)
        ->and($ids)->toContain(OverridesUnused::ID);
});
```

(If `RuleRegistry::preBuildRules()` is filtered by active level, set a high level first or confirm both rules' `level()` is within the default. Both are level 3; if the default test level excludes them, add `config()->set('openapi.lint.level', 3)` before resolving. Verify the exact accessor name against `RuleRegistry` — it is `preBuildRules()` per `LintRunner.php:193`.)

- [x] **Step 3: Run the test**

Run: `vendor/bin/pest tests/Unit/Lint/Rules/OverridesRulesRegisteredTest.php`
Expected: PASS. If it fails because `preBuildRules()` applies a level filter, adjust the test per the note above and re-run.

- [x] **Step 4: Commit**

```bash
git add src/Core/Registry/CoreRegistration.php tests/Unit/Lint/Rules/OverridesRulesRegisteredTest.php
git commit -m "feat(overrides): register override lint rules in CoreRegistration"
```

---

## Task 9: End-to-end feature test + config key

**Files:**
- Modify: `config/openapi.php`
- Create: `tests/Feature/OverridesTest.php`

- [x] **Step 1: Add the `overrides` config key**

In `config/openapi.php`, add a new top-level key. Place it near the `lint` block or after `request_payload_indirection` — match surrounding comment style. Insert:

```php
    /*
    |--------------------------------------------------------------------------
    | Operation Overrides
    |--------------------------------------------------------------------------
    |
    | A config-keyed escape hatch for mutating operation-level fields of the
    | emitted document per route — without touching controller code. Keyed by
    | exact route name or URI glob ('*' matches any run of characters,
    | including '/'). Each value is a field-array.
    |
    | Allowed fields: operationId, summary, description, tags, deprecated, and
    | any 'x-*' vendor extension. Other keys are skipped and reported by the
    | 'overrides.unknown-field' lint rule. Keys matching no route are reported
    | by 'overrides.unused'.
    |
    | Overrides beat plugin contributions and convention-derived values; a
    | code-based transformDocument() callback still wins. Set-only — there is no
    | field-removal semantics.
    |
    |   'overrides' => [
    |       'users.show' => [
    |           'operationId' => 'getCurrentUser',
    |           'tags'        => ['Identity'],
    |           'deprecated'  => true,
    |       ],
    |       'api/v1/legacy/*' => [
    |           'x-internal' => true,
    |       ],
    |   ],
    |
    */

    'overrides' => [],
```

- [x] **Step 2: Write the feature test**

Create `tests/Feature/OverridesTest.php`:

```php
<?php

/**
 * This file is part of radiergummi/laravel-openapi.
 *
 * @license MIT
 * @copyright (c) 2026 Moritz Friedrich
 */

declare(strict_types=1);

namespace Radiergummi\OpenApi\Tests\Feature;

use Illuminate\Support\Facades\Route;use Radiergummi\OpenApi\Attributes\Operation;use function Radiergummi\OpenApi\Tests\generateSpec;

uses()->group('openapi');

// region Fixture controller with an attribute-derived summary we can override
class OverridesFixtureController
{
    #[Operation(summary: 'Summary from attribute')]
    public function index(): array
    {
        return [];
    }

    public function legacy(): array
    {
        return [];
    }
}

beforeEach(function (): void {
    Route::get('/api/overrides/users', [OverridesFixtureController::class, 'index'])
        ->name('overrides.users');
    Route::get('/api/v1/legacy/thing', [OverridesFixtureController::class, 'legacy']);
});
// endregion

it('applies a config override keyed by route name and beats the attribute value', function (): void {
    config()->set('openapi.overrides', [
        'overrides.users' => [
            'summary'    => 'Summary from config override',
            'deprecated' => true,
            'tags'       => ['Identity'],
        ],
    ]);

    $spec = generateSpec();
    $op = $spec['paths']['/api/overrides/users']['get'];

    expect($op['summary'])->toBe('Summary from config override')
        ->and($op['deprecated'])->toBeTrue()
        ->and($op['tags'])->toBe(['Identity']);
});

it('applies a URI-glob override and emits x-* extensions', function (): void {
    config()->set('openapi.overrides', [
        'api/v1/legacy/*' => ['x-internal' => true],
    ]);

    $spec = generateSpec();
    $op = $spec['paths']['/api/v1/legacy/thing']['get'];

    expect($op['x-internal'])->toBeTrue();
});

it('leaves operations untouched when overrides is empty', function (): void {
    config()->set('openapi.overrides', []);

    $spec = generateSpec();
    $op = $spec['paths']['/api/overrides/users']['get'];

    expect($op['summary'])->toBe('Summary from attribute');
});
```

> **Implementer note:** Verify the `Operation` authoring attribute's namespace and constructor argument name against `src/Core/Attributes/Operation.php` before running — adjust the `use` and the `summary:` named arg if they differ. If no `summary` arg exists on the attribute, drop the attribute and instead assert the override beats a convention-derived value (e.g. set `operationId` and assert it differs from the auto-generated `overrides.users`).

- [x] **Step 3: Run the feature test**

Run: `vendor/bin/pest tests/Feature/OverridesTest.php`
Expected: PASS (3 passing). Fix attribute namespace/arg per the note if the first run errors on the fixture.

- [x] **Step 4: Commit**

```bash
git add config/openapi.php tests/Feature/OverridesTest.php
git commit -m "feat(overrides): add overrides config key and end-to-end feature test"
```

---

## Task 10: Docs + CHANGELOG

**Files:**
- Modify: `docs/` (config reference page + a short overrides section)
- Modify: `CHANGELOG.md`

- [x] **Step 1: Find the docs page index and config reference**

Run: `cat docs/README.md` and identify the config-reference page (the page documenting `config/openapi.php` keys) and where lint rule IDs are catalogued.

- [x] **Step 2: Add an `overrides` section to the config reference**

Document, in the established prose style of the surrounding config docs:
- Purpose: spec-only per-route field overrides from config, no controller edits.
- Config shape (route name OR URI glob keys; field-array values) — reuse the worked example from the design doc.
- Allowlisted fields: `operationId`, `summary`, `description`, `tags`, `deprecated`, `x-*`.
- Matching semantics: URI globs by specificity (literal-char count), ties by declaration order; exact route name wins; `*` matches across `/`.
- Precedence: overrides beat plugin + convention values; a code-based `transformDocument()` still wins.
- The two lint rule IDs: `overrides.unknown-field`, `overrides.unused`, severity-overridable like any rule.

- [x] **Step 3: Add the two rule IDs to the lint-rule catalog page** (if one exists — the `--list` output is authoritative, but mirror existing doc practice).

- [x] **Step 4: Add the CHANGELOG entry**

In `CHANGELOG.md` under `[Unreleased]`, add (match the existing Added/Changed grouping):

```markdown
### Added

- `openapi.overrides` config key: a spec-only escape hatch to set operation-level fields
  (`operationId`, `summary`, `description`, `tags`, `deprecated`, `x-*`) per route name or URI
  glob, applied as a late pipeline stage. Overrides beat plugin and convention-derived values; a
  code-based `transformDocument()` callback still wins.
- Lint rules `overrides.unknown-field` (flags non-allowlisted override fields) and
  `overrides.unused` (flags override keys matching no route or URI).
```

- [x] **Step 5: Commit**

```bash
git add docs/ CHANGELOG.md
git commit -m "docs(overrides): document the overrides config key and lint rules"
```

---

## Task 11: Full verification gate

**Files:** none (verification only)

- [x] **Step 1: Run the full test suite**

Run: `composer test`
Expected: all green.

- [x] **Step 2: Run PHPStan**

Run: `composer analyse`
Expected: no errors (level 8, CI-blocking).

- [x] **Step 3: Check Pint**

Run: `vendor/bin/pint --test`
Expected: no style violations. If any, run `vendor/bin/pint` and re-commit.

- [x] **Step 4: Sanity-check the lint catalog**

Run: `vendor/bin/pest --filter "openapi:lint --list"` if such a test exists, otherwise skip. Confirm `overrides.unknown-field` and `overrides.unused` appear in the registry (covered by Task 8's test).

- [x] **Step 5: Final commit (only if Pint or fixups changed files)**

```bash
git add -A
git commit -m "chore(overrides): satisfy pint/phpstan gates"
```

---

## Self-Review (completed during plan authoring)

**Spec coverage:**

| Spec section | Task |
|---|---|
| Config shape / allowlisted fields | Tasks 2 (allowlist), 9 (config key) |
| Matching semantics (glob specificity, route-name-wins, per-field merge) | Task 2 |
| `RouteIndex` scoped accumulator | Tasks 1, 3 |
| `OverrideMatcher` pure utility | Task 2 |
| `OverridesStage implements SpecStage` | Task 4 |
| Pipeline placement (after registry loop, before terminal) | Task 5 |
| `overrides.unknown-field` rule | Task 6 |
| `overrides.unused` rule (delegates to `unusedKeys()`) | Task 7 |
| Rule registration in `CoreRegistration::RULES` | Task 8 |
| `config/openapi.php` default | Task 9 |
| Docs + CHANGELOG | Task 10 |
| Testing (unit + feature + lint) | Tasks 1–9 |

**Type consistency:** `RouteIndex::record(string $uri, string $method, ?string $routeName)` and `routeNameFor(string $uri, string $method)` are used identically in Tasks 3 and 4. `OverrideMatcher::fieldsFor(?string, string)`, `unusedKeys(list<array{name,uri}>)`, and `isAllowedField(string)`/`ALLOWED_FIELDS` are consistent across Tasks 2, 4, 6, 7. `OverridesStage(RouteIndex, OverrideMatcher)` matches its construction in Tasks 4 and 5.

**Deviation from spec, noted deliberately:** the spec's classification rule ("a key is a route name iff a route with exactly that name exists in the discovered set") is implemented per-operation (`key === thisOperation's routeName`) rather than against a global name set, because the spec fixes `OverrideMatcher` as a pure utility constructed with only the overrides array and `fieldsFor(?routeName, uri)`. The two interpretations diverge only in the pathological case where a key equals one route's name *and* its glob form also matches a different operation's literal URI — vanishingly rare since route names are dotted and URIs are slashed. `unusedKeys()` applies the same per-key logic against all routes, so the two rules stay consistent. Flag this to the spec author if the pathological case matters.

**Placeholder scan:** none — every code step contains complete code; every run step has an expected outcome.

---

## Execution Handoff

After review, choose an execution approach (see skill guidance): subagent-driven (fresh subagent per task with review between) or inline execution with checkpoints.


Radiergummi:
## Plan reconciliation before implementation

The draft plan predates the generator-chain refactor and mis-assigns the feature to the Core plugin. Reconciled against the current tree + the architecture rule that **the Core plugin exists only to understand vanilla Laravel code patterns (FormRequests, API Resources, validation rules)** — the package must be fully functional with Core disabled. The override escape hatch is config-driven infrastructure unrelated to those concerns, so it lives in the **general package** and is always loaded.

### Deviations from the draft plan

| Concern | Draft plan (stale) | Reconciled |
|---|---|---|
| `RouteIndex` | `src/Core/Generator/` | `src/Support/Generator/RouteIndex.php` (`…\Support\Generator`) |
| `OverrideMatcher` | `src/Core/Generator/` | `src/Support/Generator/OverrideMatcher.php` |
| `OverridesStage` | `src/Core/Generator/Pipeline/` | `src/Support/Generator/Stages/OverridesStage.php` (`…\Support\Generator\Stages`) |
| Stage registration | `OpenApiRegistry::addStage()` from a plugin | **Injected into `SpecPipeline` as a fixed pre-terminal step**, mirroring `TransformersStage`. This is always-on (independent of plugins) and deterministically yields `baseline+plugin stages → OverridesStage → TransformersStage`. `addStage` from `CorePlugin` would place it *before* third-party plugin stages, violating precedence. |
| Lint rules | `src/Core/Lint/Rules/`, registered in `CorePlugin::RULES` | `src/Lint/Rules/Overrides{UnknownField,Unused}.php` (`…\Lint\Rules`), registered in **`BaselineRegistration::RULES`** (always-on, like `SpecConfigOrphaned`) |
| `CorePlugin` | modified | **not touched** |
| `PathsStage` edit | plan pasted a full `attachOperation()` body | surgical: add `RouteIndex` ctor param + one `record()` call (the real method takes a 3rd `GenerationContext` arg and calls `bindAction()` — the draft body omitted both) |

Tests relocate accordingly: `tests/Unit/Support/Generator/{,Stages/}`, `tests/Unit/Lint/Rules/`, `tests/Feature/OverridesTest.php`. Matching semantics, allowlist, precedence, and the two rule IDs (`overrides.unknown-field`, `overrides.unused`) are unchanged from the spec.

PR #60, opened 2026-06-01
feat: openapi:lint --fix (Phase 1 — infra + Tier A)

### LINKED ISSUE #20: lint --fix Phase 1: infrastructure + Tier A removals
## Summary

The first phase of `openapi:lint --fix` (epic #7): build the source-rewriting machinery and prove it end-to-end on the simplest class of findings — Tier A removals, where the fix is always "delete an attribute that does nothing." This lays the foundation that Phases 2 and 3 (#21, #22) extend.

## Motivation

`openapi:lint` already tells engineers exactly what is wrong, but every fix is currently manual. The epic's goal is instant, mechanical resolution for unambiguous findings. Phase 1 deliberately starts with the lowest-risk subset: the five Tier A rules whose only remedy is removing a redundant or conflicting PHP attribute. Getting the infrastructure right against these first means later phases (corrections, additive stubs) plug into a proven `Fixer`/`FixApplicator` path rather than debugging the framework and the harder fixers at once. This is a **Tier 1** change (bounded AST modification of a whitelist of attributes — never business logic).

## Proposed approach

This phase delivers the fix framework plus the Tier A fixers. The shared mechanism, command surface, exit codes, and the full fixable-rules inventory live in the epic — see Parent below; do not duplicate them here.

- Promote `nikic/php-parser` v5 to an explicit `require` in `composer.json` and use its format-preserving printer so only modified nodes are re-printed.
- Implement the fix primitives in `src/Lint/`: `Fix` (`file`, `description`, `ruleId`, `FixOperation`), the sealed `FixOperation` hierarchy (`RemoveLines`, `ReplaceLines`, `InsertBefore`, `ModifyAttribute`), plus `FixContext`, `FixResult`, and `FixApplicator` (groups fixes by file, applies bottom-to-top, skips overlapping later fixes, writes atomically).
- Define the extension interfaces: `Fixer { fix(Finding, FixContext): iterable<Fix>; }` and `FixableRule extends Rule { fixer(): Fixer; }`. `LintCommand` (`src/Console/LintCommand.php`) checks `$rule instanceof FixableRule` after collecting findings; `Finding` is `src/Lint/Finding.php`, collected via `FindingsCollector` (`src/Lint/FindingsCollector.php`).
- Implement the shared `RemoveAttributeFixer` and make the five Tier A rules implement `FixableRule`: `field.no-effect` (`src/Lint/Rules/FieldNoEffect.php`), `queryparam.duplicate` (`src/Lint/Rules/QueryParamDuplicate.php`), `response.duplicate-status` (`src/Lint/Rules/ResponseDuplicateStatus.php`), `tag.duplicate` (`src/Lint/Rules/TagDuplicate.php`), `link.duplicate-name` (`src/Lint/Rules/LinkDuplicateName.php`).
- Wire `--fix` and `--check` into `LintCommand` with the epic's exit codes, alongside the existing `--path` option (declared in the `$signature` at `src/Console/LintCommand.php:64`).
- After applying, run `vendor/bin/pint --dirty` on the modified files (fixers are trusted; no re-lint).

## Acceptance criteria

- `--check` exits 1 when a Tier A fix is pending and writes nothing; `--fix` removes the attribute and the finding is gone on re-run.
- Modified files pass `vendor/bin/pint --test`.
- The format-preserving printer leaves surrounding code, comments, and whitespace byte-identical.
- Add golden before/after fixtures plus unit tests for each of the five Tier A rules under `tests/Unit/Lint/Rules/` (model an assertion on a single rule on `tests/Unit/Lint/Rules/FieldDescriptionMissingTest.php`: instantiate the rule, build a `Tree/*Node` via the factories in `tests/Support/` — `OperationNodeFactory`, `ActionDescriptorFactory` — call its visitor method, `iterator_to_array(...)`, and assert on the resulting `Finding`). Cover `FixApplicator` conflict handling (overlapping fixes, bottom-to-top order) directly.

## References

Parent: #7 (shared fix mechanism, command surface, exit codes, and the full fixable-rules inventory). Sibling phases: #21 (Tier B/C), #22 (polish). Affected areas: `src/Lint`, `src/Console`, `composer.json`. A draft PR is open for this phase.

Verify: `vendor/bin/pest tests/Unit/Lint/Rules/`

### LINKED ISSUE #78: lint --fix: tag.duplicate & queryparam.duplicate detect on the deduplicated spec but fix on source attributes — unreachable in generate→lint→fix
## Summary

Two of the five Phase 1 (#20 / #60) `FixableRule`s — **`tag.duplicate`** and **`queryparam.duplicate`** — detect duplicates on the *already-deduplicated generated spec*, but their fixer removes a *source attribute*. Because the pipeline strips attribute-sourced duplicates **before** the linter sees the spec, these two rules can never fire from duplicate authoring attributes in the normal `generate → lint → fix` flow — so their `RemoveAttributeFixer`s are dead in that path.

> **Correction (post-empirical-test):** an earlier version of this issue also listed `response.duplicate-status`. That rule is **not** affected — responses are carried as a list (`[$primaryResponse, ...$additionalResponses]`, `OperationBuilder.php:162`) and survive into the document, so the rule fires and fixes correctly end-to-end. Verified on a real controller. Only `tag.duplicate` and `queryparam.duplicate` are affected.

Surfaced during Phase 1 dogfooding (11-app OSS corpus): 0 fixable findings across all apps. Planting duplicate attributes on a real controller and running `--check --only=<rule>`:

| Rule | Fires from duplicate attributes? | Mechanism |
|---|---|---|
| `tag.duplicate` | ❌ no | `OperationBuilder::mergeTags()` → `array_unique` (`src/Support/Generator/OperationBuilder.php:823`, called `:195`) |
| `queryparam.duplicate` | ❌ no | `CoreQueryParameterResolver` keys by name, last-wins: `$merged[$instance->name] = $instance` (`src/Plugins/Core/Resolvers/CoreQueryParameterResolver.php:39,45`) |
| `response.duplicate-status` | ✅ yes (works) | responses carried as a list, not pre-deduped |
| `link.duplicate-name` | ✅ yes (works) | detects via **reflection**, not the spec tree (`src/Lint/Rules/LinkDuplicateName.php:40`) |

Detection sources: both affected rules read the deduplicated spec tree — `TagDuplicate` reads `$operation->tags` (`src/Lint/Rules/TagDuplicate.php:31,35`); `QueryParamDuplicate` reads `$operation->queryParameters` (`src/Lint/Rules/QueryParamDuplicate.php:34`). `SpecTreeBuilder` copies those arrays verbatim from the generated `OA\OpenApi` document, which the pipeline has already deduplicated. The lint subsystem never lints an external/hand-authored spec — `LintRunner` always regenerates in-process (`--spec=` only *filters* which registered specs run, it does not load a file).

## The two affected rules are NOT equivalent

- **`queryparam.duplicate` — real correctness value lost.** The resolver dedups *last-wins*, so a second `#[QueryParam('q', required: true)]` silently overwrites a first `#[QueryParam('q', description: '…')]`. The user's first declaration is discarded with no warning, and the spec-level rule **cannot** see it (already collapsed). Only source/reflection detection catches this. → re-pointing to reflection (option 1 below) adds genuine value.
- **`tag.duplicate` — harmless.** Tags are plain strings; dedup loses no information. A duplicate `#[Tag]` is a true no-op. Flagging it is cosmetic, and a `level: 0` (error) severity is hard to justify for a redundancy that changes nothing in the output.

## Proposed fix (decide per rule)

1. **Re-point detection to reflection** (mirror `LinkDuplicateName`): count duplicate attributes on `$operation->descriptor->method` (+ class attributes) instead of reading the deduplicated spec arrays. Detection and fix then share the source-attribute basis and become reachable.
2. **Drop the `FixableRule` wiring** and keep as a non-fixable spec-correctness guardrail (only meaningful for user `transformDocument()` / `overrides` output, which the source-attribute fixer can't fix anyway).

Suggested split: **`queryparam.duplicate` → option 1** (catches silent overwrite of conflicting param declarations). **`tag.duplicate` → option 2**, or keep fixable but downgrade severity since it's cosmetic.

## Repro

```php
#[\Radiergummi\OpenApi\Attributes\Tag('Foo')]
#[\Radiergummi\OpenApi\Attributes\Tag('Foo')]              // deduped → tag.duplicate never fires
#[\Radiergummi\OpenApi\Attributes\QueryParam(name: 'q', description: 'A')]
#[\Radiergummi\OpenApi\Attributes\QueryParam(name: 'q', required: true)]  // last-wins drop → queryparam.duplicate never fires, 'A' silently lost
public function index() { ... }
```
`openapi:lint --check --only=tag.duplicate` / `--only=queryparam.duplicate` → "nothing to fix", exit 0. (By contrast, duplicate `#[Response(status: 404, …)]` *does* fire and fix.)

Refs #20, #60.

Radiergummi:
> **Phase 1 of EPIC #7.** The full design lives in #7; the phase scope is #20.
> Record any deviation from the plan as a PR comment.

Closes #20.

## Phase 1 — fix infrastructure + Tier A removals
- [x] Promote `nikic/php-parser` to an explicit `require` in `composer.json`.
- [x] `Fix`, sealed `FixOperation` (`RemoveLines`, `ReplaceLines`, `InsertBefore`, `ModifyAttribute`),
      `FixApplicator`, `FixContext`, `FixResult`.
- [x] `Fixer` interface + `FixableRule extends Rule`.
- [x] `RemoveAttributeFixer` (shared by all Tier A rules).
- [x] Wire `--fix` and `--check` into the lint command (exit codes: 0 / 1 / 2 per #7).
- [x] Update Tier A rules to `FixableRule`: `field.no-effect`, `queryparam.duplicate`,
      `response.duplicate-status`, `tag.duplicate`, `link.duplicate-name`.
- [x] ~~Run `vendor/bin/pint --dirty` on modified files after applying.~~ **Deviation:** no
      shellout — `--fix` prints the modified files and suggests the user run their formatter. Fixers
      emit byte-clean edits, so reformatting isn't required for correctness. See PR comment.
- [x] Golden before/after fixtures per Tier A rule; surrounding code stays byte-identical;
      modified files pass `pint --test`.

## Verification
- `--check` exits 1 on a pending Tier A fix and writes nothing.
- `--fix` removes the attribute and the source no longer contains the duplicate.
- Full suite green (1554 passed), PHPStan L8 clean, `pint --test` clean.

Closes #78 — `tag.duplicate` / `queryparam.duplicate` now detect on the source attributes via reflection (the generator deduplicates both before the spec, so the old spec-level checks could never fire), with severities corrected to match impact (tag 0→3, queryparam 0→1). `response.duplicate-status` was verified unaffected.


Radiergummi:
Phase 1 pushed (`282732a`).

**Deviation from the plan** — the issue calls for running `vendor/bin/pint --dirty` on modified files after `--fix`. We are **not** shelling out: the package CLI does not auto-invoke user-side tooling. The Tier A fixers already emit byte-clean edits (whole-line removals; surrounding bytes untouched), so reformatting is not required for correctness. Instead, `--fix` prints the modified files and suggests the user run their own formatter (`vendor/bin/pint --dirty`). This keeps `openapi:lint` free of post-command shellouts.

**Other notes**
- `nikic/php-parser` promoted to an explicit `require` (^5.0).
- `Fixer`/`FixableRule` live in `Lint\Fix\` (beside the `Lint\Visitors\*Rule` interfaces), not `Contracts\Lint\` — consistent with the existing split where only the bare `Rule` is a Contract.
- `--fix` does not re-lint after applying (fixers are trusted + golden-tested); the command reports remaining findings from the single lint pass.

Verification: full suite green (1554 passed), PHPStan L8 clean, `pint --test` clean.

PR #93, opened 2026-06-03
fix: scope extractor-emitted findings under lint --path (#50)

### LINKED ISSUE #50: lint --path leaks extractor-emitted findings
**Bug** · from dogfooding

## Summary

`openapi:lint --path=<glob>` can report findings for routes outside the glob — specifically findings
that extractors emit while the document is being generated, rather than findings produced by the tree
walker. `--path` is meant to scope an entire run to a slice of the API; generation-time findings must
honour that scope the same way tree-walk findings do.

Note before starting: the descriptor-level `--path` filter and a post-generation finding filter for
extractor-emitted findings **already exist** in `LintRunner` (see Root cause). Your first job is to
reproduce the leak against the current code and pin down which findings still slip through, then close
that specific gap — do not re-add a filter that is already there.

## Impact

`--path` is how you point lint at a slice of the API — for example, only the routes touched in a PR.
The expectation is a clean, scoped signal: findings for the matching routes and nothing else. If
extractor-emitted findings (the linter rule stubs are `request.empty` in `src/Lint/Rules/RequestEmpty.php`
and the response-side equivalents) bypass the scope, a scoped run becomes untrustworthy — you cannot tell
whether a finding belongs to your slice or to an unrelated route — which defeats `--path` for incremental,
change-focused linting. This surfaced from dogfooding the linter.

## Root cause

`--path` is resolved in `LintRunner::runWithCollector()` (`src/Lint/LintRunner.php`) and applied in two
places today:

1. The descriptor set is narrowed by glob via `LintRouteFilter::filter()` (`src/Lint/LintRouteFilter.php`,
   the `fnmatch($path, …)` branch around line 78), so only in-scope routes reach the tree walk.
2. Extractor-emitted findings are drained into the shared `ArrayFindingsCollector`
   (`src/Lint/ArrayFindingsCollector.php`, bound as the `FindingsCollector` interface in
   `src/Lint/FindingsCollector.php`) and then post-filtered: `LintRunner.php:179-187` builds an
   `$allowedRouteUris` set from the in-scope descriptors, and `LintRunner.php:269-280` drops any finding
   whose `location->routeUri` is not in that set.

The residual leak is the **keep-clause** at `LintRunner.php:276`:

```php
return $uri === null
    || isset($allowedRouteUris[ltrim($uri, '/')]);
```

A finding with `routeUri === null` is treated as "non-route-scoped" (pre-build / spec-level) and kept
unconditionally. Extractors that stamp `routeUri` on their `FindingLocation` are filtered correctly —
e.g. `RequestBodyExtractor::emitEmptyFindingIfWriteMethod()` sets `routeUri` at
`src/Support/Extraction/RequestBodyExtractor.php:80-86`, so its `request.empty` findings are scoped as
expected. The leak is any extractor-emitted, genuinely route-scoped finding that reaches the collector
**without** a populated `routeUri` (drained at `LintRunner.php:258-260` with only `withSpec()` applied —
no `withLocationDefaults()` from the route descriptor, unlike the `RouteRule` pass at
`LintRunner.php:480`). Such a finding passes the `$uri === null` gate and appears regardless of `--path`.

So this is not "no filter at all"; it is "the filter's null-route escape hatch lets un-stamped extractor
findings through." `Finding`/`FindingLocation` shape: `src/Lint/Finding.php` (`ruleId`, `level`, `message`,
`location`, `fixHint`, `context`, `spec`) and `src/Lint/FindingLocation.php` (`file`, `line`, `routeName`,
`routeMethod`, `routeUri`, `jsonPointer`).

## Proposed fix

Make every route-derived extractor finding carry a `routeUri` so the existing `$allowedRouteUris` filter
governs it, then the null-route keep-clause only ever applies to truly route-agnostic (pre-build /
spec-level) findings. Two complementary moves:

- At the per-spec drain (`LintRunner.php:258-260`), enrich each drained finding with the originating
  descriptor's `FindingLocation` (as the `RouteRule` pass already does via
  `withLocationDefaults(FindingLocation::fromDescriptor($descriptor))` at `LintRunner.php:476-481`) so a
  route-scoped finding always has `routeUri` set before the post-filter runs.
- Audit extractors that emit findings under `src/Support/Extraction/` for any emit site that omits
  `routeUri`, and stamp it (mirroring `RequestBodyExtractor.php:80-86`).

Keep a single scope decision: the `$allowedRouteUris` predicate at `LintRunner.php:269-280` stays the one
gate for generation-time findings. Do not add a parallel filter.

## Acceptance criteria

- `openapi:lint --path=<glob>` reports nothing for out-of-scope routes, including extractor-emitted
  findings, even when the emitting extractor did not stamp `routeUri`.
- Add a regression test under `tests/Feature/Lint/` (alongside the existing `LintCommandTest.php` and
  `RequestEmptyTest.php`) that registers two routes, has an extractor emit a route-scoped finding for the
  out-of-scope route, runs `openapi:lint` with a `--path` glob that matches only the other route, and
  asserts the scoped output contains no finding for the excluded route. `LintCommandTest.php` already
  exercises `--path` against `ResponseEmptyController` — model the new case on it.

## References

- `src/Lint/LintRunner.php` — `--path` resolution, descriptor filter, and the `$allowedRouteUris`
  post-filter (`:179-187`, `:258-260`, `:269-280`, `:476-481`).
- `src/Lint/LintRouteFilter.php` — the glob (`fnmatch`) descriptor filter.
- `src/Support/Extraction/RequestBodyExtractor.php:80-86` — example of an extractor that correctly stamps
  `routeUri`.
- `src/Lint/Finding.php`, `src/Lint/FindingLocation.php` — finding/location shape.
- `src/Console/LintCommand.php:64,151,160` — `--path` option declaration, read, and threading into
  `LintOptions`.

Verify: `vendor/bin/pest tests/Feature/Lint`

Radiergummi:
Closes #50

## Problem

`openapi:lint --path=<glob>` leaks generation-time (extractor-emitted) findings for routes outside the glob. The post-generation filter in `LintRunner` keeps any finding whose `location->routeUri` is `null` (treating null as "route-agnostic / spec-level"). Several generation-time emit sites never stamp `routeUri`, so their genuinely route-scoped findings slip through the `--path` gate.

Confirmed un-stamped, route-scoped emit sites:
- `ValidationRulesToSchema.php:345` — `rule.unknown` (no location)
- `ValidationRulesToSchema.php:967` — `rule.invalid-enum-value` (no location)
- `Plugins/Core/Support/SchemaFromFormRequest.php:276` — `request-body.schema-degraded` (file/line only)
- `Support/Generator/Stages/ErrorResponseInferenceStage.php:180` — `errors.resolver-failed` (no location)

(`RequestBodyExtractor` and `ThrowsErrorContributor` already stamp `routeUri`, which is why the existing `request.empty` `--path` test passes.)

## Plan

1. **Reproduce (RED).** Add a regression test under `tests/Feature/Lint/` that registers two routes, makes an out-of-scope route emit one of the un-stamped findings (e.g. `rule.unknown` via a FormRequest with an unknown `Rule` object), runs lint with a `--path` glob matching only the *other* route, and asserts the excluded route's finding is absent. Model on the existing `--path filter scopes extractor-emitted findings` test (`LintCommandTest.php:224`).
2. **Fix.** Ensure every route-derived generation-time finding carries `routeUri` so the existing `$allowedRouteUris` predicate governs it — keeping a single scope decision. Exact mechanism TBD after reproduction (per-site stamping vs. a generic route-context stamp during the per-route generation loop); recorded as a PR comment once decided.
3. **Verify.** `vendor/bin/pest tests/Feature/Lint`, full suite, Pint, PHPStan.

Note: #50 instructs to reproduce the leak first and close the specific residual gap, not re-add a filter that already exists.


Radiergummi:
## Implementation note — deviated from the issue's proposed fix

The issue proposed (1) enriching each finding at the per-spec drain with the originating descriptor's `FindingLocation`, and (2) stamping `routeUri` at extractor emit sites. Investigation showed move (1) isn't feasible as written and move (2) is semantically wrong for most of the leaking findings — so the fix takes a different shape, recorded here.

**Why the emit sites can't just stamp `routeUri`.** Three of the four leaking findings — `rule.unknown`, `rule.invalid-enum-value`, `request-body.schema-degraded` — come from `ValidationRulesToSchema` / `SchemaFromFormRequest`, which are **class-keyed and registry-cached** (`buildOnce` per schema class). A FormRequest / Data class is `$ref`'d by many routes, so there is no single route to attribute the finding to, and the per-spec drain is a flat list with no descriptor to enrich from. They are **schema-scoped, not route-scoped.** Only `errors.resolver-failed` is genuinely route-scoped (its `ErrorDescriptor` carries `$action`); that one does just stamp `routeUri`.

**Why not the route-context-stamp approach (stamp the "current route" during generation).** Schema builds are cached, so the finding would be attributed to whichever route triggered the build *first*. If that's an out-of-scope route but a later **in-scope** route also references the schema, the build doesn't re-run and the finding stays attributed to the out-of-scope route — **silently dropped**, hiding a real problem on a route you asked to lint. For a linter whose value is a trustworthy scoped signal, trading a visible false-positive for a silent false-negative is the wrong trade. (Discussed with @Radiergummi before implementing.)

**What it does instead — reachability scoping.** Each schema-derived finding now carries `source_class`. The post-filter keeps a `routeUri`-less finding iff its source class is **reachable from an in-scope route** — computed from the in-scope paths' `$ref`s, expanded transitively through component-to-component refs, mapped to classes via `componentClassMap`. A source class we can't place as a component is treated as route-agnostic and kept (confidence gate: never hide an in-scope finding). This keeps the single scope decision in the existing `$allowedRouteUris` block, as the issue required — just widened to also consult an `$allowedSchemaClasses` set.

**Tests:** `tests/Feature/Lint/LintCommandTest.php` — leak reproduction (`rule.unknown` dropped when its route is out-of-scope) + shared-schema guard (kept when an in-scope route shares the FormRequest). Fixtures: `UnknownRuleController` / `UnknownRuleFormRequest`.

**Explicitly out of scope:** the cleanest fix of all — generating only the in-scope slice under `--path` so findings scope themselves — re-architects the generation pipeline; left as a possible future simplification, not this PR.

Radiergummi:
Ran `/simplify` over the diff (reuse / simplification / efficiency / altitude). One trivial cleanup applied (dropped a redundant `$ref === null` clause subsumed by `!is_string`). Two consistency findings were deliberately deferred as they'd change behaviour or reach outside this fix — filed as follow-ups:

- #94 — unify the finding source-class context key on `Finding::CONTEXT_SOURCE_CLASS` (would also make these findings class-suppressible).
- #95 — unify the in-scope path-set derivation in `LintRunner` (removes the divergent slash handling + ordering coupling).

Marking ready for review. CI will run on the pushed commits.

PR #101, opened 2026-06-04
chore: centralize resolver fault-isolation (#100)

### LINKED ISSUE #100: Centralize resolver fault-isolation in OperationBuilder; remove ad-hoc per-resolver catches
## Summary

Centralize resolver fault-isolation in `OperationBuilder` and remove the ad-hoc, per-resolver
`try { … } catch (Throwable)` wrappers. Today the "one bad route shouldn't abort the whole run"
policy is duplicated inconsistently across individual resolvers; it should live at exactly one
seam, applied uniformly to every resolver (including third-party plugin resolvers).

## Motivation

Two problems with the current state:

**1. There is no outer safety net — the per-resolver broad catch is load-bearing.**
`OperationBuilder` iterates the primary-response resolvers with no try/catch around the loop
(`src/Support/Generator/OperationBuilder.php:123`), and `GenerateCommand` calls
`generateOne()` with no try/catch either — its only catch wraps `writeOutput`
(`src/Console/GenerateCommand.php:91`). So a single resolver throwing on one malformed route
(an exotic type, a weird docblock, a throw from `symfony/type-info` or `phpstan/phpdoc-parser`
deep in reflection) would abort the entire `openapi:generate` run. The per-resolver
`catch (Throwable)` is currently the *only* thing preventing that.

**2. Fault isolation is inconsistent across resolvers.** Different resolvers make different
guarantees:

- `catch (Throwable)`: `PaginatorResponseResolver:50`, `EloquentModelResponseResolver:52`,
  `SchemaFromFormRequest:115`, `SchemaFromDataClass:702`
- `catch (ReflectionException)` only: `DataResponseResolver:69` (SpatieData),
  `ResourceResponseResolver:60` (ApiResources), `FractalResponseResolver:77` (Fractal)

The narrow-catch resolvers would *already* crash the run on any non-`ReflectionException`.
So whether a bad route degrades gracefully or kills generation depends on which resolver
happens to handle it — an accident, not a design.

The broad `catch (Throwable)` is also a smell in its own right: it swallows genuine programming
errors (`TypeError`, logic bugs in our own resolver code) as a silent "no response," turning
defects into hard-to-localize missing-output rather than a stack trace.

## Proposed approach

Establish a single fault-isolation boundary in `OperationBuilder` around each resolver
invocation (the one seam that orchestrates *all* resolvers, including plugin-supplied ones).
With that in place:

- The "log a warning, skip this resolver for this route, continue" policy is decided once and
  applies uniformly.
- Individual resolvers drop their ad-hoc `catch` blocks and either stay exception-safe or throw
  meaningfully. Bugs in our code surface as stack traces; user-input failures still degrade
  gracefully with a logged warning.
- Plugin authors inherit the same fault isolation for free.

This is an altitude move — generalize the underlying mechanism rather than keep special-casing
it per resolver.

## Open design questions (resolve in a short brainstorm before implementing)

- **Boundary granularity:** per-resolver-call (skip just the failing resolver, let others run)
  vs. per-operation (skip the whole route). Per-resolver-call preserves the most output.
- **Log context:** the per-resolver catches currently log the route URI and resolver identity.
  The centralized boundary must preserve equivalent diagnostic context.
- **Catch breadth at the boundary:** `Throwable` (maximally robust) vs. a curated set that lets
  truly fatal errors propagate.
- **Opt-out:** should a resolver be able to declare "don't isolate me" (let my exceptions
  propagate)? Probably not needed initially — note it as a non-goal unless a case appears.
- **Scope beyond primary-response resolvers:** the same pattern exists in request-schema
  resolvers / extractors (`SchemaFromFormRequest`, `SchemaFromDataClass`). Decide whether this
  issue covers all resolver/extractor seams or just the primary-response loop, with the rest as
  follow-ups.

## Acceptance criteria

- A single fault-isolation boundary exists in `OperationBuilder` (or the agreed seam) covering
  every primary-response resolver call.
- The ad-hoc `catch (Throwable)` / `catch (ReflectionException)` wrappers in the in-scope
  resolvers are removed; those resolvers no longer self-isolate.
- A resolver throwing on one route produces a logged warning (with route + resolver context)
  and a graceful skip, while the rest of the document still generates — covered by a test that
  injects a throwing resolver.
- Generation no longer silently swallows programming errors where a stack trace is the desired
  signal (per the agreed catch-breadth decision).
- No behavior change to well-formed routes: snapshot/e2e suites stay byte-identical.

## References

Surfaced during the #18 review (PR #99): the new `EloquentModelResponseResolver` copied the
`catch (Throwable)` pattern from `PaginatorResponseResolver`, prompting the question of whether
the pattern itself is sound. It is pre-existing tech debt, not introduced by #18.

Radiergummi:
Closes #100.

## Plan

Establish a single fault-isolation boundary so one bad route can't abort generation, applied uniformly across every resolver seam (plugins included), and stop swallowing programming errors.

### 1. `ResolverFaultBoundary` (new, `src/Support/Registry/`, `@internal`)
`final readonly`, injected with `LoggerInterface`. One method:
```php
/** @template T; @param callable(): T $resolve; @return T|null */
public function isolate(string $resolver, ActionDescriptor $action, callable $resolve): mixed
```
`try { return $resolve(); } catch (\Exception $e) { warn(<resolver> failed for route <uri>: <msg>); return null; }`. `\Error`/`\TypeError` propagate.

### 2. Wrap four seams (per-resolver-call granularity)
- `OperationBuilder`: primary-response loop, query-param loop (`?? []`), ref-schema loop (`resolveRefSchema`) — thread `ActionDescriptor` through `readResponseAttributes`/`buildResponseFromAttribute`/`resolveRefSchema` for log context.
- `RequestBodyExtractor`: request-schema loop.
- Inject the boundary into both; update DI bindings in the service provider.

### 3. Remove 7 outer route-level catches + cleanup
- Primary-response: `PaginatorResponseResolver`, `EloquentModelResponseResolver` (`Throwable`); `DataResponseResolver`, `ResourceResponseResolver`, `FractalResponseResolver` (`ReflectionException`).
- Request-schema: `SchemaFromFormRequest`, `SchemaFromDataClass` (`Throwable`).
- **Keep** `SchemaFromDataClass` inner `catch (UnsupportedException)` (per-property degradation).
- Drop orphaned `logger`/`Throwable`/`sprintf` imports + `resolve()`/wrapper splits; add `@throws` where removal exposes one.

### 4. Tests (TDD)
- Unit `ResolverFaultBoundary`: success passthrough; `Exception` → null + warning (route + resolver); `Error`/`TypeError` → propagates.
- Integration: throwing primary-response resolver → route still builds (fallback 200) + warning; throwing query-param resolver → other params survive; throwing request-schema resolver → next tried; `TypeError` from a resolver → `build()` propagates.
- Migrate existing per-resolver "null on throw" tests to the boundary level.
- Snapshot/e2e stay byte-identical.

### 5. Docs/changelog
`CHANGELOG.md` `[Unreleased]`; `docs/plugin-authoring.md` note that resolvers needn't self-isolate (pipeline isolates `Exception`-level failures per route) and should let programming errors propagate.

## Decisions
Recorded on #100: all-four-seams scope, `catch (Exception)` (not `Throwable`), per-resolver-call granularity, no opt-out, boundary in `src/Support/Registry/`.


Radiergummi:
Implementation note — one deviation from the plan, decided mid-implementation (with @radiergummi):

**Request-schema catches were narrowed, not removed.** The plan said "remove 7 outer catches." On inspection, the two request-schema catches are *not* route-skip isolation like the five primary-response ones — they wrap user-code execution (`rules()` / `getValidationRules()`) deep inside schema construction and **degrade with useful output**:
- `SchemaFromFormRequest` emits a `request.schema-degraded` finding and returns a documented fallback `{type: object, description: "Schema introspection failed for X"}`.
- `SchemaFromDataClass` returns the partial schema built from the typed properties it already has.

The boundary cannot replicate that (it returns `null` → no body, no finding), so removing them would regress the malformed-`rules()` case. They are the same category as the inner `catch (UnsupportedException)` per-property blocks we agreed to keep. So both were **kept but narrowed `Throwable` → `Exception`**, which still honors the issue's intent (programming errors surface; consistent breadth with the boundary).

Net result:
- **Removed** (route-skip → boundary): 5 primary-response resolver catches.
- **Kept, narrowed to `Exception`**: 2 request-schema degrade-with-output catches.
- **Kept, unchanged**: inner `UnsupportedException` per-property catches.

Verification: `composer test` (1652 passed), `vendor/bin/pint --test` (passed), `composer lint` / PHPStan (no errors), snapshot group byte-identical.

Issue #68, opened 2026-06-01
EPIC: operationId & lint quality


Radiergummi:
## Summary
operationId derivation and a couple of lint-quality measures around it.

## Motivation
The generator can emit operationIds that its own linter then rejects, and it gives no nudge toward the typed returns that response inference depends on. Both hurt the out-of-the-box experience: a fresh run shouldn't fail its own lint, and users shouldn't be left guessing why an endpoint has no response schema.

## Shared context
operationIds are built in `PathsStage::buildOperationId(ActionDescriptor, string $method)` (`src/Support/Generator/Stages/PathsStage.php:210`). There are two branches: a **named-route** branch that uses the route name verbatim, and a **URI-fallback** branch that sanitises via `preg_replace('/[^a-zA-Z0-9]+/', '_', …)`. Validity is enforced *after* generation by the `operation.id-invalid-chars` rule (`src/Lint/Rules/OperationIdInvalidChars.php`, pattern `/^[A-Za-z][A-Za-z0-9._-]*$/`).

## Tasks
- [x] #42 — sanitise operationIds from named routes (bug)
- [x] #43 — selectable operationId strategy in config
- [x] #44 — "action missing return type" nudge lint rule

Issue #42, opened 2026-06-01
operationId sanitisation for named routes


Radiergummi:
## Summary

`openapi:generate` can emit `operationId`s that the bundled `openapi:lint` then rejects: when a route is named, its name is used verbatim, so characters that the linter forbids slip straight through.

## Impact

This was surfaced while dogfooding against Pelican, whose routes use namespaced names like `api:client.show` and bind segments like `{record}`. The `:` from route-name namespacing and the `{}` from implicit bindings both end up in the generated `operationId`, which then fails the package's own `operation.id-invalid-chars` rule. A fresh, unmodified run failing its own linter is exactly the kind of broken first impression this epic is closing — see the parent for why the named-route vs. URI-fallback split exists.

## Root cause

`PathsStage::buildOperationId()` (`src/Support/Generator/Stages/PathsStage.php:210`) has two branches that disagree on sanitisation. The named-route branch (lines 218-222) returns the route name (optionally suffixed with the HTTP method) **verbatim**, while the URI-fallback branch (line 224) runs `preg_replace('/[^a-zA-Z0-9]+/', '_', …)`. Only the fallback output is codegen-safe. Validity is checked after generation by `OperationIdInvalidChars` (`src/Lint/Rules/OperationIdInvalidChars.php`, id `operation.id-invalid-chars`, pattern `^[A-Za-z][A-Za-z0-9._-]*$`), so any forbidden character in a route name leaks through to the document and trips the rule.

## Proposed fix

Sanitise the named-route branch so its output always satisfies the `operation.id-invalid-chars` pattern, while preserving the `.`/`-`/`_` characters the pattern explicitly allows (do not reuse the fallback's stricter alphanumeric-only collapse, which would mangle legitimate dotted names like `api.users.show`). Strip or replace only the characters outside the allowed set — for example `:` and `{`/`}` — and ensure the result still begins with a letter. Keep this aligned with the rule's pattern so the two never drift again. The change is contained to `PathsStage::buildOperationId` in `src/Support/Generator/Stages/PathsStage.php` (the Core Plugin's paths stage), kept aligned with the `operation.id-invalid-chars` rule in `src/Lint/Rules/OperationIdInvalidChars.php`.

## Acceptance criteria

- A route named `api:client.show`, and one carrying a `{record}` segment, each yield a lint-clean `operationId`.
[...]

PR #102, opened 2026-06-04
feat(#68): selectable operationId strategy + return-type nudge lint rule

### LINKED ISSUE #43: Selectable operationId strategy in config
## Summary

Let users choose how `operationId`s are derived through a config option, and add a lint rule that flags duplicate `operationId`s across the document.

## Motivation

Today the derivation strategy is hard-coded in `PathsStage::buildOperationId()` (`src/Support/Generator/Stages/PathsStage.php:210`): named routes use the route name verbatim, unnamed routes fall back to `{method}_{sanitised_uri}`. That single policy is fine for most apps, but client-generator toolchains differ on what they want as a method name — some prefer the route name, others want a controller-derived `Controller@method` style or a slug built from the verb and path. There is no way to express that preference without forking the stage.

The collision half is the sharper problem: the named-route branch trusts the route name to be unique, but Laravel does not guarantee distinct names across method groups, and the URI-fallback collapses `[^a-zA-Z0-9]+` runs to `_`, so two distinct paths can sanitise to the same id. A document with duplicate `operationId`s is invalid per OpenAPI and silently breaks client generators — yet nothing in the pipeline notices. `RuleRegistry` (`src/Lint/RuleRegistry.php`) has no duplicate-id detection of any kind today.

## Proposed approach

Two independent pieces, both Tier 1.

1. **Strategy selection.** Introduce a config key (e.g. `operationid.strategy`) read in `PathsStage::buildOperationId()`. Keep the current behaviour as the default strategy so existing output is byte-stable; add at least one alternative (verb-plus-path slug) so the switch is exercised. The branch in `buildOperationId()` becomes a dispatch on the configured strategy; the named-route and URI-fallback logic stay as the default strategy's implementation. Whatever strategy is chosen, the result must still satisfy the `operation.id-invalid-chars` pattern (`/^[A-Za-z][A-Za-z0-9._-]*$/`, `src/Lint/Rules/OperationIdInvalidChars.php`) — reuse the existing sanitisation so no strategy can emit an id its own linter rejects (this is the failure mode #42 fixes for the named-route path).

2. **Collision lint rule.** Add a new rule under `src/Lint/Rules/` (stable id along the lines of `operation.id-duplicate`) that collects every `operationId` across all paths/operations and emits a `Finding` (`src/Lint/Finding.php`) for each id used more than once, with a `location` pointing at the offending operation. Register it alongside the other operationId rules so it ships active.

Documentation: add the config knob to `config/openapi.php` and document it on the relevant `docs/` page; add a `CHANGELOG.md` entry under `[Unreleased]`.

## Acceptance criteria

- The configured strategy is honoured: with the default strategy, generated `operationId`s match today's output; switching to the alternative strategy changes them as specified, and every strategy's output passes `operation.id-invalid-chars`.
- A document containing two operations with the same `operationId` produces a finding from the new rule; a document with all-unique ids produces none.
- Tests: add a `PathsStage` / generation test covering both strategies (assert ids per strategy via the `generateSpec()` helper in `tests/Pest.php`), and a model lint test under `tests/Unit/Lint/Rules/` following `tests/Unit/Lint/Rules/FieldDescriptionMissingTest.php` — instantiate the rule, build operation nodes via the factories in `tests/Support/` (`OperationNodeFactory`), call the visitor method, `iterator_to_array(...)`, and assert on the emitted `Finding`.

## References

Parent: #68 (operationId derivation surface and the `buildOperationId()` two-branch mechanism live there). Related: #42 (sanitise named-route operationIds — the validity guarantee this issue's strategies must also honour). Affected areas: `src/Core` / `src/Support/Generator/Stages/PathsStage.php`, `src/Lint`, `config/openapi.php`. Verify: `vendor/bin/pest tests/Unit/Lint/Rules`

### LINKED ISSUE #44: "Action missing return type" nudge lint rule
## Summary

Add a low-severity lint rule that flags controller actions which have neither a typed return value nor a response-declaring attribute, so a user can see *why* such an endpoint ends up with no response schema instead of just getting silence.

## Motivation

Response inference (#6) reads the action signature: a typed return value is what lets the generator synthesise a response schema. When an action returns an untyped value (or `mixed`) and carries no `#[Response]` / `#[ResponseResource]` attribute, there is nothing to infer from, and today the document simply omits the response with no diagnostic. That is a confusing dead end — the most common reason a generated operation looks "empty" is exactly this, and nothing currently points the user at the fix (annotate the return type, or add a response attribute). A documentation-quality nudge closes the gap without changing generation behaviour.

## Proposed approach

This is a Tier 1 doc-quality rule, not a hard error.

Add a new rule class under `src/Lint/Rules/` (e.g. `ActionMissingReturnType.php`) with a stable rule ID (e.g. `operation.return-type-missing`). Implement `Rule` plus the operation-level visitor interface `OperationRule` (`src/Lint/Visitors/OperationRule.php`) so the rule's `checkOperation()` method runs against each operation node. From the node, reach the `ActionDescriptor` (`src/Routing/ActionDescriptor.php`) — the same descriptor `RouteIntrospector` produces and carries on the operation. The descriptor exposes the action's `?ReflectionMethod $method` (for the return type) and `attributeInstances(string $attribute)` (for authoring attributes).

The rule should fire only when *both* are true: the return type is absent/`mixed`/`void`-like, and no response-declaring authoring attribute is present (`#[Response]` / `#[ResponseResource]` from `src/Attributes/Response.php` and `src/Attributes/ResponseResource.php`). When either signal exists, stay silent.

Emit via the `$context` collector producing a `Finding` (`src/Lint/Finding.php`) with `ruleId`, a low `level`, a `message`, and a `fixHint` steering the user toward typing the return or adding a response attribute. Register the rule alongside the other library/Core lint rules in `src/Support/Generator/BaselineRegistration.php` (the rule-class list around line 253, where `VisibilityAttributeNoOp::class` and `PublicEndpointContradictsMw::class` are registered); a config-driven severity override can later set its level via `config/openapi.lint.rules`.

Model the rule on `src/Lint/Rules/PublicEndpointContradictsMw.php` — it is an `OperationRule` (`checkOperation`) that pulls the `ActionDescriptor` off the operation node and reads attributes, which is exactly the shape this rule needs. (Note: `src/Lint/Rules/VisibilityAttributeNoOp.php` also reads an `ActionDescriptor`, but via the `RouteRule` `checkRoute` path rather than the operation node — prefer the `PublicEndpointContradictsMw` model.)

## Acceptance criteria

- An action with no return type (or `mixed`) and no response attribute emits the nudge `Finding` at a low `level` (the configured doc-quality severity).
- An action that declares a usable return type, or carries `#[Response]` / `#[ResponseResource]`, emits nothing.
- Add a unit test `tests/Unit/Lint/Rules/ActionMissingReturnTypeTest.php` covering both the fire case and the silence-when-typed / silence-when-attributed cases. Build the operation node and descriptor via the factories in `tests/Support/` — `ActionDescriptorFactory::forRoute(...)` (`tests/Support/ActionDescriptorFactory.php`) and `OperationNodeFactory` (`tests/Support/OperationNodeFactory.php`) — using small controller fixtures under `tests/Fixtures/Lint/` for the typed/untyped/attributed shapes. Call `checkOperation(...)`, `iterator_to_array(...)` the result, and assert on the `Finding` fields (`ruleId`, `level`, `message`, `fixHint`). Follow `tests/Unit/Lint/Rules/PublicEndpointContradictsMwTest.php` as the model (it wires `ActionDescriptorFactory` + `OperationNodeFactory` + `checkOperation` for an operation-level rule that reads the descriptor).

## References

Parent: #68 (operationId & lint-quality epic; shared lint-registration and severity-override context lives there). Related: #6 (response inference this rule nudges toward). New rule lives in `src/Lint` per the original spec.

Verify: `vendor/bin/pest tests/Unit/Lint/Rules/ActionMissingReturnTypeTest.php`

Radiergummi:
## Plan

Closes #43, closes #44 (both sub-issues of epic #68). #42 already merged.

### Prior state (discovered while scoping)
- #42's named-route operationId sanitisation is **already on `main`** (`PathsStage::sanitiseOperationId`).
- #43's **collision lint rule is already implemented**: `Lint\Rules\OperationIdDuplicate` (`operation.id-duplicate`, level 0) is registered in `BaselineRegistration`, tested, and in the docs catalog. So only the **strategy-selection half of #43 remains**.

### #43 — selectable operationId strategy
- New flat config key `operation_id_strategy` in `config/openapi.php` (mirrors the flat `error_envelope` scalar; consistent with `operation_id_case`).
  - `'route-name'` (default) — current behaviour, byte-stable: sanitised route name (`.method` suffix for multi-method routes), unnamed routes fall back to `{method}_{uri}`.
  - `'method-path'` — always `{method}_{sanitised_uri}`, ignoring route names.
- `PathsStage::buildOperationId()` dispatches on the config value; existing branches become the two strategy implementations. Unknown values fall back to the default. Every strategy's output satisfies `operation.id-invalid-chars`.
- Feature test asserting ids per strategy via `generateSpec()`; docs (`docs/config.md`) + CHANGELOG.

### #44 — "action missing return type" nudge
- New `Lint\Rules\ActionMissingReturnType` (`operation.return-type-missing`, level 3 doc-quality, off at default `--level 1`).
- `OperationRule`; reads the `ActionDescriptor` off the operation node. Fires only when **both**: the action's return type is absent / `mixed` / `void`-like, **and** no `#[Response]` / `#[ResponseResource]` (action- or class-level). Otherwise silent.
- Registered in `BaselineRegistration`. Unit test + small controller fixtures under `tests/Fixtures/Lint/`; docs catalog row + CHANGELOG.

### Verify
`vendor/bin/pest tests/Unit/Lint/Rules tests/Feature`, `vendor/bin/pint --test`, `composer lint`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,9 @@ All notable changes to this project are documented here.
 
 ### Added
 
+- `openapi.operation_id_strategy` config key selecting how each operation's `operationId` is derived: `'route-name'` (default — the existing behaviour: sanitised route name, or `{method}_{path}` for unnamed routes; byte-stable with prior output) or `'method-path'` (always `{method}_{path}`, ignoring route names — for client toolchains where route names aren't meaningful method names). Every strategy's output is sanitised to satisfy the `operation.id-invalid-chars` rule; unknown values fall back to `'route-name'`. (#43)
+- Lint rule `operation.return-type-missing` (level 3): nudges controller actions that have neither a usable return type (absent / `mixed` / `void`-like) nor a response-declaring attribute (`#[Response]` / `#[ResponseResource]`) — the most common reason a generated operation emits no response schema. Documentation-quality only; off at the default `--level 1`. (#44)
+
 - Response schemas are now derived from Eloquent model metadata when a controller returns a model
   (or a `Collection<Model>`) directly. The schema is built by reflection from `$casts`, the model's
   `@property`/`@property-read` docblock, typed `$appends` accessors, and `$hidden`/`$visible`;
--- a/config/openapi.php
+++ b/config/openapi.php
@@ -371,6 +371,27 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | operationId Strategy
+    |--------------------------------------------------------------------------
+    |
+    | Controls how each operation's `operationId` is derived. Whatever strategy
+    | is selected, the result is always sanitised to the codegen-safe identifier
+    | shape enforced by the `operation.id-invalid-chars` lint rule.
+    |
+    |   'route-name' (default) — use the route name (with a `.{method}` suffix
+    |       for multi-verb routes); unnamed routes fall back to `{method}_{path}`.
+    |   'method-path'          — always `{method}_{path}`, ignoring route names.
+    |       Useful when route names aren't stable or meaningful as method names
+    |       in a generated client.
+    |
+    | Unknown values fall back to 'route-name'.
+    |
+    */
+
+    'operation_id_strategy' => 'route-name',
+
+    /*
+    |--------------------------------------------------------------------------
     | Operation Overrides
     |--------------------------------------------------------------------------
     |
--- a/docs/config.md
+++ b/docs/config.md
@@ -17,6 +17,7 @@ this page is the at-a-glance summary.
 | `servers` | List of `OA\Server` entries. Default uses `APP_URL`. |
 | `tags` | Document-level tag descriptions keyed by tag name. |
 | `output_path` | Absolute path the `openapi:generate` command writes to and the spec route serves from. Defaults to `storage_path('openapi.yaml')`. |
+| `operation_id_strategy` | How each operation's `operationId` is derived: `'route-name'` (default — route name, or `{method}_{path}` for unnamed routes) or `'method-path'` (always `{method}_{path}`, ignoring route names). Output is always sanitised to satisfy `operation.id-invalid-chars`. Unknown values fall back to `'route-name'`. |
 | `exception_responses` | Maps exception FQCNs (or short names) to `['status', 'description']`. See [precedence](#exception-response-precedence) below. |
 | `middleware_responses` | 401/403/429 responses keyed by Laravel middleware alias (`auth`/`scope`/`throttle`) — only fills status codes no `@throws`-derived response already supplied. |
 | `security_schemes` | Custom OpenAPI security schemes. Merged with the auto-derived defaults (Passport's OAuth2 flows; a `sanctum` http/bearer scheme when any route uses `auth:sanctum`); config wins on key collision. See [Recipes → Declare custom security schemes](recipes.md#declare-custom-security-schemes). |
--- a/docs/linting.md
+++ b/docs/linting.md
@@ -347,6 +347,7 @@ plugin-registered rules.
 | `meta.too-many-suppressions` | 3 | Symbol carries an excessive number of suppression directives. |
 | `meta.unknown-rule` | 3 | #[IgnoreLint] references a rule ID not in the registry. |
 | `operation.id-naming-inconsistent` | 3 | operationId doesn't follow the configured operation_id_case convention. |
+| `operation.return-type-missing` | 3 | Action has no typed return value or response attribute, so no response schema can be inferred. |
 | `operation.summary-equals-description` | 3 | Operation summary and description are identical (redundant). |
 | `overrides.unknown-field` | 3 | An `openapi.overrides` block sets a field outside the allowlist (operationId, summary, description, tags, deprecated, x-*). |
 | `overrides.unused` | 3 | An `openapi.overrides` key matches no route name and no route URI. |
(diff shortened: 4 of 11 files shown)


Radiergummi:
### Implementation notes / deviations from the issue specs

1. **#43's collision rule was already shipped.** `Lint\Rules\OperationIdDuplicate` (`operation.id-duplicate`, level 0) already exists, is registered in `BaselineRegistration`, tested, and catalogued. So this PR only implements the **strategy-selection half** of #43.
2. **Config key shape.** Used a flat scalar `operation_id_strategy` (values `'route-name'` / `'method-path'`) rather than the issue's illustrative `operationid.strategy` nested form — matches the flat `error_envelope` precedent and the `operation_id_case` naming, and there's only one knob so nesting buys nothing.
3. **#44 reads the return type via `ActionDescriptor::$actionReflector`** (method *or* closure) rather than `$method` only, so closure routes are handled consistently. Attribute lookup uses `attributeInstances()`, which already covers action + controller level.
4. **#44 fires on `void`/`never` too.** A void action emits only a bare `200` today (no special 204 handling in Core), so the operation still looks "empty" — nudging is consistent. Level 3 keeps it off at the default `--level 1`.

Issue #63, opened 2026-06-01
EPIC: Security-scheme derivation


Radiergummi:
## Summary
How the generator turns authentication middleware into OpenAPI security schemes, plus the measures that extend it beyond Passport.

## Motivation
Authentication is one of the first things anyone consuming a spec checks. Today the generator only understands Passport, so a large and growing population of apps — anything on Sanctum, custom guards, or ability/scope middleware — gets a document that misleadingly looks unauthenticated. Closing this makes the generated spec trustworthy about auth.

## Shared context
`SecurityExtractor` (`src/Support/Extraction/SecurityExtractor.php`) owns this. It builds the scheme catalogue in `buildSchemes(): array<string, OA\SecurityScheme>` (from the `openapi.security_schemes` config map plus Passport), names the defaults in `defaultSchemeNames(): list<string>`, and resolves per-operation requirements in `forRoute(Route): array` from `auth:*` / `scope:*` middleware. Passport is wired via the constants `SCHEME_AUTHORIZATION_CODE = 'oauth2'` and `SCHEME_CLIENT_CREDENTIALS = 'oauth2ClientCredentials'`. Config knobs: `openapi.security_schemes`, `openapi.security_default_scheme`. Per-endpoint overrides come from the `#[Security]` and `#[PublicEndpoint]` attributes. The test model is `tests/Unit/Support/Extraction/SecurityExtractorTest.php`.

## Tasks
- [x] #32 — Sanctum scheme auto-derivation
- [x] #33 — abilities/scopes → scope list
- [x] #34 — configurable auth-middleware → scheme map

## References
Constructor-applied middleware feeds the same auth signal — see #16 (under the method-body epic #5).

Radiergummi:
All three children shipped: #32 (Sanctum scheme auto-derivation, #91), #33 (abilities/`ability:` → scope list, #90/#103), #34 (configurable auth-middleware → scheme map, #103). Validated across six Sanctum apps in the OSS-survey dogfooding (Pelican, Koel, Vito, AdvisingApp, Speedtest Tracker, AureusERP) — re-running the survey confirmed 7 apps now emit a security scheme where they previously showed none. Constructor-applied middleware as an auth signal remains tracked under #16 (method-body epic #5). Closing.

Issue #32, opened 2026-06-01
Sanctum security-scheme auto-derivation


Radiergummi:
**Tier 0** · formerly OAPI-065 · **v1.0 candidate**

## Summary

A Sanctum-protected API (`auth:sanctum`) generates a document with zero security schemes and empty per-operation `security`. This teaches the generator to recognise Sanctum and emit a bearer scheme, the same way it already does for Passport.

## Motivation

Sanctum is now the most common Laravel API-auth mechanism, so this is a broad miss — the surveyed apps Pelican and Koel both emitted zero schemes despite being authenticated. A spec that silently omits auth is actively misleading to anyone consuming it, which undermines trust in the whole document.

## Proposed approach

Tier 0 — middleware reflection, small, and mirrors the existing Passport code.

Root cause: `SecurityExtractor::buildSchemes()` (`src/Support/Extraction/SecurityExtractor.php:88`) derives schemes only for Passport plus the `openapi.security_schemes` config map, so a Sanctum app gets an empty catalogue and `defaultSchemeNames()` (`:326`) resolves to `[]` — even though `forRoute()` (`:210`) already detects the `auth:*` middleware. Mirror the Passport branch (constants `SCHEME_AUTHORIZATION_CODE`/`SCHEME_CLIENT_CREDENTIALS` at `:59`/`:61`): when `auth:sanctum` appears in the discovered route middleware, register an `http` / `bearer` scheme and add it to the defaults, so `forRoute()` then emits the per-operation requirement unchanged. The parent epic holds the full `SecurityExtractor` surface and config-knob context — see #63. Constructor-applied middleware feeds the same signal via #16.

## Acceptance criteria

- A route with `auth:sanctum` yields a bearer security scheme plus a per-operation `security` requirement.
- The existing Passport path is unchanged.
- Add cases for Sanctum detection (and a Passport-still-works guard) to `tests/Unit/Support/Extraction/SecurityExtractorTest.php`.
- Override hatch unchanged: the `#[Security]` / `#[PublicEndpoint]` attributes.

## References

Parent: #63 (the `SecurityExtractor` surface, config knobs, and shared auth context live there). Related: #16 (constructor middleware), #33. Milestone: v1.0. Verify: `vendor/bin/pest tests/Unit/Support/Extraction/SecurityExtractorTest.php`

Radiergummi:
## Third-wave OSS dogfooding evidence (2026-06-01)

[...]

PR #103, opened 2026-06-04
feat(#63): Sanctum abilities→scopes + configurable auth-middleware→scheme map

### LINKED ISSUE #33: `abilities:`/`scopes:` Sanctum middleware → scopes
## Summary

Sanctum's `abilities:` and `ability:` route middleware should populate the scope list on a route's per-operation security requirement, exactly the way Passport's `scope:` / `scopes:` middleware already does.

## Motivation

Once Sanctum gets its own security scheme (#32), the next gap is granularity: a route guarded by `abilities:read,write` carries meaningful authorization information that today's spec throws away. Sanctum abilities are the conventional Sanctum analogue of Passport scopes, so an app that scopes its tokens ends up with a document that says "a token is required" but never says *which* abilities — under-describing the contract for anyone consuming it. This is a small, mechanical Tier 0 follow-on that makes the Sanctum story complete rather than half-done.

## Proposed approach

This is a focused extension of `SecurityExtractor` (`src/Support/Extraction/SecurityExtractor.php`). Scope extraction lives in `extractScopes()`, which currently recognises the `scope:` and `scopes:` prefixes and collects the comma-separated values into a deduplicated list. Add `abilities:` and `ability:` to that same prefix handling so Sanctum's ability middleware feeds the identical scope list. The downstream path is unchanged: `forRoute()` already calls `extractScopes()` and hands the result to `requirementForScopes()`, so once the abilities are extracted they flow onto the per-operation `['schemeName' => [...scopes]]` requirement with no further wiring. As with Passport, a route with auth but no abilities still emits an empty scope list (token required, no specific ability), distinct from `security: []` (public).

This builds directly on #32 — that issue establishes the Sanctum scheme; this one fills in its scopes. Tier 0.

## Acceptance criteria

- A route with `abilities:read,write` lists `read` and `write` as scopes on its security requirement; `ability:admin` contributes `admin`. Multiple/duplicate abilities deduplicate like the Passport path.
- The existing Passport `scope:` / `scopes:` behaviour is unchanged.
- Add cases to `tests/Unit/Support/Extraction/SecurityExtractorTest.php`.

## References

Parent: #63 (the `SecurityExtractor` surface, the `forRoute()` / `requirementForScopes()` mechanism, config knobs, and the `#[Security]` / `#[PublicEndpoint]` override hatch all live there). Builds on #32 (Sanctum scheme auto-derivation). Related: #16 (constructor-applied middleware feeds the same auth signal). Milestone: v1.0. Verify: `vendor/bin/pest tests/Unit/Support/Extraction/SecurityExtractorTest.php`

### LINKED ISSUE #34: Configurable auth-middleware → scheme map
## Summary

Add a config map from custom guard-middleware names to security schemes, so apps using guards beyond the built-in `auth:sanctum` / `auth:api` still get per-operation security in the generated document.

## Motivation

Plenty of real apps name their auth middleware something the generator does not recognise — a custom guard alias, a bespoke `auth:` guard, or an entirely project-specific middleware. Sibling issues #32 and #33 teach the extractor about Sanctum and ability/scope middleware, but there will always be a long tail of project-specific guard names we cannot enumerate. Today those routes emit no security at all, so a protected endpoint looks open. A small config escape hatch lets the operator point any middleware name at a scheme they have declared, keeping the spec honest without us trying to special-case every guard convention in existence.

## Proposed approach

Tier 0. The work lives in `SecurityExtractor` (`src/Support/Extraction/SecurityExtractor.php`) and `config/openapi.php`. The parent epic (#63) describes the extractor surface — `buildSchemes()`, `defaultSchemeNames()`, and `forRoute()` resolving `auth:*` / `scope:*` middleware — so this issue only covers the new mapping layer.

Add a config map (e.g. `openapi.security_middleware_map`) of guard-middleware name to the name of a scheme already declared in `openapi.security_schemes`. In `forRoute()`, when a route's middleware matches a key in that map, emit the mapped scheme's per-operation requirement — alongside (not replacing) the existing `auth:*` / `scope:*` handling. The schemes themselves continue to come from the `openapi.security_schemes` catalogue via `buildSchemes()`; this issue adds only the middleware-name to scheme-name lookup, so a mapped entry must reference an already-defined scheme. The `#[Security]` / `#[PublicEndpoint]` attributes remain the per-endpoint override.

## Acceptance criteria

- A route carrying a middleware name configured in the map yields the mapped scheme plus its per-operation security requirement.
- The existing Passport / `auth:*` / config-scheme paths are unchanged for routes not covered by the map.
- Add cases to `tests/Unit/Support/Extraction/SecurityExtractorTest.php` covering: a mapped custom middleware producing the requirement, and a route with no matching map entry staying as-is.

## References

Parent: #63 (the `SecurityExtractor` surface and the `openapi.security_schemes` / `openapi.security_default_scheme` config knobs live there). Related: #32 (Sanctum auto-derivation), #33 (abilities/scopes to scope list), #16 (constructor-applied middleware as the same auth signal). Milestone: v1.0. Verify: `vendor/bin/pest tests/Unit/Support/Extraction/SecurityExtractorTest.php`

Radiergummi:
## Plan

Closes #33, closes #34 (sub-issues of epic #63). #32 already merged.

Both are focused Tier-0 extensions of `Support\Extraction\SecurityExtractor`.

### #33 — abilities/scopes Sanctum middleware → scope list
- Extend `extractScopes()` to recognise `abilities:` and `ability:` (comma-separated, deduplicated) alongside the existing `scope:` / `scopes:`. Downstream (`forRoute()` → `requirementForScopes()`) is unchanged, so abilities flow onto the per-operation scope list exactly like Passport scopes.

### #34 — configurable auth-middleware → scheme map
- New config map `openapi.security_middleware_map` (middleware name → an already-declared `security_schemes` name).
- In `forRoute()`, a route whose middleware matches a map key emits the mapped scheme's per-operation requirement, alongside the existing `auth:*` / `scope:*` handling (deduped against it). Routes with no matching key are unchanged. Mapped schemes must reference a scheme already in `security_schemes`.

### Tests / docs
- New cases in `tests/Unit/Support/Extraction/SecurityExtractorTest.php` for both.
- `config/openapi.php` doc for the new map; `docs/` (security) + `CHANGELOG.md`.

### Verify
`vendor/bin/pest tests/Unit/Support/Extraction/SecurityExtractorTest.php`, `vendor/bin/pint --test`, `composer lint`.


Radiergummi:
### Implementation notes

- **#34 matching is on the full middleware token**, including any guard argument — so `'auth:partner' => 'partner'` matches the `auth:partner` token, and a bare alias like `'verify-api-key'` matches that token. Exact-match keeps it predictable.
- **"Alongside, not replacing"** is realised as: mapped schemes are emitted first, then the existing `auth:*`/`scope:` default-scheme resolution is appended *deduped by scheme name* (so a mapped scheme that also happens to be the project default isn't listed twice). A route whose only auth signal is the mapped middleware emits just the mapped scheme.
- **#33**: both `abilities:` and `ability:` are treated as comma-separated lists (Sanctum allows lists for both), feeding the identical deduplicated scope list as Passport scopes. No downstream wiring changed.

Radiergummi:
### Code review (high) — follow-ups

**Fixed in `23c6f18`:**
- `OperationNode::hasAuthMiddleware()` now recognises `abilities:`/`ability:`, so the `operation.security-missing` backstop fires for an abilities-guarded route with no derivable scheme (the #33 change otherwise left it silently looking public).
- `extractScopes()` filters empty-string segments — an argument-less `ability:` / trailing-comma token no longer leaks an empty scope.
- Added tests: mapped-scheme + auth co-emission, dedup-on-name-collision, empty-ability.

**Surfaced but not auto-fixed (need a decision):**

1. **Sanctum abilities vs `security.invalid-scope` (level 1, on by default).** Abilities now flow into the scope list, but `SecurityInvalidScope` validates *every* scope against Passport's registered scope set (`LintRunner` populates `registeredScopes` from `Passport::scopes()` only). In a hybrid Passport+Sanctum app (Passport installed *with* registered scopes), valid Sanctum abilities would be flagged as undefined scopes. The clean fix lives in the lint rule (only validate scopes for Passport oauth2 schemes), which is outside this diff and warrants its own change/tests — suggest a follow-up issue.

2. **Mapped `auth:*` guard + auto-derivation → multi-scheme OR.** With the documented `'auth:partner' => 'partner'` example and Passport installed, an `auth:partner` route emits `[partner, oauth2, oauth2ClientCredentials]` (an OR), because the mapped scheme is emitted *alongside* the auto-derived default per #34's "alongside (not replacing)" wording. For the common case (custom non-`auth:` middleware) this is clean. If "the map should take precedence over auto-derivation for that route" is preferred, that's a small change — but it reinterprets the spec, so leaving it for your call.

Radiergummi:
### F3 resolved → precedence per-route (`b173a46`)

`forRoute()` now treats an explicit `security_middleware_map` entry as authoritative for that route: it emits only the mapped scheme(s) and no longer appends the auto-derived default, so the documented `auth:partner` example yields `[partner]` instead of `[partner, oauth2, oauth2ClientCredentials]`. Route scopes still attach to the mapped scheme; routes with no map entry are unchanged. The name-collision dedup is gone (precedence makes it unnecessary). Config comment, docs, CHANGELOG, and tests updated.

### F1 → tracked as #104

Follow-up issue opened for the `security.invalid-scope` false-positive on Sanctum abilities in Passport+Sanctum apps (the fix belongs in the lint rule, not this diff).

Issue #106, opened 2026-06-04
OAPI-017 opportunity sizing across 11 OSS apps — prioritises #12 + an under-tracked lever-2 (body-level Resource type resolution)


Radiergummi:
## Summary

Re-running the OSS dogfooding exercise against the 11 locally-available apps produced a hard quantification of the **OAPI-017 response-fidelity gap** — sizing exactly how much of it is reachable by the planned Tier-1 work (#5 / #6 / #12) versus genuinely beyond signature/whitelist inference. The headline: of the response-schema gap, **the slice reachable by better inference is small and concentrated, but #12 alone only captures ~40% of it — a second, currently-under-tracked lever (resolving *which* Resource/model a return wraps from the method body) is needed to reach the rest.**

This is a prioritization/measurement finding, not a new feature request. It informs the build order of #12, #14, #13 under epic #5.

## Method

Booted each app exactly as `openapi:generate` does (PHP 8.4), reflected every route's controller action to categorize its declared return shape, then cross-referenced the generated spec for whether we actually emitted a **substantive** response schema (a 2xx schema that resolves — through `$ref` and a `{data: …}` envelope unwrap — to ≥1 property; **thin empty-`$ref` envelopes count as not-substantive**). Resource `toArray()` bodies and controller return expressions were analyzed with `nikic/php-parser` (+ `NameResolver` for FQCNs). Analyzer scripts are in the dogfood scratch workspace (`$WS/harness/{oapi017,resource_toarray,lever2}.php`).

## Macro picture — 3,980 operations across 11 apps

| Bucket | Ops | % | Lever |
|--------|----:|----:|-------|
| **RESOLVED_RICH** — already emits a substantive response schema | 82 | 2.1% | — (working; 80 of 82 are Lychee's Spatie Data) |
| **ADDRESSABLE** — typed to a concrete Resource/Data/Model, but we emit nothing / a thin envelope | 185 | 4.6% | Tier-1 (#12 + lever-2 below) |
| **IMPERATIVE** — typed `JsonResponse`/`array`/`Response` (shape only at runtime) | 1,821 | 45.8% | Tier-2 (refused) → authoring attributes |
| **NO_SIGNAL** — no return type at all | 1,892 | 47.5% | authoring attributes only |

[...]

PR #118, opened 2026-06-04
fix: convert nested #[Response(schema:)] literal to OA object graph (#116)

### LINKED ISSUE #116: #[Response(schema:)] can't express nested properties/items (raw array → swagger-php validation fails)
## Symptom

`#[Response(status: 200, schema: [...])]` cannot express a nested object/array literal. `OperationBuilder::buildResponseFromAttribute()` does `new OA\Schema($attribute->schema)` — passing the raw PHP array straight to swagger-php — so any literal containing a `properties` (object) or `items` (array) key trips swagger-php (`properties is an object literal`, etc.) and fails generation. Only a flat `{type: 'object'}` passes, which isn't substantive.

## Reproducer

`#[Response(status: 200, schema: ['type' => 'object', 'properties' => ['logs' => ['type' => 'string']]])]` on an action returning a bare `response()->json([...])`. Surfaced on Vito's `SiteController::showDeploymentScript` / `showEnv` (the full-spec proof, #106) — had to fall back to a `transformDocument()` callback to inject the nested schema as `OA\Property` objects.

## Expected

`#[Response(schema: [...])]` accepts a one-level (ideally nested) object/array literal and emits a substantive response schema.

## Suggested fix

In `buildResponseFromAttribute()`, deep-convert array literals into `OA\Property` / `OA\Items` (recursively) instead of handing the raw array to `OA\Schema`. Mirrors how request/resource field schemas are already assembled.

Surfaced by the Vito full-spec attribute proof (#106), exemplar domain. Related to #14 (inline `response()->json` inference) — this is the attribute-side authoring path.

Radiergummi:
## What
`#[Response(schema: [...])]` with a nested `properties`/`items` literal now validates on swagger-php 5.x. Closes #116.

## Why
`buildResponseFromAttribute()` passed the raw array to `new OA\Schema(...)`, leaving nested arrays unconverted. swagger-php **5.x** rejects that (`@OA\Schema()->properties is an object literal`), failing `openapi:generate`; **6.x** silently tolerated it — which is why a 6.x-only check (mine, initially) missed it and the dedicated **swagger-php 5.8 CI job** correctly caught it.

## How
The literal is recursively converted into an `OA\Property` / `OA\Items` object graph, valid on both `^5.8` and `^6.1.2`.

## Verification
- Reproduced RED on swagger-php **5.8.3** locally, GREEN after the fix; full 5.8 suite (excl. snapshots) 1672 passed.
- GREEN on **6.1.2**: 1679 passed, Pint + PHPStan clean.
- Regression test runs a full document `validate()` through the generator (the real `GenerateCommand` path), plus the serialized-shape assertions.

Surfaced by the Vito full-spec attribute proof (#106).


Radiergummi:
Updated: this is now a **fix**, not a refutation. The swagger-php 5.8 CI job correctly failed — a nested literal `#[Response(schema:)]` is rejected by swagger-php 5.x's validator (`properties is an object literal`), only tolerated by 6.x. `buildResponseFromAttribute()` now recursively converts the literal into an `OA\Property`/`OA\Items` graph. Verified RED→GREEN on 5.8.3 and green on 6.1.2; the test validates a full document through the generator. Closes #116.

Issue #129, opened 2026-06-05
Lint: response.description-missing false-fires on $ref responses (the bundled components/responses/* the generator itself emits)


Radiergummi:
## Problem (lint correctness bug)

`response.description-missing` (level 0 — a "correctness", not "nicety", rule) **false-fires on responses that are `$ref`s to `components/responses/*`**, even though the referenced component carries a valid `description`. The rule reads `$response->description` on the un-dereferenced node, which is `null` for a `$ref` response.

The generator itself emits the bundled shared error responses **as `$ref`s** — e.g. an auto-derived Sanctum `401` is:

```json
"401": { "$ref": "#/components/responses/Unauthorized" }
```

and the component is correct:

```json
"components": { "responses": {
  "Unauthorized": { "description": "Unauthenticated" },
  "ValidationFailed": { "description": "Validation failed" },
  "TooManyRequests": { "description": "Too many requests" }
}}
```

So the document is valid (OAS 3.1 allows a Response to be a Reference Object; the referenced object has the description), but `ResponseDescriptionMissing::checkResponse()` flags it because `ResponseNode.description` is `null` — the `SpecTreeBuilder`/`ResponseNode` does not resolve the `$ref` to the component before the description check.

## Impact

This fires on **every operation that gets an auto-derived 401/422/429** — i.e. essentially every Sanctum-protected route and every FormRequest-validated write. On the AdvisingApp `/api` surface it was **28 of 34** total findings, none of them authored by app code. It makes `lint --level=1` unreachable for any app relying on the library's own auto-derived error responses + shared response components.

## Repro

AdvisingApp 1.126.2, lib `9dca7b2`: `openapi:generate` then `openapi:lint --path='api/*'` → 26× "Response 401 … has no description" + 2× "Response 422 …", all pointing at `$ref` responses whose components have descriptions.

## Workaround

Declaring an explicit `#[Response(status: 401, description: 'Unauthenticated.')]` on each op replaces the `$ref` with an inline response carrying a description (verified — no duplicate). But that is per-op boilerplate that also defeats the shared-component design; the real fix is to resolve `$ref` responses (and headers/parameters) before description checks.

## Proposed

Resolve component `$ref`s in `SpecTreeBuilder` (or in the description rules) so `description-missing` (and the analogous header/parameter rules) see the referenced object's `description`.

## Tier / area

tier-0, area:lint. Correctness bug.

_Evidence: AdvisingApp full-spec attribute proof._

Issue #128, opened 2026-06-05
`ResourceField` / `RequestField` items: cannot express array-of-$ref (collections of a nested Resource) or inline nested-object items


Radiergummi:
## Problem

`#[ResourceField]` and `#[RequestField]` cannot express an **array whose items are a `$ref`** (a collection of a nested Resource), nor an inline nested-object item schema. `items:` is treated as a **scalar type string only**.

`SchemaFromResource::buildProperty()` resolves a `$ref` solely from `type:` (a single object):

```php
if ($type !== null && class_exists($type)) {
    $ref = $this->resolveClassRef($type);     // single $ref
    ...
}
// else: scalar branch → descriptor()->applyTo()
```

and `Support/Generator/SchemaDescriptor::itemsSchema()` only ever does:

```php
return $this->items !== null
    ? new OA\Items(['type' => $this->items])   // scalar type — no class-string/$ref branch
    : new OA\Items([]);
```

So `#[ResourceField('email_addresses', type: 'array', items: ProspectEmailAddressResource::class)]` emits `items: { type: "AdvisingApp\\…\\ProspectEmailAddressResource" }` — a bogus non-JSON-Schema type, **not** `items: { $ref: … }`. Single nested refs (`type: SubResource::class`) work; **array-of-$ref does not**. This is the natural follow-on to #115 (which added scalar `items:`).

## Affected (AdvisingApp full-spec proof)

Every nested **collection** field in the Prospect/Student resources:
- `ProspectResource.email_addresses` → `ProspectEmailAddressResource` collection
- `StudentResource.email_addresses` → `StudentEmailAddressResource`, `StudentResource.phone_numbers` → `StudentPhoneNumberResource`

Symmetrically on the request side: `#[RequestField('email_addresses', type:'array')]` / `#[RequestField('enrollments', type:'array')]` / `#[RequestField('programs', type:'array')]` cannot express a per-item object sub-schema, so nested array-of-objects request bodies collapse to a prose description.

## Severity / workaround

Not a hard completeness blocker in AdvisingApp because all these fields are `conditional` (`whenLoaded`) — the parent response stays substantive on its scalar fields. **Workaround used:** `type:'array', items:'object'` + a prose `description:` naming the item class. This loses the item shape (no `$ref` reuse), so it is a fidelity gap.

## Proposed

[...]

Issue #108, opened 2026-06-04
Tier-1: resolve concrete Resource/model type from the controller return expression (`X::collection()`, `->toResource()`, `new JsonResource($typed)`)


Radiergummi:
## Summary

Resolve the concrete API-Resource (or wrapped model) type from a controller's **return expression** when it isn't on the signature — `FooResource::collection($q)`, `FooResource::make($x)`, `new FooResource($x)`, `$model->toResource()`, `new JsonResource($typedModel)`, or an `@return AnonymousResourceCollection<FooResource>` generic. This is the prerequisite that lets #12 (Resource `toArray()` reader) and #6/#18 (model→schema) actually fire on the common case where the action's declared return type is the **base** `JsonResource` / a bare `ResourceCollection` / `AnonymousResourceCollection`.

## Motivation

#12 assumes the concrete Resource is already known (via `ResourceClassLocator`, i.e. the signature). The OSS survey (#106) shows that assumption misses most of the opportunity: across the 11 apps, of the 179 addressable concrete-Resource ops, **107 return a base/anonymous Resource whose concrete type is only in the method body** — almost always `FooResource::collection(…)` on an `index`, or `->toResource()` / `new JsonResource($model)`. The signature says `AnonymousResourceCollection` / `JsonResource`; the useful type is one `::collection()` call away.

Measured realism on those 107 ops (bounded scan of the return expression, no dataflow):

| Body shape | Ops |
|------------|----:|
| `X::collection()` / `X::make()` / `new X()` names a concrete Resource | 77 |
| `new JsonResource($typed)` / `->toResource()` on a Model-typed value | 3 |
| runtime-only (no statically-nameable type) | 28 |

**~74% are statically resolvable.** Worked example: Vito's 42 collection endpoints all do `FooResource::collection(…)` over the same 21 concrete Resources that #12 already reads — so this issue + #12 together take Vito from 0 → 111 ops of real response fidelity. Provenance and full breakdown: #106.

## Proposed approach

**Tier-1**, per epic #5's bounded-AST contract — match a narrow whitelist on the method's `return` expression (and the simple `$x = X::collection(...); return $x;` two-statement case), no variable tracking across calls, no following values into services:

- `X::collection(...)` / `X::make(...)` / `new X(...)` where `X` is a concrete `JsonResource` subclass → item/resource = `X`; hand off to #12 to read its `toArray()`.
[...]

Issue #19, opened 2026-06-01
Response status-code inference


Radiergummi:
## Summary

Teach the generator to emit the *actual* success status code for an operation — `201` on a create, `204` on a delete — instead of flattening every success response into a single `200`/default.

## Motivation

Today every successful response is documented as a flat `200`/default, regardless of what the controller really returns. That is wrong for the most common Laravel CRUD shapes: a `store` action returns `201 Created`, a `destroy` action returns `204 No Content`, and a `->noContent()` call is plainly empty. Consumers reading the spec see a body where there is none, or the wrong code on resource creation — a small but pervasive fidelity gap across surveyed apps. This is one of the Tier-0 response-fidelity sources tracked under the parent epic.

## Proposed approach

Two complementary sources, matching the two tiers:

- **Tier 0 — resource convention.** Read the route's `apiResource`/`resource` semantics and map the action to its conventional success code: `store` → `201`, `destroy` → `204`, the rest → `200`. The action/route relationship is already available where operations are assembled in `src/Core` — the response resolvers in `src/Plugins/Core/Resolvers/` (e.g. `PaginatorResponseResolver.php`) are the pattern to follow, and operations are wired together in `src/Support/Generator/OperationBuilder.php`. This branch needs no method-body parsing. Coordinate the action-name detection with #25.
- **Tier 1 — explicit literal status argument.** Recover a literal status from the controller body where it is unambiguous: `response()->json($x, 201)`, `->setStatusCode(201)`, and `->noContent()` (→ `204`, empty body). This reads only literal arguments and depends on the body-scan utility from #5; non-literal arguments are left to the convention default.

Precedence: an explicit literal status (Tier 1) overrides the convention default (Tier 0). `#[Response]` (`src/Attributes/Response.php`) and the other authoring attributes in `src/Attributes/` remain the escape hatch when neither source is correct.

## Acceptance criteria

- A `store` action, or a body containing `response()->json($x, 201)`, emits a `201` success response.
- A `destroy` action, or a `->noContent()` call, emits a `204` with an empty body.
- Other actions still emit `200` when no literal status is present.
- Add a new feature test (e.g. `tests/Feature/ResponseStatusCodeTest.php`) covering both the resource-convention path and the literal-argument path, exercised through a fixture controller in the style of `tests/Fixtures/StandardResponsesFixtureController.php` (extend it or add a sibling fixture).
[...]

PR #131, opened 2026-06-05
Pre-Tier-1 foundation: fix $ref response.description-missing (#129) + dogfood re-baseline

Radiergummi:
## Purpose

Land the **pre-Tier-1 foundation**: clear the one correctness defect that would otherwise pollute every Tier-1 dogfood validation, and establish a clean lint baseline so Tier-1 inference work (#9 / #108 / #12) can be judged against a known-good reference. This is **not** a Tier-1 PR — it builds nothing inferential. It removes a false-positive and re-baselines.

Surfaced by the two full-spec attribute proofs (Vito, then AdvisingApp). On AdvisingApp's `/api` surface, **28 of 34** lint findings were a single false-positive (see Task 1); until it's fixed, no Sanctum app can reach `lint --level=1`, and Tier-1 dogfood runs would be drowning in noise.

`Closes #129.`

---

### Problem (verified)
The generator emits its bundled shared error responses **as Reference Objects**:

```json
"401": { "$ref": "#/components/responses/Unauthorized" }
```

and the component is valid (OAS 3.1 permits a Response to be a Reference Object; the referenced object carries the description):

```json
"components": { "responses": {
  "Unauthorized":     { "description": "Unauthenticated" },
  "ValidationFailed": { "description": "Validation failed" },
  "TooManyRequests":  { "description": "Too many requests" }
}}
```

But `ResponseDescriptionMissing` (level **0** — a correctness rule, `src/Lint/Rules/ResponseDescriptionMissing.php`) reads `ResponseNode->description`, which is `null` for a `$ref` response because **the tree builder never dereferences the component**. So the rule fires on the library's *own* auto-derived `401`/`422`/`429` — every Sanctum-protected route and every FormRequest-validated write.

### Where (exact)
`src/Lint/Tree/SpecTreeBuilder.php`, `buildResponses(OA\Operation $operation)` (~L694–800). Line ~714:

```php
$description = SchemaAccessor::undefinedToNull($response->description);
```

For a `$ref` response, swagger-php sets `$response->ref` (a `#/components/responses/{name}` string) and leaves `$response->description` UNDEFINED → `$description` is `null`. There is no resolution step.

### Fix
When a response node is a Reference Object pointing at `#/components/responses/{name}`, resolve the referenced `OA\Response` and use **its** `description` for the `ResponseNode` (so the rule sees the real description). Minimum viable fix is the description; consider also surfacing the referenced `content`/`headers` so any other response-shape rule sees through the ref too (see "Scope" below).

Implementation notes:
- `buildResponses()` has no components in scope. `build()` (top of file) holds `$spec`; build a `name → OA\Response` lookup from `$spec->components->responses` once and thread it in (constructor/instance property, mirroring how component **schemas** are already indexed — see `componentsByName` in `src/Lint/TreeIndex.php` and `buildComponents()` in this file). Prefer an instance property set during `build()` over changing every `buildResponses` call site if cleaner.
- Read the ref off the response with the existing pattern: `$response->ref` guarded by `is_defined(...)` (see `SchemaAccessor::extractRef()` at `src/Lint/Tree/SchemaAccessor.php:28` for the schema analog — there may be room for a small `extractResponseRef()` helper, or reuse the `#/components/(responses)/(.+)` regex already in `TreeIndex.php:126`).
- Resolve `#/components/responses/{name}` → the `OA\Response` in the lookup → use its `description` (run it through `SchemaAccessor::undefinedToNull`). If the ref is dangling (name not in components), leave `description` null and do nothing — a dangling `$ref` is a *different* rule's job (`spec.invalid` / ref-undefined); do **not** crash.
- Only override when the local node's own `description` is absent (a response that is *both* a ref and carries an inline description — unusual — should keep its inline one).

### Scope — audit the same pattern for headers & parameters
The identical "read `->description` off an un-dereferenced node" shape exists for:
- `buildHeader()` (~L833) → `HeaderDescriptionMissing`
- parameter building (~L293 / L397) → `ParameterDescriptionMissing`

`$ref`'d headers/parameters would false-fire the same way. **Audit both.** If the generator can emit `$ref`'d headers/params (check whether anything emits `components.headers` / `components.parameters`), fix them consistently in the same PR. If it provably cannot today, add a short code comment noting the latent issue and leave them — do not expand scope speculatively. State which you found in the PR description when you finish.

### Tests (in-repo — this is the real gate)
- Add a feature/lint test: assemble (or generate) a document whose operation has a `$ref` response into `components.responses.Unauthorized` (description present), run the lint walker, assert **`response.description-missing` does NOT fire** for that response. There is existing lint-rule test scaffolding under `tests/` (find the current `ResponseDescriptionMissing` coverage and the shared-error-response fixtures — the `error_envelope = 'none'` path is what emits these shared `components.responses`; see `docs/recipes.md` → "Choosing an error envelope").
- Add a negative test: an operation with an **inline** response that genuinely has no description still fires (guard against over-correcting).
- Add a dangling-`$ref` test: a response `$ref`ing a non-existent component does not crash the builder and does not silently pass the description check.
- Run the **full** suite + the byte-exact snapshot group — shared-response emission is snapshot-covered, so expect snapshot churn only if you change emission (you should NOT — this is lint-side only).

### Gates (must be green before un-drafting)
- `composer test` (Pest), `vendor/bin/pint --test`, `composer lint` (PHPStan L8) — all clean.
- The CI matrix incl. the swagger-php **5.8** job and the Snapshot job.

---

## Task 2 — Re-baseline the dogfood lint (Tier-1 reference point)

After Task 1 lands, capture the corrected baseline so Tier-1 work has a clean reference:
1. Using the dogfood harness (`/Users/moritz/Projects/laravel-openapi-dogfood`, `harness/run.sh`, `harness/completeness.php`), regenerate + lint a Sanctum app on a **pristine** (un-annotated) checkout — Vito or a fresh AdvisingApp — with the library symlinked to this branch.
2. Confirm the auto-derived `401`/`422`/`429` `$ref` responses **no longer** trip `response.description-missing` (the count should drop by exactly the number of `$ref`'d shared error responses).
3. Record the new black-box lint baselines in the dogfood survey notes. This is the legible "before" for Tier-1.

This task is verification only — **no library changes**. Keep all dogfood-app edits uncommitted scratch; never commit inside `laravel-openapi-dogfood`.

---

## Explicitly DEFERRED — not pre-Tier-1 (do **not** scope-creep into this PR)

- **#128 (array-of-`$ref` via `items:`)** — a real attribute-surface capability gap, but independent of Tier-1 and not blocking. Separate PR.
- **#19 (204 / status-code inference)** — this *is* Tier-0/1 inference work; it belongs **in** the Tier-1 program, not before it.
- **SpatieData real-world datapoint** — the plugin is already well covered in-suite (`tests/Feature/Plugins/SpatieData/*`, `tests/Unit/Plugins/SpatieData/*`, and the `examples/spatie-data/` worked example with a snapshot). The only gap is that no *surveyed OSS app* exercised it — a nice-to-have dogfood datapoint, not a correctness prerequisite. Skip for now.
- **operationId recipe (PR #130)** — already open separately; unrelated.

---

## Context for the implementing agent
- **Never edit to make a dogfood app pass** — the library is the system under test. App-side annotations (if any) stay uncommitted scratch in `laravel-openapi-dogfood`.
- Project workflow: squash-merge into `main`, gated on green CI + Moritz's review; keep history linear. Behaviour changes need a `CHANGELOG.md` `[Unreleased]` entry — Task 1 is a lint correctness fix (observable), so **add a CHANGELOG entry**. Update `docs/linting.md` only if the rule's documented behaviour description changes (it shouldn't — the rule's *intent* is unchanged; only its `$ref` resolution is fixed).
- Evidence trail: issue #129 (full repro), and the AdvisingApp full-spec runbook at `laravel-openapi-dogfood/apps/AdvisingApp/full-spec-runbook.md`.


Radiergummi:
## Implementation complete — both tasks done, all gates green

### Task 1 — `response.description-missing` $ref false-positive (#129)
`SpecTreeBuilder` now indexes `components.responses` (`indexComponentResponses()`, mirroring the existing `indexComponentSchemas()`) and dereferences a `$ref` response to the referenced component's `description` (`resolveReferencedResponseDescription()`), applied only when the local node has no inline description. Added `SchemaAccessor::extractResponseRef()` (the response analog of `extractRef()`). A dangling `$ref` leaves the description absent — `ref.broken`'s job — rather than crashing or silently passing.

**Headers/parameters audit (Scope section):** the generator **provably cannot** emit `$ref`'d headers or parameters today — there is no `registerNamedHeader`/`registerNamedParameter`, nothing writes `components.headers`/`components.parameters`, and headers/params are only ever inlined onto operations (`OperationBuilder`). Per the plan's instruction I left them unfixed and added a short code comment at each site (`buildHeader()`, `buildParameters()`) noting the latent identical bug and where to wire the dereference if that changes. **No speculative scope expansion.**

**Tests** (`tests/Unit/Lint/Tree/SpecTreeBuilderRefResponseDescriptionTest.php`, end-to-end through `SpecTreeBuilder` → `SpecTreeWalker`):
- `$ref` response → referenced component description present → rule does **not** fire ✅
- inline response with genuinely no description → still fires (no over-correction) ✅
- dangling `$ref` → no crash, does **not** silently pass ✅

**Gates:** full Pest suite **1686 passed** (incl. byte-exact snapshot group — no churn, lint-side only), Pint clean, PHPStan L8 clean. CHANGELOG `[Unreleased] → Fixed` entry added. `docs/linting.md` unchanged (rule intent unchanged).

### Task 2 — dogfood re-baseline (verification only, no lib changes)
AdvisingApp (Sanctum, pristine/un-annotated), PHP 8.4, clean before/after isolation on the same generated spec:

| library state | total | `response.description-missing` | `response.no-error` |
|---|---:|---:|---:|
| `main` (pre-fix) | 109 | **30** | 79 |
| this branch | 79 | **0** | 79 |

The fix removes **exactly** the 30 false-positives and nothing else. The spec carries 32 `$ref` shared-error responses (18 ValidationFailed + 13 Unauthorized + 1 TooManyRequests); 30 false-fired, the other 2 sit outside the linter's default path scope. Corrected pristine baseline recorded in dogfood scratch (`apps/AdvisingApp/pre-tier1-rebaseline.md`); `$LIB` git-clean. The residual 79 are all `response.no-error` — the genuine Tier-1 signal, now legible without the `$ref` noise.

Left as **draft** pending review.

Issue #124, opened 2026-06-05
Performance: `openapi:generate` runs swagger-php `Analysis::validate()` unconditionally (~8% of generate self-time)


Radiergummi:
## Summary

`openapi:generate` unconditionally runs swagger-php's `Analysis::validate()` pass on the
generated document before serialising it — on **every** invocation, with no flag to skip it.
Profiling shows this is one of the larger single cost centers in generation.

## Evidence

Profiled `openapi:generate --format=json` against the **Vito** dogfood app (457 routes) on
2026-06-05 with Xdebug profile mode (cachegrind), parsed for self-time. Xdebug inflates absolute
time (~800×), so only **relative %** and **call counts** are meaningful; wall-clock baseline is
~0.47s (much of which is Laravel bootstrap).

Top non-bootstrap cost centers in generate are swagger-php's annotation model:

| self % | calls | function |
|---|---|---|
| 5.85% | 3834 | `OpenApi\Annotations\AbstractAnnotation::jsonSerialize` (serialisation — necessary) |
| 4.92% | 3834 | `OpenApi\Analysis::collectAnnotations` (**validate pass**) |
| 3.34% | 3834 | `OpenApi\Annotations\AbstractAnnotation::validate` (**validate pass**) |
| 1.75% | 11347 | `OpenApi\Annotations\OpenApi::versionMatch` |
| 1.65% | 9780 | `OpenApi\Context::isVersion` |

`collectAnnotations` + `validate` (~8.3% combined, plus a share of the version-match traffic)
come from the validation step, **not** from serialisation — `toJson()`/`toYaml()` are a plain
`json_encode($this)` and don't validate.

## Call site

`src/Console/GenerateCommand.php:93` → `validate()` at `:149-162`:

```php
if (!$this->validate($document)) {
    return self::FAILURE;
}
// ...
$valid = $analysis->validate();   // unconditional, every run
```

## Proposal (for discussion)

Make the validation pass opt-out (or opt-in):
- a `--no-validate` flag (validate-by-default preserved), **or**
- skip validation when emitting to a trusted target, **or**
- only validate at higher verbosity.

The doc is machine-constructed from our own descriptors, so a clean validate result is the
expected case; the cost is paid on every generate regardless. Lint already validates the spec
(see companion issue) so a user who wants the guarantee has a second path.

## Notes / non-findings

- This is **not urgent** — at 457 routes generate is ~0.47s wall and the robustness gate is green.
[...]

PR #160, opened 2026-06-06
feat(#125): `--no-validate` flag for `openapi:lint`

### LINKED ISSUE #125: Performance: `openapi:lint` runs full `opis/json-schema` meta-schema validation every run (~10%+ of lint self-time)
## Summary

`openapi:lint` validates the **entire generated document against the OpenAPI 3.1 meta-schema**
via `opis/json-schema` on every run (the `SpecInvalid` rule). Profiling shows the opis validation
is the dominant non-bootstrap cost in linting, ahead of our own tree walk.

## Evidence

Profiled `openapi:lint --format=json` against the **Vito** dogfood app (457 routes) on 2026-06-05
with Xdebug profile mode (cachegrind), self-time. Relative %/call-counts only (Xdebug inflates
absolute time ~800×); wall baseline ~0.58s.

| self % | calls | function |
|---|---|---|
| 4.79% | 27609 | `Opis\JsonSchema\Info\DataInfo::fromContext` |
| 1.78% | 14605 | `Opis\JsonSchema\Schemas\ObjectSchema::applyKeywords` |
| 1.53% | 11625 | `Opis\JsonSchema\Schemas\ObjectSchema::validate` |
| 1.20% | 11625 | `Opis\JsonSchema\ValidationContext::pushSharedObject` |
| 1.13% | 2644 | `Opis\JsonSchema\Keywords\PropertiesKeyword::validate` |

That's **~10%+ of lint self-time** in opis/json-schema alone. By comparison our own
`Lint\AnnotationWalker::doWalk` is 3.13% and `SpecTreeWalker::enrichAll` 0.93%.

Separately, lint **re-runs the full generation pipeline** before walking (`OperationBuilder::build`
×454, `RouteIntrospector`, `SecurityExtractor::forRoute` ×454 all appear in the lint profile) —
expected, since lint needs a spec to lint, but worth noting the meta-schema validation is *on top*
of a full generate.

## Call site

`src/Lint/Rules/SpecInvalid.php:96-99`:

```php
$validator = new Validator();
// ...
$result = $validator->validate($specJson, $schemaJson);
```

## Proposal (for discussion)

- Make the `SpecInvalid` meta-schema validation **opt-in / gated** (e.g. only at a higher
  `--level`, or behind a flag), since it is the most expensive single rule, **or**
- cache the compiled schema across the run (verify it isn't recompiled per-call), **or**
- scope validation to the parts of the doc the other rules don't already cover.

## Notes

- **Not urgent** — lint is ~0.58s wall at 457 routes; filed for a later targeted fix.
- This is the single biggest lever if lint runtime ever becomes a concern on larger specs
  (AdvisingApp 539, Invoice Ninja 530 routes).
- Companion to the generate-side validation pass (see #124).

Radiergummi:
Closes #125.

## What

`openapi:lint` runs the OAS 3.1 meta-schema validation (the `spec.invalid` rule) on every invocation — profiled as the single most expensive rule on large specs (~10%+ of lint self-time via `opis/json-schema`; Vito, 457 routes). This adds an opt-out `--no-validate` flag that drops `spec.invalid` from the active rule set for that run.

## Why a dedicated flag (not the existing --skip)

`spec.invalid` is deliberately **non-suppressible** — `LintRunner::resolveSkip()` strips it from `--skip` and `lint.disabled_rules`, and `#[IgnoreLint]` can't touch it. That invariant stays. The new flag is a separate, explicit, discoverable escape hatch (you can't silently lose spec-validity checking through config drift). Validation stays **on by default**; the flag mirrors `openapi:generate --no-validate` (#124).

The issue's "cache the compiled schema" option is a dead end: the schema is decoded + validator built once per run already (`spec.invalid` is an `ApiRule`, visiting the single `ApiNode` once). The cost is the one full-document traversal against the meta-schema, which caching can't reduce.

## How

- New signature option `--no-validate` on `openapi:lint`.
- `LintOptions` gains `bool $validateSpec = true`; `LintCommand::buildOptions()` sets it from the flag.
- `LintRunner` filters `spec.invalid` (`RuleRegistry::EXEMPT_RULE_ID`) out of the resolved rule set when `validateSpec` is false — right after `forLevel()`, clearly separate from the `--skip`/`disabled_rules` exemption.

## Tests (TDD)

`LintCommandTest` (corrupting the generated spec via a `transformDocument` hook that sets `openapi` to a value the meta-schema's version pattern rejects):
- **default**: the invalid document makes `spec.invalid` (level 0) fire → exit 1.
- **`--no-validate`**: the same invalid document → exit 0 (the rule is skipped; the clean path has no other level-0 findings).

## Verification

`composer test` (1740 passed), `composer lint` (PHPStan L8, no errors), `vendor/bin/pint --test` (passed). CHANGELOG `[Unreleased]` + `docs/linting.md` updated.


src/Console/LintCommand.php
```diff
@@ -60,6 +60,7 @@ class LintCommand extends Command
         {--path= : Restrict to routes matching this URI glob}
         {--diff= : Restrict to routes touched since git-ref (default: merge-base with the repository default branch)}
         {--no-suppress : Ignore #[IgnoreLint] attributes}
+        {--no-spec-validation : Skip the OAS 3.1 meta-schema validation (the spec.invalid rule); faster on large specs}
```
Radiergummi: Given that #156 adds a `--no-validate` flag to the `openapi:generate` command, should we perhaps rather use that flag name here too for symmetry?

src/Lint/LintRunner.php
```diff
@@ -205,6 +205,16 @@ private function runWithCollector(
 
         $rules = $this->registry->forLevel($level, only: $only, skip: $skip);
 
+        // --no-spec-validation drops the (otherwise non-suppressible) OAS meta-schema validation.
+        // This is a deliberate, explicit opt-out — distinct from --skip / disabled_rules, which
+        // cannot disable spec.invalid (see resolveSkip()).
```
Radiergummi: No need to mention this all over the place, the code is clearly readable without this comment already. Drop it.

src/Console/LintCommand.php
Radiergummi: Renamed to `--no-validate` for symmetry with `openapi:generate --no-validate`. Propagated through `LintCommand`, `LintOptions` PHPDoc, the test, `docs/linting.md`, and the changelog. (`validateSpec` stays as the internal `LintOptions` field name — clearer than `validate` in the lint context.)

src/Lint/LintRunner.php
Radiergummi: Dropped.

Issue #163, opened 2026-06-06
Custom-key route binding ({param:field}) does not surface the bound field — description reports the model's default route key


Radiergummi:
**Tier 0** · bug

## Summary

A route binding that uses Laravel's custom-key syntax — `/posts/{post:slug}` (and the scoped-nested form `/users/{user}/posts/{post:slug}`) — is mishandled in two compounding ways:

1. **The path parameter is dropped entirely.** `ActionDescriptor::resolveUriParameters()` matches signature parameters to URI segments by literal text (`src/Routing/ActionDescriptor.php:184`):
   ```php
   if (str_contains($uri, "{{$name}}") || str_contains($uri, "{{$name}?}")) {
   ```
   For `posts/{post:slug}` the signature parameter is named `post`, so `str_contains("posts/{post:slug}", "{post}")` is `false`. The parameter never enters the list, so no `OA\Parameter` is emitted for it.

2. **The emitted OpenAPI path template is invalid.** `PathsStage::normalisePath()` (`src/Support/Generator/Stages/PathsStage.php:264`) only prepends a leading slash, so the path key is rendered verbatim as `/posts/{post:slug}`. The `:slug` suffix is not a valid OpenAPI path-template expression — the segment should be `{post}`, with the custom field surfaced as parameter metadata, not baked into the template.

This bites hardest exactly where custom keys are most common: **scoped/nested bindings** (`Route::scopeBindings()` with `{parent}/{child:field}`), where the custom-keyed child silently disappears from the documented parameters.

## Motivation

Custom-key bindings are a first-class, widely-used Laravel routing feature (slug/uuid lookups, nested scoped resources). Today they produce a spec that is both incomplete (missing a required path parameter) and structurally invalid (an illegal path template) — strictly worse than the bare-string fallback every other binding gets. All the information needed to fix it is already on the route at Tier 0: `Route::bindingFields()` returns `['post' => 'slug']`, and the field can equally be stripped from the URI string with a small regex. No method-body analysis.

## Proposed approach (Tier 0)

Two coordinated fixes:

[...]

Issue #24, opened 2026-06-01
Enum-backed route binding → enum path parameter


Radiergummi:
## Summary

When a route path segment is bound to a PHP `BackedEnum` (implicit enum binding), the generated path parameter should carry an `enum` schema drawn from the enum's cases instead of being emitted as a bare string. The descriptor that backs the parameter already resolves the cases; today the extractor only emits them for `whereIn`-constrained segments, so a plain enum-typed binding silently loses them.

## Motivation

Enum-backed route bindings are a first-class Laravel routing feature, and the enum cases are the parameter's complete set of valid values. Emitting such a segment as a plain string throws that information away: consumers and codegen tools can't validate inputs or offer the closed set of choices. Recovering it is essentially free — the binding already names the enum class, the resolver already reflects its cases, and the extractor merely needs to surface them. This is a **Tier 0** item (reflection and signatures only, no method-body analysis).

## Proposed approach

The relevant infrastructure already exists and does most of the work:

- `UriParameterResolver` (`src/Support/Routing/UriParameterResolver.php`) calls `resolveEnumCases()` unconditionally for any `BackedEnum`-typed parameter and stores the case values on `UriParameterDescriptor::$enumCases` (`src/Support/Routing/UriParameterDescriptor.php`). So by the time extraction runs, the cases are already present on the descriptor regardless of whether a `#[Where*]` / `whereIn` constraint exists.

- `UriParametersExtractor` (`src/Support/Extraction/UriParametersExtractor.php`) builds each `OA\Parameter` from a `UriParameterDescriptor`. The bug lives in `applyWhereKindOverrides()`: the `enum` schema property is only assigned inside the `WhereKind::In` arm. A segment typed as a `BackedEnum` with no route-level `where` constraint has `whereKind === null`, so its `enumCases` are never written onto the schema.

The fix is to apply `descriptor->enumCases` to the schema whenever they are present (independent of `whereKind`), not only on the `WhereKind::In` branch. Backing type follows the enum: a string-backed enum keeps `type: string`, an int-backed enum should produce `type: integer` (the case values are stored as strings on the descriptor and may need coercion for the integer case). This sits alongside the sibling route-model-binding work (#23), which branches on `descriptor->modelClass` in the same descriptor/extractor path.

[...]

PR #170, opened 2026-06-06
feat(#23): typed path parameters from route-model binding

### LINKED ISSUE #23: Route-model binding → typed path parameters
## Summary

Path parameters like `/users/{user}` are emitted by name only, with no type. When a route uses route-model binding (`show(User $user)`) or an explicit `Route::bind`, the bound model already reveals the type — so the generated parameter should be typed (and described) from the model's route key instead of left bare.

## Motivation

An untyped path parameter is the single most common fidelity gap in generated specs: every model-bound route hits it, and a `{user}` with no schema is close to useless for client/codegen tooling. The type is sitting right there in the route binding — Laravel already resolves `{user}` to a `User` via its route key, so emitting `string`-or-nothing throws away information we can read for free.

## Proposed approach

The reflection groundwork lives in `src/Support/Routing/UriParameterResolver.php`: `resolveModelBinding()` already detects an `UrlRoutable` bound type and reads `getRouteKeyName()` (default `id`), recording it on the `UriParameterDescriptor` (`src/Support/Routing/UriParameterDescriptor.php`) as `routeKeyName`. The descriptor also already carries any `#[Where*]` / `$route->wheres[]` constraint as `whereConstraint` plus a semantic `whereKind` (the `WhereKind` enum in `src/Support/Routing/WhereKind.php`, with cases `Number`, `Uuid`, `In`, `Custom`). The work for this issue is to make `UriParametersExtractor::buildParameter()` (`src/Support/Extraction/UriParametersExtractor.php`) turn that binding into a typed `OA\Parameter` schema:

- Map the route key to a JSON-Schema type — `integer` when `whereKind` is `WhereKind::Number` (numeric key), `string` with `format: uuid` when `whereKind` is `WhereKind::Uuid` (UUID-keyed model), `string` otherwise — and add a description naming the bound model and key.
- When the segment is not model-bound, keep the current untyped/`string` fallback.

The explicit `Model::findOrFail($id)` / `firstOrFail()` body form is the Tier-1 complement for routes with no bound parameter, and feeds the implicit-404 inference in #29 — out of scope here.

## Acceptance criteria

- A route-model-bound `{user}` emits a typed parameter derived from the model's route key.
- Backed-enum route keys are #24's concern, not this issue's.
- Add tests covering int, uuid, and string route keys plus the unbound fallback. The natural home is a unit test alongside the resolver/extractor: `tests/Unit/Support/Routing/UriParameterResolverTest.php` for `UriParameterResolver` and/or `tests/Unit/Support/Extraction/UriParametersExtractorTest.php` for `UriParametersExtractor`, building descriptor/parameter pairs via the factories in `tests/Support/` (e.g. `ActionDescriptorFactory`).

## Attribute fallback

A `#[PathParam]` authoring attribute (`src/Attributes/PathParam.php`) is the escape hatch for non-bound `{}` segments; `UriParametersExtractor` already reads it for `description`/`example`.

## References

Parent: #64 (shared `RouteIntrospector` / `ActionDescriptor` context and the parameter-fidelity rationale live there). Enum-backed binding: #24. Body-derived params and implicit 404: #29, epic #5 (#11, #15). Shared enum component: #35.

Verify: `vendor/bin/pest tests/Unit/Support/Routing tests/Unit/Support/Extraction`

Radiergummi:
Closes #23.

## Context

The issue's original body-scope — mapping a path parameter's type from the route's `where*` constraint (`whereUuid()` → `format: uuid`, `whereNumber()` → `type: integer`) — already shipped in `UriParametersExtractor::applyWhereKindOverrides()` alongside the #163/#165 binding work. This PR implements the **refinement** Moritz added in the issue comments: derive the key type/format from the **bound model's metadata** when no explicit `where*` constraint is present.

## Plan

- [x] **`ModelPrimaryKey` VO** (`src/Support/Routing/`) — carries the model's primary-key `name`, JSON-schema `type` (`integer`/`string`), and optional `format` (`uuid`).
- [x] **`UriParameterResolver::resolveModelBinding()`** — when the bound `UrlRoutable` is an Eloquent `Model`, read `getKeyName()` / `getKeyType()` and detect `HasUuids` (→ `format: uuid`) off the constructor-less instance it already builds. Non-Eloquent routables and reflection failures leave the VO null (graceful degradation).
- [x] **`UriParameterDescriptor`** — new `?ModelPrimaryKey $modelPrimaryKey` field.
- [x] **`UriParametersExtractor`** — the `null` arm of `applyWhereKindOverrides()` applies the model's key type/format, **guarded** by `(bindingField ?? routeKeyName) === modelPrimaryKey->name` so a custom `{post:slug}` binding (or an overridden `getRouteKeyName()`) does not inherit the primary key's type. Explicit `where*` constraints keep precedence (they occupy the non-null match arms).

## Boundaries

- ULID keys → `string` with no `format` (no standard OAS format; `getKeyType()` already yields `string`).
- Reading the key column's `$casts` is out of scope — trait detection covers the idiom.
- Enum-backed route keys remain #24's concern.

## Verify

`vendor/bin/pest tests/Unit/Support/Routing tests/Unit/Support/Extraction`, `composer lint`, `vendor/bin/pint --test`.


Radiergummi:
Implementation note — deviation from the plan worth recording:

The plan said to read the key type from `getKeyType()`. That turned out to be **unreliable on the constructor-less instance** the resolver builds (`newInstanceWithoutConstructor()`, kept to tolerate bound models with required ctor args). For `HasUuids`/`HasUlids` models, `getKeyType()` returns `string` only when the key is in `uniqueIds()`, which itself depends on the `usesUniqueIds` flag set exclusively by `initializeHasUniqueStringIds()` — i.e. by the constructor. Bypassing the constructor leaves the flag `false`, so `getKeyType()` wrongly reports `int` for a UUID model.

Fix: detect `HasUuids`/`HasUlids` by **trait membership** (`class_uses_recursive`) for the string+uuid / string cases, and fall back to `getKeyType()` only for plain models, where it reads the `$keyType` *property default* (reliable without the constructor). The `TypedPathParameterTest` UUID case is the regression guard — it fails under the naive `getKeyType()` approach.

Also: the body-level `Model::findOrFail($id)` form (the example apps’ `string $flight` + `findOrFail` pattern) is Tier-1 and out of scope here, so those params correctly remain `string` — no example snapshot changed.

Verified locally: `pest` 1788 passed, `composer lint` clean, `vendor/bin/pint --test` passed.

src/Support/Routing/ModelPrimaryKey.php
Radiergummi, line 1: should this really be "primary" key, or rather `RoutingKey`?

src/Support/Routing/UriParameterDescriptor.php
```diff
@@ -31,5 +33,6 @@ public function __construct(
         public ?string $routeKeyName,
         public ?array $enumCases,
         public ?string $bindingField,
+        public ?ModelPrimaryKey $modelPrimaryKey,
```
Radiergummi: We have separate route key and primary key parameters now; this feels wrong. Can we have a single structure for the key information?

src/Support/Routing/UriParameterResolver.php
```diff
@@ -167,34 +173,70 @@ private function isLiteralAlternation(string $regex): bool
      * `BackedEnumType` extends `ObjectType` in symfony/type-info — check it first to avoid treating
      * an enum as a model binding.
      *
-     * @return array{class-string, string}|array{null, null}
+     * @return array{class-string, string, ?ModelPrimaryKey}|array{null, null, null}
```
Radiergummi: A three-item tuple is taking things too far. There must be a more elegant abstraction available here.

Radiergummi (commented):
the architecture is unsound. Needs more work.

PR #177, opened 2026-06-07
feat(#25): apiResource convention → summaries + status codes

### LINKED ISSUE #25: `apiResource` convention → summaries + status codes
## Summary

`Route::apiResource(...)` / `resource(...)` declares the seven CRUD actions with well-known semantics, but the generator emits a flat `200` and no derived summary for them. This derives the conventional success status code and a default human summary from the resourceful route plus the action method name.

## Motivation

Resourceful controllers are idiomatic Laravel, and their semantics are entirely predictable: `store` returns `201`, `destroy` returns `204`, `index` returns a collection, and so on. Emitting a flat `200` for those actions is simply wrong, and leaving summaries blank makes the generated docs harder to navigate — both fixable for free from information already on the route, with no body parsing.

## Proposed approach

Tier 0 — read the route definition and the action method name. When a route belongs to an `apiResource`/`resource` group, map the action to its conventional success code (`store` → `201`, `destroy` → `204`, `index`/`show`/`update` → `200`, etc.) and derive a default summary from the method name (`index` → "List …", `show` → "Show …"). Route introspection already yields the `ActionDescriptor` (`src/Support/Routing/RouteIntrospector.php`), and the operation is assembled in `PathsStage` / `OperationBuilder`. An explicit DocComment summary or a `#[Response]` attribute must always win over the convention.

## Acceptance criteria

- A `store` action emits `201`, `destroy` emits `204`, and the remaining CRUD actions their conventional codes; each gains a method-name-derived default summary.
- An explicit DocComment summary or `#[Response]` overrides the convention.
- A feature test under `tests/Feature` covers each CRUD action and the override path.

## References

Tier 0. Closely related to #19 (response status-code inference from other signals — coordinate so the two don't double-emit) and the response epic #6. Override hatch: `#[Response]` or a DocComment summary. Verify: `vendor/bin/pest tests/Feature`.

Radiergummi:
Closes #25

## What

Derive the conventional success status code and a default human summary for resourceful controller actions from Tier-0 signals (action method name + HTTP verb) — no body parsing. `store` → `201`, `destroy` → `204` (empty body), `index`/`show`/`update` → `200`; each gains a `List Posts` / `Show Post` / `Create Post` / `Update Post` / `Delete Post` summary. Explicit author intent (`#[Response]`, `#[Summary]`/`#[Operation]`, DocComment) always wins.

This also delivers the Tier-0 branch of #19; #19 is left with only its Tier-1 literal-status work (depends on #5).

## How

A new registry extension point, mirroring `PrimaryResponseResolver`, gated behind the Core Plugin (Core off → plain `200`/no summary, per the package's Core-optional contract):

- **`Contracts\Registry\OperationConventionResolver`** + value object **`OperationConvention`** (`?int successStatusCode`, `?string summary`). One `resolve(ActionDescriptor): ?OperationConvention` — single detection pass yields both.
- **`OpenApiRegistry`** gains `operationConventionResolvers` + `addOperationConventionResolver()`; wired into `OperationBuilder` by the service provider like the other resolver lists.
- **`Plugins\Core\Resolvers\ResourceConventionResolver`** — the convention table, verb-gated (route's methods must intersect the action's conventional verb). Resource noun derived from controller short name (consistent with `TagDeriver`), pluralized for `index`.
- **`OperationBuilder::build()`** consults it once (fault-boundary-wrapped):
  - **status** — when no explicit `#[Response(2xx)]` override, rewrites the resolved primary response's status + reason phrase; `204` drops `content`. Layers on top of whatever body a resolver produced (a `store` returning a Model stays `201` *with* its schema).
  - **summary** — lowest-precedence fallback after attr → DocComment → class-attr.

### Precedence
`#[Response]`/`#[Summary]`/`#[Operation]` → DocComment → class-level attr → **convention** → today's bare `200`/no-summary.

### Detection
Action method ∈ {index, show, store, update, destroy} **and** route verb matches the convention. Generalizes beyond literal `apiResource()` to hand-written resourceful routes (`Route::post('/flights', [FlightController::class, 'store'])`).

## Tests
- New feature test `tests/Feature/ResourceConventionTest.php`: status codes, summaries, verb gate, and the three override paths.
- Example-flavor snapshots regenerated (every flavor with Flight/Booking resource controllers shifts: `store`→`201`, `destroy`→`204` body-less, summaries added). `ExamplesTest` stays byte-match + OAS 3.1 valid + lint-clean.
- `CHANGELOG.md` `[Unreleased]` entry; docs touch on the responses/inference page.


Radiergummi:
**Implementation note — vanilla example demo relocation**

The convention surfaced an interaction with the vanilla example: `FlightController::destroy` carried `#[IgnoreLint('operation.summary-missing')]` to demo suppression-with-reason on a deliberately summary-less endpoint. The convention now fills that summary (`Delete Flight`), making the suppression a no-op.

Resolved by relocating the `#[IgnoreLint('operation.summary-missing')]` demo onto a new **non-resourceful** action, `GET /flights/{flight}/manifest` (action name `manifest` is outside the CRUD set, so the convention supplies no summary and `summary-missing` genuinely fires). `destroy` drops the IgnoreLint and now shows its `Delete Flight` convention summary. Vanilla README updated to match. This keeps the IgnoreLint showcase alive while demonstrating the new convention.

Issue #123, opened 2026-06-05
Docs: positioning vs L5-Swagger + end-to-end migration guide


Radiergummi:
## Summary

Add docs that position laravel-openapi against the annotation-driven incumbent (`darkaonline/l5-swagger` / raw swagger-php) and spell out the migration story, so users arriving with the "annotations are the source of truth" mental model understand the inversion and have a clear on-ramp.

## Motivation

The incumbent's model is explicit and ubiquitous: you hand-write `#[OA\*]` / `@OA` blocks and the package scans them. Ours is the inverse — infer from typed code, annotate only the gaps — which is strictly more capable but also less obvious to someone whose instinct is "where do I write the annotation?". Two things follow:

- We should state the contrast plainly ("type your returns or annotate them to get response schemas") so the mental-model switch is deliberate, not a surprise.
- We should document the migration path off L5-Swagger end to end, because adoption breadth, not technical lead, is where the incumbent wins — and a clear migration story is how we convert that install base.

This is the user-facing companion to the interop machinery (#17 harvester + the migration lint rules/fixers issue).

## Proposed approach

Docs only.

- A short "How this differs from L5-Swagger / swagger-php" section: source-of-truth inversion, inference tiers, attributes as escape hatch, plus what we add (linting, edit-time PHPStan validation, plugins).
- A migration guide tying the pieces together: enable the harvester (#17) so the existing spec stays correct → run `openapi:lint` to see which annotations inference/our attributes already cover → `openapi:lint --fix` to remove/rewrite the redundant ones → optionally drop the harvester once the codebase runs on inference.
- Be honest about the floor: we target modern, well-typed Laravel 12/13 + PHP 8.4 codebases by design.

## Acceptance criteria

- A positioning section and a migration guide exist under `docs/` and are linked from `docs/README.md`.
- The migration guide references the harvester (#17) and the migration lint rules/fixers, and reflects their actual behaviour once those land (or is staged behind them).

## References

Companion to #17 (annotation harvester) and the migration lint-rules/fixers issue. Prompted by the `darkaonline/l5-swagger` comparison. Affected: `docs/`.

Radiergummi:
Tracks the user-facing companion to #17 (annotation harvester) and #122 (migration lint rules + fixers).

Radiergummi:
[...]

PR #184, opened 2026-06-07
feat(#17): @OA / swagger-php annotation harvester

### LINKED ISSUE #17: `@OA` / swagger-php annotation harvester
## Summary

Pick up the OpenAPI annotations apps already write — `#[OA\*]` attributes and `@OA` docblock annotations on their models and controllers — and merge those authored schemas into the generated document.

## Motivation

A large class of Laravel apps is already annotated for L5-Swagger / swagger-php, yet today the generator ignores that surface entirely: an app with hand-authored `#[OA\Schema]` or `@OA\Schema` gets **zero** pickup. That is a glaring miss because swagger-php is already a dependency and the schemas in question are *authored by the developer* — there is no inference risk, no false positives. Two surveyed apps make the payoff concrete: Coolify carries `#[OA\*]` attributes on roughly 16 models, and Invoice Ninja has on the order of 5,042 `@OA` annotations. Harvesting them is the single highest-ROI response measure for the already-annotated population.

## Proposed approach

Tier 0 — pure reflection over annotations that are already parseable elsewhere in the codebase; no AST, no method-body parsing.

Add a new bundled plugin under `src/Plugins/` (modelled on the existing `src/Plugins/SpatieData/` plugin — see `SpatieDataPlugin.php`) that reuses the swagger-php annotation parsing the lint tree-builder already depends on (`src/Lint/Tree`, fed by the swagger-php object model that `src/Support/Generator/OpenApiGenerator.php` also produces). The plugin scans discovered models/controllers for `#[OA\Schema]` / `@OA\Schema` (and the operation-level `@OA` annotations), and contributes the resulting schemas into the components pool and per-operation responses during generation.

Because it depends on swagger-php conventions a host app must opt into, it ships as a plugin registration in `config/openapi.plugins` (the array at `config/openapi.php:388`) — and is off by default, like `QueryBuilder` and `Fractal`. Follow the SpatieData precedent and no-op via a `class_exists` guard in `register()` so listing the plugin imposes no hard dependency.

## Acceptance criteria

- A model or controller carrying `#[OA\Schema]` / `@OA\Schema` contributes that schema to the document's components and responses.
- A Coolify-shaped fixture (`#[OA\*]` on models) and an Invoice-Ninja-shaped fixture (`@OA` docblock annotations) gain response schemas in the generated spec.
- The harvester is inert unless its plugin is registered in `config/openapi.plugins`.
- Add a feature test under `tests/Feature/` exercising a fixture model/controller carrying both attribute and docblock annotations, asserting the harvested schema appears in the generated document; add a flavor to `tests/Feature/ExamplesTest.php` so the plugin's output is snapshot byte-matched, OpenAPI 3.1-validated, and lint-clean.

## References

Parent: #6 — the shared Tier-0 response-fidelity rationale and the "recover schemas that already exist, via reflection" framing live there; this issue is §3.1. Related siblings: #18 (model → schema), #19 (status-code inference). `#[ResponseField]` (`src/Attributes/ResponseField.php`) remains the escape hatch for computed/virtual fields.

Verify: `vendor/bin/pest tests/Feature/ExamplesTest.php`

Radiergummi:
## Implementation plan — `@OA` / swagger-php annotation harvester

Closes #17. Spec lives in the issue; this is the *how*.

### Scope (settled while brainstorming)
- Harvest **authored component schemas** (`#[OA\Schema]` / `@OA\Schema` on models/DTOs) **and** authored **operation responses** (`@OA\Get`/`@OA\Post`… with their `@OA\Response`s on controller methods).
- Our generated operation stays the **spine**: authored annotations only contribute response bodies + component schemas. We do **not** harvest authored paths/parameters/request bodies in this iteration.
- Scan `app_path()` **unconditionally** — no user-facing config key.
- Mechanism: a **single late `SpecStage`** (not the resolver-pair the issue sketched), chosen for fewer moving parts, fidelity to all authored status codes, and to avoid the component-flush timing problem (a plugin stage runs after `ComponentsStage`, so writing schemas straight into the document is the clean path).

### New unit layout
```
src/Plugins/SwaggerPhp/
  SwaggerPhpPlugin.php                       # register() → addStage(HarvestAuthoredAnnotationsStage)
  Support/AuthoredAnnotationScanner.php      # runs swagger-php once; exposes 3 indexes
  Stages/HarvestAuthoredAnnotationsStage.php # merges authored responses + schemas into the document
```

### `AuthoredAnnotationScanner` (scoped singleton, memoized)
- Runs `(new \OpenApi\Generator($captureLogger))->generate($scanPaths, validate: false)` **once** per generation; `Generator::$context` is already pinned to 3.1.0 by `SpecPipeline`.
- `$scanPaths` is an **injected internal value**, default `[app_path()]` — *not* a config key. The example service provider rebinds it to the fixture directory for tests (verified seam, see Testing).
- Builds three lookups from the resulting `OA\OpenApi` (each annotation keeps its `_context` exposing `class` / `method` / `fullyQualifiedName()`; `OA\Schema->schema` holds the authored name — both confirmed against installed swagger-php 6.1.2):
  - `classToSchema`: declaring FQCN → `OA\Schema`
  - `nameToSchema`: authored schema name → `OA\Schema`
  - `methodToOperation`: (controller FQCN, method) → `OA\Operation`
- Returns empty indexes when nothing is annotated. Scanner warnings/parse errors are captured by the injected logger and logged — never fatal (Tier-0 graceful degradation).

### `HarvestAuthoredAnnotationsStage` (`SpecStage`)
Runs after `ErrorResponseInferenceStage` + `ComponentsStage`, before the fixed `OverridesStage`/`TransformersStage`. For each operation in the built document:
1. Resolve the `ActionDescriptor` via `context->actionFor($operation)`.
2. **If** `methodToOperation` has an authored operation for that controller method → copy its `@OA\Response`s onto `operation->responses`, **authored winning per status code**; our inferred responses for *other* statuses remain.
3. **Else if** the action's return type resolves to a class in `classToSchema` → set the `200` response body to a `$ref` — only when we'd otherwise emit a bare `200 OK`.
4. For every schema **name** referenced by the responses just added, register that authored `OA\Schema` (and its transitive authored refs) directly into `$document->components->schemas`, **preserving the exact authored name** so authored cross-refs resolve. Transitive walk is cycle-guarded (visited set).

### `SwaggerPhpPlugin`
- `register()` → `$registry->addStage(HarvestAuthoredAnnotationsStage::class)`.
- **No `class_exists` guard** — swagger-php is a hard dependency, so the guard the issue mentions would be dead code. Opt-in is the config registration itself.
- Shipped **commented-out** in `config/openapi.plugins` (like QueryBuilder/Fractal), with a short note in `config/openapi.php`.

### Error handling (Tier-0)
- Malformed annotation → dropped + logged, generation continues.
- Authored `@OA\Response` referencing a schema name absent from `nameToSchema` → **drop that ref + log** rather than emit a dangling `$ref` (keeps the spec valid + lint-clean).
- Cycle-guarded transitive schema registration.

### Testing
- **Feature test** (`tests/Feature/`): fixture with one model carrying `#[OA\Schema]` (attribute / Coolify shape) + one controller method carrying a `@OA\Get` + `@OA\Response` docblock (Invoice-Ninja shape). Assert the harvested component schema and the operation response `$ref` both appear in the generated document.
- **Plugin-inert test**: with the plugin unregistered, the same fixture yields no harvested schemas.
- **New `swagger-php` flavor**: `examples/swagger-php/` + entry in `examples/_shared/Flavors.php`; byte-exact snapshot, OpenAPI-3.1 validated, lint-clean. Auto-excluded from the swagger-php 5.x CI job per the snapshot-group rule.

### Docs (per CLAUDE.md)
- New `docs/` page for the plugin + page-index entry in `docs/README.md`.
- `CHANGELOG.md` `[Unreleased]` entry.
- Commented plugin entry + note in `config/openapi.php`.
- Per the issue comment: once the migration docs page (#123) lands, update the README "How this compares" parenthetical to link it — leave a `TODO(#123)` marker rather than inventing the link now.

### Build sequence
1. `AuthoredAnnotationScanner` + its three indexes, with a focused unit test over a fixture dir.
2. `HarvestAuthoredAnnotationsStage` (responses merge + component registration) + the plugin class.
3. Feature test (attribute + docblock fixture) and plugin-inert test → green.
4. New `swagger-php` example flavor + snapshot; `ExamplesTest` green.
5. Docs + changelog + config wiring.
6. `composer test`, `vendor/bin/pint --test`, `composer lint` all green.

### Deviations from the issue text
- Single `SpecStage` instead of `RefSchemaResolver` + `PrimaryResponseResolver`.
- No `class_exists` guard (swagger-php is mandatory).


Radiergummi:
## Step 1 done — `AuthoredAnnotationScanner` (4eeb63c)

Scanner runs swagger-php's analyser over the configured paths once and exposes three indexes: schema-by-name, schema-by-declaring-class, operation-by-(controller, method). Lazy + memoised; degrades to empty indexes on scan failure. 7 unit tests green; full suite (1834) green; Pint + PHPStan clean.

### Deviation from the plan: `doctrine/annotations` dependency
While implementing I confirmed swagger-php only parses **`@OA` PHPDoc annotations** when **`doctrine/annotations`** is installed (`DocBlockParser::isEnabled()` → `class_exists(Doctrine\\Common\\Annotations\\DocParser)`). It is an *optional* swagger-php dependency and was **not** installed here, so the entire docblock path — the Invoice-Ninja case — was silently dead. `#[OA\*]` **attributes** (Coolify case) work without it.

Decision (agreed): add `doctrine/annotations ^2.0` to **`require-dev`** (so our CI exercises the docblock path) and **`suggest`** (so consumers opt in). The plugin works attribute-only without it, attributes+docblocks with it. This is the genuine optional-dependency the issue's `class_exists`-guard note was reaching for (the guard on *swagger-php itself* remains dead code, since swagger-php is a hard dep). `doctrine/annotations` is flagged abandoned but has no security advisory; acceptable for an optional/dev dep that swagger-php itself recommends.

The plugin docs page (step 5) will document: *`@OA` PHPDoc harvesting requires `composer require doctrine/annotations`.*

Radiergummi:
## Steps 2–3 done — harvester, plugin, feature test, example flavor

- **Stage + plugin** (c9e87d4): `HarvestAuthoredAnnotationsStage` merges authored responses onto operations + registers authored schemas (transitively, authored names) into the document; `SwaggerPhpPlugin` registers it; scanner bound over `app_path()`; commented-out config entry.
- **Feature test + example flavor** (185f33b): end-to-end `tests/Feature/SwaggerPhpHarvesterTest.php` (harvest + `toYaml()` serialize + OpenAPI 3.1 `Analysis::validate()` + plugin-inert) and a new `examples/swagger-php/` flavor covering both shapes — a `#[OA\Schema]` attribute model and an `@OA\Get` / `@OA\Schema` PHPDoc-annotated controller. ExamplesTest: snapshot byte-matches, validates 3.1, lints clean.

Full suite **1846 passed**; Pint + PHPStan clean.

### Refinement beyond 'schemas + response refs': authored operation metadata
Building the example surfaced that when a controller's docblock **is** the `@OA\Get` annotation (the Invoice-Ninja norm), the library's own PHPDoc parser leaks the raw annotation text into the operation's `summary`/`description` — a pre-existing issue for any `@OA`-docblock app, plugin or not. Since we're harvesting that operation, I extended the merge to **adopt the authored operation's `summary` / `description` / `operationId` / `tags`** (the annotation is the source of truth for the operation it describes). `summary`/`description` replace ours outright; `operationId`/`tags` only when the author set them. This is a small step beyond the literal 'schemas + response refs' scope but necessary for a coherent result; easily revertible if you'd rather keep operation prose strictly library-owned.

### CI watch
- The byte-exact snapshot case is in the `snapshot` group → auto-excluded from the swagger-php 5.8 job per the matrix rule. The harvested schemas are real `OA\Schema` objects (not raw nested arrays), so they should clear 5.8 validation, but worth a glance when CI runs.
- `doctrine/annotations` is in `require-dev`, so the docblock path is exercised in every CI cell.

Ready for review from my side — implementation, tests, docs, changelog all in.

PR #185, opened 2026-06-08
feat(#122): migration removal lint rule + fixers — redundant #[OA] / @OA → inference

### LINKED ISSUE #122: Migration lint rules + fixers: redundant `#[OA]` annotations → inference / our attributes
## Summary

Ship lint rules — each with a `--fix` fixer — that detect hand-written `#[OA\*]` / `@OA` annotations which the generator can now produce on its own (via inference) or express more concisely (via one of our authoring attributes), and rewrite or remove them. This is the **migration cleanup** half of the swagger-php interop story: #17 teaches the generator to *read* existing annotations; this issue gives users a guided, mechanical path to *delete* them as inference and our attributes take over.

## Motivation

The single biggest barrier for an app coming from `darkaonline/l5-swagger` is the annotation-as-source-of-truth model: their codebase is full of `#[OA\*]` / `@OA` blocks, and "rip them all out and trust inference" is an unacceptable leap. The harvester (#17) softens the landing — keep your annotations, we honour them — but it leaves the codebase permanently annotation-heavy. The natural next step is a lint pass that, annotation by annotation, tells the user exactly which blocks are now redundant *because the generator already derives the same schema* (or because a shorter `#[ResponseField]` / `#[QueryParam]` / type-hint says the same thing), and offers a one-command fix. That turns migration from a cliff into a ratchet:

1. enable the harvester (#17) — spec is correct immediately,
2. run `openapi:lint` — it flags every annotation inference/our attributes already cover,
3. `openapi:lint --fix` — removes the redundant ones and rewrites the replaceable ones,
4. eventually the harvester plugin can be turned off and the app runs on inference alone.

## Proposed approach

Tier 1 — bounded AST whitelist, building on the fixer infrastructure from the `--fix` epic (#7) and Phase 1 (#20: `FixableRule` / `Fixer` / `Fix` / `FixApplicator`, format-preserving `nikic/php-parser` v5 printer), and on the harvester (#17) for parsing `#[OA\*]` / `@OA` into the same object model the generator uses.

The decision is a **structural-equivalence** check, never a heuristic guess — only fire when the annotated schema is provably equal to (or strictly subsumed by) what the generator would emit without it. Two finding families:

- **`migration.oa-redundant-with-inference`** — the generator already infers an equivalent schema/parameter/response from types, FormRequest rules, PHPDoc, route binding, etc. Fixer: **remove** the `#[OA\*]` attribute / `@OA` docblock block.
- **`migration.oa-replaceable-by-attribute`** — inference alone wouldn't recover it, but one of our authoring attributes expresses the same thing more concisely (e.g. an `#[OA\Property]` description → `#[ResponseField(description:)]`, an `#[OA\Parameter]` → `#[QueryParam]`). Fixer: **rewrite** the OA block into the equivalent laravel-openapi attribute.

Both fire only when the comparison is exact, so `--fix` never deletes information the generator can't reproduce. Where the annotation carries something inference *cannot* recover (a runtime shape, a human description with no Tier-0/1 source), no finding is emitted — that annotation is load-bearing and stays.

Because these rules depend on swagger-php annotation parsing a host app must opt into, they ship as **lint rules registered by the harvester plugin from #17** (off by default, `class_exists` guarded), keeping `src/Core/` and `src/Support/` free of plugin/third-party coupling per the architecture rules. Rule IDs and fixers follow the conventions in #7's fixable inventory.

## Acceptance criteria

- `migration.oa-redundant-with-inference` and `migration.oa-replaceable-by-attribute` exist as registered rules, emitted only when the annotated and generated schemas are structurally equivalent (golden fixtures prove no false positives on load-bearing annotations).
- Each rule returns a `Fixer` that produces exactly the documented edit (clean removal, or a rewrite to the equivalent `#[ResponseField]` / `#[QueryParam]` / type-hint), preserving surrounding comments and whitespace.
- The rules are inert unless the #17 harvester plugin is registered.
- Before/after golden fixtures under `tests/Unit/Lint/Rules/` (modelled on the existing rule tests) plus a feature flavor proving an annotated fixture lints clean after `--fix` and still produces the same generated document.
- Fixed fixtures pass `vendor/bin/pint --test`.

## References

Complements #17 (the swagger-php annotation harvester — parses the annotations these rules act on). Builds on #7 (fixer architecture / `FixableRule` epic) and #20 (Phase 1 fixer infra); related to #21 (Tier B/C fixers). Escape-hatch attributes these rules rewrite *into* live in `src/Attributes/`. Affected area: `src/Lint`, the #17 plugin. Verify: `vendor/bin/pest tests/Unit/Lint/Rules/`

Radiergummi:
Closes #122.

## What this delivers

`migration.oa-redundant-with-inference` — a lint rule + `--fix` (registered by the off-by-default **SwaggerPhp** plugin) that flags hand-authored `#[OA\Schema]` / `@OA\Schema` annotations the generator now reproduces on its own, and removes them. The first step in turning a swagger-php / l5-swagger codebase off annotations and onto inference.

```bash
php artisan openapi:lint --only 'migration.*'          # report redundant annotations
php artisan openapi:lint --only 'migration.*' --fix     # …and remove them
php artisan openapi:lint --skip 'migration.*'           # opt the family out of a high-level audit
```

## The redundancy decision (provenance-based, sound)

For a class with an authored schema, the rule compares **what the author wrote** (`AuthoredAnnotationScanner`) against **what inference produces for the same class** in an inference-only control document — matched **by source class, ignoring serialized names**.

- **Subsumption, not equality.** Inference enriches schemas (a synthesised `example`, a discovered property, a derived `format`), so strict equality almost never holds. The rule fires only when inference *subsumes* the annotation — reproduces everything it said and possibly more. A `description` or `additionalProperties: false` inference can't derive is a key it lacks, so subsumption fails and the annotation is correctly **kept** as load-bearing.
- **Whole-class granularity.** The unit is the entire `#[OA\Schema]`+`#[OA\Property]` set / `@OA\Schema` docblock block. Per-property removal is unsound when the authored schema survives independently.
- **Dangling-`$ref` guard.** A schema another surviving authored annotation still `$ref`s by name is never removed.
- **"Same generated document" = same API surface.** Removal may consolidate a duplicate component to inference's name; the documented surface is preserved.

## How it's built

- **`SpecPipeline::withoutStage(...)`** — a subtractive pipeline seam (clean on top of #190). The control document is a normal generation with the harvester stage excluded.
- **`InferenceControlDocument`** (scoped) — runs that control generation with fresh scoped state (`forgetScopedInstances`, a throwaway findings collector), memoized per spec, and exposes `schemaForClass()`.
- **`SchemaEquivalence::subsumes()`** — normalized, order-insensitive recursive containment over the swagger-php object trees (drops `UNDEFINED`, internal `_*`, and the serialized name).
- **`RedundantOaAnnotationFixer` / `DocblockAnnotationRemover`** — remove the `#[OA\*]` attributes (AST) or the `@OA\Schema` docblock block (line-based, paren-depth-aware, quote-safe).

## No `--migrate` flag — a `migration.*` rule family

Earlier revisions gated this behind a dedicated `--migrate` flag; that was bolted-on. Instead, migration rules are ordinary rules identified by the **`migration.` ID prefix**, parked at **level 4** (cleanup tier) so their second, inference-only generation stays unpaid on default-level runs. This added general **`*`-glob support** to `--only` / `--skip` / `enabled_rules` / `disabled_rules` (`LintRunner::expandRulePatterns`), so any family is selectable/disable-able at once — not a migration-specific mechanism. Removed the `MigrationRule` marker, `LintOptions::$migrate`, the `--migrate` flag, and the level-bypass special-case.

## Scope: removal only

This PR ships **removal**. The rest of the OA-annotation migration surface is tracked as sub-issues of #122:

- #195 — rewrite OA annotations → our authoring attributes (the replacement half), with the full mapping whitelist.
- #196 — operation-level annotation removal (`@OA\Get`/`@OA\Response`/…).
- #197 — per-property removal within still-needed schemas.
- #198 — `$ref`-following in subsumption.
- #199 — document-level `@OA\Info`/`@OA\Server`/`@OA\SecurityScheme` → config.
- #200 — residual constructs with no equivalent (cookie params, `additionalProperties`, encoding, XML, …) — decision issue.

## Tests & gates

- Unit: `SpecPipeline::withoutStage`, `SchemaEquivalence::subsumes`, the two fixers (golden before/after fixtures for both attribute and docblock shapes).
- Feature: `InferenceControlDocument` (incl. scope-isolation), detection under `--only 'migration.*'`, inert on default-level runs, family-disable via `--skip 'migration.*'`, no false positive on a non-inferable class, and the same-API-surface invariant.
- Green: PHPStan (level 8), `vendor/bin/pint --test`, full suite (1871 passed).


Radiergummi:
## Implemented — decisions & deviations from the plan

Shipped on top of #190 (clean `SpecPipeline::withoutStage()` exclusion) and the merged #17 harvester. **Gates green:** PHPStan clean, `vendor/bin/pint --test` passes, full suite 1870 passed.

### Scope: removal-only (rewrite deferred)

This PR ships only `migration.oa-redundant-with-inference` (removal). The rewrite rule `oa-replaceable-by-attribute` (the open question in the plan) is **deferred to #195** — removal is the core, lower-risk migration value; rewrite is a follow-up.

### Soundness: provenance-based subsumption, not name-based equality

The plan framed the oracle as per-annotation structural *equivalence* against the control node. Implementing it surfaced two corrections, agreed with @Radiergummi:

1. **Subsumption, not equality.** Inference *enriches* schemas (synthesised `example`, discovered properties, derived `format`), so strict equality almost never holds. The rule fires when inference **subsumes** the authored annotation — `authored ⊆ inferred` (`SchemaEquivalence::subsumes()`). A `description` or `additionalProperties: false` inference can't derive is a key it lacks, so subsumption fails and the annotation stays load-bearing.
2. **Provenance, not name.** Comparison is by **source class**, ignoring serialized component names (the `schema` name key is dropped from the canonical form). "Produces the same generated document" therefore means **same API surface** — removal may consolidate a duplicate component to inference's name, which is the desired migration outcome. The real safety check is a **dangling-`$ref` guard** (`AuthoredAnnotationScanner::isSchemaReferencedByOtherAuthored()`), not name-matching.
3. **Granularity: whole-class.** Per-property removal is unsound when the authored schema survives independently. The fixer removes the whole `#[OA\Schema]`+`#[OA\Property]` set or the entire `@OA\Schema` docblock block.

### Infrastructure added beyond the file plan

- `Contracts\Lint\MigrationRule` marker + `LintOptions::$migrate` + `--migrate` flag. Migration rules run **only** under `--migrate` (they trigger a 2nd, inference-only generation) and their findings **bypass the level preset** (like `spec.invalid`), since the user explicitly opted in.
- `LintContext::$specName` (general, nullable) so a rule can key per-spec generation state off the spec being walked.
- `DocblockAnnotationRemover` tracks parenthesis depth across lines and ignores parens inside quoted strings, so a `description="see (note)"` doesn't unbalance the block.

Radiergummi:
## Update: dropped `--migrate` in favour of a `migration.*` rule family

Per review, the dedicated `--migrate` flag was bolted-on and overfit. Replaced with a general convention (commit `4700887`):

- Migration rules are ordinary rules identified by the **`migration.` ID prefix**, parked at **level 4** (cleanup tier) so their second, inference-only generation stays unpaid on default-level lint runs — the level system does the cost-gating the flag used to.
- Removed the `Contracts\Lint\MigrationRule` marker, `LintOptions::$migrate`, the `--migrate` flag, and the level-bypass special-case.
- Added **`*`-glob support** to `--only` / `--skip` / `enabled_rules` / `disabled_rules` (`LintRunner::expandRulePatterns`), so the family is selectable/disable-able in one go and the mechanism generalises to any family (`field.*`, future importers, …):

```bash
openapi:lint --only 'migration.*'   # report (and --fix) just the migration findings
openapi:lint --skip 'migration.*'   # disable the family
```

Consequence: a high-level audit (`--level max`) with the SwaggerPhp plugin on now runs the rule (and pays the 2nd generation); `--skip 'migration.*'` opts out. Default-level runs and plugin-off configs stay cost-free.

This commit also folds in the earlier cleanup pass (reuse `AnnotationWalker` for `$ref` walking; drop dead sort canonicalisation in `SchemaEquivalence`). Gates green: PHPStan, `pint --test`, full suite (1871 passed).

Issue #190, opened 2026-06-08
Consolidate stage registration into one ordered sequence in the provider (+ back unordered pools with container tags)


Radiergummi:
**Umbrella issue.** Tracks consolidating stage registration. The work splits into two independently-trackable proposals, now broken out as sub-issues:

- [x] **#194 — Proposal A:** one ordered stage sequence in the provider (+ registry sealing + the #17 harvester direct-write cleanup). The load-bearing consolidation.  
      *Implemented by PR #192.*
- [x] **#193 — Proposal B:** back the registry's unordered pools (resolvers, rules, …) with container tags. Optional, splittable; may close as won't-do if it doesn't clearly simplify.

---

## Context

Stage registration was spread across more places than it should be, and the ordering of the stage pipeline was expressed three different ways at once:

1. **`Support\Generator\BaselineRegistration`** — added the load-bearing baseline stages, runs *first*.
2. **The `OpenApiRegistry` factory closure** — runs `BaselineRegistration`, then the plugin loop, then config lint rules.
3. **`SpecPipeline`** — held `OverridesStage` and `TransformersStage` as injected fields and ran them *after* the registry loop.

So a contributor asking "when does my stage run?" had to reconcile three mechanisms. That was the smell. It also produced a concrete blocker in #17 (the SwaggerPhp harvester ran *after* the `ComponentsStage` flush and had to write `$document->components` directly).

The proposals and their full rationale, trade-offs, and validation now live in #194 (A) and #193 (B). See those sub-issues for the detail.

## References

- #17 — SwaggerPhp annotation harvester (the direct-write Proposal A removes).
- #189 — related core ADR on the in-memory model.

PR #192, opened 2026-06-08
feat(#190): consolidate stage registration into one ordered sequence

### LINKED ISSUE #194: Consolidate stage registration into one ordered sequence in the provider
## Goal

Fewer registration sites, with stage order controlled in one readable place — the service provider — and the `ComponentsStage` flush positioned so late contributors (the #17 harvester, and any future one) participate normally.

## Proposal A — one ordered stage sequence in the provider

Collapse the three ordering mechanisms (`BaselineRegistration`'s baseline stages, the plugin loop, and the two terminal steps `SpecPipeline` held as injected fields) into a single explicit sequence inside the registry factory closure. The order *is* the list of `addStage` calls, top to bottom:

```php
// pre-plugin: build the operation skeleton + emit contributions
$registry->addStage(RootStage::class);
$registry->addStage(PathsStage::class);
$registry->addStage(ErrorResponseInferenceStage::class);

// plugins: operation mutators + schema contributors
foreach ($plugins as $pluginClass) { $app->make($pluginClass)->register($registry); }

// post-plugin: flush the registry, then finalize
$registry->addStage(ComponentsStage::class);    // flush — now sees plugin contributions
$registry->addStage(SecurityStage::class);
$registry->addStage(OverridesStage::class);     // was a SpecPipeline field
$registry->addStage(TransformersStage::class);  // was a SpecPipeline field
```

Consequences:
- `SpecPipeline` becomes a **pure executor** — loop the registered stages, nothing else.
- `BaselineRegistration` keeps only its lint-rule registration.
- **#17's direct-write dissolves**: with the flush after the plugin loop, the harvester registers authored schemas via `ComponentSchemaRegistry::registerNamed()` and gets O(1) dedup + schema-transformer dispatch for free.

### Trade-off (decided)

Retires the *structural* "runs after all stages regardless of registration order" guarantee for the terminal stages, downgrading it to *positional*. Accepted: all plugins (incl. third-party) register via `Plugin::register()` inside the loop, before the terminal append, so they cannot register past the terminals. To enforce the funnel the positional guarantee assumes, `OpenApiRegistry::seal()` is added (called last in the closure); out-of-band `addX()` on the built registry throws an unchecked `RegistrySealedException`.

## Risks & validation

- `ComponentsStage` repositioning is snapshot-affecting — only SwaggerPhp-enabled runs actually move (Core/Spatie/ApiResources register no stages), but it requires the byte-exact `snapshot` group + SwaggerPhp suite.
- Tier-0: pure wiring/reflection, no method-body parsing.

## Status

Implemented by PR #192 (sealing folded in per maintainer decision).

Refs #190. References #17 (the direct-write this removes).

Radiergummi:
## Plan

Closes #194 — Proposal A of the #190 epic: one ordered stage sequence in the provider + the #17 harvester direct-write cleanup. Proposal B (container tags) is split to #193.

### Decisions (confirmed with maintainer)
- **Scope:** Proposal A (one ordered stage sequence in the provider) + the #17 harvester direct-write cleanup. Proposal B (back unordered pools with container tags) becomes a follow-up issue.
- **Terminal-stage guarantee:** accept the structural→positional downgrade. All plugins (incl. third-party) register via `Plugin::register()` inside the loop, *before* the terminal append, so they cannot register past the terminals. The only thing lost is protection against out-of-band `addStage()` on an already-built registry — an unsupported path.
- **Sealing:** fold in registry sealing to make the funnel an *enforced* invariant rather than a convention. `seal()` + the adder guards are `@internal`. Turns the one accidental-misuse path (resolve registry in a later `boot()`, append a stage) from a silent ordering bug into a loud exception.

### Changes
1. **`OpenApiRegistry`** — add `private bool $sealed` + `seal()` (`@internal`); every `addX` guards on it and throws once sealed.
2. **`OpenApiServiceProvider::registerRegistries`** — inline the single ordered sequence in the factory closure: pre-plugin stages (`Root → Paths → ErrorResponseInference`) → plugin loop → post-plugin stages (`Components → Security → Overrides → Transformers`) → rules + envelope resolver → `seal()`.
3. **`BaselineRegistration`** — drops the stage portion; keeps the lint-rule registration (`register` → `registerRules`).
4. **`SpecPipeline`** — becomes a pure executor: loop `registry->stages`, nothing else. The two injected `SpecStage` fields and the terminal-precedence comment go away (Context pinning stays).
5. **`HarvestAuthoredAnnotationsStage` (#17)** — now an ordinary pre-flush contributor: registers authored schemas via `ComponentSchemaRegistry::registerNamed()` (O(1) dedup + schema-transformer dispatch for free); the bespoke `registerSchema`/`components()`/linear-dedup code is removed.

### Why this is lower-risk than the issue feared
Only **one** plugin registers a stage (SwaggerPhp's harvester) — Core/Spatie/ApiResources register resolvers/rules but no stages. So for the default config the stage list stays byte-identical (`Components` still precedes `Security`; terminals still last). Only SwaggerPhp-enabled runs see `Components` move after the harvester — exactly the intent.

### Validation
- Unit: sealing throws; `SpecPipeline` runs registry stages in order (updated, terminal fields removed).
- `snapshot` group (byte-exact) + SwaggerPhp suite for the harvester move.
- `composer test`, `vendor/bin/pint --test`, `composer lint` (PHPStan L8).

### Docs
`CLAUDE.md` (Registry/plugins + terminal-precedence sections → single-sequence model), `docs/architecture.md`, `docs/plugin-authoring.md`, `CHANGELOG.md` `[Unreleased]`.


Radiergummi:
**Decision record — sealing folded in (not in the original issue text).**

Per discussion, the structural→positional downgrade for the terminal stages is accepted. To keep the positional guarantee from being merely conventional, `OpenApiRegistry::seal()` is folded into this PR: the factory closure calls it last, and out-of-band `addX()` on the built registry throws `RegistrySealedException`.

Rationale: sealing isn't defence against deliberate misuse (reflection defeats any seal) — it's defence against *accident*. Without it, a contributor/user who resolves the built registry in a later `boot()` and appends a stage gets a silent bug (their stage runs after Overrides/Transformers, clobbering user intent); with it, a loud, actionable exception. It enforces the exact funnel Proposal A's positional guarantee assumes. Cost is one flag + one guard.

`RegistrySealedException extends LogicException` and is marked **unchecked** in `phpstan.neon` (`exceptions.uncheckedExceptionClasses`) so it does not leak into the public `Plugin::register()` `@throws` contract — `LogicException` can't be blanket-unchecked because `InvalidArgumentException` (used with `@throws` in the provider) extends it.

**Proposal B split to #193** as the issue allowed.

Issue #197, opened 2026-06-08
Migration: per-property redundant `#[OA\Property]` / `@OA\Property` removal within still-needed schemas


Radiergummi:
Parent: #122. Untracked gap surfaced while building the schema-level removal rule (PR #185).

## Summary

`migration.oa-redundant-with-inference` is **whole-class, all-or-nothing**: it fires only when the *entire* authored schema is subsumed by inference. A class where some properties merely restate inferred types but others carry load-bearing descriptions gets **nothing** flagged — the redundant subset can't be cleaned. Add per-property granularity: flag individual `#[OA\Property]` / `@OA\Property` entries inference subsumes, even when the enclosing `#[OA\Schema]` / `@OA\Schema` must stay.

## Soundness caveat (the hard part)

Whole-class was chosen deliberately because per-property removal is unsound when the authored schema survives independently under its own name: deleting one of its properties silently mutates *that* emitted component. A per-property rule must account for the harvester's name-collision / supersession semantics — only remove a property when doing so provably does not change any emitted schema. This is the same subtlety documented in the design notes for #122.

## Acceptance criteria

- Per-property findings on `#[OA\Property]` / `@OA\Property` inference subsumes, within a schema that is otherwise kept.
- A golden fixture proving a class with a mix of redundant and load-bearing (`description`) properties keeps the load-bearing ones and removes only the redundant ones, with the emitted schema unchanged.
- `docs/linting.md` row + `CHANGELOG.md` entry.

Radiergummi:
🤖 **[lead]** admitted → planning (sequential cluster, #195/PR #399 merged → foundation ratified). Per-property OA\Property removal granularity, building on #195's now-merged scanner accessors + the whole-class migration.oa-redundant-with-inference removal rule (#185).

Radiergummi:
🤖 **[lead]** **Deferred — not feasible as specified for Spatie Data** (Moritz's call). Per the vetted blocker on #400 (coder + reviewer1 both reproduced empirically), per-property `#[OA\Property]` removal targets an empty niche on Data: authored member `#[OA\Property]` is inert there, the scanner only sees it under a class-level `#[OA\Schema]`, and any such schema yields a defined name that trips the independent-component guard — net guaranteed no-op. The whole-class `migration.oa-redundant-with-inference` rule (#185) already covers these classes.

[...]

PR #204, opened 2026-06-08
feat(#196): operation-level OA redundancy lint rule + fixer

### LINKED ISSUE #196: Migration: remove redundant operation-level `@OA` / `#[OA\*]` annotations (OperationRule)

## Summary

`migration.oa-redundant-with-inference` is a `ComponentSchemaRule` — it only evaluates **class-level** `#[OA\Schema]` / `@OA\Schema`. Operation-level authored annotations are never checked for redundancy. Add an `OperationRule` (sibling rule, or extend the existing one) that evaluates a controller method's authored annotations against the inference-only **control operation** and removes the ones inference reproduces.

In scope: `@OA\Get`/`@OA\Post`/… method annotations and their `@OA\Response` / `@OA\Parameter` / `@OA\RequestBody`, plus authored `summary` / `description` / `operationId` / `tags`. The harvester already reads all of these (`HarvestAuthoredAnnotationsStage::mergeAuthoredOperation()` / `copyAuthoredMetadata()`), so they exist in real apps and have a control counterpart.

## Why this is the highest-value next step

l5-swagger / swagger-php apps are predominantly annotated at the **operation docblock** level on controllers, not at the model `#[OA\Schema]` level. Today's rule cleans the smaller share; this covers the bulk.

## Approach

Reuse the infrastructure from #122/#185: `InferenceControlDocument` (already builds the full control doc, not just schemas), `SchemaEquivalence::subsumes()`, and the same dangling-`$ref` guard. Compare each authored operation field against the matching control operation (same path+method, available via `OperationNode::$descriptor`). Fire only on subsumption; never remove a human `description`/`summary` inference can't derive. Fixer removes the redundant `@OA\Response`/etc. block (docblock) or `#[OA\Response]` attribute, reusing `DocblockAnnotationRemover` / the attribute-removal path.

## Out of scope / depends-on

`@OA\Response` / `@OA\Parameter` **component definitions** (`#/components/responses/*`, `#/components/parameters/*`) — the harvester does not yet harvest these (it logs + skips response-component `$ref`s), so they can't be migrated until harvesting lands. Track separately if needed.

## Acceptance criteria

- An `OperationRule` flags operation-level authored annotations inference reproduces, gated the same way (`migration.*`, level 4).
- `--fix` removes the redundant operation annotation, preserving surrounding comments/whitespace.
- Before/after golden fixtures + a same-API-surface feature flavor (operation unchanged after removal).
- Dangling-`$ref` guard honored; `docs/linting.md` row + `CHANGELOG.md` entry.

Radiergummi:
Closes #196.

## What this delivers

`migration.oa-redundant-operation-with-inference` — the operation-level sibling of the schema-level removal rule shipped in PR #185. It flags hand-authored controller-method annotations (`@OA\Get`/`#[OA\Get]`/… and their nested `@OA\Response` / `@OA\Parameter` / `@OA\RequestBody`, plus authored `summary` / `description` / `operationId` / `tags`) whose contribution inference now reproduces on its own, and `--fix` removes them.

This is the **higher-volume** half of the migration story: l5-swagger / swagger-php apps are predominantly annotated at the operation-docblock level on controllers, not at the model `#[OA\Schema]` level (which #185 covered).

## The redundancy decision (same as #185, applied to operations)

Provenance-based, sound, whole-operation granularity:

- The author's side is the operation the harvester read (`AuthoredAnnotationScanner::operationForMethod($class, $method)`, keyed off `OperationNode.descriptor`).
- The control side is the matching operation in the **inference-only document** (harvester stage excluded), looked up by route key (`METHOD pathUri`).
- Fires only when inference **subsumes** every authored field — scalar metadata by equality (`summary`/`description`/`operationId`/`tags`), schema-bearing parts (`responses`/`parameters`/`requestBody`) via `SchemaEquivalence::subsumes()`. Anything inference can't recover (a human description, a runtime-only response shape) keeps the block load-bearing → no finding.
- **Whole-operation, all-or-nothing** (mirrors #185's whole-class choice for soundness). Per-field operation cleanup is the operation analog of #197 and stays out of scope here.
- Dangling-`$ref` guard reused: never remove an annotation another surviving authored annotation still `$ref`s by name.

## How it's built

- **`LintContext::$inferenceOperationsByKey`** — the runner indexes the inference-only document's operations by route key alongside the existing `inferenceSchemasByClass`, populated from the same `InferenceOnlyGeneration` (which already carries the full control `document`). No second generation; the schema rule's inference doc is reused.
- **`OaRedundantOperationWithInference`** (`OperationRule`, `FixableRule`, `NeedsInferenceDocument`) — new sibling rule, level 4, `migration.*`, registered by the off-by-default SwaggerPhp plugin, `excludedStages() = [HarvestAuthoredAnnotationsStage]`.
- **Fixer** — removes the operation-level `#[OA\*]` attributes (AST) / `@OA\Get`…`@OA\Response` docblock block (line-based remover, reusing/generalizing `RedundantOaAnnotationFixer` + `DocblockAnnotationRemover`).

## Plan / checklist

1. `LintContext` + `LintRunner`: index & expose inference-only operations → verify: unit test on the index keying.
2. `OaRedundantOperationWithInference` rule (subsumption verdict, dangling-`$ref` guard) → verify: golden fixtures, including a mixed load-bearing case that is **kept**.
3. Operation-level fixer (attribute + docblock) → verify: before/after fixtures pass `vendor/bin/pint --test`.
4. Register in `SwaggerPhpPlugin`; feature flavor: annotated fixture lints clean after `--fix` **and** produces the same generated document.
5. `docs/linting.md` catalog row + `CHANGELOG.md` `[Unreleased]` entry.

Gates: `composer test`, `vendor/bin/pint --test`, `composer lint` (PHPStan level 8).


Radiergummi:
### Implementation notes / deviations from the plan

Two supporting fixes surfaced during implementation that the plan didn't anticipate; both touch shared code, so flagging them here:

1. **Scanner no longer runs swagger-php's `OperationId` processor.** swagger-php synthesises a hash `operationId` on every operation lacking an explicit one. The scanner was capturing that hash as if the author had written it, so the redundancy comparison could never match inference's readable `operationId` and the rule fired ~never. Dropping the `OperationId` processor (`Generator::withProcessorPipeline(... ->remove(OperationId::class))`) makes the scanner report only author-written content — an author-written `operationId` is preserved (the processor only fills missing ones). This also stops the harvester from injecting hash ids into the generated document, a latent #17 wart; verified no harvester/snapshot regressions.

2. **`SchemaEquivalence` generalised for operation trees.** Two bugs only response/media-type trees hit: it dropped the `schema` key unconditionally (meant for the component *name* on `OA\Schema`, but on `OA\MediaType` that property is the nested schema holding the `$ref`), and `contains()` couldn't match a keyed-map `content` against a list-shaped `content` (swagger-php serialises one side each way). Both fixes are scoped so schema-rule behaviour is unchanged — confirmed by the existing `SchemaEquivalenceTest` + schema-migration suite.

### Granularity decision

Whole-operation, all-or-nothing (mirrors #185's whole-class choice). Per-field operation cleanup is the operation analog of #197 and stays out of scope. Subsumption is the test, so removal only ever *enriches* toward inference (never loses author-written content) — consistent with #185's documented "inference is free to say more" stance.

Gates green locally: `composer test` (1923 passed), `vendor/bin/pint --test`, `composer lint` (PHPStan L8).

Issue #14, opened 2026-06-01
`response()->json([...])` literal inference


Radiergummi:
## Summary

Teach the generator to read a literal `response()->json([...])` (or `response()->json($model)`) call in a controller method and emit a response schema from it, so controllers that return manual JSON without a typed return value are no longer documented as empty.

## Motivation

Manually-built JSON responses are everywhere in real Laravel apps — a controller that does `return response()->json([...])` instead of returning a typed DTO is idiomatic, not exotic. Because the generator only reads typed return values today, every one of these endpoints documents an empty response body, which is silently wrong for anyone consuming the spec. The surveyed apps BookStack and Coolify both lean on untyped manual responses, so this is a recurring miss rather than an edge case.

## Proposed approach

**Tier 1**, per the parent epic's bounded-AST contract: we only match a literal shape, and a built-up variable is refused. Concretely, scan the first N statements of the method body (via the shared body-scan utility introduced under `src/Plugins/Core/Support/` — see #5) for a `response()->json(...)` call and inspect its first argument:

- An **array literal** → build the response schema's properties from the literal's keys.
- A **model expression** → resolve the model's schema (via #6).
- A **`$data` variable** (or any other non-literal/non-model expression) → refuse and emit a generation-log note; this is Tier-2 dataflow and belongs to the `#[Response]` attribute.

Status-code reading is handled in tandem with #19 — this issue covers the body shape only. The resolved schema feeds the existing response-assembly path; an explicit `#[Response]` attribute (`src/Attributes/Response.php`) always wins over the inferred result.

## Acceptance criteria

- A `response()->json([...])` literal argument yields a response schema whose properties come from the array keys.
- A `response()->json($model)` argument resolves to the model's schema.
- A `response()->json($data)` variable argument is skipped and logged, falling back to today's behaviour.
- Add a unit test covering all three cases (literal, model, refused variable) under `tests/Unit/`, plus a flavor fixture exercised by `tests/Feature/ExamplesTest.php` (snapshot byte-match + OpenAPI 3.1 validate + lint-clean) for the literal path.

## References

Parent: #5 (shared bounded-AST body-scan utility, the Tier-1/Tier-2 boundary, graceful-degradation + attribute-fallback contract). Depends on #6 (model schema resolution). Pairs with #19 (response status-code inference). Attribute fallback: `#[Response]`. Verify: `vendor/bin/pest tests/Feature/ExamplesTest.php`

Radiergummi:
[...]

Issue #11, opened 2026-06-01
Query params from `$request->*()` accessors


Radiergummi:
## Summary

Teach the generator to discover query parameters that a controller reads directly off the request inside its method body (`$request->query('sort')`, `->boolean('active')`, and friends), so they show up in the spec instead of being silently dropped.

## Motivation

Filter, sort, and search parameters are almost never declared in a typed request DTO — they are pulled straight off the request in the method body. Because the generator reads signatures only (OAPI-017), the default query-parameter resolver never sees them, so list/index endpoints across surveyed apps ship with an incomplete (often empty) parameter list. These reads are also one of the concrete examples the parent epic calls out as invisible-at-runtime, which makes them a high-value facet to close.

## Proposed approach

Tier 1 — a fixed accessor-name whitelist, no dataflow. Reuse the shared bounded body-scan utility introduced by the first epic child (see Parent for the mechanism) to walk the first N statements of the controller method and match a small set of request-accessor call shapes, mapping each to a query parameter whose type comes from the accessor:

- `$request->query('sort')` and `$request->input('q')` → `string`
- `$request->string('name')` → `string`
- `$request->integer('page')` → `integer`
- `$request->boolean('active')` → `boolean`

The default name is the accessor's first string argument. Feed the discovered parameters through the default query-parameter resolver in `src/Plugins/Core/Resolvers/CoreQueryParameterResolver.php` so they merge with whatever the signature already produced.

For **GET** routes, inline `validate()` keys are query parameters rather than a request body — coordinate with #9 (the inline-`validate()` scanner) so that on GET its extracted keys are routed to this resolver as query params instead of a body schema.

As with every epic child, an explicit `#[QueryParam]` attribute always wins, and anything off the whitelist degrades to today's behaviour with a generation-log note.

## Acceptance criteria

- Each accessor shape (`query`/`input`/`string`/`integer`/`boolean`) yields a query parameter with the correct inferred type and the first argument as its name.
- On a GET route, inline-`validate()` keys surface as query parameters, not a request body.
- An explicit `#[QueryParam]` overrides an inferred parameter of the same name.
- Add a unit test (a new `tests/Unit/Core/Resolvers/CoreQueryParameterResolverTest.php`, mirroring the existing Core resolver/extractor tests such as `tests/Unit/Core/Extractors/FormRequestRequestSchemaResolverTest.php`) exercising each accessor shape plus the GET inline-validate case, with before/after fixtures per matcher.

## References

[...]

Issue #10, opened 2026-06-01
`abort()` / `abort_if` / `abort_unless` → 4xx responses


Radiergummi:
## Summary

Teach the generator to read `abort()`, `abort_if()`, and `abort_unless()` calls in a controller method body and emit the corresponding 4xx responses, so hand-rolled authorization and not-found branches stop being invisible in the document.

## Motivation

Authorization and not-found logic in Laravel controllers is overwhelmingly expressed as `abort(403)` or `abort_unless($user->can(...), 403)` inside the method body. Because the generator reads only signatures (typed request DTOs and typed return values) and never the method body, every one of these 4xx outcomes is currently undocumented — a consumer reading the spec has no idea the endpoint can return a 403 or 404. This is one facet of the broader body-inference gap tracked by the parent epic (#5), and it is a high-value one: these guard clauses are nearly universal in real apps.

## Proposed approach

Tier 1 — a narrow call-shape whitelist, no dataflow. Match three call shapes in the first N statements of the method body, reusing the shared bounded-AST body-scan utility from the parent epic (#5):

- `abort(403)` and `abort(404, 'msg')`
- `abort_if($cond, 404, 'msg')`
- `abort_unless($cond, 403, 'msg')`

The cleanest fit for the existing pipeline is a new `ErrorResponseContributor` modelled on `src/Plugins/Core/ErrorContributors/ThrowsErrorContributor.php`. Like the `@throws` contributor, it takes the bound `ActionDescriptor` and returns a list of `ErrorDescriptor`s (`status`, `description`, `action`) — here built from each matched `abort*` call: the literal integer becomes the status, and the message argument, when present and a literal string, becomes the response description. Register the new contributor in `src/Plugins/Core/CorePlugin.php` alongside the existing `$registry->addErrorResponseContributor(...)` calls (currently `ThrowsErrorContributor`, `MiddlewareErrorContributor`, `ValidationErrorContributor`).

`ErrorResponseInferenceStage` (`src/Support/Generator/Stages/ErrorResponseInferenceStage.php`) already drives the contributor chain, deduplicates by status (first contributor wins), and **drops any status already declared by an explicit `#[Response]` attribute** — so the "attribute wins" guarantee comes for free from the stage and this child does not re-implement it.

[...]

Issue #5, opened 2026-06-01
EPIC: Controller method-body inference (Tier-1 AST whitelists)


Radiergummi:
## Summary

The generator infers request bodies and responses purely from method *signatures*, so anything a controller expresses only inside its method body is invisible to the spec. This epic closes the high-value facets of that gap with bounded AST whitelists, without introducing general dataflow analysis.

## Motivation

Today the pipeline derives request bodies and responses from **signatures** — a typed request DTO (Spatie Data or `FormRequest`) and a typed return value (see `OperationBuilder` at `src/Support/Generator/OperationBuilder.php` and the resolvers it runs). It never reads the **body** of a controller method, so anything expressed only at runtime is invisible to the generated document: inline `$request->validate([...])`, `abort(403)`, `response()->json([...])`, an API Resource's `toArray()`, a Fractal transformer's `transform()`, `$request->query(...)` reads, `QueryBuilder::for(...)` chains, and constructor `$this->middleware(...)`.

In real-world Laravel apps these idioms are pervasive, not edge cases, which is why this is tracked as the headline known gap (OAPI-017 — see `docs/internal/known-gaps.md`). Surveyed apps lean heavily on them: controllers that validate inline or return `response()->json()` produce specs that silently omit their request schema and responses. This epic recovers those facets with the bounded Tier-1 approach proven by Scribe's `GetFromInlineValidatorBase` and by Scramble — matching a small set of well-known call shapes in the first N statements of a method, with no variable tracking across calls.

## Shared context

**Mechanism — Tier-1, bounded AST whitelist.** Each child matches a narrow, well-known call shape in the first N statements of a method body. No variable tracking across calls, no following values into services or through conditionals.

**Boundary (do NOT cross).** General **Tier-2 dataflow** — tracking values across calls, into services, through conditionals — stays **refused**. Where only Tier-2 would close a case, that case is by definition the authoring attribute's job (`#[RequestBody]`, `#[Response]`, `#[QueryParam]`, … in `src/Attributes/`). Every child here must degrade gracefully — fall back to today's signature-only behaviour plus a generation-log note — when the body shape isn't on its whitelist, and an explicit attribute must always win.

[...]

PR #219, opened 2026-06-10
feat(#9): inline `validate()` + `$rules` property request-body extraction

### LINKED ISSUE #9: Inline `validate()` + `$rules` property extraction
## Summary

Recover a request-body schema from controllers that declare their validation rules outside a `FormRequest` — either through inline `validate([...])` calls or a `protected $rules` property/`rules()` method — so these endpoints stop being documented as having no request body.

## Motivation

Plenty of real-world Laravel APIs predate the typed-DTO/`FormRequest` style and express their request shape only at runtime. Today the generator can't see those rules, so the operation emits an empty request body and the linter raises `request.empty` (rule at `src/Lint/Rules/RequestEmpty.php`). The impact is concrete: BookStack produces **0 request schemas**, and the private consumer app survey turned up **`request.empty` ×72**. Two idioms dominate that surveyed population, and both are bounded enough to read safely without any dataflow analysis.

## Proposed approach

**Tier:** the `protected $rules` property is **Tier 0** (a literal class member); the inline calls are a narrow **Tier-1** call-shape whitelist — no variable tracking across statements. Build on the shared bounded-AST body-scan utility from the parent epic (see #5); reuse the existing `ValidationRulesToSchema` at `src/Support/Extraction/ValidationRulesToSchema.php` to turn the recovered rules into a JSON-Schema — do **not** reimplement that mapping.

Two recovery shapes:

1. **Inline validator calls** — scan only the first ~10 statements of the method body for the four common call shapes (the set Scribe recognises): `$request->validate([...])`, `$this->validate($request, [...])`, `Validator::make($request->all(), [...])`, and `Request::validate([...])`. Pull the rules array literal; trailing comments on each entry become field descriptions.
2. **`protected $rules` property** or **`protected function rules()`** on the controller (the BookStack shape) — read the rules array directly, without requiring a `FormRequest` subclass.

For **GET** routes, inline-validate keys describe **query parameters**, not a request body — coordinate the hand-off with #11. Non-literal or dynamically-built rules fall back to today's behaviour (empty body) and emit a generation-log note rather than guessing.

## Acceptance criteria

- A plain `Request` controller using inline `validate([...])` yields a request-body schema.
- A `protected $rules` array is read without a `FormRequest` subclass.
- Dynamically-built rules degrade gracefully and emit a generation-log note; `ValidationRulesToSchema` is reused, not reimplemented.
- Tests cover both idioms plus the degrade path under `tests/Feature/` (an examples flavor exercising a plain-`Request` controller and a `$rules`-based controller, asserting the generated request-body schema), with the rule-to-schema mapping covered via `tests/` unit coverage against `src/Support/Extraction/ValidationRulesToSchema.php`.

## References

Parent: #5 (shared bounded-AST body-scan mechanism, the Tier boundary, and graceful-degradation contract live there). Related: #11 (GET-route keys become query parameters). Attribute fallback for dynamically-built rules: `#[RequestBody]` (`src/Attributes/RequestBody.php`). Verify: `vendor/bin/pest tests/Feature/ExamplesTest.php`

Radiergummi:
## Plan

Recover a request-body schema from controllers that validate outside a FormRequest, per the spec in #9 and the Tier-1 mechanism/boundary contract in #5.

### Shared bounded-AST scan utility — `src/Support/MethodBody/` (`@internal`)

The reusable core later epic children (#10, #11, #14, …) build on:

- **`MethodBodyScanner`** — parses a controller file with `nikic/php-parser` (+ `NameResolver`), locates the `ClassMethod` node for a `ReflectionMethod` by line range, and returns its **first N top-level statements**. Honors the epic's perf invariants: **per-file AST cache** (parse count tracks files, not routes; scoped lifecycle so the cache is per-run) and it is only invoked on a **Tier-0 miss** (the resolver runs last in the request-schema chain, after the FormRequest/`#[RequestField]` resolvers). Also exposes a trailing-`//`-comment lookup per source line (sourced from the same cached file read).
- **`AstLiteralEvaluator`** — evaluates a php-parser expression to a PHP literal (scalars, `true/false/null`, nested arrays, `::class`), throwing `NonLiteralValueException` for anything dynamic. No variable tracking, ever.

### Validate-specific matcher — `src/Plugins/Core/Support/InlineValidatorRulesReader` (`@internal`)

Scans the first **10** statements for exactly the four whitelisted shapes (the Scribe set):

1. `$request->validate([...])` — `$request` must be the method's `Illuminate\Http\Request`-typed parameter
2. `$this->validate($request, [...])`
3. `Validator::make($request->all(), [...])` (facade, alias-resolved; chained `->validate()` fine)
4. `Request::validate([...])` (facade)

The **rules argument** may be:
- an **array literal** → evaluated entry-wise; trailing `// comments` on entries become field descriptions; a non-literal *element* inside a field's rule list (e.g. `Rule::unique(...)`) is skipped, keeping the literal rest
- **`$this->rules`** (property default via reflection — Tier 0), **`$this->rules()`** (zero-arg method, invoked on an uninstantiated-constructor instance, fault-guarded), or either **subscripted with a literal key** (`$this->rules()['create']` — the actual BookStack idiom, verified against the dogfood checkout)
- anything else → **degrade**: no schema (today's behaviour) + a generation-log note naming the action and the escape hatches (`#[RequestBody]` / `#[RequestField]`)

### Resolver — `src/Plugins/Core/Resolvers/InlineValidationRequestSchemaResolver`

Registered by `CorePlugin` **after** `FormRequestRequestSchemaResolver` (first-non-null chain = Tier-0 gate). Maps recovered rules through the existing `Support\Extraction\ValidationRulesToSchema` (not reimplemented), synthesises examples via `FakerExampleSynthesiser`, registers the component as `{Controller}{Method}Request` via `ComponentSchemaRegistry::registerNamed()`, picks `multipart/form-data` when a file rule is present. The `#[RequestBody]` attribute override continues to win unconditionally (applied later in `OperationBuilder::applyRequestBodyOverride()`), and `#[RequestField]` wins by resolver order.

### GET hand-off (#11)

The resolver fires for **POST/PUT/PATCH only**. On GET, inline-validate keys describe query parameters — that routing belongs to #11's query-parameter resolver; this PR suppresses body emission on GET and leaves the wiring to #11 (decision recorded as a PR comment).

### Tests / docs / changelog

- `tests/Unit/Support/MethodBody/MethodBodyScannerTest.php` — statement bounding, per-file cache, comment lookup
- `tests/Unit/Core/Support/InlineValidatorRulesReaderTest.php` — each call shape (before/after fixtures), `$rules` property / `rules()` method / keyed subscript, trailing-comment descriptions, statement limit, degrade paths
- `tests/Feature/InlineValidationSchemaTest.php` — examples-flavor: plain-`Request` controller with inline `validate()`, a `$rules`-based controller, the degrade path (no body + log note), GET suppression
- `docs/request-bodies.md` — new "Inline validation in the controller" section; `CHANGELOG.md` under `[Unreleased]`

### Notes

- `nikic/php-parser` is **already an explicit `require`** (`^5.0`) in composer.json — the epic's "promote from transitive" item is already satisfied; no composer change needed.

Closes #9


Radiergummi:
**GET-route hand-off decision (#11):** the resolver fires for **POST/PUT/PATCH only** and returns `null` before scanning on any other verb, so a GET route with an inline `validate()` keeps today's behaviour (no request body, no log noise). The keys-to-query-params routing belongs to #11; `InlineValidatorRulesReader::read()` is deliberately verb-agnostic, so #11's query-parameter resolver can call the exact same reader and route the recovered keys through `CoreQueryParameterResolver` without new scan machinery. A feature test pins the GET suppression (`tests/Feature/InlineValidationSchemaTest.php`).

Radiergummi:
**Implementation notes / deviations from the spec text**

1. **`$rules` property / `rules()` method reading is tied to a whitelisted validate call.** The issue lists the BookStack shape as a standalone Tier-0 source, but applying a class-level `$rules` array to *every* write action of a controller without seeing it used would be guessing (different actions, different rules). Instead, the rules **argument** of a matched call may be `$this->rules`, `$this->rules()`, or either subscripted with a literal key. I verified the actual BookStack idiom against the dogfood checkout: it is `$this->validate($request, $this->rules()['create'])` — a keyed subscript over a `rules()` *method* — which this PR supports (property default via reflection; method invoked on a constructor-less instance, fault-guarded).
2. **Log-note granularity:** a generation-log note (`notice`) is emitted only when a whitelisted call **was** found but its rules couldn't be read (or individual fields were dropped). A method with no matched shape stays silent — logging every non-validating action would be noise.
3. **Partial literals:** a dynamic *element* inside a field's rule list (`Rule::unique(...)`, `new Enum(...)`) drops just that element, keeping the literal rest; a fully dynamic field value drops that field (named in the log note); a dynamic key or the whole argument being non-literal degrades the entire body. Plain-string `in:a,b` rules still produce enums; literal-args `Rule::in([...])` reconstruction is a possible follow-up, not done here.
4. **Conditional statements are not descended into** — a `validate()` inside an `if` branch is conditional and reading it would be guessing. Only top-level expression/return statements within the first 10 participate (chains and assignments are unwrapped within those).
5. **`composer.json` untouched:** `nikic/php-parser` is already an explicit `require` (`^5.0`) on `main` — the epic's "promote from transitive" note is stale.
6. **Perf invariants (epic #5 comment):** per-file AST cache in `MethodBodyScanner` (parse count tracks files, not routes — unit-tested via node identity), and the Tier-0 gate falls out of resolver ordering (registered after the FormRequest resolver; first-non-null short-circuits before any parse).
7. **Known theoretical ordering caveat:** Core's inline resolver precedes the SpatieData plugin's resolver in the chain, so an action with *both* a typed `Data` param *and* a matching inline validate would now resolve from the inline rules. In practice the shapes can't co-occur (without a `Request`-typed param, shapes 1–3 can't match; Data-typed actions don't call `Request::validate`), so no extra gate was added.
8. **Example snapshots regenerated** (`vanilla`, `query-builder`, `combined`): the scan fires on our own example controllers — their inline-validated write ops now document request bodies, which is the feature doing its job. The `#[RequestBody]`/`#[RequestField]` precedence is unchanged machinery (override stage + resolver order).

Radiergummi:
## Independent review + empirical survey — **VERDICT: REQUEST CHANGES**

Reviewed `51fc5e9` against the #9 spec and the epic #5 Tier-1 contract; ran all gates myself and a full Layer-A corpus pass (baseline = `main` @ 04194eb, the PR's merge base) on both sides. The feature delivers exactly the promised coverage — BookStack goes 0 → 19 API request bodies (completeness 72.1 → 94.2), zero regressions anywhere — but the call-shape matcher violates the project's own "degrade, never guess" contract on three edges, all empirically demonstrated below. They're narrow fixes in one file; everything else is in good shape.

### Gates (run locally, PHP 8.4.16)

| Gate | Result |
|---|---|
| `composer test` | 1999 passed (one cold-run parallel flake traced to a **pre-existing** `tests/TestCase.php` setUp race, not this PR → filed #222; serial + 2 further parallel runs all green) |
| `composer lint` (PHPStan L8) | No errors |
| `vendor/bin/pint --test` | Passed |

### Findings

**1. [blocker] The scan descends into conditional expressions and closures — including `if` statements *inside* closures — contradicting the declared Tier boundary.**
`src/Plugins/Core/Support/InlineValidatorRulesReader.php:102-105` matches via `NodeFinder::findFirst($expression, …)`, which traverses *every* subnode of a top-level statement. The comment above it (lines 90-91), deviation note 4 on this PR, and `docs/request-bodies.md` ("A `validate()` call inside an `if` branch … is not picked up") all promise otherwise. Probed against the branch build:

| Shape | Result |
|---|---|
| `if ($req->has('x')) { $req->validate([…]); }` (statement-level) | correctly skipped |
| `$x ? $request->validate(['c'=>'required']) : []` (ternary) | **matched, rules read** |
| `$req->has('guard') && $req->validate(['d'=>'required'])` (short-circuit) | **matched** |
| `DB::transaction(function () use ($request) { if (…) { return $request->validate(['e'=>'required']); } … })` | **matched — a doubly-conditional validate documented as the request body** |

A branch-specific rules array documented as *the* request body is a guess. Fix: replace the unconditional descent with a traversal that refuses to enter `Closure` / `ArrowFunction` / `Ternary` / logical `BinaryOp` / `Match_` nodes (if you decide straight-line closure bodies à la `DB::transaction(fn () => $request->validate([…]))` should stay matchable, that's defensible — but conditionals inside them must not be). And pin the boundary with tests — none of the four rows above is covered today, which is how this slipped through.

**2. [important] `facadeMatches()` written-name fallback misidentifies non-facade classes.**
`src/Plugins/Core/Support/InlineValidatorRulesReader.php:194-204` — the docblock claims the bare-short-name fallback applies only "in a namespace where it does not resolve to anything else", but no such check exists: after the FQCN comparison fails, *any* call written as `Validator::make(...)` / `Request::validate(...)` matches via `originalName`, even when the import resolves it to an unrelated class. Probe: `use App\Support\Validator; Validator::make($data, ['not-rules' => 'just-options'])` → **rules recovered, bogus request body fabricated**. Fix: only take the fallback when the resolved name is the namespace-fallback resolution (no matching import) — e.g. accept resolved `\Validator`/`\Request` root aliases and the current-namespace resolution of an unimported short name, but never a name an import explicitly bound elsewhere.

**3. [minor] Shape 2 places no constraint on argument 0.**
`InlineValidatorRulesReader.php:152-154` — any two-argument `$this->validate(x, y)` matches. Probe: a controller's own `private function validate(array $data, array $options)` helper matched and emitted a degrade notice (log noise); with string-valued options it would fabricate a body. Caveat discovered in the survey: Bagisto's real idiom is `$this->validate(request(), [...])` — **no `Request` param in the signature** — so the fix must accept `{request-typed param var, request() helper call}` for arg 0 rather than requiring the parameter, or Bagisto's +14 bodies vanish.

**4. [important] Component-key collision: same-short-name controllers silently share one schema.**
`InlineValidationRequestSchemaResolver.php:179-182` keys components as `{ControllerShortName}{Method}Request`; `ComponentSchemaRegistry::registerNamed()` returns early on an existing key, so the second controller's (potentially different) schema is silently discarded and its operation `$ref`s the first one. This is live in the survey output: in Bagisto, `CartControllerStoreCouponRequest` is shared by `Webkul\Admin\…\Sales\CartController::storeCoupon` **and** `Webkul\Shop\…\API\CartController::storeCoupon` — two different classes. Their rules happen to be identical today, so nothing visibly breaks, but that's luck; V1/V2-style apps with diverging rules get a wrong schema with no warning. The pattern is inherited from `RequestFieldRequestSchemaResolver`, but this PR turns it from a rare explicit-attribute path into a pervasive inferred one. Suggest qualifying on collision (the registry already knows namespace-disambiguation via `deriveKey()`/`isKeyTaken()`), here and in the `RequestField` twin.

**5. [minor] `AstLiteralEvaluator` ships without dedicated unit tests.**
21.95 % patch coverage (codecov above): `::class`, `UnaryMinus`, const-fetch, unpack/by-ref rejection and non-scalar-key paths are never exercised. Epic #5's acceptance says the shared utility is "unit-tested under `tests/Unit/`" — `MethodBodyScanner` is, the evaluator isn't.

**6. [nit]** `__invoke` actions produce `AgentController__invokeRequest` (double underscore, seen in Vito's output) — worth a small special-case in `componentKey()`. **7. [nit]** `docs/request-bodies.md` "On GET, inline-validate keys describe query parameters, not a body" reads like shipped behaviour; the query-param routing is #11, not yet built.

### Survey Layer A — 11-app corpus, baseline (`main` @ 04194eb) → PR (`51fc5e9`)

| App | reqBodies | completeness % | respSchemas | lint total | gen exit |
|---|---|---|---|---|---|
| AdvisingApp | 15 → 15 | 96.8 → 96.8 | 31 → 31 | 31 → 31 | 0 → 0 |
| AureusERP | 36 → 37 | 90 → 91.1 | 90 → 90 | 57 → 57 | 0 → 0 |
| **Bagisto** | **4 → 18** | **49 → 59.8** | 76 → 76 | 457 → 457 | 0 → 0 |
| **BookStack** | **0 → 19** | **72.1 → 94.2** | 86 → 86 | 129 → 129 | 0 → 0 |
| Coolify | 0 → 1 | 55.2 → 55.8 | 154 → 154 | 134 → 134 | 0 → 0 |
| InvoiceNinja | 251 → 255 | 88.1 → 88.9 | 522 → 522 | 199 → 199 | 0 → 0 |
| Koel | 54 → 54 | 82.4 → 82.4 | 170 → 170 | 62 → 62 | 0 → 0 |
| Lychee | 92 → 92 | 97.6 → 97.6 | 202 → 202 | 104 → 104 | 0 → 0 |
| Pelican | 72 → 75 | 94.3 → 96.2 | 158 → 158 | 21 → 21 | 0 → 0 |
| SpeedtestTracker | 0 → 1 | 87.5 → 100 | 8 → 8 | 3 → 3 | 0 → 0 |
| Vito† | 40 → 41 | 86.8 → 87.5 | 144 → 144 | 44 → 44 | 0 → 0 |

The expected #9 signal **materialised**: BookStack (the issue's headline case) goes from zero request schemas to 19 API-prefixed bodies / 52 component schemas (54 bodies including web routes); Bagisto +14. No app crashes, no response-schema or lint movement anywhere, `request.empty` was 0 in this corpus on both sides (the ×72 app, a private consumer app, isn't in the pinned corpus). †Vito's tree carries leftover B1-lift annotations (uncommitted), so its absolute numbers aren't pristine — the delta over the identical tree is still valid; its +1 (`AgentController__invokeRequest`) is genuinely inline-derived.

*(Process note: both full-corpus runs silently dropped SpeedtestTracker + Vito from `results.json` — a pre-existing `corpus.sh` stdin bug, filed as #221; both apps were backfilled with `--only` runs on each side.)*

### Schema spot-checks (correctness, not just presence)

- **BookStack `POST /api/books`** (`BookApiController::create`, the `$this->validate($request, $this->rules()['create'])` idiom): all 6 fields recovered, `name` required + `maxLength: 255`, `description` 1900, `image` → `format: binary` + nullable and the body correctly keyed `multipart/form-data` — the invoked `rules()` even resolved the `array_merge(['nullable'], $this->getImageValidationRules())` entry at runtime. **Correct.**
- **BookStack `POST /books`** (web `BookController::store`, array literal): 4 of 5 fields recovered; `image => array_merge(…)` correctly **dropped** as a dynamic element with the promised generation-log NOTICE (`…dropped field(s) image — their rules are not statically readable`). **Correct degrade.**
- **Vanilla example snapshot** (`FlightController::store`): `size:3` → `minLength`/`maxLength` 3, `date` → `format: date`, `sometimes` update variant correctly emits no `required` list. **Correct.**

### Spec-compliance verdict in brief

All four whitelisted shapes work; statement limit enforced (pinned by test); `$rules` property + keyed `rules()` (BookStack idiom) work end-to-end; `ValidationRulesToSchema` reused, not reimplemented; GET suppressed; `#[RequestBody]` override path unchanged (applied after extraction in `OperationBuilder::applyRequestBodyOverride()`); per-file AST cache pinned by node-identity test and `#[Scoped]` keeps it Octane-safe; no variable tracking across statements anywhere — **no Tier-2 leakage in the dataflow sense**. The defects above are all in the *guess-resistance* of the matching, not in the architecture. Fix 1-4 (1 is the gate) and this is a merge.

Radiergummi:
## Review response — all findings addressed in `f9362c0`

Gates re-run locally on the new head: `composer test` ×2 in parallel mode (2043 passed both runs; the known cold-run flake is #222), `composer lint` (PHPStan L8, no errors), `vendor/bin/pint --test` (passed).

**1. [blocker] Conditional-context descent — fixed, as a reusable policy-driven helper.**
The blind `NodeFinder::findFirst` is gone. New shared primitive `Support\MethodBody\StatementNodeFinder` (`findFirst(array $statements, ConditionalContextPolicy $policy, Closure $predicate)`) with a `ConditionalContextPolicy` enum:

- `SkipConditionalContexts` (what #9 uses): only expression/return statements participate, and the depth-first search refuses to enter `Closure` / `ArrowFunction` / `Ternary` / `Match_` / `&&` / `||` / `and` / `or` / `??` / `??=` nodes — wholesale, including their unconditionally-evaluated parts (a ternary's condition, a short-circuit's left operand); matching there would document bizarre code. Argument lists, array literals, call chains, and assignments are still descended (they execute unconditionally as part of their statement).
- `IncludeConditionalContexts`: full descent — built for #10, where an `abort(403)` inside an `if` is exactly the thing to find.

All four probe rows from the review are pinned twice: at the helper level (`tests/Unit/Support/MethodBody/StatementNodeFinderTest.php`, 14 cases including the `DB::transaction(function () { if (…) { …validate(…) } })` double-conditional and the arrow-function variant, plus two inclusive-policy cases for the #10 path) and end-to-end through the reader against real fixture methods (`InlineValidatorRulesReaderTest`: ternary, short-circuit, closure-if → all `null`). On the defensible-judgment note: straight-line closure bodies (`DB::transaction(fn () => $request->validate([…]))`) do **not** stay matchable — the closure boundary is also a deferred/maybe-never execution context (a transaction callback can retry or not run), so the strict policy refuses all closure bodies uniformly rather than special-casing straight-line ones.

**2. [important] `facadeMatches()` impostor fallback — fixed.**
The written-name fallback is removed. The *resolved* name (NameResolver pass) must be the facade FQCN itself (import/alias) or the bare root-namespace short name — which is precisely Laravel's `Validator`/`Request` root alias (`\Validator::make`, `use Validator;`, or global-namespace code). Slight deviation from the sketched fix: the "current-namespace resolution of an unimported short name" case is **not** accepted — in a namespaced file, unimported `Validator::make` resolves at runtime to `CurrentNamespace\Validator`, which either is a real (non-facade) class we must not match, or doesn't exist and the code fatals before it could ever validate anything. So that branch can only ever produce wrong matches. Pinned: `InlineValidationImpostorController::viaImportedValidator` imports a userland `…\Validation\Validator` and calls `Validator::make($request->all(), ['not-rules' => 'just-options'])` → no match (`src/Plugins/Core/Support/InlineValidatorRulesReader.php:213-225`).

**3. [minor] Shape 2 argument 0 — constrained, Bagisto idiom kept.**
`$this->validate(x, y)` now matches only when `x` is the method's `Illuminate\Http\Request`-typed parameter variable or a zero-argument `request()` helper call (`isRequestArgument()`, `InlineValidatorRulesReader.php:189-211`). Judgment call on the untyped-parameter option: rejected — an untyped variable could hold anything, and accepting it re-opens exactly the `$this->validate($data, $options)` hole; documented in the method docblock. Tests: `viaRequestHelper` (`$this->validate(request(), […])`, the Bagisto +14 path) recovers rules; the impostor controller's own `validate(array $data, array $options)` helper with a literal second argument no longer matches at all (previously: degrade-notice noise at best, fabricated body at worst).

**4. [important] Component-key collisions — fixed via the registry's existing machinery; twin filed as #224.**
`InlineValidationRequestSchemaResolver::componentKey()` now reserves a synthetic class-like name — `{ControllerNamespace}\{ControllerShortName}{Method}Request` — through `ComponentSchemaRegistry::reserveKey()`, so the existing `deriveKey()`/`isKeyTaken()` namespace-segment disambiguation applies: the Bagisto pair becomes `CartControllerStoreCouponRequest` + (e.g.) `APICartControllerStoreCouponRequest` — deterministic, readable, and only qualified when actually colliding. `registerNamed()` then replaces the synthetic owner with the named-key sentinel, so the synthetic name never leaks into `componentClassMap()`/the lint suppression walk. Feature test: two same-short-name `CouponController`s (Admin/Shop fixtures) with diverging rules → two distinct `$ref`s, each with its own fields. Per the surgical-changes rule, `RequestFieldRequestSchemaResolver`'s identical pattern is untouched here — filed as **#224** with the same fix sketched.

**5. [minor] `AstLiteralEvaluator` unit tests — added.**
`tests/Unit/Support/MethodBody/AstLiteralEvaluatorTest.php`: scalars (string/int/float/bool/null, negative int/float), nested arrays with mixed keys, `::class`, and the rejection paths — variables, function/method/static calls, instantiation, string concatenation (unsupported by design), named constants, non-`::class` class constants, spread, by-ref, negated non-numerics, non-scalar array keys — all asserting `NonLiteralValueException`.

**6. [nit] `__invoke` keys — normalised.** No method segment for `__invoke`: Vito's case now yields `AgentControllerRequest` instead of `AgentController__invokeRequest`. Pinned by a feature test (`InlineValidationInvokableControllerRequest`). Example snapshots unaffected (no invokable or same-short-name inline-validating controllers in the example apps; `ExamplesTest` green).

**7. [nit] Docs GET sentence — reworded.** `docs/request-bodies.md` now says the scan stays off on GET and the keys are simply not documented today, with query-parameter routing explicitly *planned* (#11); the conditional-context boundary bullet now spells out ternary/match/short-circuit/closure alongside `if`. The CHANGELOG entry's matching phrase got the same one-line tightening.

**Re-review note for the survey pass:** the matcher is strictly narrower now. Expected movement: any operation in the corpus whose body was recovered through a conditional context (ternary/short-circuit/closure-wrapped validate), an impostor short-name class, or a non-request `$this->validate(...)` helper will lose its (incorrect) body — by design. The headline paths are all pinned as still working: BookStack's `$this->validate($request, $this->rules()['create'])`, Bagisto's `$this->validate(request(), […])`, facade imports/root aliases. Bagisto's shared `CartControllerStoreCouponRequest` will split into two distinct components.

Radiergummi:
## Re-review of `f9362c0` — **VERDICT: APPROVE**

All seven findings from the [original review](https://github.com/Radiergummi/laravel-openapi/pull/219#issuecomment-4667075028) verified fixed — the blocker and both matchers empirically, against fresh probes independent of the committed tests. Gates re-run green; full 11-app corpus re-run on the new head shows the tightening cost **zero** request bodies and produced exactly the two predicted improvements. Ready for Moritz's merge review.

### Per-finding verification

**1. [blocker] Conditional-context descent — fixed, verified empirically.** `Support\MethodBody\StatementNodeFinder` + `ConditionalContextPolicy::SkipConditionalContexts` replaces the blind `NodeFinder::findFirst`. I re-ran my original probes plus six shapes the committed suite doesn't cover, against the branch build:

| Probe | Result |
|---|---|
| ternary `$x ? $req->validate([…]) : []` | refused |
| short-circuit `$req->has('g') && $req->validate([…])` | refused |
| `if`-guarded validate inside `DB::transaction(function () {…})` | refused |
| elvis `$req->input('x') ?: $req->validate([…])` | refused |
| `??=` coalesce-assign | refused |
| `or` / `and` keyword short-circuits | refused |
| `match` arm | refused |
| straight-line arrow closure `DB::transaction(fn () => $req->validate([…]))` | refused (documented judgment: closure = deferred context, uniform refusal) |
| straight-line `$req->validate([…])` | still matches |

The helper is pinned at both levels: 14 unit cases in `tests/Unit/Support/MethodBody/StatementNodeFinderTest.php` (including two for the `IncludeConditionalContexts` policy pre-built for #10 — genuinely unit-tested, not dead code) and end-to-end reader tests for ternary/short-circuit/closure-if. Docs and CHANGELOG now state the boundary precisely.

**2. [important] `facadeMatches()` impostor fallback — fixed, verified empirically.** The written-name fallback is gone; only the resolved FQCN or the bare root-namespace short name match. Probes: imported userland `Validation\Validator` → no match; **bare unimported `Validator::make` in a namespaced file → no match** (the deviation from my sketched fix — correctly reasoned: that name resolves to `CurrentNamespace\Validator`, which is either a real non-facade class or a fatal); `\Validator::make(…)` root alias → matches; bare `Validator::make` in a global-namespace file → matches. Impostor pinned in `InlineValidationImpostorController` + `tests/Fixtures/Validation/Validator.php`.

**3. [minor] Shape-2 argument 0 — fixed.** `isRequestArgument()` accepts only the `Illuminate\Http\Request`-typed parameter variable or a zero-argument `request()` call. Probes: `$this->validate(request(), […])` (the Bagisto idiom) → matches; typed param → matches; **untyped variable → refused** (right call — accepting it re-opens the hole); private `validate($data, $options)` helper → no match, no degrade noise.

**4. [important] Component-key collisions — fixed via `reserveKey()`.** Verified the sentinel ownership: `componentKey()` reserves the synthetic FQCN, `registerNamed()` immediately (no early return between them) replaces the owner with `NAMED_KEY_OWNER`, which `componentClassMap()` filters — so the non-existent synthetic class never reaches the lint suppression walk or `SpecTreeBuilder`'s sourceClass map. Confirmed empirically: corpus lint totals unchanged in all 11 apps, and Bagisto's spec now carries `CartControllerStoreCouponRequest` **and** `APICartControllerStoreCouponRequest`, each `$ref`'d by its own operation (`/admin/sales/cart/{id}/coupon` vs `/api/checkout/cart/coupon`). Same-key idempotency for one action on two routes also holds (`reserveKey` + `registerNamed` both idempotent). The `RequestField` twin is properly tracked as #224.

**5. [minor] `AstLiteralEvaluator` tests — added.** 8 scalar cases, nested mixed-key arrays, `::class`, and 12 rejection paths incl. spread/by-ref/negated-string/non-scalar-key. Covers everything I flagged.

**6. [nit] `__invoke` — fixed.** No method segment; Vito's spec now emits `AgentControllerRequest`, zero `__`-keys in the whole document. Pinned by feature test.

**7. [nit] GET docs wording — fixed.** `docs/request-bodies.md` now says the scan stays off on GET with query-param routing explicitly *planned* (#11); conditional-boundary bullet spells out ternary/match/short-circuit/closure; CHANGELOG matched.

### Gates (re-run locally, PHP 8.4)

| Gate | Result |
|---|---|
| `composer test` | 2043 passed (5023 assertions) |
| `composer lint` (PHPStan L8) | No errors |
| `vendor/bin/pint --test` | Passed |

### Survey — three-way, 11-app corpus (baseline `04194eb` → `51fc5e9` → `f9362c0`)

| App | reqBodies | completeness % | respSchemas | compSchemas | lint | gen exit |
|---|---|---|---|---|---|---|
| AdvisingApp | 15 → 15 → 15 | 96.8 → 96.8 → 96.8 | 31 | 44 → 59 → 59 | 31 | 0 |
| AureusERP | 36 → 37 → 37 | 90 → 91.1 → 91.1 | 90 | 17 → 18 → 18 | 57 | 0 |
| Bagisto | 4 → 18 → 18 | 49 → 59.8 → 59.8 | 76 | 25 → 107 → **108** | 457 | 0 |
| BookStack | 0 → 19 → 19 | 72.1 → 94.2 → 94.2 | 86 | 0 → 52 → 52 | 129 | 0 |
| Coolify | 0 → 1 → 1 | 55.2 → 55.8 → 55.8 | 154 | 2 → 3 → 3 | 134 | 0 |
| InvoiceNinja | 251 → 255 → 255 | 88.1 → 88.9 → 88.9 | 522 | 367 → 375 → 375 | 199 | 0 |
| Koel | 54 → 54 → 54 | 82.4 | 170 | 80 | 62 | 0 |
| Lychee | 92 → 92 → 92 | 97.6 | 202 | 234 | 104 | 0 |
| Pelican | 72 → 75 → 75 | 94.3 → 96.2 → 96.2 | 158 | 101 → 104 → 104 | 21 | 0 |
| SpeedtestTracker | 0 → 1 → 1 | 87.5 → 100 → 100 | 8 | 0 → 1 → 1 | 3 | 0 |
| Vito | 40 → 41 → 41 | 86.8 → 87.5 → 87.5 | 144 | 67 → 81 → 81 | 44 | 0 |

`51fc5e9 → f9362c0` is a clean null delta on every count except Bagisto's **+1 component schema — exactly the coupon-controller split** (rules happen to be identical today, but the silent share is broken). **Zero request-body tick-downs anywhere**: none of the corpus's recovered bodies rode the now-closed edges (conditional contexts, impostor classes, non-request `$this->validate`), so the guess-resistance tightening was free on real-world code. The headline paths held: BookStack 19 API bodies / POST `/api/books` still `multipart/form-data`, Bagisto 18, all `$this->validate(request(), …)` recoveries intact. No crashes, no response-schema or lint movement.

### Verdict

**APPROVE.** The matcher now honours "degrade, never guess" on every edge I could construct, the shared `StatementNodeFinder`/`ConditionalContextPolicy` primitive is the right seam for #10 (its inclusive policy is already tested), and the corpus confirms the narrowing cost nothing. Ready for Moritz's merge review.

Radiergummi:
Stack rebased onto main @45c3bfb — 2026-06-11. All branches force-pushed with `--force-with-lease`; gates verified bottom and top (`composer test`, PHPStan, Pint, snapshot group), `composer test` green on every intermediate branch, and CI is green on all 11 PRs. No reconciliation commits were added: the 2-arg `JsonSchemaFromType` constructor reconciliation (`tests/Support/SchemaFromResourceFactory.php` + `EloquentModelPropertyLookupTest` on #247, four Fractal tests on #261) was amended into the rebased commits that introduced those files. Merge order unchanged: #219 → #225 → #230 → #235 → #241 → #247 → #251 → #253 → #256 → #258 → #261, retargeting each successor to main after its base squash-merges.

Radiergummi:
## Stack rebased onto main @45c3bfb — 2026-06-11

All 11 branches of the Epic-#5 stack were rebased bottom-up and force-pushed with `--force-with-lease`. Reconciliation during the rebase: the `JsonSchemaFromType` two-argument constructor change from main was folded into the #247 branch's `SchemaFromResourceFactory`, and main's #243 parameter dedup now applies to the #241/#256 query-param composition (later-resolver-wins, as verified in the #256 review). Docs/CHANGELOG conflicts were resolved keeping both sides' entries.

**CI: all 11 PRs fully green post-rebase (11/11 checks each, incl. the swagger-php 5.8 and byte-exact snapshot jobs).**

Merge order unchanged — squash-merge bottom-up, retargeting each successor to `main` after its base merges:
#219 → #225 → #230 → #235 → #241 → #247 → #251 → #253 → #256 → #258 → #261

Epic status map: https://github.com/Radiergummi/laravel-openapi/issues/5#issuecomment-4680371894

Radiergummi:

**1. [important — cross-PR with the error machinery] Inline `validate()` infers the request body but never the 422 the same call guarantees.**
`ValidationErrorContributor::contribute()` returns `[]` unless the action has a FormRequest-typed parameter (`src/Plugins/Core/ErrorContributors/ValidationErrorContributor.php:42-44`), so the parity this PR establishes on the request side breaks on the response side. Probe (two POST twins, identical rules): `storeForm(SomeFormRequest $r)` documents `422 → #/components/responses/ValidationFailed`; `storeInline` with `$request->validate(['name' => 'required|string'])` documents the request body but **no 422 at all**. The #241 GET hand-off has the same gap (query validation throws 422 too). The fix is nearly free: `InlineValidatorRulesReader` already answers "does this method run a whitelisted validate?" — the contributor just never asks it. Not recorded anywhere (no 422 mention in the PR decisions, issue #9/#11, or the known-issues list), so this reads as an accidental omission rather than a scope cut.

**2. [minor — shared with #241] `requestParameterName()` is a verbatim ~15-line duplicate** in `InlineValidatorRulesReader.php:~230` and `RequestQueryAccessorReader.php:~236` — same plugin, same namespace; will drift on the first subtlety (variadics, unions). See also the impostor-discipline inconsistency noted on #241 (the `request()` helper guard differs between the request-side and response-side scanners).

Radiergummi:
Fix landed for the stack-review 422 finding: `ValidationErrorContributor` now consults `InlineValidatorRulesReader` on a FormRequest miss, so a whitelisted inline `validate()` documents the same 422 `ValidationFailed` response a FormRequest does — including degraded scans (the throw is a fact of the call). Unit + feature coverage added; the combined example snapshot gains the 422. Commit `1f2620b` (this branch also merged current main, reconciling #266/#269/#270).

Issue #221, opened 2026-06-10
survey: corpus.sh loop loses apps after a bootstrap consumes stdin (results.json silently incomplete)


Radiergummi:
## Summary

`tools/survey/corpus.sh` iterates the corpus with `while IFS= read -r name; do … done <<< "$names"`, but the commands inside the loop (bootstrap scripts, `run.sh`, artisan subprocesses) inherit the loop's stdin. During the 2026-06-10 review runs for PR #219, an app bootstrap (Pelican's, which runs `php artisan p:upgrade` among other things) consumed the remaining heredoc lines, so the loop ended after **Pelican** and **SpeedtestTracker and Vito were silently dropped** from `results.json` — the file contained 9 of 11 apps with no warning, while `manifest.json` (built from a separate pass) still listed all 11.

## Evidence

- Two independent full-corpus runs (library = `main` @ 04194eb, and `feat/9-inline-validate-rules` @ 51fc5e9) both produced a 9-entry `results.json` ending at Pelican; `$WS/apps/SpeedtestTracker/run.json` kept its previous-run mtime, `$WS/apps/Vito/run.json` did not exist.
- `corpus.sh --only SpeedtestTracker` / `--only Vito` immediately afterwards ran fine — nothing wrong with the apps themselves.

## Fix sketch

Redirect the inner block's stdin (`… done <<< "$names"` → feed the loop over fd 3, or end the loop body with `< /dev/null` on the composite command), or collect names into a bash array and use a `for` loop. Also worth a guard: after aggregation, compare `count(results)` against the corpus size and warn loudly on mismatch.

Issue #29, opened 2026-06-01
Implicit framework errors from existing convention


Radiergummi:
## Summary

Emit the standard framework error responses (404, 422, 401, 403) for each operation by inferring them from signals the generator already reads — route-model binding, validation presence, and auth middleware — without parsing controller bodies.

## Motivation

Every Laravel API returns a predictable set of errors that almost nobody documents by hand: a 404 from route-model binding, a 422 from validation, a 401/403 from auth and authorization. Inferring these from convention is the single biggest error-side fidelity win in the package, and it is effectively free because all the signals are already in hand by the time an operation is built. This is the load-bearing half of the convention-derived errors work; the shared envelope it reuses lands in #30.

## Proposed approach

Tier 0 — everything here is derived from binding types, validation presence, and middleware already visible to the generator (the one small Tier-1 add is `findOrFail`/`firstOrFail` detection).

The decoration target already exists: `ErrorResponseInferenceStage` (`src/Support/Generator/Stages/ErrorResponseInferenceStage.php`) decorates each `OA\Operation` and already carries the status -> reason map (401 Unauthorized, 403 Forbidden, 404 NotFound, 422 ValidationFailed). The work is to feed it the inferred per-operation error signals:

- Route-model binding on the action (and, as the Tier-1 add, `findOrFail`/`firstOrFail`) -> **404**.
- A `FormRequest` or inline `validate()` on the action -> **422**, carrying Laravel's `{message, errors}` shape produced by `ValidationRulesToSchema` (`src/Core/Support/ValidationRulesToSchema.php`).
- Auth middleware -> **401**, and `authorize()` / policy / `can:` middleware -> **403**. Reuse the auth detection already performed by `SecurityExtractor::forRoute()` (`src/Support/Extraction/SecurityExtractor.php`) rather than re-deriving the middleware signal.

Each inferred response reuses the standard error-envelope component from #30 rather than inlining its own schema. Keep this in `src/Core` (the error-response inference path); do not read controller method bodies — body-level `abort()` inference lives under #10 (epic #5).

## Acceptance criteria

[...]

PR #225, opened 2026-06-10
feat(#10): abort() / abort_if / abort_unless → 4xx error responses

### LINKED ISSUE #10: `abort()` / `abort_if` / `abort_unless` → 4xx responses

Radiergummi:
> [!IMPORTANT]
> **STACKED on #219** (`feat/9-inline-validate-rules`, issue #9). This PR reuses the shared `Support\MethodBody` scan infrastructure introduced there. After #219 squash-merges, this PR must be **retargeted to `main` and rebased** before review/merge.

Closes #10

## Implementation plan

Second child of epic #5 (Tier-1 bounded AST whitelists): read `abort()`, `abort_if()`, and `abort_unless()` calls in the first 10 top-level statements of a controller action and emit the corresponding 4xx/5xx error responses.

### 1. `StatementNodeFinder::findAll()` (shared primitive, `src/Support/MethodBody/`)

A method can contain several `abort*` calls with different statuses, but `findFirst()` returns one node. Add a `findAll(array $statements, ConditionalContextPolicy $policy, Closure $predicate): list<Node>` sibling using the same policy machinery (full `NodeFinder::find()` under `IncludeConditionalContexts`; the existing conditional-context-refusing walk, collecting instead of short-circuiting, under `SkipConditionalContexts`). Unit-tested in `tests/Unit/Support/MethodBody/StatementNodeFinderTest.php`.

### 2. `AbortErrorContributor` (`src/Plugins/Core/ErrorContributors/`)

Modelled on `ThrowsErrorContributor`; implements `Contracts\Registry\ErrorResponseContributor`. Per action:

- Scan the first **10** top-level statements (`MethodBodyScanner::firstStatements()`, same limit convention as #9) with **`IncludeConditionalContexts`** — an `abort(403)` inside an `if` guard *is* exactly what we want to document.
- Whitelist: `abort(status[, message])`, `abort_if(cond, status[, message])`, `abort_unless(cond, status[, message])`. The condition argument is **never** analysed. First-class callables (`abort(...)`) and namespaced non-global functions are off-whitelist.
- Name matching happens on the **NameResolver-resolved** name: a fully-qualified `abort`/`abort_if`/`abort_unless` matches; an unqualified call in a namespaced file (php-parser leaves it unresolved with a `namespacedName` attribute, mirroring PHP's global fallback) matches **unless** a same-namespace function of that name is actually defined (`function_exists` on the `namespacedName`) — then it's the user's function, not Laravel's helper.
- Status argument via `AstLiteralEvaluator`: literal `int` in **400–599** → an `ErrorDescriptor(status, exceptionClass, description, action)`. Literal int outside that range (e.g. `abort(302)` redirects) → silently skipped (readable, intentionally out of scope). Non-literal status (variable, `abort($exception)`, constants) → call skipped + **generation-log NOTICE** (abort-shaped call found but unreadable; `#[Response]` is the escape hatch).
- Message argument: literal non-empty string → response description. Absent or non-literal → default description from Symfony's status texts (`Response::$statusTexts`), status still emitted.
- `exceptionClass` mirrors what `abort()` actually throws: `NotFoundHttpException` for 404, Symfony `HttpException` otherwise — so error-envelope resolvers treat these like every other contributor's descriptors.

### 3. Registration (`CorePlugin`)

After `ThrowsErrorContributor` (explicit `@throws` stays most specific), **before** the convention-derived contributors (Middleware/Validation/RouteModelBinding) — a body-authored `abort(403, '…')` message beats a generic middleware-map description. `ErrorResponseInferenceStage` already dedups by status (first contributor wins) and drops statuses declared by `#[Response]` — attribute-wins comes free, not re-implemented.

### 4. Tests

- Unit: `tests/Unit/Support/MethodBody/StatementNodeFinderTest.php` (`findAll` both policies, ordering), `tests/Unit/Core/Inference/AbortErrorContributorTest.php` (each call shape, message handling, non-literal skip + notice, range filter, namespaced impostor, statement limit) against new fixture controllers in `tests/Fixtures/`.
- Feature: `tests/Feature/AbortResponsesTest.php` — 4xx responses (with descriptions) in the generated document, `#[Response]` precedence, skip path + log note.
- `examples/` flavors contain no `abort*` calls — snapshots unaffected (verified by grep + full suite).

### 5. Docs + changelog

- `docs/auto-derivation.md`: extend the *Error responses* row of the derivation table.
- `CHANGELOG.md` `[Unreleased]` entry.

### Gates

`composer test`, `composer lint` (PHPStan L8), `vendor/bin/pint --test` — all green before un-drafting.


Radiergummi:
## Decisions & deviations from the issue text

**1. Status range: 400–599, not just 4xx.** The issue title says 4xx, but a literal `abort(500, 'Upstream failure')` is exactly as readable and as much a documented outcome as a 403, so any literal status in **400–599** becomes a response. Literal statuses *outside* that range — `abort(302)` redirects being the realistic case — are **silently skipped**: they're readable and intentionally out of scope (a redirect is not an error response), so no log noise. `abort()` with a `Response`/`Responsable` argument falls out naturally as a non-literal status (see 3).

**2. Non-literal message, literal status → emit with the default description.** The status is a fact worth documenting even when the message is built at runtime; dropping the whole response over presentation text would under-document. The descriptor uses the standard reason phrase (Symfony's `Response::$statusTexts`, e.g. `Not Found`) and no log note is emitted — only an unreadable *status* degrades loudly.

**3. Log-note convention reconciled with #9's, not the issue's wording.** The issue says "when nothing on the whitelist matches … a generation-log note records the skip", which taken literally would log on every abort-free action — pure noise. Following the convention #9 established: **NOTICE only when an abort-shaped call was found but its status isn't statically readable** (dynamic status, `abort($exception)`, named/unpacked args); fully silent when nothing matches. The note names the action and points at `#[Response]`.

**4. Namespaced-function resolution.** Matching happens on the NameResolver-resolved name. Fully-qualified `\abort` matches; a `use function` alias to some other package's `abort` resolves to that FQCN and never matches. An *unqualified* call in a namespaced file is left unresolved by php-parser (mirroring PHP's runtime fallback) — it matches as Laravel's global helper **unless `function_exists()` confirms a same-namespace `abort`/`abort_if`/`abort_unless`** (then it's the user's function; covered by the `AbortImpostor` fixture). Caveat: functions aren't autoloadable, so an impostor in a Composer-`files`-loaded helper is detected, but one in a never-loaded file is not — best effort, consistent with Tier-1's bounded ambitions.

**5. Deviation: one surgical change to `ErrorDescriptor` + the stage's component-sharing condition.** The stage shares description-only error responses via named per-status components (`components.responses.Forbidden`), which are **first-write-wins**. A route-authored `abort(403, 'Admins only')` message hoisted there would leak into every other operation's 403 (or be silently swallowed by an earlier registration). `ErrorDescriptor` gains `shareableDescription` (default `true`, so every existing contributor is untouched), and `ErrorResponseInferenceStage` inlines non-shareable descriptions per operation. Default reason-phrase descriptions still share the component. Covered by a dedicated feature test (`keeps route-authored messages inlined instead of poisoning the shared status component`).

**6. Contributor ordering.** Registered after `ThrowsErrorContributor` (an explicit `@throws` is the author's word) and **before** the convention-derived Middleware/Validation/RouteModelBinding contributors: a body-authored abort message is more specific than a generic middleware-map description for the same status.

**7. `exceptionClass` mirrors runtime behaviour.** `abort(404)` throws `NotFoundHttpException`, every other status a plain `HttpException` — the descriptors say exactly that, so envelope resolvers branch on them like on any other contributor's output.

**8. Examples unaffected.** No `abort*` calls exist in any `examples/` flavor (verified by grep); the full suite, including the byte-exact snapshot group, passes unchanged.

Radiergummi:
Observation (pre-existing, not fixed here): the contributor list in `docs/architecture.md` step 4 was already missing `RouteModelBindingErrorContributor` (added in #29). I added `AbortErrorContributor` to the list per the behaviour-change checklist but left the pre-existing omission alone — flagging it rather than folding an unrelated fix into this PR.

Radiergummi:
## Review — stacked diff `feat/9-inline-validate-rules` (f9362c0) → `feat/10-abort-error-responses` (dff3667)

Empirical review: local gates, throwaway probe fixtures against the contributor (removed afterwards), and a full Layer-A corpus run (11/11 apps) compared against the #219-head baseline (`results-pr-219-r2.json`), so the delta isolates this PR.

**Verdict: APPROVE** — spec-compliant, degradation contract holds in the wild, survey delta clean with zero regressions. One docs nit worth fixing before merge (finding 1); everything else is a follow-up or observation.

### Gates (run locally at dff3667)
- `composer test` — **2072 passed** (5096 assertions). One unrelated first-run flake (fixture-destination `mkdir` race under parallel Pest) that passed on immediate re-run.
- `composer lint` (PHPStan L8) — no errors.
- `vendor/bin/pint --test` — pass.

### Findings

1. **MINOR (docs — fix in this PR).** The bundled-contributor enumerations in `docs/plugin-authoring.md` (≈line 57: "Bundled by Core: `ThrowsErrorContributor`, `MiddlewareErrorContributor`, `ValidationErrorContributor`") and the ASCII pipeline diagram in `docs/architecture.md` (≈lines 46–48) were not updated — only the step-4 prose list was. Two one-line additions. (The pre-existing `RouteModelBindingErrorContributor` omission you flagged is now #228.)
2. **OBSERVATION (follow-up filed: #227).** Class-constant statuses — `abort(Response::HTTP_FORBIDDEN, …)` — are non-literal to `AstLiteralEvaluator` → NOTICE + skip. Empirically this is the *dominant* idiom in Koel: **46 NOTICEs in its generation log and zero documented-error lift**, while every plain-int-literal app lifted as expected. The constant is deterministically resolvable (`constant()` on the NameResolver-resolved FQCN — reflection, no dataflow), so this is a high-value, cheap follow-up.
3. **OBSERVATION (no action requested).** Registering Abort *before* `ValidationErrorContributor` means an `abort(422, …)` on a FormRequest route displaces the canonical 422. In the wild: Lychee `POST /api/v2/Contact` now documents `422 "Incorrect answer to the security question."` (inlined) instead of `$ref ValidationFailed` — and the FormRequest's field-validation 422 outcome becomes invisible. With an envelope resolver configured, the `HttpException` descriptor would also resolve a generic body where the `ValidationException` one is richer. Decision 6 is defensible (OpenAPI's one-response-per-status forces a pick, and the authored message is the more specific text), but it's a real fidelity tradeoff worth having on record.
4. **NIT.** No `examples/` flavor exercises the feature, so the new inlined authored-description response shape never passes the byte-exact + OpenAPI-3.1-validate + lint-clean `ExamplesTest` path that issue #10's acceptance criteria pointed at (#9 did add example routes). `AbortResponsesTest` pins the spec content and the risk is low (the inline shape is a description-only response object), but an example route would double as living documentation.
5. **NIT (fine as-is).** A literal named-argument call (`abort(code: 404)`) is treated as unreadable → NOTICE, per decision 3. Rare enough not to matter.

### Probed edge cases (throwaway fixtures, all behave as designed)

- `abort(403, 'msg')` inside a nested closure in `DB::transaction(...)` → found, inlined ✓
- `abort($code)` / `abort_unless($cond, $status, 'msg')` → skipped + NOTICE naming the action ✓
- `abort(302)` → silent skip ✓ · `abort(new HttpException(…))` / `abort(response()->json(…))` → skipped + NOTICE ✓
- two aborts, same status, different messages in one method → first in source order wins via the stage's `??=`; deterministic, no collision possible ✓
- `__invoke` controller → 403 with message end-to-end ✓
- abort at statement 12 → invisible, silent (limit) ✓
- same-namespace impostor `abort()`: loaded → correctly ignored ✓; defined in a **never-loaded** helper file → matched as the global helper — which is also exactly what PHP's runtime fallback would do, so the generated doc is *correct* in that case. The only genuinely wrong case is a helper the app loads at runtime but artisan doesn't — narrower than decision 4's own caveat suggests. The `function_exists` check is sound in the generator's runtime (same composer bootstrap as the app).

### `shareableDescription` deviation — sound

The problem is real: the per-status response components are first-write-wins, so a route-authored `abort(403, 'Admins only')` hoisted there would either leak into every other operation's 403 or be silently swallowed. The fix is minimal (one default-`true` flag + one condition in the stage), and the zero-impact claim for existing contributors holds — none of them pass the flag, and the stage condition is an identity when it's `true`. Inlining beats suffixed component names: one-off messages would otherwise churn the component namespace while the canonical per-status component stays meaningful. Probed both registration orders: an authored-message route does not poison the shared component for a later plain abort, and vice versa. Same-operation collisions are impossible (status-level `??=` dedup, source order).

### Survey Layer A — f9362c0 → PR #225

Expectation formed *before* the run by grepping the corpus for `abort*` in routed controllers: lift in Vito/Bagisto/AdvisingApp/InvoiceNinja/Coolify/SpeedtestTracker (plain int literals), ~zero in Koel (Symfony constants — see finding 2), zero in AureusERP/Pelican (no aborts). **Matched exactly.**

| App | `response.no-error` | Other metrics |
|---|---|---|
| AdvisingApp | 31 → **22** | stable |
| AureusERP | 57 | stable |
| Bagisto | 422 → **401** | reqBodies 18 ✓ |
| BookStack | 108 → **107** | reqBodies 19 ✓ |
| Coolify | 12 → **11** | stable |
| InvoiceNinja | 26 → **21** | stable |
| Koel | 62 (0Δ) | 46 NOTICEs → #227 |
| Lychee | 104 (0Δ) | 422 desc changed (finding 3) |
| Pelican | 21 | stable |
| SpeedtestTracker | 3 → **2** | stable |
| Vito | 8 → **2** | stable |

No crashes (`generateExit` 0 everywhere); paths / operations / apiOperations / responseSchemas / requestBodies / componentSchemas / completeness all byte-identical to baseline; the **only** lint delta across the corpus is `response.no-error` dropping. Saved as `$WS/results-pr-225.json`.

Spot-checks:
- **SpeedtestTracker** `GET /prometheus` — invokable `MetricsController`, `abort(404)` inside an `if` guard → `404: $ref NotFound` (no message → shared component). Right status, right operation ✓
- **Vito** `UserServerProviderController::show` — has both `#[Response(404, 'Server provider not found or…')]` *and* `abort(404, 'Server provider not found')` → the **attribute** description appears in the doc. Attribute-wins verified in the wild, not just in the fixture ✓ (also confirms the scan does not follow `$this->validateRoute()` into a private method — Tier-2 stays refused)
- **Negative:** Koel `PUT /api/users/{user}` (`abort(Response::HTTP_FORBIDDEN, …)` in a catch block) → **no** 403 emitted, NOTICE logged ✓

Harness note: #221 struck again (9/11 apps in `results.json`), and backfilling via `--only` turned out to *overwrite* the aggregate with a single-app array — filed as #229; the 11-app result was reassembled deterministically via `metrics.php` over the on-disk artifacts.

Radiergummi:
Review fixes pushed as ac379a7:

**Finding 1 (docs):** `AbortErrorContributor` added to the bundled-contributor enumeration in `docs/plugin-authoring.md` and to the `ErrorResponseInferenceStage` block of the pipeline diagram in `docs/architecture.md`, in registration order (after `ThrowsErrorContributor`). The pre-existing `RouteModelBindingErrorContributor` omission is left to #228.

**Finding 4 (examples coverage):** the vanilla flavor's `FlightController::destroy` now guards with `abort_if($model->status === 'departed', 409, 'Departed flights can no longer be cancelled.')`. Snapshot regenerated via `composer examples:vanilla`; the diff is exactly the new inlined authored-description response on `DELETE /flights/{flight}`:

```diff
         '404':
           $ref: '#/components/responses/NotFound'
+        '409':
+          description: 'Departed flights can no longer be cancelled.'
```

so the inline shape now passes the byte-exact + 3.1-validate + lint-clean `ExamplesTest` path, and contrasts nicely with the shared `$ref` Conflict on `bookings.store` (README bullet added).

Gates at ac379a7: `composer test` 2072 passed (5096 assertions), `composer lint` no errors, `vendor/bin/pint --test` pass.

Radiergummi:
Independent stack review (holistic pass over the full epic delta) — findings attributable to this PR:

**1. [minor] `abort()` inside a closure passed to a callback emits the response, and the residual over-claim is unrecorded.**
Probe: `$this->process(function () { abort(404, 'Gone away'); });` → the operation documents `404`. Consistent with the documented `IncludeConditionalContexts` policy, and correct for request-lifecycle closures (`DB::transaction`) — but closures that never execute in this request (queued callbacks, `streamDownload` bodies) over-claim, and that residual is written down nowhere (not in the docblock's policy rationale, not in #238). Worth one sentence in the contributor docblock so the boundary is a decision rather than an accident.

**2. [important — epic-level, shared with #241; noted here because the two ungated scanners are this PR's `AbortErrorContributor` and #241's accessor reader] Perf invariant 2 ("gate AST parsing on Tier-0 miss; well-annotated apps pay ≈ zero") is structurally violated.** `AbortErrorContributor::contribute()` and `CoreQueryParameterResolver::accessorParameters()` call `firstStatements()` for every controller route on every verb with no Tier-0 gate, so **every controller file in any app — fully typed or not — is parsed**. Confirmed empirically: all 8 fully-typed example flavors parse files. The per-file cache bounds the cost (~+0.4–0.6 s at 700 routes), but the epic's economic story and `MethodBodyScanner`'s own docblock ("Callers must only invoke the scanner on a Tier-0 miss") are false as written. Either document the decision on the epic or add a cheap pre-gate (`str_contains($cachedSource, 'abort')` / `'->query('` etc. before walking — ~free). Full numbers in the epic comment.

**3. [minor] Cross-contributor error precedence has single-point test coverage.** The first-wins-per-status dedup is protected only by `ErrorResponseInferenceStageTest` (~line 166); mutating it to last-wins leaves `AbortErrorContributorTest` + `AbortResponsesTest` green. Defensible (the stage test owns the contract), but one feature-level assert pinning e.g. abort-beats-validation end-to-end would harden the registration-order comment in `CorePlugin.php:70-80` against silent reorderings.

Issue #229, opened 2026-06-10
survey: corpus.sh --only clobbers results.json instead of merge-updating it


Radiergummi:
## Summary

Sibling of #221. The natural backfill workaround for #221's dropped apps — `corpus.sh --only <App>` — doesn't merge into the existing `results.json`; it **overwrites the file with a single-app array** (the aggregator only collects entries from the current loop pass before writing).

## Evidence

During the PR #225 review run, the full corpus pass produced 9/11 apps (#221), and backfilling via `--only SpeedtestTracker` then `--only Vito` left `results.json` containing **only Vito**. The full 11-app result had to be reassembled by invoking `metrics.php` per app over the on-disk artifacts (which works, since the extractor is deterministic over `generated-spec.json` / `lint.json` / `run.json`).

## Fix sketch

When `--only` is given, read the existing `results.json` (if any), replace/insert the entry for the named app, and write the merged array back — plus #221's proposed count-vs-corpus-size guard.

PR #230, opened 2026-06-10
feat(#227): resolve class-constant statuses in the body-scan literal evaluator

### LINKED ISSUE #227: Resolve class-constant HTTP statuses (`Response::HTTP_*`) in body-scan whitelists
## Summary

The #10 abort body scan treats a class-constant status — `abort(Response::HTTP_FORBIDDEN, 'Cannot update a user prospect.')` — as non-literal: `AstLiteralEvaluator` only resolves `::class` constants, so the call is skipped with a generation-log NOTICE.

## Motivation (empirical, from the PR #225 review survey)

Symfony's `Response::HTTP_*` constants are the *dominant* abort idiom in at least one corpus app: **Koel** has 46 abort-shaped calls skipped as unreadable (46 NOTICEs in its generation log) and got **zero** documented-error-response lift from #10, while every plain-int-literal app (Vito, Bagisto, AdvisingApp, InvoiceNinja, SpeedtestTracker…) lifted as expected. AdvisingApp also mixes the idiom in (`abort_unless($user->has_enabled_public_profile, Response::HTTP_NOT_FOUND)`).

## Proposed approach

A class constant on a NameResolver-resolved class name is deterministically readable without any dataflow: the scanner already resolves the FQCN, and `constant('Symfony\\…\\Response::HTTP_FORBIDDEN')` (guarded by `class_exists` + `defined`) is plain reflection on an autoloadable class — the same Tier-0 trust we extend to `function_exists` in the impostor check. Either:

- extend `AstLiteralEvaluator::evaluateClassConstant()` to resolve non-`::class` constants when the class is loadable (benefits every body-scan consumer: #9 rules arrays, #14 `response()->json(…, Response::HTTP_CREATED)`), or
- resolve it locally in `AbortErrorContributor` if widening the evaluator's literal contract is unwanted.

Failure stays graceful: unresolvable class or undefined constant → existing NOTICE path.

## References

Sibling: #10 (PR #225 — review identified this gap). Entry points: `src/Support/MethodBody/AstLiteralEvaluator.php`, `src/Plugins/Core/ErrorContributors/AbortErrorContributor.php`.

Radiergummi:
Closes #227.

> [!IMPORTANT]
> **Stacked PR:** this branch is based on `feat/10-abort-error-responses` (#225), which itself stacks on `feat/9-inline-validate-rules` (#219). Merge order: #219 → retarget+rebase #225 → retarget+rebase this PR onto `main`. Only the commits past #225's head belong to this change.

## Why

PR #225's review survey: Koel writes `abort(Response::HTTP_FORBIDDEN, '…')` — 46 abort-shaped calls skipped as non-literal, **zero** documented-error-response lift, while every plain-int-literal app lifted as expected. A class constant on a NameResolver-resolved FQCN is deterministically readable (Tier-0 trust: `class_exists` + `defined` + `constant`), no dataflow involved. #14 (`response()->json(…, Response::HTTP_CREATED)`) hits the same wall.

## Plan

1. **Evaluator-level** (the issue's preferred option): extend `AstLiteralEvaluator::evaluateClassConstant()` to resolve non-`::class` constants when the class part is a `Name` that loads (`class_exists`/`interface_exists`) and the constant is `defined()`. Benefits every body-scan consumer: #10's abort statuses, #9's rules arrays, the upcoming #14 — no contributor-local special case.
2. **Literal contract guard:** `constant()` may yield non-literal values — an enum case is an object, and array constants may *contain* enum cases. The resolved value is accepted only when it is recursively scalar/null/array; anything else stays `NonLiteralValueException` → the caller's existing graceful path.
3. **Unresolved class names** (`self::`/`static::`, dynamic `$class::CONST`, dynamic constant names) stay non-literal — no special-casing; `class_exists('self')` is simply false.
4. **TDD:** failing unit tests in `AstLiteralEvaluatorTest` first (Symfony `Response::HTTP_FORBIDDEN` → 403, imported+aliased class, undefined constant, missing class, enum case, array constant, array-with-enum-case, dynamic class/name); then end-to-end coverage in `AbortErrorContributorTest` + `AbortResponsesTest` via a `abort(Response::HTTP_FORBIDDEN, …)` fixture method, plus one inline-validate test proving #9's rules path benefits.
5. Docs touch-up where the literal contract is described (`docs/auto-derivation.md`, `docs/request-bodies.md`) + `CHANGELOG.md` under `[Unreleased]`.

Gates: `composer test`, `composer lint`, `vendor/bin/pint --test`; verify examples snapshots unaffected.


Radiergummi:
## Decisions & deviations from the plan

**1. Evaluator-level, as the issue preferred.** No `AbortErrorContributor` change was needed at all — once the evaluator resolves the constant, the abort path (and #9's rules path) lift for free. The contributor diff is zero lines.

**2. Literal-value guard is recursive.** `constant()` can return non-literal values: an enum case is an object, and since PHP 8.1 an *array* constant may contain enum cases. Scalars, `null`, and arrays thereof (recursively) count as literal; anything else throws `NonLiteralValueException` and takes the caller's existing graceful path. Plain array constants are accepted — they are compile-time literals, and the inline-validate consumer legitimately wants them. Both sides pinned by unit tests (`NESTED_ARRAY` accepted, `CONTAINS_ENUM_CASE` rejected).

**3. `defined()` pre-check replaced by `catch (Error)`.** PHPStan (checked-exceptions mode) flags `constant()`'s `Error` even behind a `defined()` guard. Catching it is the cleaner shape anyway: one lookup instead of two, and the undefined-constant edge is *guaranteed* to degrade to the NOTICE path rather than crash. `class_exists`/`interface_exists` stays as the explicit intent gate (and is what keeps `self::`/`static::`/never-resolved names out — no special-casing, `class_exists('self')` is simply false).

**4. Autoloading is accepted, unguarded — matching the codebase.** Every other Tier-0 read (`ThrowsErrorContributor`, `RouteModelBindingErrorContributor`, `SuppressionCollector`, …) calls bare `class_exists()` on app-named classes; a throwing autoloader would already break those. No extra try/catch around the existence check.

**5. Interface constants resolve too** (`interface_exists` alongside `class_exists`, same pairing as `ThrowsErrorContributor`) — `DateTimeInterface::ATOM` is as deterministic as a class constant.

**6. #9's path benefits with zero extra changes** — verified by a new feature test: `'title' => RuleSets::TITLE` (whole-ruleset string constant) and `['required', RuleSets::STATUS]` (constant array element) both produce the full schema (`maxLength: 120`, `enum: [draft, published]`).

**7. One pre-existing test dataset renamed.** `AstLiteralEvaluatorTest`'s rejection entry `'non-class class constant' => ['Status::Draft']` still rejects, but now because the class is unloadable, not because non-`::class` constants are categorically refused — renamed to `'constant on an unloadable class'` so the pin stays honest.

**8. Examples unaffected** — the full suite including the byte-exact `snapshot` group passes unchanged (examples use plain int statuses).

Radiergummi:
## Review of the #230 delta (empirical, gates + corpus)

Reviewed `origin/feat/10-abort-error-responses..origin/feat/227-class-constant-statuses` only (#219 and #225 previously approved). Production change is confined to `AstLiteralEvaluator::evaluateClassConstant()` as planned; the contributor diff is genuinely zero lines.

**Gates** (PHP 8.4.16, fresh worktree at `c210895`): `composer test` 2087 passed (5121 assertions) · `composer lint` no errors · `vendor/bin/pint --test` passed.

### Findings

1. **[info] `self::`/`static::` behaviour verified empirically — the claim holds.** php-parser's NameResolver preserves special class names (probed through the real `MethodBodyScanner` + `AbortErrorContributor`: a controller with `abort(self::FORBIDDEN_STATUS, …)` keeps the `ClassConstFetch` class node as literal `self`/`static`), so `class_exists('self')` is false and the call degrades to the NOTICE path exactly as decision 3 describes. Pinned by the `'unresolved self reference'` unit case. The contributor *does* hold the enclosing `ReflectionClass`, so substituting it for `self`/`static` would be a cheap deterministic follow-up — but no corpus app uses the idiom, so I'm not filing it.
2. **[info] `catch (Error)` breadth probed — narrow enough.** Undefined constant → graceful; a *lazily evaluated* constant expression that throws at fetch time (e.g. `const BOOM = self::OK % 0;` → `DivisionByZeroError`) → also caught → graceful, which is the right outcome and slightly more than the "undefined constant on a loaded class" comment claims (cosmetic). Private/protected constants → `Error` → graceful skip (probed). An autoloader throwing a plain `Exception` during lazy constant resolution escapes, but that matches the accepted bare-`class_exists` posture codebase-wide (decision 4). No unrelated-Error swallowing path found: by the time `constant()` runs, the class is loaded, so the catch scope is precisely constant resolution.
3. **[minor] Array-valued *whole-ruleset* constants don't actually reach the schema.** Decision 2 says array constants are accepted because "the inline-validate consumer legitimately wants them" — but `InlineValidatorRulesReader::fieldRulesOf()` only passes through **strings** in the whole-ruleset slot: probed `'title' => RuleSets::TITLE_ARRAY` (`public const array TITLE_ARRAY = ['required', 'string', 'max:120']`) and the field lands in `skippedFields` while its inline-written equivalent works. Only string constants (tested) and constant *elements* benefit today. Not blocking — the abort path is unaffected and the degraded path is graceful — but either accept literal arrays in `fieldRulesOf()` (looks like a two-line widening) or soften the decision-comment claim.
4. **[nit] CHANGELOG mechanism wording is stale:** "`class_exists` + `defined` + `constant`" — the implementation deliberately dropped `defined()` for `catch (Error)` (decision 3). Cosmetic.
5. **[ok] Dataset rename (decision 7) is honest** — `Status::Draft` in the resolver-less unit context is an unloadable class, so `'constant on an unloadable class'` is the correct pin. Consumers handle the widened contract safely: the abort path gates on `is_int` + 400–599, rules keys on `is_string`. Docs deltas (auto-derivation.md, request-bodies.md) match observed behaviour.

### Survey (Layer A) — but first, an infrastructure caveat

The previously saved aggregates were **not trustworthy**: 8 of 11 apps' vendor symlinks still pointed at the **#219 review worktree** (lock pinned `dev-f9362c0`) because `composer require` answers "Nothing to modify in lock file" when relinking, and four apps additionally had corrupted `composer.json` files blocking bootstrap entirely. Filed as #233 with the full forensics. I therefore produced **both** runs fresh from reset composer state, with per-app link verification (`readlink` + locked sha):

- **base** = `ac379a7` (#225 head) → `results-base-230.json` (11 apps)
- **PR** = `c210895` (this head) → `results-pr-230.json` (11 apps; both runs needed the known #221 stdin-drop backfill for SpeedtestTracker + Vito, merged manually per #229)

**Aggregate delta (base → PR):** every column identical for all 11 apps — paths, responseSchemas, requestBodies, completeness%, lint findings, `generateExit` all unchanged, no new crashes. The lift is invisible to these aggregates by construction: `response.no-error` doesn't drop because the affected operations already carried a middleware-derived 401/404. The real signal is in the specs:

| app | abort NOTICEs base → PR | documented 4xx/5xx base → PR | new statuses |
|---|---|---|---|
| **Koel** | **46 → 0** | **367 → 386 (+19)** | 400×2, 403×10, 404+4, 409×1, 500×1, 501×1 |
| **AdvisingApp** | **18 → 4** | **206 → 212 (+6)** | 401+6 |
| plain-int apps | — | metrics byte-identical | — |

**Koel spot-checks (3/3 correct against controller source):**
- `PUT /api/users/{user}` → `403: "Cannot update a user prospect."` — from `abort(Response::HTTP_FORBIDDEN, …)` in `UserController::update`, and a nice double-proof: it sits inside a **catch block** (conditional-context reach from #10) and resolves `Illuminate\Http\Response::HTTP_FORBIDDEN` via **inheritance** from the Symfony base class.
- `GET/DELETE /api/invitations`, `POST /api/invitations/accept` → `404: "The invitation token is invalid."` — `UserInvitationController`.
- `POST/DELETE /api/playlists/{playlist}/songs` → `403: "Smart playlist content is automatically generated"` — `abort_if($playlist->is_smart, Response::HTTP_FORBIDDEN, …)`.

**Remaining NOTICEs:** Koel is at **zero**. AdvisingApp's residual 4 are all the **named-arguments** idiom (`abort_if(boolean: …, code: 403, message: '…')` in the OTP controllers) — literal status, blocked only by the positional-args assumption. Filed as #234 (cheap, deterministic).

### Verdict

**APPROVE.** Finding 3 is worth a small follow-up commit or a softened decision comment, 4 is a one-word CHANGELOG touch-up; neither blocks. The headline empirical claim of #227 is fully borne out — Koel converts all 46 skipped aborts into documented responses with the right statuses and messages.

Radiergummi:
Review fixes pushed in 00e10ab:

**Finding 3 (array-valued whole-ruleset constants):** took the widening — it stayed surgical. `InlineValidatorRulesReader::fieldRulesOf()` now passes a non-array expression through when it evaluates to a **list of strings**, not just a string, so `'title' => RuleSets::TITLE_ARRAY` reaches `ValidationRulesToSchema` (which already accepted array rule lists from the FormRequest path). Anything else a constant resolves to (non-list arrays, non-string elements) still skips the field gracefully. Pinned end-to-end: `LiteralConstantsFixture::BODY_RULES` (`['required', 'string', 'max:5000']`) added to the constant-rules fixture action, with the feature test asserting `type: string`, `maxLength: 5000`, and `required` — decision-comment 2's claim now actually holds.

**Finding 4 (CHANGELOG wording):** corrected to `class_exists` + `constant()` (undefined constant's `Error` caught, degrades to the skip) — no more phantom `defined()`.

Gates: `composer test` 2087 passed (5125 assertions) · `composer lint` no errors · `vendor/bin/pint --test` passed · `snapshot` group 9 passed (examples byte-identical).

Radiergummi:

**1. [minor] Path-dependent literal semantics: the same value yields a different schema written inline vs reached through a class constant.**
Probe-verified against `SchemaDefinitionFromLiteral`:

- `[]` → `{type: array}` via the AST branch (`fromValue`, ~line 110) vs `{}` via this PR's evaluated branch (`fromLiteralValue`) — the evaluated answer matches the class docblock, which the AST branch contradicts.
- `[1, 'a']` → items-less `{type: array}` via the AST branch (`listDefinition` runs a homogeneity check and refuses to claim items) vs `{type: array, items: {type: integer}}` via the evaluated branch — **no homogeneity check**, trusting the first element and fabricating an items claim the AST path deliberately refuses.

So a ruleset/body written inline and the identical structure moved into a class constant document differently. `fromLiteralValue` should reuse `listDefinition`'s homogeneity rule, and the two empty-array answers should agree. (Both branches also emit the items-less `{type: array}` crash shape — that family is consolidated on issue #265.)

Issue #234, opened 2026-06-10
Read named arguments in the `abort()` body scan (`code:` / `message:`)


Radiergummi:
## Summary

The #10 abort body scan only reads **positional** arguments: `AbortErrorContributor::literalValueAt()` skips any `Arg` with `name !== null`. A named-argument call with a perfectly literal status is therefore skipped with a generation-log NOTICE.

## Motivation (empirical, from the PR #230 review survey)

After #227, AdvisingApp's remaining abort NOTICEs (4 of the original 18) are **all** this idiom:

```php
// app-modules/authorization/src/Http/Controllers/OtpLoginCodeController.php:48
abort_if(
    boolean: now()->greaterThanOrEqualTo($otpCode->created_at->addMinutes(20))
        || $otpCode->used_at !== null,
    code: 403,
    message: 'This OTP link has expired or has already been used. Please request a new one.'
);
```

(`OtpLoginCodeController::__invoke` and `VerifyOtpLoginCodeController::__invoke`.) The status is a plain int literal; only the positional-argument assumption blocks it. Koel is at zero NOTICEs after #227, so this is the next (small) idiom on the corpus.

## Proposed approach

Tier-0 trust on the helper signatures: map argument names to positions for the known whitelist shapes — `abort(code, message)`, `abort_if(boolean, code, message)`, `abort_unless(boolean, code, message)`. `literalValueAt()` (or a sibling) resolves an argument by position **or** by name; unknown names keep the graceful skip. Deterministic, no dataflow. The same treatment would benefit #14's `response()->json(..., status: 201)` when that lands.

## References

Siblings:

- #10 (PR #223)
- #227 (PR #230 — review surfaced this as the residual idiom)

Entry point: `src/Plugins/Core/ErrorContributors/AbortErrorContributor.php` (`literalValueAt`).

Issue #233, opened 2026-06-10
survey: bootstrap trusts stale workspace state — path-repo relink silently no-ops, apps keep running an old library commit


Radiergummi:
Found while empirically reviewing PR #230. Three related failure modes, all rooted in the bootstrap trusting whatever composer state the previous run left behind:

## 1. Relinking `$LIB` silently keeps the previous library commit (the serious one)

`survey_link_library` re-points the composer path repo and runs `composer require radiergummi/laravel-openapi:@dev`, but when the lock already contains the package, composer answers:

```
Running composer update radiergummi/laravel-openapi
Updating dependencies
Nothing to modify in lock file
```

— and the vendor symlink keeps pointing at the **old** worktree. Observed today: 8 of 11 apps had `vendor/radiergummi/laravel-openapi -> …/laravel-openapi-wt/review-9/` with the lock pinned to `dev-f9362c0` (the **#219** head), even though several "PR runs" for #225/#230 had been executed since. Every aggregate produced in that window (`results-pr-225.json`, the 10:30 `results-pr-230.json`) partially measured stale library code — which is why those delta tables showed zero movement everywhere.

## 2. Corrupted `composer.json` (trailing extra `}`) blocks bootstrap, run.sh still "produces data"

Coolify, InvoiceNinja, Koel, and Lychee all had an identical corruption signature — one extra `}` appended after the closing brace. Mechanism unconfirmed (likely an interrupted `composer require` revert from an aborted run). Consequences: every `composer config`/`require` fails at read, bootstrap records `blocked-compat`, but `run.sh` still runs `openapi:generate` against the leftover vendor state — yielding either `RuntimeException: Unable to detect application namespace` (crash recorded against the PR under test) or, worse, quietly valid-looking results from a stale library.

## 3. Stale `bootstrap/cache/packages.php` chicken-and-egg

With the package absent from vendor (pristine lock) but `bootstrap/cache/packages.php` still listing `OpenApiServiceProvider`, Koel's `pre-update-cmd` (`@php artisan clear-compiled`) dies with `Class "Radiergummi\OpenApi\OpenApiServiceProvider" not found`, so `composer require` reverts and the app stays blocked.

## Repair procedure that worked (used for the #230 review runs)

Per app, before each corpus run:

```bash
git -C "$repo" checkout -- composer.json composer.lock
[...]

PR #235, opened 2026-06-10
feat(#14): infer response schemas from response()->json([...]) literals

### LINKED ISSUE #14: `response()->json([...])` literal inference

### LINKED ISSUE #240: Resource-action convention status silently overrides a body-scan literal 2xx status
## Context

`OperationBuilder` layers the resource-action convention status onto whatever primary response a resolver produced (`applyConventionStatus`, the "convention claims status, resolver claims body" split):

```php
if ($primaryOverride === null && $convention?->successStatusCode !== null) {
    $primaryResponse = $this->applyConventionStatus($primaryResponse, $convention->successStatusCode);
}
```

Since #14 (PR #235), the `InlineJsonResponseResolver` can produce a primary response whose status was **read from a literal the author wrote** — and the convention rekey silently overwrites it. Probed at `509e5db`:

```php
public function store(): JsonResponse
{
    return response()->json(['id' => 1], 200);   // documented as 201 Created
}

public function index(): JsonResponse
{
    return response()->json(['items' => []], 201); // documented as 200 OK
}
```

Both document the literal body under the **convention** status, not the literal status the code actually returns.

## How this surfaced

Re-review of PR #235 (lost-op audit, https://github.com/Radiergummi/laravel-openapi/pull/235#issuecomment-4669430299). At the pre-fix head `a1bdde1` this interaction was pathological: InvoiceNinja `GET /webcron` (action named `index`, terminal `return response()->json(['message' => 'Invalid secret'], 403)`) had its **error body rekeyed under `200 OK`** by this path. The PR's 2xx-only guard fixed that worst case (the scan now refuses non-2xx literals before the convention layer sees them), so what remains is bounded to a 2xx↔2xx relabel — wrong status, right body.

## Suggested direction

A literal status read from the body scan is Tier-1 *evidence*, not a default: when the resolver's response carries an explicitly-read status, the convention should defer (or at minimum the conflict should emit a generation-log NOTICE instead of silently relabelling). The convention rekey stays correct for resolvers that never read a status (typed-return resolvers default to 200, where `store()` → 201 is exactly the convention's job).

Relevant to #238: when non-2xx `response()->json()` literals are routed into the error machinery, they must not pass through this rekey path either.

Radiergummi:
Closes #14. Closes #240.
## Stack

This PR is the fourth element of the body-scan stack and **must merge last**:

`main` ← #219 (`feat/9-inline-validate-rules`) ← #225 (`feat/10-abort-error-responses`) ← #230 (`feat/227-class-constant-statuses`) ← **this PR** (`feat/14-response-json-literal`)

Rebase/retarget order: when #230 merges into #225's branch (or onward), retarget this PR's base accordingly and rebase. Diff is meaningful only against `feat/227-class-constant-statuses`.

## Plan

1. **New `InlineJsonResponseResolver`** (`src/Plugins/Core/Resolvers/`), a `PrimaryResponseResolver` running the shared Tier-1 bounded scan (`MethodBodyScanner`, first 10 statements, `StatementNodeFinder` under `SkipConditionalContexts` — a conditional `response()->json()` is not the canonical 2xx).
2. **Call shape**: the global `response()` helper (zero-argument) followed by `->json(...)`, matched with the same guess-resistant name resolution as the abort scan. The `Response` facade and `new JsonResponse(...)` stay out of scope (the spec names the helper; narrow by design).
3. **Argument reading is name-aware** from day one (per the #234 corpus finding), against the helper signature `json($data = [], $status = 200, ...)` — both `response()->json(data: [...], status: 201)` and positional calls resolve.
4. **Array-literal body** → object/array schema from the literal structure (nested literals recurse; scalar literals type their property). Partial literals: a dynamic *value* under a literal key emits the property with an unconstrained (empty) schema — dropping a response property is silently wrong for consumers, the opposite trade-off from the request side in #9. A dynamic *key* degrades the whole call.
5. **Status**: a literal (or class-constant, free since #230) `status` argument becomes the response status; absent → 200. A *non-literal* status degrades the whole call with a NOTICE — documenting the body under a guessed status would be wrong. No further #19 machinery (no `setStatusCode()`/`noContent()` chains) in this PR.
6. **Anything else** (variable, method call, `compact()`, model expression) → refuse + generation-log NOTICE (found-but-unreadable convention from #9/#10; silent when no `response()->json` is present at all). The `response()->json($model)` path from the issue spec is deferred — without dataflow there is no Tier-1-resolvable model expression; recorded on issue #14.
7. **Chain placement**: registered last among Core's primary-response resolvers, guarded to fire only when the declared return type carries no schema information (untyped or an HTTP-response class). Typed returns (Model/Data/Resource/Paginator) keep their Tier-0 resolution; an explicit `#[Response(2xx)]` attribute keeps winning via `OperationBuilder`'s primary-override path — verified end-to-end.
8. **Shared conversion**: extract `OperationBuilder`'s private array-definition→`OA\Schema` converter into a shared `Support\Generator` helper used by both the `#[Response(schema: [...])]` path and this resolver.
9. **Tests**: unit (literal incl. nested/list/partial, named args, literal + class-constant + non-literal status, conditional-context refusal, variable refusal, >10-statement miss), feature (end-to-end literal path, attribute-wins, typed-return-wins), and an idiomatic `response()->json([...])` route in the vanilla examples flavor (snapshot + 3.1-validate + lint-clean).
10. **Docs**: auto-derivation row + section, CHANGELOG.


Radiergummi:
## Implementation decisions (deviations & boundaries)

**1. Model-argument path deferred (spec change, also recorded on #14).** The issue's second bullet — `response()->json($model)` resolving via #6's model→schema machinery — is not implementable without dataflow: by the time a model reaches `json()` it is a `$variable`, a `::query()->…` chain, or a `new Model(...)` expression, none of which the Tier-1 contract lets us evaluate (and `EloquentModelResponseResolver`, the #18 machinery, keys off the *return type*, which is `JsonResponse` here). Every model expression therefore takes the non-literal degradation path (NOTICE + `#[Response]` hint). If a Tier-0-resolvable model expression shape ever materialises, it can slot into `bodyDefinition()` without disturbing the literal path.

**2. Partial literals keep their property (opposite of #9's request side).** A dynamic *value* under a literal key emits the property with an unconstrained (empty) schema: on the response side, dropping a property documents the response as *not containing* something it always contains — silently wrong for consumers — whereas an unconstrained property is honest ("present, type unknown"). On the request side (#9) a dropped field merely under-documents accepted input, so dropping was the right call there. A nested literal array that is itself unreadable (dynamic key inside) gets the same unconstrained-property treatment; a dynamic key or spread at any level of the *matched* array degrades the whole call (a guessed key set is a guessed shape).

**3. Status handling — the exact boundary.** A literal or class-constant `status` argument (positional or named) becomes the response status; absent → 200; non-literal → the **whole call** degrades with a NOTICE, because documenting the body under a guessed status is worse than not documenting it. Per the scope note on #14/#19, no further #19 machinery: `->setStatusCode(…)` and `->noContent()` chains are not read (follow-up filed, see below). Also out of scope: precedence against the #25 resource-convention status. `OperationBuilder::applyConventionStatus()` still layers the convention code over any resolver's response, so a (rare) resource-conventional action whose literal status disagrees with the convention — `store` + `response()->json($x, 200)` — documents the convention's 201. That is the pre-existing #25 semantics shared by every response resolver; #19's "explicit literal beats convention" rule would need a resolver→builder signal and belongs to the follow-up, not this PR.

**4. `json()` / `json([])` / `json(null)` skip silently.** They are perfectly readable — the body is `[]`/`null` — but carry no shape worth a schema, and they are *everywhere* in fixtures and real code as "return something JSON-ish". Emitting `type: array` for them would be technically true and practically worthless. The found-but-unreadable NOTICE convention stays reserved for calls we found and could not read.

**5. Scalar literal bodies are in.** `response()->json('ok')` / `json(true)` are valid JSON documents and fall out of the evaluator for free → `{type: string}` / `{type: boolean}`. Class-constant *bodies* (e.g. `json(SomeClass::PAYLOAD)`) also resolve, via the same evaluated-value → definition mapping.

**6. Lists derive `items` from the first element.** Literal response lists are homogeneous in practice (`[1,2,3]`, list-of-objects); when a later element disagrees on type, `items` is omitted (unconstrained) rather than guessing.

**7. No `example` values from the literals.** Tempting (the values are real authored data), but the response-side machinery (`EloquentModelToSchema`, resource schemas) emits no synthesised examples anywhere — `FakerExampleSynthesiser` is request-only. Emitting examples here would be novel behavior for one resolver; left out for consistency.

**8. Facade / `new JsonResponse` out of scope** (per the PR plan): the spec names the helper, and the helper is the corpus idiom. The matcher uses the same guess-resistant name resolution as the abort scan (fully-qualified `\response` matches; an unqualified name matches unless a same-namespace `response()` function actually exists).

**9. Two lint fixtures adjusted.** `ResponseEmptyController` / `SuppressedResponseEmptyController` relied on `response()->json(['ok' => true])` producing no schema to trigger `response.success-empty-body`; the scan now documents that literal, so the fixtures switch to a variable payload (which the scan refuses) — their lint semantics are unchanged, and the fixture docblocks now say why.

**10. Snapshot delta.** Exactly one addition: the new `GET /api/status` operation in the vanilla flavor (attribute-free, fully inferred — schema `status: string` / `read_only: boolean` / `incidents: integer`). No other flavor changed; `examples/combined`'s `response()->json(['status' => 'ok'])` sits behind `#[Hide]` and stays out.

Radiergummi:
## Review: empirical pass on the #235 delta (code probes + full survey re-run)

Gates on `a1bdde1`: `composer test` ✅ (2113 passed; one transient parallel-run fixture-mkdir race, clean on re-run), `composer lint` ✅, `vendor/bin/pint --test` ✅.

**Verdict: REQUEST CHANGES** — the literal scan itself is solid and the corpus lift is real (+116 substantive 2xx schemas, see below), but two precedence/correctness holes surfaced, one of them visible in the corpus.

### Findings

**1. MAJOR — a straight-line literal 4xx/5xx becomes the *primary* response and evicts the 200.** `resolveStatus()` accepts any int, so the pervasive "guarded success + terminal error fallback" idiom gets documented as an error-only operation:

```php
public function verify()   // InvoiceNinja StripeController
{
    $user = auth()->user();
    if ($user->isAdmin()) {
        return $company_gateway->driver(new Client())->verifyConnect(); // real success, conditional
    }
    return response()->json(['message' => 'Unauthorized'], 403);        // straight-line ⇒ matched
}
```

Baseline documented `{200, 429}`; this PR documents `{403, 429}` — **the operation now has no success response at all**. 12 InvoiceNinja `/api/v1` ops regress this way (`stripe/verify`, `postmark_webhook`, `oauth_login`, …); it's why InvoiceNinja's survey `responseSchemas` drops 522→507 and completeness 88.9→87.5. `response.no-success` (level 2) would flag these, but the regression shouldn't ship. Suggest: only a 2xx (or 2xx/3xx) literal status claims the primary; a straight-line non-2xx literal degrades (silently or NOTICE — it's arguably #10-adjacent error material, but that's a follow-up).

**2. MAJOR — explicit `#[ResponseResource]` is defeated by the scan.** The return-type guard keeps the scan off *typed* Tier-0 territory, but not off *attribute-driven* resolution that later plugin resolvers own. Probe (ApiResources enabled):

```php
#[ResponseResource(ProbeArticleResource::class)]
public function show(): JsonResponse
{
    return response()->json(['data' => $this->buildResource()]);
}
```

documents `{data: {}}` (the partial-literal result) — the resolver chain runs Core before ApiResources, so `InlineJsonResponseResolver` claims the route and `ResourceResponseResolver` never runs. Counterfactual confirmed: with a body the scan refuses (`json($payload)`), the same action documents `{data: $ref ProbeArticleResource}` — the attribute *is* resolvable, the scan shadows it. That violates the epic's "an explicit attribute must always win". `#[ResponseResource]` lives in the general `Attributes\` namespace, so the resolver can refuse when the action (or controller) carries it without crossing the plugin boundary; alternatively the scan needs a truly-last chain position.

**3. MEDIUM — `->setStatusCode(201)` chain documents the body under 200.** Probe: `response()->json(['created' => true])->setStatusCode(201)` → documented as **200** `{created: boolean}` while the runtime status is 201. I know decision #3 defers *reading* these chains to #236, but the interim behavior is actively wrong documentation, not absence. A narrow interim guard — degrade with NOTICE when the matched call is the receiver of a `setStatusCode(...)` chain — keeps #236's scope intact without shipping wrong statuses. (`->header(...)` chains are correctly matched and harmless, as pinned.)

**4. MINOR — explicit sequential int keys document an object where runtime emits a JSON array.** `json([0 => 'a', 1 => 'b'])` → `{type: object, properties: {"0": …, "1": …}}`, but `json_encode` emits `["a","b"]`. The *evaluated*-literal path (a class-constant body) handles the same value correctly via `array_is_list()`; the AST path in `definitionFromArrayNode()` doesn't. Serialization itself is fine (checked YAML and JSON on both swagger-php lines — keys survive), it's the semantics that are wrong. Edge shape, but a one-check fix for AST/evaluated-path consistency.

**5. MINOR — multi-call behavior is unpinned and picks a non-returned match.** Probe: `$first = response()->json(['first' => true]); return response()->json(['second' => true], 201);` documents `{first}` @ 200 — the assignment wins over the returned call. Real code rarely does this, but "first match wins regardless of `return`" deserves either a `Return_`-statement preference or at least a pinning test + docblock sentence.

**6. NIT — no status-range sanity.** `json($x, 999)` documents a `'999'` response ("HTTP 999"); `json(['gone' => true], 204)` documents a 204 *with* a body. Both are faithful-to-code, but a 100–599 clamp (mirroring the abort scan's 400–599) would be cheap. Fixing #1 with a 2xx restriction subsumes most of this.

### Verified explicitly (no action)

- **Unconstrained `{}` partial-literal property**: verified through full document serialization on swagger-php **6.1.2 and 5.8.3** — identical YAML (`logs: {}`), `Analysis::validate()` passes on both. The claim holds.
- **Lint fixture change** (`ResponseEmptyController` → variable payload): semantically right, not masking anything. A literal `json()` body now *has* a documented schema, so `response.success-empty-body` not firing on it is the correct new behavior; the fixtures keep testing the rule via a refused payload.
- **Degradation contract**: probes for `compact()`, spread, dynamic key, mixed list+map, dynamic status → refuse + NOTICE; `json()` / `json([])` / `json(null)` → silent; userland same-namespace `response()` impostor → silently not matched; statement 11+ → silent miss; closure-only → conditional NOTICE. All consistent with #9/#10.
- **Typed-return guard**: Model/Data/Resource-typed actions untouched; `#[Response]` 2xx attribute wins via the primary override (pinned in the feature tests).
- **Extraction**: `SchemaFromArrayDefinition` is a faithful move of `OperationBuilder`'s converter (same algorithm, static), `#[Response(schema:)]` path unchanged.
- **Snapshot delta**: exactly the one `GET /api/status` vanilla route, as claimed.

### Survey (Layer A, full corpus re-run at `dev-a1bdde1`)

State verified per #233 before and after (all 11 vendor symlinks → the PR worktree, `composer.lock` ground truth `dev-a1bdde1…`); #221 hit again (run dropped post-Pelican; SpeedtestTracker + Vito backfilled via `--only`, aggregates merged manually per #229). Baseline = `results-pr-230.json` @ `c210895`; no request-side deltas anywhere (`requestBodies` flat in all 11 apps), so nothing to disambiguate against the `00e10ab` rule-widening.

**API ops gaining a substantive 2xx schema** (own spec-level diff; the headline `responseSchemas` metric counts bare-200s and is mostly insensitive):

| app | gained | lost | predicted | note |
|---|---|---|---|---|
| InvoiceNinja | **+66** | 0 | 40–80 ✓ | plus the 12 ops *regressing* per finding 1 |
| Coolify | +16 | 0 | 25–40 ✗ over | most literals sit inside `if`/beyond stmt 10 |
| AureusERP | +14 | 0 | 0 ✗ under | controllers under `plugins/`, outside my grep — restore/force-delete `{message}` ops |
| Koel | +8 | 0 | 4–6 ~ | |
| Vito | +6 | 0 | 6–9 ✓ | `/api/health`, git-hooks — the ops issue #14 was filed about |
| BookStack | +3 | 0 | 3–6 ✓ | `/api/system`, `/api/search`, recycle-bin restore |
| Lychee | +2 | 0 | 0 ✗ | v2 controllers outside `app/Http` |
| SpeedtestTracker | +1 | 0 | 0–1 ✓ | |
| AdvisingApp | 0 | 0 | 1–3 ✗ over | its literals are on non-`/api` routes |
| Bagisto / Pelican | 0 | 0 | 0 ✓ | Bagisto gains 2 partial `{data:{}}` docs (below) |
| **total** | **+116** | **0** | ~120–190 | |

Aggregate metric deltas vs pr-230: flat everywhere except **InvoiceNinja respSchemas 522→507 / completeness 88.9→87.5** — 12 of the 15 are finding 1 (real regression); the other 3 (+ Bagisto's 15→13) are ops that gained a single-key `{data: {}}` partial schema the survey's substantive test doesn't credit — the spec actually got marginally richer there, metric artifact, fine.

**Spot-checks (positive)**: AureusERP `POST …/tags/{id}/restore` → `{message: string}` @ 200 matches `['message' => 'Tag restored successfully.']` ✓; InvoiceNinja `GET /api/v1/ping` → `{company_name: {}, user_name: {}}` partial-literal keys match source ✓; Coolify `GET /api/v1/servers/{uuid}/validate` → 201 `{message: string}` matches the straight-line `json([...], 201)` while the *conditional* 422/404 literals above it are correctly not taken as primary ✓.

**Negative checks**: Coolify `GET /api/v1/teams` (`json($teams)` variable) gains nothing ✓; Coolify `POST /api/v1/deploy` (literal jsons conditional / beyond stmt 10) gains nothing ✓.

### Wrap-up

Findings 1 and 2 are the blockers; 3 is strongly recommended; 4–6 at your discretion. The lift is exactly where #14 predicted it (helper-literal envelopes on ops convention couldn't touch), degradation is disciplined, and nothing fabricated showed up anywhere in 11 apps — once the primary-response slot can't be claimed by an error fallback or stolen from an explicit attribute, this is a clear merge.

Artifacts: `results-pr-235.json` in the survey workspace (all 11 apps at `dev-a1bdde1`).

Radiergummi:
## Review fixes — all six findings addressed in `509e5db`

Gates on `509e5db`: `composer test` ✅ (2124 passed), `composer lint` ✅, `vendor/bin/pint --test` ✅. The vanilla example (`GET /api/status`, plain 200 literal) is unaffected — no example/snapshot output changed (the byte-exact `snapshot` group is green and `git status` shows no `examples/` delta).

### 1. MAJOR — non-2xx literal evicting the success response: fixed, 2xx-only

`resolveStatus()` now refuses any literal status outside 200–299. The call is readable but out of the primary slot's scope, so it degrades with a **NOTICE** pointing at `#[Response]` (your suggested call) — silent felt wrong because the author wrote a real, readable response the generator is declining to document; the note explains why and names the escape hatch. The InvoiceNinja shape (guarded success in an `if` + terminal `json(['message' => 'Unauthorized'], 403)`) is pinned both at resolver level and end-to-end: the operation keeps its bare 200, no 403 appears (`guardedSuccessWithTerminalError` fixture).

**Follow-up filed: #238** — route straight-line non-2xx `response()->json()` literals into the error-response machinery like #10's aborts (`enhancement, spec, tier-1, area:responses`), including the envelope-vs-literal-body question and the 12 regressing InvoiceNinja ops as the acceptance corpus.

### 2. MAJOR — `#[ResponseResource]` defeated: fixed via a shared authoring marker

I went with your shared-guard suggestion rather than a per-resolver attribute list, shaped as a **marker interface**: `Contracts\Attributes\PrimaryResponseAuthoringAttribute`, implemented by `#[ResponseResource]` and `#[FractalResponse]` (the two attributes a later primary resolver consumes — Spatie Data and the paginator resolve from types/generics, no attribute; `#[Response]` 2xx already wins via the `OperationBuilder` override and must *not* blanket-suppress the scan, since a lone `#[Response(404)]` shouldn't cost the inferred 200). The check lives on `ActionDescriptor::declaresAttributeImplementing(string $interface)` — generic, reads the existing attribute buckets, and crucially lets Core step aside for `#[FractalResponse]` **without referencing the Fractal namespace**; third-party plugin attributes opt in by implementing the marker (documented in `docs/plugin-authoring.md`). The scan steps aside *silently* — same convention as the return-type guard: deferring to explicit authoring is correct behaviour, not degradation. #11/#12 can reuse the same marker+method pair. Pinned end-to-end: `#[ResponseResource]` + literal `{data: …}` envelope → `$ref` to the resource component, not `{data: {}}`; plus silent-refusal unit tests for both attributes.

### 3. MEDIUM — `->setStatusCode(201)` documented under 200: fixed, whitelist guard

The matched call's fluent chain is now walked outwards; any link **not** in a known-harmless whitelist (`header`, `withHeaders`, `cookie`, `withCookie` — decoration that cannot touch status or body) degrades with a NOTICE naming the chained method. I checked `JsonResponse`'s surface and inverted your status-only suggestion into a whitelist per degrade-never-guess: `setStatusCode` and `setNotModified` mutate status, but `setData`/`setJson`/`setContent`/`setEncodingOptions` mutate the *body*, and a blacklist would silently rot when Symfony grows a method. A dynamic chain name (`->{$m}(...)`) also degrades. `->header(...)` stays matched (still pinned). #236's scope (actually *reading* the chain) is untouched.

### 4. MINOR — integer-keyed lists on the AST path: fixed

`definitionFromArrayNode()` now keeps evaluated keys under PHP's native key coercion and applies `array_is_list()` before falling back to an object — `json([0 => 'a', 1 => 'b'])` documents as `{type: array, items: {type: string}}` on both paths. Pinned twice: AST literal and class-constant (`LIST_WITH_EXPLICIT_INTEGER_KEYS`) fixtures.

### 5. MINOR — multi-call selection: returned beats assigned, pinned

`findJsonCall()` does two passes: first over `Return_` statements only (first match wins), then the previous any-statement pass as fallback — so `$x = response()->json(['first']); return response()->json(['second'], 201)` documents `{second}` @ 201, while the assignment-only idiom (`$response = response()->json([...]); return $response;`) keeps working. Documented in the class docblock and `docs/auto-derivation.md`.

### 6. NIT — status sanity: subsumed + 204 handled

`999` falls to finding 1's 2xx guard (pinned). `json([...], 204)` now documents as **`204 No Content` without a body schema** — Symfony's `Response::prepare()` strips the body at runtime, so documenting the literal would be the unfaithful choice (mirrors the destroy-convention's 204 handling). Pinned through serialization (no `content` key).

Docs (`auto-derivation.md` boundaries + `plugin-authoring.md` marker note) and the CHANGELOG `#14` entry updated for everything observable.

Radiergummi:
## Re-review: `509e5db` (fix commit for the six findings)

Gates on `509e5db`: `composer test` ✅ (2124 passed, 5246 assertions), `composer lint` ✅, `vendor/bin/pint --test` ✅. PR CI fully green, including the swagger-php 5.8 job and the byte-exact snapshot job.

**Verdict: APPROVE** — all six findings verified fixed empirically (15 fresh probes + full corpus re-run), no regressions, no new findings. Ready for Moritz's merge review.

### Per-finding verification

**1. 2xx-only primary — fixed.** Re-probed the InvoiceNinja shape (guarded success in an `if` + terminal `json(['message' => 'Unauthorized'], 403)`): the operation keeps its bare 200, no 403 appears, NOTICE emitted naming the action and the status. `999` and `302` literals also claim nothing (NOTICE each); the 100–599 question from finding 6 is subsumed. Corpus-confirmed: all **12 regressed InvoiceNinja ops recovered** (table below), and the refusal NOTICEs show up in the app log for exactly the right controllers (`StripeController::verify`, `PostMarkController::webhook`, `SystemLogController::*`, …) — #238's acceptance corpus is real and well-scoped (proper context/spec/tier, labelled `enhancement, spec, tier-1, area:responses`).

**2. Authoring guard — fixed, and the shape is right.**
- *Behavior:* re-ran the `#[ResponseResource]` + literal-envelope probe → documents `{data: $ref ProbeR2ArticleResource}`, not `{data: {}}`. Pinned end-to-end in the PR's tests plus silent-refusal unit tests for both attributes.
- *Placement:* `Contracts\Attributes\PrimaryResponseAuthoringAttribute` is consistent with the house rule (cross-plugin seams live in `Contracts\`; this is an attribute-side seam, so a new `Contracts\Attributes\` sub-namespace is the right parallel to `Contracts\Registry\`). It is documented extension surface (plugin-authoring.md tells plugin authors to implement it), so no `@internal` tag is correct.
- *Completeness:* checked `src/Attributes/` + all plugin attribute namespaces exhaustively against the six `PrimaryResponseResolver`s. `#[ResponseResource]` (consumed by ApiResources' `ResourceClassLocator` *and* Core's `PaginatorResponseResolver`) and `#[FractalResponse]` are indeed the only action-level attributes a later primary resolver consumes. SpatieData and the paginator/model resolvers are return-type-driven (the return-type guard covers them); `#[ResponseField]`/`#[ResourceField]` are payload-class field attributes, `#[ResponseExample]`/`#[ResponseHeader]` decorate already-resolved responses in `OperationBuilder` — composing with a scan-claimed 200 is correct for those. No `#[ResponseFile]` exists. Complete.
- *Not marking `#[Response]`:* sound. A lone `#[Response(404)]` must not suppress the inferred 200, and the 2xx-`#[Response]`-wins path is still pinned (`prefers an explicit #[Response] attribute over the inferred body`).
- *Silent step-aside:* consistent — same convention as the silent return-type guard; deferring to explicit authoring is correct behavior, not degradation. Both pinned with empty-log assertions.

**3. Chain guard — fixed.** `->setStatusCode(201)` degrades with a NOTICE naming the method (no body under 200, no 201); `->header('X','y')` stays matched. Whitelist contents assessed against `JsonResponse`'s full public fluent surface: the four whitelisted methods are genuinely status/body-preserving; everything else degrades. Only arguably-harmless omissions are `withoutCookie`/`withoutHeader` (and cache-header decoration like `setEtag`) — degrading on them is safe-conservative and they didn't occur in the corpus, fine as-is. `->withCallback()` (JSONP, body-wrapping) correctly degrades — probed. The whitelist-over-blacklist inversion is the right call per degrade-never-guess.

**4. Integer-keyed lists — fixed.** `json([0=>'a',1=>'b'])`, `json(['a','b'])`, and the class-constant path all document `{type: array, items: {type: string}}`. Pinned semantics at the edges are correct, too: non-sequential int keys (`[0=>'a',2=>'b']`) document as an object with `"0"`/`"2"` properties — exactly `json_encode` parity — and a list+map mix still degrades.

**5. Returned-call preference — fixed and sanely pinned.** Assignment-then-return documents the returned call (201/`{second}`). If/else two-conditional-returns degrades (conditional-only NOTICE path). Two sequential top-level returns: first wins — dead code, faithful to runtime, fine.

**6. 204 — fixed.** `json([...], 204)` documents `'204': {description: 'No Content'}` with no `content:` key in the serialized YAML (verified through full document serialization on the 6.x worktree; the 5.8 CI job is green for the serialization path).

Docs delta (auto-derivation boundaries, plugin-authoring marker note) and the CHANGELOG entry accurately describe all of the above. Vanilla example/snapshot untouched, as claimed.

### Survey (Layer A, full corpus re-run at `dev-509e5db`)

State verified per #233 before and after (all 11 vendor symlinks → the PR worktree; per-app `composer.lock` ground truth `dev-509e5db…` after a pre-run composer-state reset; manifest `libraryCommit` matches). #221 hit again — run dropped after Pelican; SpeedtestTracker + Vito backfilled via `--only` and merged manually per #229.

Three-way (`results-pr-230.json` @ #230 head → `results-pr-235.json` @ `a1bdde1` → `results-pr-235-r2.json` @ `509e5db`):

| app | respSchemas | completeness % | reqBodies | genExit |
|---|---|---|---|---|
| AdvisingApp | 31 → 31 → 31 | 96.8 → 96.8 → 96.8 | 15 → 15 → 15 | 0→0→0 |
| AureusERP | 90 → 90 → 90 | 91.1 → 91.1 → 91.1 | 37 → 37 → 37 | 0→0→0 |
| Bagisto | 15 → 13 → 13 | 34.1 → 29.3 → 29.3 | 12 → 12 → 12 | 0→0→0 |
| BookStack | 86 → 86 → 86 | 94.2 → 94.2 → 94.2 | 19 → 19 → 19 | 0→0→0 |
| Coolify | 154 → 154 → 154 | 55.8 → 55.8 → 55.8 | 1 → 1 → 1 | 0→0→0 |
| **InvoiceNinja** | **522 → 507 → 519** | **88.9 → 87.5 → 88.3** | 255 → 255 → 255 | 0→0→0 |
| Koel | 170 → 170 → 170 | 82.4 → 82.4 → 82.4 | 54 → 54 → 54 | 0→0→0 |
| Lychee | 202 → 202 → 202 | 97.6 → 97.6 → 97.6 | 92 → 92 → 92 | 0→0→0 |
| Pelican | 158 → 158 → 158 | 96.2 → 96.2 → 96.2 | 75 → 75 → 75 | 0→0→0 |
| SpeedtestTracker | 8 → 8 → 8 | 100.0 → 100.0 → 100.0 | 1 → 1 → 1 | 0→0→0 |
| Vito | 144 → 144 → 144 | 87.5 → 87.5 → 87.5 | 41 → 41 → 41 | 0→0→0 |

- **Recovery:** InvoiceNinja regains exactly the 12 finding-1 ops — per-op spec diff `a1bdde1` → `509e5db` shows precisely 12 status-set changes, every one a non-2xx eviction reversed (`403/400/401/422-eviction/500 → bare 200 back, error status gone`), zero collateral. The remaining 522−519 = 3 (and Bagisto's 15→13) are the round-1-adjudicated metric artifact: ops whose new single-key `{data: {}}` partial schema the substantive test doesn't credit — the spec is strictly richer than the #230 baseline there, the headline number just can't see it.
- **Gains hold:** per-op spec diff for the other 10 apps shows **0 gained / 0 lost / 0 status-set changes** between `a1bdde1` and `509e5db` — the full +116 substantive-2xx lift survives the narrowing untouched (no app outside InvoiceNinja had a straight-line non-2xx literal, status-mutating chain, or 204 literal in scope).
- **Lost-op audit:** 0 operations lost a substantive 2xx schema corpus-wide.
- **Spot-checks:** `POST /api/v1/stripe/verify` and `POST /api/v1/postmark_webhook` → `{200 (bare), 429}`, 403 gone ✓; AureusERP `POST …/tags/{id}/restore` keeps `{message: string}` @ 200 ✓; Vito `GET /api/health` keeps `{success: boolean, version: {}}` ✓ (the op #14 was filed about).
- No new crashes; requestBodies and error responses flat everywhere.

Artifacts: `results-pr-235-r2.json` + archived per-app specs for both heads in the survey workspace.

Radiergummi:
## Addendum: whole-document lost-op audit (beyond the API prefixes)

The verdict above stands (**APPROVE**). Completing the lost-op audit at *whole-document* scope (the table and per-op diff above are API-prefix-scoped) surfaced three data points worth recording:

**1. Three more recoveries outside the API prefixes** — all the same correct narrowing as InvoiceNinja's 12: Bagisto `POST /admin/tinymce/upload` (`['400'] → ['200']`) and `POST /install/api/run-seeder` (`['500'] → ['200']`), Coolify `POST /forgot-password` (`['400','429'] → ['200','429']`). Corpus-total recoveries: **15**, zero collateral.

**2. Exactly one op corpus-wide lost a substantive 2xx schema — and the loss is correct.** InvoiceNinja `GET /webcron` (`a1bdde1`: `200` with `{message: string}` → `509e5db`: bare `200`). The lost schema was *wrong documentation*: the action is named `index`, so at `a1bdde1` the straight-line `return response()->json(['message' => 'Invalid secret'], 403)` was matched, and `OperationBuilder`'s convention-status layer (`applyConventionStatus`) then **rekeyed the 403 error body under `200 OK`**. The 2xx guard now refuses the literal before the convention layer sees it (refusal NOTICE confirmed in the app log for `WebCronController::index`), leaving the honest bare 200.

**3. The rekey path itself has a residual, pre-existing mislabel** — now bounded to 2xx↔2xx by this PR's guard, but still real: a scan-read literal `json([...], 200)` in a `store()` action documents as `201 Created`, and `json([...], 201)` in an `index()` documents as `200 OK` (both probed at `509e5db`). The literal the author wrote is Tier-1 evidence and shouldn't lose silently to a convention default. Filed as #240 (also flagged there for #238's design: error-routed literals must bypass this rekey).

Nothing here changes the merge readiness — findings 1–3 above are strictly *more* evidence the 2xx guard was the right fix.

Radiergummi:
Folded the #240 fix into this PR (it's a bug in this feature's interaction with the resource-action convention #25, not reachable on `main`).

**#240** — the resource convention silently relabelled a 2xx status the body scan read from the author's `response()->json([...], <status>)` literal (`store` 200→201, `index` 201→200). The resolver now marks a response whose status was read **explicitly** (status argument present, not the helper's 200 default) via a transient vendor-extension marker on `OA\Response`; `OperationBuilder` reads and **strips** that marker before the response is emitted and defers the convention when it's set. An *absent* status stays a default and is still subject to the convention, so a conventional `store` with no explicit status is still promoted to 201.

Tests: new `OperationBuilderConventionStatusTest` (store literal-200 / index literal-201 beat the convention; default-status store still → 201; marker doesn't leak into the serialized response) — the first two fail on the pre-fix head. CHANGELOG (#14 entry) + `docs/auto-derivation.md` precedence notes updated. Full suite green, Pint + PHPStan clean.

Radiergummi:
Independent stack review (holistic pass over the full epic delta) — findings attributable to this PR:

**1. [important] A `json()` literal wrapped in an enclosing call is documented as the primary response.**
`findJsonCall()` (`src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php:163-183`) matches the first `response()->json(...)` node *anywhere inside* a scanned statement; the chain guard (`chainPreservesResponse`) covers only fluent method chains hanging off the call. Probe: `return $this->decorate(response()->json(['leak' => true]));` documents `200 → {leak: boolean}` — but `decorate()` (or `tap()`, an error-wrapper helper, a macro) may change status or body arbitrarily. This is exactly the transformation risk the `->setStatusCode()` chain guard exists to refuse, reached through an argument position instead of a chain — and it's asymmetric with #253's `ReturnExpressionResourceReader`, which refuses any unrecognised enclosing shape with a NOTICE. Not covered by #236/#238/#240. Suggested fix: require the matched call (after whitelisted chain links) to *be* the return expression / statement expression, mirroring #253.

**2. [minor — shared with #261] The "Tier-0 resolvers stay authoritative regardless of chain order" docblock claim is false for convention-based later resolvers.**
The return-type guard protects *signature*-keyed resolvers only. Fractal's `$entity_transformer` convention (#261) targets **untyped** methods — exactly where this scan also fires. An untyped method that *assigns* a straight-line `response()->json([literal])` to a variable (never returning it — `findJsonCall` falls back to non-returned matches) and then `return $this->itemResponse(...)` gets the json literal as primary; the Fractal resolver is never consulted. Narrow in practice, but the comment encodes a safety property the code doesn't have. Weaken the comment or drop the non-returned-match fallback.

**3. [minor] Untested guard — survived a mutation test.** Deleting the `$receiver->getArgs() !== []` check in `isResponseHelperCall` (`InlineJsonResponseResolver.php:215`) — the line that rejects `response('content')->json()`, where the receiver is a `Response`, not the factory — leaves all 37 relevant tests green; no test or fixture anywhere exercises that shape. One unit test with a `response('x')->json([...])` fixture closes it.

(A related finding — path-dependent semantics between this resolver's AST literal path and the #230 class-constant path — is posted on PR #230.)

Radiergummi:
Fix landed for the stack-review items-less-array family (#265): `SchemaFromArrayDefinition::apply` now guarantees an `items` object on every array-typed node — one guard at the definition→`OA\Schema` chokepoint covers `'tags' => []`, heterogeneous lists, and every future literal path on both swagger-php majors. Unit-pinned in the new `SchemaFromArrayDefinitionTest`. Commit `850d1c6`.

Issue #224, opened 2026-06-10
RequestFieldRequestSchemaResolver: same-short-name controllers silently share one request-body component


Radiergummi:
## Summary

`RequestFieldRequestSchemaResolver::componentKey()` (`src/Plugins/Core/Resolvers/RequestFieldRequestSchemaResolver.php:78-81`) keys request-body components as `{ControllerShortName}{Method}Request` and registers them via `ComponentSchemaRegistry::registerNamed()`, which is first-wins idempotent. Two controllers with the same short name in different namespaces (e.g. `Admin\…\CartController::storeCoupon` and `Shop\…\CartController::storeCoupon`, both carrying `#[RequestField]` attributes) therefore collide: the second controller's schema is silently discarded and its operation `$ref`s the first controller's component — a wrong schema with no warning when their fields diverge.

## Context

Found during the #219 re-review (PR comment findings, item 4). The inline-validation resolver introduced in #219 had the identical pattern and was fixed there by routing the key through the registry's existing collision machinery: reserve a synthetic class-like name (`{ControllerNamespace}\{ControllerShortName}{Method}Request`) via `ComponentSchemaRegistry::reserveKey()`, which disambiguates with namespace segments on collision (see `InlineValidationRequestSchemaResolver::componentKey()` after #219). The `RequestField` path was deliberately left untouched there (surgical-changes rule) — this issue tracks applying the same fix.

## Proposed fix

Mirror the #219 approach in `RequestFieldRequestSchemaResolver::componentKey()`: derive the key through `reserveKey()` with the controller-namespace-qualified synthetic name, and add a feature test with two same-short-name controllers carrying differing `#[RequestField]` sets. While there, consider the `__invoke` method-segment normalisation (`AgentController__invokeRequest` → `AgentControllerRequest`) for parity.

## Acceptance criteria

- Two same-short-name controllers with differing `#[RequestField]` declarations produce two distinct components, each `$ref`ed by its own operation.
- Existing non-colliding keys are byte-stable.

Issue #222, opened 2026-06-10
tests: parallel-run race in TestCase::setUp fixture copy — rare 'Failed to create fixture destination' flake


Radiergummi:
## Summary

One flaky failure in `composer test` (Pest `--parallel`, 12 workers) on a cold cache: a worker threw `RuntimeException: Failed to create fixture destination: …/tests/Fixtures/OpenApi/example_payloads` from `tests/TestCase.php::setUp()` (throw at line 71, `mkdir` at line 68). Subsequent runs — serial, parallel-warm, and parallel with `.cache/pest` cleared — all passed (1999/1999), so this is a parallel-setup race, not a deterministic break.

Observed on `feat/9-inline-validate-rules` @ 51fc5e9 during the PR #219 review, but `tests/TestCase.php` is untouched by that PR (last change: 6279390) — pre-existing.

## Analysis

Every test's `setUp()` does:

```php
if (
    !is_dir($destination)
    && !mkdir($destination, 0o777, true)
    && !is_dir($destination)
) {
    throw new RuntimeException(...);
}
```

With 12 workers racing the same path, the `!is_dir / mkdir / !is_dir` guard can still throw: the recursive `mkdir` can fail when a sibling wins the race on an intermediate segment, and the re-check can be defeated by PHP's per-process stat cache returning the stale pre-`mkdir` result. The subsequent unsynchronized `copy()` loop over the same target files is also racy (harmless on APFS today, but worth noting).

## Fix sketch

`@mkdir($destination, 0o777, true); clearstatcache(true, $destination); if (!is_dir($destination)) { throw … }` — or copy fixtures once per process (e.g. guard with a static flag / `beforeAll`) instead of in every `setUp()`.

Radiergummi:
Second symptom of the same race, observed during the PR #241 review (branch `feat/11-request-accessor-query-params` @ 5b15aa6, `tests/TestCase.php` untouched): roughly 1 in 15 parallel runs fails with

```
FAILED  Tests\Feature\ExampleFileLoaderTest > it OAPI-0…  RuntimeException
OpenAPI example file contains invalid JSON (…/laravel/tests/Fixtures/OpenApi/example_payloads/create_project.json): Syntax error
```

[...]

PR #241, opened 2026-06-10
feat(#11): query parameters from $request->*() accessor reads + GET inline-validate hand-off

### LINKED ISSUE #9: Inline `validate()` + `$rules` property extraction

### LINKED ISSUE #11: Query params from `$request->*()` accessors

Radiergummi:
Closes #11.

**Stack:** #219 (`feat/9-inline-validate-rules`) ← #225 (`feat/10-abort-error-responses`) ← #230 (`feat/227-class-constant-statuses`) ← #235 (`feat/14-response-json-literal`) ← **this PR**. Retarget/rebase bottom-up after each merge; this PR must be retargeted to its new base (or `main`) once #235 lands.

## Plan

Two halves, one resolver:

### 1. Accessor scan (Tier-1 whitelist)

New `Plugins\Core\Support\RequestQueryAccessorReader` (`@internal`, `#[Scoped]`), built on the shared `Support\MethodBody` primitives (`MethodBodyScanner::firstStatements`, limit 10; `StatementNodeFinder::findAll`; `AstLiteralEvaluator`):

- Whitelist (spec's five, nothing more): `query`/`input` → `string`, `string` → `string`, `integer` → `integer`, `boolean` → `boolean`.
- **Receiver discipline:** the call's receiver must be the method's `Illuminate\Http\Request`(-subclass)-typed parameter variable, or a zero-argument `request()` helper call (the Bagisto precedent, same as #9's shape 2). An Eloquent builder's `->where()`, a `$builder->query('…')`, or a same-named variable in a request-less method never match.
- **Name:** first string-literal argument — positional or named (`key:`), per the #234 named-args precedent. A non-literal name degrades that call with a generation-log NOTICE (found-but-unreadable, per #9/#10 convention).
- **Conditional policy: `IncludeConditionalContexts`.** Unlike #9's request-body case there is no guess risk: a read is a read — `$request->boolean('active')` inside an `if` (or a `->when(…)` closure) still proves the parameter is consumed; and the strict policy is blind to `try {}` bodies (per the #235 review), which would hide the pervasive try-wrapped controller. The scan claims nothing beyond "this parameter exists and has this type".
- **Defaults:** the accessor's second argument (positional or `default:`), when literal *and* type-matching the inferred type, becomes the schema `default` — it falls out free via `AstLiteralEvaluator`; a non-literal or type-mismatched default is simply omitted (no notice — the parameter itself is still documented).

### 2. GET inline-validate hand-off (closes #9's deliberate gap)

On **GET/HEAD** routes, `CoreQueryParameterResolver` calls the verb-agnostic `InlineValidatorRulesReader::read()` (built by #9 for exactly this) and maps recovered rules through `ValidationRulesToSchema::process()` into parameters — name, schema (type/format/constraints/enum), `required` from the rules, trailing-comment descriptions. The body side already suppresses GET (pinned in #219), so nothing is documented twice.

**Nested/array keys** map to the PHP/Laravel wire format (the same bracket notation the QueryBuilder plugin already emits):
- `filter.name` → parameter `filter[name]` (recursive for deeper nesting); required only when the leaf *and* all ancestors are required.
- `ids` + `ids.*` (scalar items) → parameter `ids[]` with an array schema.
- Arrays **of objects** (`items.*.price`) and the bare `*` (additionalProperties) rule cannot be expressed honestly as a parameter name → those keys are dropped with a generation-log NOTICE.

### 3. Merging & precedence (in `CoreQueryParameterResolver`)

Layered name-keyed map, later layers replace same names, different names compose:
1. accessor-scan parameters (typed accessors `string()`/`integer()`/`boolean()` beat untyped `query()`/`input()` for the same name; first wins among equals),
2. GET-validate parameters (win over accessor reads on the same name — rules know required/constraints),
3. `#[QueryParam]` attributes (win **entirely**, class-level then method-level as today — no duplication, no partial merge).

This is deliberately *not* the response-side exclusive-claim model (`PrimaryResponseAuthoringAttribute`): query parameters from different sources compose; only same-name collisions resolve by precedence.

### Verb discipline (decision, see PR comment)

`query()` matches on every verb — it can only ever read the query string. `input()`/`string()`/`integer()`/`boolean()` read the *merged* body+query input, so they match only on GET/HEAD routes; on body-carrying verbs they overwhelmingly mean body fields (which #9 already documents) and emitting them as query parameters would be a guess.

## Tests

- `tests/Unit/Core/Resolvers/CoreQueryParameterResolverTest.php` + `tests/Fixtures/QueryAccessorFixtureController.php`: each accessor shape, named args, literal defaults, non-literal name (NOTICE), conditional read (matches, per policy), >10-statement miss, receiver discipline (impostor `$builder->query()`, request-less `$request` variable), `request()` helper receiver, dedup (typed beats untyped), `#[QueryParam]` same-name wins / different-name composes, GET-validate hand-off (required, constraints, nested `filter.name`, `ids.*`, array-of-objects degrade, dynamic-rules NOTICE), verb gating.
- Feature test (mirrors `InlineValidationSchemaTest`): end-to-end spec assertions for both halves.
- Examples: vanilla `BookingController::index` gains idiomatic accessor reads; snapshots regenerated; diff verified to be exactly this feature.

## Docs & changelog

- `docs/auto-derivation.md`: new "Query parameters" table row + a dedicated section (mirroring the Inline JSON responses section).
- `docs/request-bodies.md`: the GET bullet ("routing into query parameters is planned") flips to shipped, with a link.
- `CHANGELOG.md` under `[Unreleased]`.

## Gates

`composer test`, `composer lint` (PHPStan L8), `vendor/bin/pint --test` — green before ready.


Radiergummi:
## Decisions (as built, with reasoning)

**(a) Conditional policy — `IncludeConditionalContexts`.** The accessor scan descends into `if` branches, ternary/match arms, short-circuits, and closures. Rationale: unlike the request-body scan (#9) and the response scan (#14), an accessor read makes no claim that needs straight-line provenance — it asserts only "this parameter is consumed, with this type", and a `$request->boolean('compact')` inside an `if` (or the pervasive `->when($request->query('seat'), …)` builder idiom — see the vanilla example) consumes the parameter exactly as much as a top-level read. The strict policy is also blind to `try {}` bodies (per the #235 review finding), which would silently hide try-wrapped controllers — an unacceptable miss rate for a claim this weak. Same reasoning family as #10's inclusive abort scan. Pinned by `conditionalRead` (unit + fixture).

**(b) Defaults — taken, because they fell out free.** The accessor's second argument (positional or `default:`) goes through the existing `AstLiteralEvaluator` (one try/catch); a literal default becomes the parameter schema's `default` **only when its PHP type matches the accessor's inferred type** — `integer('per_page', 25)` carries `25`, but `query('page', 1)` does *not* pin an integer default onto a string parameter (a type-mismatched schema would be self-contradictory; the read still documents the parameter itself). `null`, non-literal, or mismatched defaults are silently omitted — no notice, nothing was lost worth telling. Pinned by `withDefaults`.

**(c) Named arguments — handled from day one** (`query(key: 'sort')`, `integer(default: 10, key: 'limit')`), per the #234 precedent; one shared `argument(position, name)` helper resolves positional-or-named for both the key and the default. Pinned by `namedArguments`.

**(d) Whitelist — exactly the spec's five.** `get()`, `has()`, `filled()`, `date()`, `enum()`, `float()` stay off, stated in the reader's whitelist docblock and in the docs boundary list. No additions felt forced during implementation; `enum()`/`date()` would be the first candidates if the corpus demands them (they carry real type information), but that's a future issue, not a silent widening.

**(e) Verb discipline (not in the spec — flagging explicitly).** `query()` matches on **every** verb — it can only ever read the query string. `input()` / `string()` / `integer()` / `boolean()` read the *merged* body+query input, so they match only on **GET/HEAD** routes: on POST/PUT/PATCH those reads overwhelmingly mean body fields (think `$request->boolean('notify')` in a `store()`), which the #9 scan already documents — emitting them as query parameters would be a guess, and a visibly wrong one (`title` as both body property and query param). On body verbs the input-family reads are simply not matched (no notice — they're not failed query-param reads, they're body reads). Pinned by the POST/HEAD unit tests and the e2e POST test.

**(f) Nested/array rule keys → PHP wire notation** (the same convention the QueryBuilder plugin already emits for `filter[…]`):
- `filter.name` → parameter `filter[name]` (recursive for deeper nesting); `required` only when the leaf **and** every ancestor are required — OpenAPI can't express "required iff parent present", so anything weaker degrades to `required: false`.
- `ids` + `ids.*` (scalar items) → parameter **`ids[]`** with an array schema; with default `style: form, explode: true` semantics that is byte-for-byte the PHP wire format (`ids[]=1&ids[]=2`).
- Arrays **of objects** (`rows.*.price`) and the bare `*` (additionalProperties) rule have no honest parameter-name representation — `deepObject` is only defined one level deep and nothing covers indexed object arrays — so those keys are **dropped with a NOTICE** naming them. Dotted *accessor* keys get the same bracket conversion (`input('filter.name')` → `filter[name]`); a wildcard-containing accessor key counts as unreadable (NOTICE).

**(g) Merge/precedence — layered name-keyed map, deliberately *not* the response-side exclusive-claim model.** In `CoreQueryParameterResolver`, ascending precedence with later layers replacing same names and different names composing: accessor reads (typed accessor beats untyped for the same name, first wins among equals) < GET-validate rules (they know `required` and constraints) < `#[QueryParam]` attributes (win **entirely** — no partial merge of inferred type under an authored description, no duplication; class-level then method-level as before). No `PrimaryResponseAuthoringAttribute`-style suppress-marker: a `#[QueryParam]` documents *one name*, not the operation's whole parameter list. Note `OperationBuilder` concatenates across resolvers without dedup — all three sources live in this one resolver, so the map covers every collision class the spec names; cross-plugin collisions (e.g. QueryBuilder's `filter[x]`) were already possible before this PR and are unchanged.

**(h) Receiver discipline** — same as #9's shape 2: the method's `Illuminate\Http\Request`(-subclass)-typed parameter variable, or a zero-argument `request()` helper call. An impostor object with `query()`/`input()`/`integer()` methods, and a `$request`-named local in a request-less method, are pinned as non-matches.

**(i) Scan gating / perf.** Unlike the body/response scans there is no Tier-0 signal to gate on (a signature never reveals accessor reads), so the accessor scan runs for every controller-method route. The per-file AST cache in `MethodBodyScanner` means each controller file is still parsed at most once per run, and most files are already parsed by the #9/#10/#14 scans.

**Suite/corpus expectations:** the full suite (2151 tests incl. the byte-exact snapshot group) was green *before* the examples change — i.e. no existing snapshot moved implicitly; the vanilla `FlightController::index` (`#[QueryParam('page')]` + `$request->query('page', 1)`) is a live precedence proof, its output is byte-identical. The only snapshot diff is the deliberate showcase: `GET /flights/{flight}/bookings` gains `seat` (string) and `limit` (integer, default 50) from accessor reads. On the dogfood corpus, expect parameter counts to rise on list/index endpoints; `parameter`-related lint movement should be additive only.

Radiergummi:
## Review: empirical pass on the #241 delta (`feat/14-response-json-literal..feat/11-request-accessor-query-params`)

Gates, run locally on a detached worktree @ 5b15aa6: `composer test` 2151 passed; `composer lint` (PHPStan L8) clean; `vendor/bin/pint --test` clean. One non-deterministic failure surfaced across repeated runs (~1/15): the `ExampleFileLoaderTest` invalid-JSON read of `create_project.json` — the unsynchronized fixture-`copy()` race already analysed in #222, untouched by this PR (new symptom noted there).

### Findings

1. **[minor] Multi-verb routes leak the GET-gated inferences onto their POST twin.** `ActionDescriptor::httpMethod` reads `route->methods[0]`, and `PathsStage` fans one built operation out per verb — so on a `Route::match(['get', 'post'])` route every verb gate in this PR evaluates GET for both operations. Live on the corpus: Coolify's `POST /api/v1/applications/{uuid}/start` now documents `force` (boolean, default `false`) and `instant_deploy` as query parameters sourced from `$request->boolean(…)` reads — exactly what decision (e) says must not happen on a body verb. The mirror defect is pre-existing (#9's `WRITE_METHODS` gate means the same route's POST op never documents its inline-validate body), and the root cause sits outside this diff, so: filed as #245, not blocking here.
2. **[minor] Receiver discipline: a closure that re-declares the receiver name leaks its reads.** Probe: a method with `Request $request` containing `collect([])->each(function (Request $request): void { $request->query('inner'); })` documents **`inner`** as a query parameter of the route — the inner `$request` is a different object, matched purely by variable name. This deviates from decision (h) as stated ("the method's Request-typed parameter variable"). A bounded, dataflow-free fix exists: when descending into a `Closure`/`ArrowFunction` whose own parameter list re-declares the receiver name, skip that subtree. (The twin probe — method without a Request parameter, closure declaring one — correctly matches nothing.) Rare idiom; fix or pin as a known boundary, your call.
3. **[nit] The non-literal-name NOTICE ignores the verb gate.** Probe: a POST action with `$request->integer($this->dynamicKey())` logs "Request accessor read(s) … have a non-literal parameter name; those query parameters are not documented" — but per decision (e), on POST that read is a body read, not a failed query parameter. The unreadable-accessor notice should be filtered by the same verb discipline as the reads (`query()` unreadables always notice; input-family only on GET/HEAD).
4. **[nit] DELETE falls between the chairs, silently.** Inline `validate()` on a DELETE route is documented nowhere — the body scan gates on POST/PUT/PATCH, the hand-off on GET/HEAD — and no generation-log note fires (probed e2e: no body, no parameters, empty log). Defensible (DELETE may carry either), but it deserves either a docs boundary line or a decision; today it just vanishes.
5. **[info] #243 confirmed empirically.** A second `QueryParameterResolver` emitting `sort` next to Core's inferred `sort` puts two `(sort, in: query)` entries into the document — invalid OpenAPI. With stock plugins the corpus shows **zero** duplicate `(name, in)` pairs across all 11 full specs, so it stays theoretical until QueryBuilder-style resolvers meet inferred reads; rightly parked in #243.

### Verified (probes, removed after)

All five accessor shapes with correct types; named args (`key:`, `default:`, reordered); type-matched defaults incl. `boolean('archived', false)` serializing as `default: false` (and `query('page', 1)` correctly unpinned); Request-**subclass** receiver matches; impostor `$builder->query()`, `Facade::query()` static call, `$request = new Lookalike()` local, and `request('bag')->query()` all correctly refused; zero-arg `request()` helper matches; reads inside `if`/arrow-fn closure/`try {}` all match per the conditional policy; >10-statement miss; dotted keys → `filter[name]`; GET-validate hand-off with `required`, constraints, descriptions, `ids[]` array schema; bare `*` and `rows.*.price` dropped with the NOTICE; GET request body stays suppressed; POST keeps the validate body with no query leak; precedence (attribute wins entirely, validate beats accessor, typed beats untyped); `FlightController::index` byte-identical (examples diff is exactly `seat`+`limit` on the bookings op). The schema surface (`type`/`default`/`enum`/`items`) is all long-modeled in swagger-php 5.x — 5.8-safe.

**Verb discipline — position: agree.** On GET/HEAD, `input()` and the typed accessors literally read the query bag (`getInputSource()` *is* the query bag on those verbs), so the GET-side claim is exact, not a heuristic; on POST they predominantly read body fields that #9 already documents, and emitting them as query params would double-document and often be wrong. The conservative under-claim (a POST endpoint paginating via the query string) is the right default — that's `#[QueryParam]`'s job. The one wrinkle is finding 1: the gate is only as good as `httpMethod`, which lies on multi-verb routes.

### Survey (Layer A, all 11 apps)

Baseline `results-pr-235-r2.json` (#235 head 509e5db) → `results-pr-241.json` (5b15aa6). State verified per #233 before and after: every app's vendor symlink → the review worktree, `composer.lock` at `dev-5b15aa6…`, repaired composer state. #221 reproduced (loop died after Pelican); SpeedtestTracker + Vito backfilled `--only` and the 11-app aggregate reassembled per-app via `metrics.php` (#229).

| App | apiOps | query params b→p | ops w/ query b→p | reqBodies | respSchemas | completeness | lint | genExit |
|---|---|---|---|---|---|---|---|---|
| AdvisingApp | 31 | 83 → 83 | 11 → 11 | = | = | = | = | 0 |
| AureusERP | 90 | 0 → 0 | 0 → 0 | = | = | = | = | 0 |
| Bagisto | 41 | 0 → 1 | 0 → 1 | = | = | = | = | 0 |
| BookStack | 86 | 0 → 5 | 0 → 3 | = | = | = | = | 0 |
| Coolify | 154 | 0 → **30** | 0 → 17 | = | = | = | = | 0 |
| InvoiceNinja | 522 | 0 → **17** | 0 → 15 | = | = | = | = | 0 |
| Koel | 170 | 0 → 5 | 0 → 5 | = | = | = | = | 0 |
| Lychee | 205 | 0 → 4 | 0 → 1 | = | = | = | = | 0 |
| Pelican | 158 | 0 → **16** | 0 → 15 | = | = | = | = | 0 |
| SpeedtestTracker | 8 | 0 → 0 | 0 → 0 | = | = | = | = | 0 |
| Vito | 144 | 2 → 8 | 2 → 8 | = | = | = | = | 0 |

(`=` — byte-identical to baseline: requestBodies, responseSchemas, completenessPercent, lint totals all unmoved on every app.) **+84 query parameters across 65 newly-parameterized API operations**, everything else stable, all generate exits 0, and **zero duplicate `(name, in)` pairs in any full spec** — validity holds corpus-wide.

**Prediction vs actual:** grep-first predicted top-3 gainers InvoiceNinja / Pelican / Koel — actual top-3 **Coolify (+30)** / InvoiceNinja (+17) / Pelican (+16). 2 of 3; Coolify's 49 typed-accessor reads sit squarely in GET API actions and made it the runaway winner, while Koel's reads mostly live in POST handlers (+5 only).

**Spot-checks (3):** Coolify `GET /api/v1/deploy` — all six params match the `input()`/`string()` reads, the non-literal `input('pull_request_id', $request->input('pr'))` default correctly omitted *and* the nested `input('pr')` read correctly documented as its own parameter. Pelican `GET /api/client` — `query('per_page', '50')` carries default `"50"`, the `query('per_page') ?? 50` idiom elsewhere correctly carries none, and the zero-arg `->appends($request->query())` bag read is ignored. Coolify `GET /api/v1/deployments/applications/{uuid}` — the GET-validate hand-off: `'skip' => ['nullable','integer','min:0']` → `skip` with `type: [integer, null]`, `minimum: 0`, `required: false`. **Negative checks:** Vito's `POST …/ssh-keys` (`input()` reads in the handler) gained no query parameters; all 31 of AdvisingApp's `#[QueryParam]`-annotated API ops have byte-identical parameter lists — no duplication, no override.

**Perf:** full 9-app corpus pass 225 s wall (incl. composer relink churn), backfills 15 s each — within noise of the #235 baseline run (~240 s for 11). The ungated scan's cost is absorbed by the shared per-file AST cache as decision (i) predicted; no regression signal.

### Verdict

**APPROVE.** Findings 1–4 are minor/nits (1 is pre-existing machinery, filed #245; 2 and 3 are small in-PR polish items worth a follow-up commit or an acknowledged boundary; 4 needs a one-line docs decision). The feature does exactly what the spec and the decision log claim, the precedence story holds under collision probes, and the corpus delta is the cleanest kind: large additive gains on exactly the list endpoints the issue motivated, with every stability metric frozen.

Radiergummi:
Review polish landed in ce8c1a8 — per finding:

**Finding 2 (shadowed-closure receiver leak) — fixed as suggested.** `RequestQueryAccessorReader` no longer runs a blanket `NodeFinder` pass; it walks the statement subtrees itself, and a `Closure`/`ArrowFunction` whose **own parameter list** re-declares the receiver variable name nulls the receiver for that subtree. Exact semantics: parameter lists are the only shadow source (closures capture only explicitly — `use` imports the outer variable, it does not shadow — and arrow functions capture implicitly); inside a shadowing subtree the variable receiver never matches but the zero-argument `request()` helper still does; nesting inherits the shadow. Pinned both ways: the leak shape (`->each(function (Request $request) { $request->query('inner'); })` must not match, the outer read before it still does) and the legit captures (`fn() => $request->query('captured')` and `function () use ($request)` must match).

**Finding 3 (verb-gate-aware notice) — fixed.** The unreadable-accessor list is filtered by the same verb discipline as the reads before the NOTICE fires: on a body-carrying verb only `query()` unreadables note (it can only read the query string); a non-literal `integer($key)` on POST is a body read, not an undocumented query parameter. Pinned three ways: POST+`integer($dyn)` → silent, GET+`integer($dyn)` → notice, POST+`query($dyn)` → notice.

**Finding 4 (DELETE hole) — NOTICE added + documented; gates unchanged.** Took your suggestion: found-but-not-documented fits the NOTICE convention, so inline `validate()` on a DELETE route now logs *"Inline validation in X is not documented: a DELETE route may carry the validated fields in either the request body or the query string. Annotate the action with #[QueryParam] or #[RequestBody] to document them."* and emits nothing — the verb gates stay put. Boundary documented in `docs/request-bodies.md` (inline-validation boundaries) and `docs/auto-derivation.md` (query-parameter boundaries), plus the CHANGELOG entries. **No spec issue filed:** below Tier-2 there is no signal to disambiguate body vs query on DELETE — any widening would be a guess, and the NOTICE + attribute hatches are the durable answer. If a corpus idiom ever makes one side dominant, that's the moment to file.

Gates re-run on ce8c1a8: `composer test` 2155 passed (4 new pins), `composer lint` clean, `vendor/bin/pint --test` clean, `--group=snapshot` 9 passed (byte-identical).

Radiergummi:
Independent stack review (holistic pass over the full epic delta, `origin/main...feat/13-fractal-transform-inference`) — findings attributable to this PR:

**1. [blocker] The rebase commit `87ace67` deletes main's `[Unreleased]` CHANGELOG entries.**
`git diff origin/main...HEAD -- CHANGELOG.md` (merge-base IS current main tip `45c3bfb`, so the diff is exact) shows the stack removing the entire `### Documentation` subsection (#123 migration guide, #158 CI guide, the accuracy sweep) plus the #119 Swagger UI, #35 shared-enum-component, #243 query-param dedup, #187/#186 SwaggerPhp harvester, and #224 RequestField-collision entries — and reverting the #24 entry to its pre-#35 wording ("When the shared enum component (#35) lands…", contradicting `docs/auto-derivation.md`, which documents #35 as shipped in this same tree). Per-commit attribution: the `feat(#11)` rebase commit `87ace67` removes 13 lines that no later commit restores — a conflict resolution that took the stack's stale changelog hunk. Squash-merging the stack as-is clobbers main's changelog. Fix: restore main's entries on this branch (every successor PR inherits the fix on retarget).

**2. [important — shared with #219/#235] `docs/troubleshooting.md:76-82` states the opposite of three shipped features.**
"This is the **method-body inference** limitation. The generator reads signatures only. Inline `validate()`, `request()->validate()`, and ad-hoc `response()->json([…])` calls are invisible." All three are now the headline features of this stack (#219, #235, #241). The most likely troubleshooting entry a new user hits tells them the features don't exist. The "Response is a bare 200 OK" section below it likewise predates #235/#253. Eleven PRs each updated auto-derivation.md but none swept troubleshooting.md.

**3. [minor] The three-source merge drops the accessor-observed default.**
`CoreQueryParameterResolver` (precedence accessor < GET-validate < `#[QueryParam]`) replaces wholesale by name. Probe: `$request->integer('per_page', 10)` + `validate(['per_page' => 'integer|max:100'])` emits `{type: integer, maximum: 100}` — the author-written default `10` (real signal from the accessor read) is silently dropped instead of merged into the winning schema. Contrast with the documented "compose, don't discard" framing of this same merge.

**4. [minor — shared with #219] Inconsistent impostor discipline for the global `request()` helper.**
`AbortErrorContributor::helperName` and `InlineJsonResponseResolver::isResponseHelperCall` guard via `namespacedName` + `function_exists` so a same-namespace user function never matches; `InlineValidatorRulesReader::isRequestArgument` (~line 207) and `RequestQueryAccessorReader::isAccessorCall` (~line 226) accept any unresolved `request()` call without that guard. Same question, opposite answers, no docblock claims intent — looks accidental, and it's the request-side scanners (the ones emitting schemas) that have the weaker check.

**5. [nit] The two #241 CHANGELOG entries are the only Added entries without a trailing issue ref** (`(#11)`), unlike every sibling.

Radiergummi:
Fixes landed for the stack-review blocker + docs findings: main's `[Unreleased]` CHANGELOG entries are restored (the two #11 entries re-inserted with their issue refs; the #24 entry carries main's wording again), `docs/troubleshooting.md`'s inline-validate entry now describes the shipped body scans, their degrade conditions, and the escape hatches (plus a body-scan bullet in the bare-200 section), and `docs/auto-derivation.md`'s blanket 'method-body inference is not done' claim is narrowed to the still-true `Model::find()` case (#97). Commit `b6b6b23`. Note the head's CHANGELOG now deletes zero lines vs current main (verified `git diff origin/main...feat/13 -- CHANGELOG.md`).

Issue #98, opened 2026-06-03
ApiResources: derive wrapped-model schema via EloquentModelToSchema


Radiergummi:
## Summary

When an Eloquent API Resource wraps a model (`UserResource::make($user)`, or an action returning `UserResource`), derive a baseline response schema from the **wrapped model's** metadata by reusing the shared `EloquentModelToSchema` reader that #18 lands in `src/Support/Extraction/`.

## Context

Spun out of #18. #18's third source — "a resource wrapping a model" — overlaps the dedicated **ApiResources plugin**, so rather than have Core reach into `JsonResource` territory, #18 places the model→schema reader in `Support` (plugin-agnostic) and the ApiResources plugin consumes it here.

See the design comment on #18: https://github.com/Radiergummi/laravel-openapi/issues/18#issuecomment-4613636060

## Scope

- In the ApiResources plugin's response resolver, when the resolved resource class wraps a known `Model` (via `#[ResponseResource]`, a `@mixin`/generic, or convention), call `EloquentModelToSchema` to produce the wrapped model's component and reference it.
- This is the Tier-0 baseline; `toArray()` body inference (#12) remains the Tier-1 refinement layered on top.

## References

Parent epic: #6. Depends on: #18 (shared reader). Related: #12 (`toArray()` Tier-1 inference), #25 (apiResource action-name detection). Verify: `composer test`.

Radiergummi:
Deferred — folding this into #12 as its base case (decided with @radiergummi, 2026-06-04).

## Why

#12 (Tier-1 `toArray()` parsing) and this issue do not overlap — they **partition** the resource-response space:

- Resource **overrides** `toArray()` with an explicit list → #12 parses the body.
- Resource **does not override** it (`return parent::toArray($request)` passthrough) → there is no body to parse; the only possible source is the wrapped model. That is this issue, and #12 already lists it as its degradation path ("degrade to the model schema").

So body parsing never makes this redundant — but the two cases share one unit (resource→model resolution), and that is better designed **once**, against the full requirement, as part of #12 rather than shipped now as a standalone Tier-0 slice plus a `#[UseResource]` follow-up plus doc churn that #12 would then reconcile. The passthrough slice is also the narrow end of the resource story; the fat part (custom `toArray()`) is #12. Nothing regresses by waiting: #18 already covers direct `return $user` returns, and passthrough resources keep emitting the empty schema + the `resource.fields-undeclared` nudge until #12 lands.

[...]

PR #247, opened 2026-06-10
feat(#12): infer Resource response schemas from the toArray() array literal

### LINKED ISSUE #12: Eloquent API Resource `toArray()` inference
## Summary

Read the array literal an Eloquent API Resource returns from its `toArray()` so that controllers returning a `JsonResource` or `ResourceCollection` get a real response schema, instead of nothing, when they declare no `#[ResourceField]` attributes.

## Motivation

API Resources build their payload imperatively inside `toArray()`, and the generator only reads signatures — never method bodies — so today a resource with no `#[ResourceField]` annotations produces an empty response schema. This is common in the wild: the surveyed apps Bagisto, Koel, and Lychee all return untyped `*Resource` classes, so their responses currently document nothing useful. Provenance: OAPI-064 (Eloquent/Resource response-schema inference) and §3.3 of the body-inference plan.

## Proposed approach

Tier 1: bounded to the case where `toArray()` returns a single array literal — anything more dynamic degrades to Tier 0 (the model schema). Use the shared bounded-AST body-scan utility from the parent epic (see #5) to locate the `return [...]` statement; the keys of that literal become the response properties. Resolve each `$this->field` reference against the wrapped model's schema (delivered by #6, §3.2). When `toArray()` is not a plain array literal, degrade to the model schema (#6) or the envelope only, and emit a generation-log note.

This work lives in the `ApiResources` plugin (`src/Plugins/ApiResources/`). The natural seams are `ResourceResponseResolver.php` (which currently resolves the response for a resource return type) and `SchemaFromResource.php` (which builds the schema); `ResourceClassLocator.php` already resolves the resource class to inspect. An explicit `#[ResourceField]` must always win over inferred keys.

## Acceptance criteria

- A resource whose `toArray()` returns an array literal yields a response schema carrying exactly those keys.
- `$this->field` references inside the literal resolve against the wrapped model's schema.
- A dynamic (non-literal) `toArray()` degrades to the model schema / envelope plus a generation-log note, with no error.
- `#[ResourceField]` continues to override inferred fields.
- Add coverage alongside the existing plugin tests in `tests/Feature/Plugins/ApiResources/` (extend `tests/Feature/Plugins/ApiResources/ApiResourceResponseTest.php`, or add a sibling test there), with before/after fixtures for the literal, `$this->field`, and dynamic-degradation cases; assert the resulting response schema. If a resource example flavor applies, also exercise an end-to-end case in `tests/Feature/ExamplesTest.php`.

## References

Parent: #5 (shared bounded-AST body-scan utility, the Tier-1 boundary, and graceful-degradation contract live there). Depends on: #6 (model→schema, §3.2). Tier 1. Attribute fallback: `#[ResourceField]` (already exists, in `src/Plugins/ApiResources/Attributes/`). Verify: `vendor/bin/pest tests/Feature/Plugins/ApiResources/` (or `composer test` for the full suite).

Radiergummi:
## What & why

Closes #12 (epic #5 — Tier-1 bounded body scans; sized in #106 as the single highest-value child: 72 operations across the surveyed apps, Vito 69, are signature-typed to concrete Resource classes whose `toArray()` is one static `return [...]` literal, currently documented as nothing).

Reads the array literal an Eloquent API Resource returns from `toArray()` so concrete-Resource endpoints get a real response schema without hand-restating every key as `#[ResourceField]` (~264 attributes in the Vito full-spec proof, ~90 in AdvisingApp — the largest single annotation cost both times). Also owns the **passthrough base case** folded in from #98: a resource that does not override `toArray()` documents as its wrapped model's schema.

**Stack:** #219 ← #225 ← #230 ← #235 ← #241 ← this (base: `feat/11-request-accessor-query-params`).

## Scope

Bounded case (Tier-1, per epic #5's contract): `toArray()` is overridden and its body — within the first 10 top-level statements — contains exactly one `return` statement, at top level, whose expression is an array literal. The survey found zero dynamic cases among 24 concrete classes; anything else (early returns, `return $data`, `array_merge(...)`) degrades to today's behaviour plus a generation-log note.

The literal's **string keys** become response properties; each **value** resolves best-effort, refusing per value:

| Value shape | Resolution |
|---|---|
| `$this->field` / `$this->resource->field` | the wrapped model's property schema (casts, `@property`, appends — via the #18/#99 `EloquentModelToSchema` reader) |
| literal scalar / literal array | typed via the shared literal→schema mapping (extracted from #14's resolver) |
| `new OtherResource(...)` / `OtherResource::make(...)` | `$ref` to the nested resource's schema |
| `OtherResource::collection(...)` | `array` of `$ref` items |
| `$this->when($cond, $value)` | `$value` resolved recursively, property **optional** (not in `required`) — condition never analysed |
| `$this->whenLoaded('relation')` | the model's relation property (`$ref` when `@property-read` typed), **optional** |
| `$this->merge([...])` / `$this->mergeWhen($cond, [...])` | literal-array payload keys inlined at top level (optional for `mergeWhen`) |
| `$this->whenCounted('x')` | `integer`, optional |
| anything else (method calls, ternaries, …) | property kept with an unconstrained `{}` schema — dropping a response property is silently wrong (#235 precedent) — listed in one summarising note per resource |

Precedence & fallback:

- `#[ResourceField]` wins **per field**; inferred fields it doesn't cover compose alongside (declared fields first, preserving today's output order, then inferred extras in literal order).
- When neither attributes nor the literal yield any field and a wrapped model resolves (`@mixin` → generic `@extends` precedence, #98 design), the response references the **model component** directly — no empty `*Resource` component. Covers both the passthrough and the unreadable-dynamic case (the latter with a note), exactly as the issue's degradation path describes.
- Envelopes (`data` / `data+links+meta`) unchanged — this fills the item schema only.
- `resource.fields-undeclared` goes quiet when inference or the model fallback produces a non-empty schema (its "the response schema is empty" message would be false).

## File structure

| File | Action |
|---|---|
| `src/Support/MethodBody/SchemaDefinitionFromLiteral.php` | create — literal AST → schema-definition mapping, **extracted** from `InlineJsonResponseResolver` (#14) so both consumers share it |
| `src/Support/MethodBody/SingleReturnArrayLiteralFinder.php` | create — generic "the single `return [...]` literal of a method" primitive (plugin-agnostic: #13 Fractal `transform()` needs the same shape) |
| `src/Support/Extraction/EloquentModelToSchema.php` | modify — expose a memoised per-property lookup (`propertyFor()`) for `$this->field` resolution; no hidden/visible filtering on lookup (the resource controls output visibility) |
| `src/Plugins/ApiResources/Support/WrappedModelLocator.php` | create — resource class → wrapped `Model` class-string (`@mixin` → `@extends Base<Model>`), shared by builder + lint (#98 design) |
| `src/Plugins/ApiResources/Support/ResourceToArrayReader.php` | create — the value-resolution whitelist above |
| `src/Plugins/ApiResources/Support/SchemaFromResource.php` | modify — compose attribute + inferred fields; model-component fallback in `buildRef()` |
| `src/Plugins/ApiResources/Lint/Rules/ResourceFieldsUndeclared.php` | modify — quiet when inference/model fallback applies |
| tests: `tests/Unit/Support/MethodBody/…`, `tests/Unit/Plugins/ApiResources/…`, `tests/Feature/Plugins/ApiResources/ResourceToArrayInferenceTest.php` (+ fixtures) | create |
| `examples/api-resources/` (new inferred-only `BookingResource` + route), snapshot regenerated | modify |
| `docs/plugins.md` (ApiResources), `docs/auto-derivation.md` row, `CHANGELOG.md` | modify |

## #108 seam

`SchemaFromResource::buildRef(class-string<JsonResource>): string` stays the single "concrete Resource class → item schema `$ref`" entry point (now fallback-aware). #108 only has to widen `ResourceClassLocator` to resolve the concrete class from the controller's return expression and hand it to the existing `ResourceTarget` flow — nothing in this PR assumes the class came from a signature.

## Implementation plan (TDD, one commit per step)

1. **Extract `SchemaDefinitionFromLiteral`** from `InlineJsonResponseResolver` — pure refactor, #14's tests stay green.
2. **`SingleReturnArrayLiteralFinder`** — unit tests: single literal return, early-return shape refused, `return $var` refused, beyond-limit refused.
3. **`WrappedModelLocator`** + `EloquentModelToSchema::propertyFor()` — unit tests: `@mixin`, `@extends`, non-model rejected, none → null; property lookup across casts/`@property`/appends.
4. **`ResourceToArrayReader` + `SchemaFromResource` composition** — feature tests: plain literal; `$this->field` typed from model; nested resource `$ref` (+ self-reference cycle); `::collection()` value; `when`/`whenLoaded`/`merge`/`mergeWhen`/`whenCounted`; unresolvable value → `{}` + one note; dynamic `toArray()` degrade + note; `#[ResourceField]` per-field override; passthrough → model `$ref`; lint quieting.
5. **Examples + docs + CHANGELOG** — extend `examples/api-resources` with an attribute-free `BookingResource`, regenerate the snapshot, verify the diff is exactly this feature.

Verify: `composer test && composer lint && vendor/bin/pint --test`.


Radiergummi:
## Implementation decisions (vs. the plan / spec)

**Reader is pure and registry-free; the schema builder owns all notes and `$ref`s.** The plan had `ResourceToArrayReader` resolving nested-resource refs through a closure into `SchemaFromResource`. Implemented instead with a small VO (`InferredResourceField`): nested resources come back as a *class reference* the builder resolves via its own `buildRef()`. Two wins: no construction cycle/closure plumbing, and the `resource.fields-undeclared` lint rule can consult the reader with **zero side effects** (no component registration, no log spam at lint time). All generation-log notes moved to `SchemaFromResource`, which builds each resource component exactly once per run — natural once-per-class note dedup.

**Note strategy (the "26-line log spam" question).** One summarising NOTICE per resource listing the keys that came out unconstrained — but only keys actually *used* from inference: a key covered by a `#[ResourceField]` never appears in a note, so a fully-declared resource with an exotic `toArray()` (our own `FlightResource` example: `->toIso8601String()` calls) generates **no** noise. Dynamic-body notes likewise fire only when the resource declares no fields (inference was the only hope); with attributes present the scan steps aside silently.

**`when*` family.** `when`/`whenLoaded` resolve their inner value and mark the key optional (condition never analysed — the `abort_if` rule); bare `whenLoaded('relation')` resolves the relation against the wrapped model (`@property-read Author $author` → conditional `$ref`); `whenCounted` → optional `integer`. Beyond the spec list, any **other** `when`-prefixed resource method (`whenHas`, `whenNotNull`, `whenPivotLoaded`, …) yields an unconstrained **optional** property: claiming a runtime-conditional key as `required` would over-promise, and the `when` prefix is a reliable Laravel-conditional signal. Each spec idiom has its own test + doc line.

**`merge`/`mergeWhen` partial-keep.** A literal payload inlines its keys (optional for `mergeWhen`). A *non-literal* payload skips just that entry with a note and keeps the other keys — everything still stated is true (OpenAPI objects are open by default), whereas degrading the whole literal would throw away the readable majority. A plain unkeyed entry / spread / dynamic key still refuses the whole literal — there the key *set* itself is unknowable.

**Dynamic `toArray()` fallback.** Issue wording allowed "model schema *or* envelope only". Implemented: model schema when `@mixin`/`@extends` resolves (+ note), today's behaviour otherwise (+ note when nothing was declared). Fallback is strictly **fallback-only** per the folded #98 design — any declared or inferred field makes the resource authoritative and the model is ignored.

**`propertyFor()` ignores `$hidden`/`$visible`.** Those govern the *model's own* serialization; a resource that names `$this->internal_notes` in its literal has decided the key is output. Pinned by a unit test.

**Required semantics.** Every always-present key is `required` — including nullable ones (`subtitle: ?string` → required + nullable), which is what the runtime array actually does. Only conditional wrappers (and `mergeWhen` payloads, and declared `conditional: true`) drop out of `required`. A readable-but-empty `return [];` is honoured as a genuinely empty object schema (no model fallback — the author wrote an empty body).

**Model discovery scope.** `WrappedModelLocator` reads the resource class's own docblock only (no ancestor walk) — `@mixin` wins over `@extends`, `Model` subclasses only, memoised. Matches where real apps put the tag (Vito/AdvisingApp style).

**Recursion guard.** No new depth limit: nested/self-referencing resources terminate through `ComponentSchemaRegistry::buildOnce()`'s existing `inProgress` reservation (test: `SelfReferencingCategoryResource` emits a self-`$ref`).

**Perf.** One parse per resource class per run, memoised in the reader and backed by the scanner's per-file AST cache; the parse runs only when a resource component is actually built (i.e. the class is reachable from a response — the Tier-0-miss discipline at the resource level). Note: `read()` also runs for resources fully covered by attributes, because composition can't know the attribute set covers the literal without reading it; that is one cached parse per class, file-level cache shared with the other body scans.

## The #108 seam

`SchemaFromResource::buildRef(class-string<JsonResource>): string` is the single "concrete Resource class → item-schema `$ref`" entry point, now fallback-aware (model component for passthrough/dynamic). #108 only needs to widen `ResourceClassLocator` to resolve the concrete class from the controller's **return expression** and hand it to the existing `ResourceTarget` flow; nothing in this PR assumes the class came from a signature. The lint-quieting and note machinery compose unchanged.

## Also in this PR (small shared-infra touches)

- `Support\MethodBody\SchemaDefinitionFromLiteral` — extracted verbatim from `InlineJsonResponseResolver` (#14) so both literal walkers share one mapping; the resolver now delegates.
- `Support\MethodBody\SingleReturnArrayLiteralFinder` — the generic single-return-literal primitive (plugin-agnostic; #13 Fractal `transform()` is the second consumer). Returns inside nested closures/anonymous classes don't count against the single-return rule.
- `Support\Generator\SchemaFromArrayDefinition::buildProperty()` — the property-shaped twin of `build()`.
- `EloquentModelToSchema` — internals reshaped around a memoised `metadataFor()` (behaviour of `build()` unchanged, all existing tests green) to expose `propertyFor()`.

Radiergummi:
## Review: empirical test report (#247 delta only; #219…#241 taken as approved)

**Verdict: APPROVE.** Gates green, refactors byte-faithful, every epic-contract probe passed, and the controlled survey experiment reproduces the #106 prediction exactly (+69 Vito ops). Findings below are non-blocking.

### Gates (worktree @ af95f95, PHP 8.4.16)

- `composer test`: 2187 passed (5450 assertions). One first-run flake — `tests/TestCase.php:71` "Failed to create fixture destination" under `--parallel`; pre-existing (TestCase untouched by this PR), green on re-run.
- `composer lint` (PHPStan level 8): no errors. `vendor/bin/pint --test`: passed.

### Findings

1. **[Minor] The "lint can consult the reader with zero side effects (no component registration)" claim doesn't fully hold.** `read()` → `EloquentModelToSchema::propertyFor()` → relation-tagged `@property-read` → `build()` → `ComponentSchemaRegistry::buildOnce()`. Proved empirically: a cold `read(ConditionalArticleResource::class)` against a fresh registry registers the `Author` component (and `schemaFromTagType()` can emit logger warnings on unresolvable tags). In the real `openapi:lint` flow the scoped reader/registry are warm from generation, so there's no observable impact — but the decisions-comment claim should be softened, or relation refs resolved lazily, so the next person doesn't build on the stronger guarantee. Fine to fix as a doc-comment touch-up.
2. **[Nit] Duplicate literal keys keep the first read; PHP runtime keeps the last.** `['a' => 1, 'a' => 'x']` documents `a: integer` while the wire value is a string. Vanishingly rare; mention only because the comment in `buildSchema()` states the choice as if it matched runtime.
3. **[Nit] Divergent int-key semantics between the two literal walkers.** `SchemaDefinitionFromLiteral::fromArrayNode()` treats explicit sequential integer keys as a JSON list (json_encode semantics); `ResourceToArrayReader::fieldsFromArrayNode()` turns the same shape into properties named `"0"`, `"1"`. A `toArray()` returning a list isn't an object schema either way, so this is academic today — but worth one line of doc-comment if #13 reuses the reader pattern.
4. **[Nit] Conditional-wrapper edge over-/under-claims.** `when($cond, $value, $default)` (3-arg) is always-present at runtime but documented optional (safe direction); `whenCounted('x', fn ($c) => …)` with a transforming callback is still typed `integer` regardless of the callback's return (unsafe direction, extremely rare). Neither blocks; noting for the record.

### Refactor fidelity

- `SchemaDefinitionFromLiteral` extraction: verbatim (instance→static only); `InlineJsonResponseResolver` delegates cleanly. `SchemaFromArrayDefinition::buildProperty()` is purely additive. Snapshot suite: the only example diff is the new `BookingResource` flavor — no unrelated snapshot lines moved.
- `EloquentModelToSchema`: `schemaFor()` still applies `$hidden`/`$visible` (direct-model returns unchanged); memoisation is metadata-level, so `propertyFor()` returns fresh `OA\Property` instances per call — the reader's `$property->property = $name` rename is aliasing-safe.
- **Hidden-fields position: agreed.** `$hidden` governs the model's own serialization; a resource naming `$this->internal_notes` in its literal genuinely outputs it, so documenting it is faithful, not a leak. Pinned by `EloquentModelPropertyLookupTest`.
- Plugin independence: no `Plugins\Core` import anywhere in ApiResources / `Support\MethodBody` / `EloquentModelToSchema`.

### Empirical probes (throwaway fixtures, removed)

| Probe | Result |
|---|---|
| A→B→A mutual resource cycle (`new B(...)` / `A::make(...)`) | terminates; mutual `$ref`s; `whenLoaded`-wrapped side optional ✓ |
| Same-short-name resources, two namespaces | distinct components, distinct `$ref`s per op (#219 lesson holds for inferred resources) ✓ |
| `whenHas` / `whenPivotLoaded` / 3-arg `when` | unconstrained optional, `required` untouched ✓ |
| Lint: fully-inferred resource / passthrough+model | quiet ✓ / quiet ✓ |
| Lint: dynamic body, no model, no attributes | still fires `resource.fields-undeclared` ✓ |
| Cold lint-time `read()` registry writes | **registers relation model component** (finding 1) |

Shipped tests already cover literal/model/nested/self-ref/conditional/merge/dynamic/passthrough/precedence — good suite.

### Survey (11 apps, #233 procedure)

State verified: all 11 `vendor/radiergummi/laravel-openapi → …/review-12/`, lock `dev-af95f95` (composer state reset per #233 repair before the run). #221 reproduced (SpeedtestTracker/Vito dropped after Pelican — this branch predates the main-side corpus.sh fix); backfilled via `--only` and hand-merged (#229) → `$WS/results-pr-247.json`.

**Like-for-like vs `results-pr-241.json`** (note: baseline @ 5b15aa6, merge-base is ce8c1a8 — no query-param deltas appeared anywhere, so the polish commit is a non-factor in this table):

| app | aOps | respΔ | compΔ | lintΔ | gen_exit |
|---|---:|---:|---:|---:|---:|
| AdvisingApp | 31 | +0 | +0 | +0 | 0 |
| AureusERP | 90 | +0 | +0 | +0 | 0 |
| Bagisto | 41 | +0 | +0 | **−2** | 0 |
| BookStack | 86 | +0 | +0 | +0 | 0 |
| Coolify | 154 | +0 | +0 | +0 | 0 |
| InvoiceNinja | 522 | +0 | +0 | +0 | 0 |
| Koel | 170 | +0 | +0 | +0 | 0 |
| Lychee | 205 | +0 | +0 | +0 | 0 |
| Pelican | 158 | +0 | +0 | +0 | 0 |
| SpeedtestTracker | 8 | +0 | +0 | +0 | 0 |
| Vito | 144 | +0 | +0 | +0 | 0 |

The flat table is explainable, not a miss: the workspace Vito/AdvisingApp carry the full-spec-proof `#[ResourceField]` scratch (the 241 baseline measured Vito at 144/144 response schemas — saturated), Bagisto's two STATIC resources (`TransactionResource` 8 props, `TaxCategoryResource` 5 props — both now substantive) sit under `/admin`, outside the survey's `apiPrefix`, and Bagisto's lint −2 is exactly `resource.fields-undeclared` 2→0.

**Controlled clean-Vito pair** (annotation scratch stashed; base = ce8c1a8 worktree vs PR head, same tree otherwise):

| | base ce8c1a8 | PR af95f95 |
|---|---:|---:|
| substantive 2xx-schema ops (of 146 `/api`) | 10 | **79 (+69)** |
| `resource.fields-undeclared` | 69 | **0** |
| other lint rules | — | byte-stable (incl. `resource.response-ambiguous` 42 — the #108 half, untouched as expected) |
| `openapi:generate` wall-clock | 1.21 s | 1.26 s |

**Prediction vs actual:** Vito **+69 — exactly the #106 number**. Bagisto +2 components-substantive (invisible to the apiOperations counter only because of the `/admin` prefix). Pelican predicted +1 did **not** materialize and on inspection #106 overcounted: its one "STATIC" class is `Daemon\ServerConfigurationCollection`, whose literal lives inside a `->map()` closure (`return $this->collection->map(...)->toArray()`) — the reader correctly refuses it. AdvisingApp +0 as predicted (UNRESOLVED-signature, #108 territory).

**Spot-checks (clean Vito):** `SiteResource` 26/26 keys, `ServerResource` 25/25, `ServiceResource` 15/15 — no fabricated keys; `server` → optional `$ref: ServerResource` via `whenLoaded` ✓; model-typed scalars right (`port: integer`, `domain: string`, nullable `@property` → 3.1 type unions); method-call values (`status->getText()`, `getUrl()`, `webserver()->id()`) → `{}` as designed (the #126 pairing). Negative case: no concrete corpus resource has a dynamic `toArray()` (consistent with #106's "zero dynamic"), so the degrade path is exercised by this PR's own fixtures only — verified there.

**Perf / log volume:** no crashes anywhere (Bagisto's hundreds of resources parse clean, all 11 `gen_exit=0`); +0.05 s on the 329-path Vito run; exactly **one** summarising NOTICE per resource (24 notices for 24 resources on clean Vito; 0 on annotated Vito — attribute-covered resources stay silent as claimed; Bagisto: 2 resources → 2 notices per artisan invocation).

Two recurring unconstrained-key patterns dominate the notices and are pre-existing reader gaps, filed separately: #249 (`array`/`json` casts document JSON lists as `type: object` — `aliases`, `ssh_users`, …) and #250 (`created_at`/`updated_at` resolve to no metadata on conventional models — they appear in **every** Vito notice).

### #108 readiness

The seam is as advertised: `buildRef()` is the single fallback-aware entry point, nothing assumes signature provenance, and the 42 `resource.response-ambiguous` ops on clean Vito are sitting right behind a widened `ResourceClassLocator`. #249/#250 are worth landing before or alongside #108 — they multiply across every op the wider locator unlocks.

Radiergummi:
Independent stack review — finding attributable to this PR:

**1. [minor] `docs/auto-derivation.md:248-255` "Known limitations" contradicts this PR's own delivery.** "**Method-body inference is not done.**" is false in a file that documents two body scans seven lines later; only the narrow `Model::find()` (#97) claim should remain. And the second bullet — "`JsonResource` wrapping a model … (tracked as #98); the model schema and the resource schema are currently independent" — is exactly the passthrough base case this PR delivered (per its own CHANGELOG entry and `docs/plugins.md:96-102`). Also `docs/recipes.md:74-76` still presents `#[ResourceField]` as the only way to document resource keys ("output keys are arbitrary toArray() entries … each key is declared with … #[ResourceField]"), which post-#247 is the escape-hatch-restating-inference smell the project's own philosophy warns about.

Radiergummi:
Fixes landed on this branch from the stack review: (1) the path-dependent literal semantics flagged on PR #230 — `SchemaDefinitionFromLiteral::fromLiteralValue` now answers like the AST branch (empty array → array-of-unknown-items; list items only claimed when every element agrees, via the shared `listDefinition` homogeneity rule) — and (2) the stale auto-derivation 'Known limitations' bullet claiming the resource and model schemas are independent (#98) now describes the shipped passthrough/field-resolution behaviour. Commit `0ef02c3`.

Issue #142, opened 2026-06-05
Route registered for all verbs (`#[Any]` / `Route::any`) fans out into 6 identical operations; no way to restrict emitted verbs


Radiergummi:
## Summary

A route registered for **all HTTP verbs** (Spatie `#[Any(...)]`, or `Route::any()`) fans out in `PathsStage` into one operation per verb (GET/POST/PUT/PATCH/DELETE/OPTIONS), all sharing the single built operation (same summary/body/responses). There is no way to restrict which verbs emit.

## Context

Surfaced by the Vito full-spec proof (parent #113). `GitHookController` is registered via `#[Any('api/git-hooks')]`, so the spec gains six identical operations for one URI even though the webhook is only ever `POST`ed.

This is **not a library defect** — the route genuinely matches all verbs. The honest fix is usually upstream (change `#[Any]` to `#[Post]` in the app). But there is no authoring attribute or filter to say "only emit POST for this route" when the app legitimately registers `any`.

## Affected domains (Vito proof)

1 domain (GitHookController). Low frequency / low priority.

## Proposed direction (open question)

Decide whether an authoring surface is warranted, e.g.:

- a route/method-level attribute restricting which verbs are documented, or
- a `RouteFilter` recipe / docs note steering users to narrow the route upstream.

Given it affects a single domain and the upstream fix is trivial, this may resolve as a **docs note** rather than a feature.

Parent: #113

Radiergummi:
This should just be documented for clarity; the behaviour itself is correct IMHO

Issue #186, opened 2026-06-08
SwaggerPhp harvester: authored schema name colliding with a convention component silently keeps the wrong schema


Radiergummi:
## Summary

The `SwaggerPhp` harvester (#17) registers authored `#[OA\Schema]` / `@OA\Schema` definitions **directly** into `$document->components->schemas`, deduplicating by authored name. Because the harvest runs in a `SpecStage` that executes *after* `ComponentsStage` has already flushed `ComponentSchemaRegistry`, it never sees the registry's disambiguation machinery. When an authored schema name collides with a convention-derived component of the same key, the collision is resolved silently and possibly wrongly.

## Problem

`HarvestAuthoredAnnotationsStage::registerSchema()` (`src/Plugins/SwaggerPhp/Stages/HarvestAuthoredAnnotationsStage.php`) does:

```php
foreach ($components->schemas as $existing) {
    if ($existing->schema === $schema->schema) {
        return; // already present — first wins
    }
}
$components->schemas[] = $schema;
```

Two failure modes:

1. **Authored vs. convention collision.** An app has both a convention component keyed `Invoice` (e.g. a Spatie `Data` class or `JsonResource` named `Invoice`, already written by `ComponentsStage`) **and** a hand-authored `@OA\Schema(schema="Invoice")`. The authored operation's `@OA\Response` references `#/components/schemas/Invoice`, but `registerSchema` finds the convention `Invoice` already present and drops the authored definition. The response now advertises the **convention** shape, not the shape the author wrote and intended — silently, with no warning (the harvester only warns when a name is *absent*, not when it collides).

2. **Two different authored schemas sharing a name** collapse to the first, again silently.

`ComponentSchemaRegistry` was built precisely to handle this (key reservation via `registerNamed()`, `deriveKey()` disambiguation, the named-key-owner sentinel). The harvester routes around all of it.

## Proposed approach

Tier 0; no new parsing. Options, roughly in increasing order of effort:

- **Minimum:** make the collision non-silent — when `registerSchema` finds an existing component under the authored name that is **not** the authored schema object, emit a `warning` (and ideally a lint finding) so the drift is diagnosable.
[...]

PR #248, opened 2026-06-10
fix(#245): verb-conditional inference gates on the emitted verb, not route->methods[0]

### LINKED ISSUE #245: Verb-conditional inference gates read `route->methods[0]` only — multi-verb routes mis-gate both ways
## Summary

`ActionDescriptor::httpMethod` is derived as `HttpMethod::fromString($this->route->methods[0])` (`src/Routing/ActionDescriptor.php`), while `PathsStage` fans a multi-verb route out into one operation per verb sharing the single built operation. Every verb-conditional inference therefore evaluates only the *first* registered verb, and a `Route::match(['get', 'post'], …)` route mis-gates in both directions:

- **GET-only inferences leak onto the POST twin.** The #241 accessor scan matches `input()`/`string()`/`integer()`/`boolean()` only on GET/HEAD — but on a `['get', 'post']` route the descriptor says GET, so the POST operation carries the same inferred query parameters. Observed live on the survey corpus (Coolify, PR #241 review run): `POST /api/v1/applications/{uuid}/start` documents `force` (boolean, default `false`) and `instant_deploy` as query parameters sourced from `$request->boolean(…)` reads, which on POST are body reads per #241's own verb-discipline decision. Same for the GET/HEAD-gated inline-validate → query-parameter hand-off.
- **POST-only inferences never fire.** The mirror: the #9 body scan gates on `WRITE_METHODS` (POST/PUT/PATCH), so the POST twin of a `['get', 'post']` route never documents its inline-`validate()` request body — `httpMethod` reports GET.

Verb order also matters: `Route::match(['post', 'get'], …)` flips both failure modes.

## How widespread

Coolify alone registers ~10 `Route::match(['get', 'post'])` API endpoints (deploy/start/stop/restart). Single-verb routes — the overwhelming majority — are unaffected (`Route::get` yields `[GET, HEAD]`, where GET-first is correct).

## Direction

The verb-conditional resolvers need the verb of the operation actually being emitted, not the route's first verb — either build operations per verb (pass the verb through `OperationBuilder` / the descriptor), or split multi-verb routes into per-verb descriptors in `RouteIntrospector`. Related but distinct: #142 (multi-verb fan-out emits identical operations; closed as upstream-fix).

## References

Found during the empirical review of PR #241. Entry points: `src/Routing/ActionDescriptor.php` (`httpMethod` getter), `src/Support/Generator/Stages/PathsStage.php` (per-verb fan-out), gates in `src/Plugins/Core/Resolvers/CoreQueryParameterResolver.php` (`BODYLESS_METHODS`) and `src/Plugins/Core/Resolvers/InlineValidationRequestSchemaResolver.php` (`WRITE_METHODS`).

Radiergummi:
## Problem

`ActionDescriptor::httpMethod` is `HttpMethod::fromString($this->route->methods[0] ?? '')` — a single value per route. But `PathsStage::attachOperation()` builds the operation **once** (`OperationBuilder::build($action, …)`) and then fans the *same* built operation out across every verb in `route->methods()`. So on a `Route::match(['get', 'post'], …)` route, all verb-conditional inference is evaluated against the *first* registered verb only, and the result is attached to both operations.

Concrete failure modes (both observed on the survey corpus during PR #241 / #9 review):

- **GET-only inference leaks onto the POST twin.** `Route::match(['get','post'])` reports `httpMethod === GET`, so the #241 `$request->*()` accessor scan and the GET/HEAD-gated inline-validate→query hand-off run, and their query parameters are attached to the POST operation too. Live: Coolify `POST /api/v1/applications/{uuid}/start` documents `force` / `instant_deploy` as **query** params sourced from `$request->boolean(…)` body reads.
- **POST-only inference never fires.** The mirror: #9's body scan and `RequestBodyExtractor` gate on the write verbs (`POST/PUT/PATCH`); on a `['get','post']` route `httpMethod` reports `GET`, so the POST twin never documents its inline-`validate()` request body.

Verb order flips which way it breaks (`Route::match(['post','get'])` inverts both). Single-verb routes are unaffected: `Route::get` yields `[GET, HEAD]`, where GET-first is correct, and HEAD is skipped in the fan-out.

Coolify alone registers ~10 `Route::match(['get','post'])` API endpoints (deploy/start/stop/restart).

## Root cause

The operation is built once, before the per-verb fan-out loop — `httpMethod` is a property of the *route*, not of the *operation being emitted*. Every verb-conditional resolver therefore reads a verb that may not be the one it's documenting.

`httpMethod` consumers that this affects: `RequestBodyExtractor` (write-verb gate), `ThrowsErrorContributor` + `FindingLocation` (location metadata), and the in-flight body-scan resolvers from #9 / #11. Also surfaces in `GenerateCommand` / `WhyCommand` display.

## Plan

**Recommended — thread the emitted verb through `build()` (surgical):**

1. Give `ActionDescriptor` an explicit, nullable per-operation verb that `httpMethod` returns when set, falling back to `methods[0]` otherwise; add a cheap `withHttpMethod(HttpMethod)` cloner.
2. In `PathsStage::attachOperation()`, move `OperationBuilder::build(...)` *inside* the verb loop and build with `$action->withHttpMethod($method)` for each non-HEAD verb. Every `$descriptor->httpMethod` consumer then sees the verb actually being emitted — no resolver changes needed.
   - Common case is unchanged: `Route::get` → `[GET, HEAD]`, HEAD skipped → builds once for GET, exactly as today. Only genuine multi-verb routes build more than once (rare; acceptable).

**Alternative considered — split multi-verb routes into per-verb descriptors in `RouteIntrospector`.** Makes "one descriptor = one verb" an invariant so every consumer is correct for free, but ripples into the `PathItem` grouping, the fan-out loop, and the `count($methods) > 1` operationId path. Heavier surface for the same outcome; will confirm direction in review before implementing.

## Verification

- **Reproduce first (TDD):** a `Route::match(['get','post'])` fixture where the GET path infers a query param and the POST path infers a request body — assert the param is GET-only and the body is POST-only. Add the `['post','get']` order variant to lock the order-independence.
- Regression: single-verb GET / POST / resourceful routes unchanged (snapshot group).
- `composer test`, `vendor/bin/pint --test`, `composer lint` (PHPStan L8) all green.
- Cross-check against the in-flight #241 PR's Coolify output: `POST …/start` no longer carries `force` / `instant_deploy` as query params.

## Notes

- Distinct from #142 (multi-verb fan-out emitting identical operations; closed as upstream-fix) — this is about *which verb the inference gates on*, not duplicate emission.
- Unblocks the correctness of #241 (#11) and #9; worth landing alongside that cluster.

Closes #245


Radiergummi:
Implemented the **recommended (surgical)** option from the plan — threading the emitted verb through `build()` — not the per-verb-descriptor split:

- `ActionDescriptor` gains a private `httpMethodOverride` + a `withHttpMethod(HttpMethod)` cloner; the `httpMethod` hook returns the override when set, else `methods[0]` (unchanged default).
- `PathsStage::attachOperation()` now derives the tag once, then builds **inside** the verb loop with `$action->withHttpMethod($method)` for each non-HEAD verb. `bindAction` and `OperationContext` also receive the verb-specific descriptor, so the location metadata / transformer consumers see the emitted verb too.

Single-verb routes build exactly once as before (`Route::get` → `[GET, HEAD]`, HEAD skipped); only genuine multi-verb routes build per verb. Re-building registers the same `$ref` components twice for those routes, which the `ComponentSchemaRegistry` dedups (transitive same-object re-registration, per #186) — confirmed no snapshot drift.

Reproduction is a `Route::match(['get','post'])` fixture with one `$request->validate(['name' => …])`: the GET twin documents `name` as a query parameter (no body), the POST twin as a request body (no query param). Both verb orders are covered to lock order-independence.

Verified: `composer test` (2418 passed), `composer lint` (0 errors), `vendor/bin/pint --test` (passed).

Radiergummi (commented):
The comments repeat the same thing over and over, and also reference the issue. Don't do that. Add comments when they help understand a concrete piece of code, sparingly, when absolutely necessary, and don't mention issues unless they refer to a very specific edge case explicitly necessary reading material to understand some unconventional choice. that is *clearly* not the case here.

Radiergummi:
Addressed — trimmed the comments in `ActionDescriptor` and `PathsStage` to one concrete line each and dropped the issue numbers. Recorded the preference (sparing, concrete comments; no issue refs in code) so it carries forward.

PR #251, opened 2026-06-10
fix(#249, #250): timestamp columns + list-shaped array casts in EloquentModelToSchema

### LINKED ISSUE #249: Eloquent 'array'/'json' casts document JSON lists as type:object — descend into @property generics to disambiguate
Found while spot-checking PR #247's inferred Vito resources, but the mechanism predates that PR — it is the #18/#99 `EloquentModelToSchema` cast mapping.

## Problem

`castToProperty()` maps the `array`, `json`, `object`, and `collection` casts uniformly to `{type: object}`. For a JSON **list** column this documents the wrong type: Vito's `Site::$aliases` is cast `array`, declared `@property array<int, string> $aliases`, and serializes as `["a.example.com", "b.example.com"]` — but the generated schema says `aliases: {type: object}` (observed in the inferred `SiteResource` component).

Because `propertyFor()`/`schemaFor()` check `$casts` **before** the `@property` tag, the (more precise) docblock never gets a say for cast columns — and even if it did, `schemaFromTagType()` deliberately refuses `GenericTypeNode`/`ArrayTypeNode` ("Tier-0 doesn't parse deep generics"), so `array<int, string>` falls back to the untyped property anyway.

## Proposal (Tier 0)

For the ambiguous casts (`array`, `json`, `collection` — `object` can stay `object`), consult the model's `@property` tag for the same name before settling:

- `array<int, T>` / `list<T>` / `T[]` → `{type: array}` (+ `items` when `T` is a scalar keyword) — one level only, no deep generic descent;
- `array<string, T>` / no tag → today's `{type: object}`.

This is signature/docblock reading only, no new parsing machinery — `TypeNodeResolver` already holds the node.

## Impact

Every JSON-list column on a model documents as `object` today; with #247 merged these now surface widely through inferred Resource fields (`aliases`, `ssh_users`, `config_paths`, `tags`-style columns across the survey corpus).

### LINKED ISSUE #250: Model timestamps (`created_at` / `updated_at`) resolve to no property metadata — type them date-time when `$timestamps` is on
Found while spot-checking PR #247's inferred Vito resources; the gap is in the #18/#99 `EloquentModelToSchema` metadata reader, so it predates that PR.

## Problem

`metadataFor()` unions `$casts` + `@property` tags + `$appends` + `$fillable`. Laravel's `created_at` / `updated_at` live in none of these on a conventional model (they are framework-managed via `$timestamps`/`getDates()`, not `$casts`), so `propertyFor('created_at')` returns null and `schemaFor()` never lists them.

Observed effect on clean (unannotated) Vito under #247: **every one** of the 24 inferred resources lists `created_at`/`updated_at` among its unconstrained keys — they are by far the most common entries in the per-resource degradation notices, despite being the most conventional, most predictable columns Laravel has.

## Proposal (Tier 0)

When the instantiated model reports timestamps enabled (`$model->usesTimestamps()`), add `created_at` and `updated_at` (via `getCreatedAtColumn()` / `getUpdatedAtColumn()` — respects renames) to the known-property union typed as `{type: string, format: date-time}` (nullable, matching runtime). An explicit cast or `@property` tag for the same name should keep winning.

Pure reflection on the already-instantiated model — no new machinery. Would remove the dominant noise term from #247's unconstrained-key notices and put two real properties on nearly every model/resource schema in the corpus.

Radiergummi:
## Plan

Closes #249
Closes #250

Two pre-existing `Support\Extraction\EloquentModelToSchema` metadata gaps, filed by the #247 review and combined here because they touch the same reader and multiply across every operation #108 will unlock.

**Stack:** #219 ← #225 ← #230 ← #235 ← #241 ← #247 ← this (base: `feat/12-resource-toarray-inference`).

### #250 — timestamps

- `metadataFor()` gains a `timestamps` entry: when `$model->usesTimestamps()`, the non-null `getCreatedAtColumn()` / `getUpdatedAtColumn()` values (respects `CREATED_AT`/`UPDATED_AT` renames).
- Both `schemaFor()` (name union + property loop) and `propertyFor()` resolve a timestamp column to `{type: [string, 'null'], format: date-time}` — nullable, matching runtime (`{Model}::freshTimestamp()` rows exist, but pre-save instances and nullable columns do not guarantee a value).
- Precedence preserved: an explicit `$casts` entry or `@property` tag for the same column wins; the timestamp default fills in only below them. Timestamp-only columns never enter `required` (no non-nullable tag).
- No `deleted_at`/SoftDeletes handling — #250 scopes to `created_at`/`updated_at`.

### #249 — array/json/collection casts as lists

- New `TypeNodeResolver::listValueType()`: returns the element type node when a docblock node denotes a JSON list — `list<T>`, `array<int, T>`, `T[]` (one level, leading nullable unwrapped) — null otherwise. Reuses the existing node machinery; no new parsing.
- `castToProperty()` consults the model's `@property` tag for `array` / `json` / `collection` casts: list shape → `{type: array}` (+ `items` when the element is a scalar keyword); `array<string, T>`, single-arg `array<T>`, or no tag → today's `{type: object}` (conservative default, documented). `object` cast stays `object` unconditionally.

### Verification

1. Unit tests beside `EloquentModelToSchemaTest` / `EloquentModelPropertyLookupTest` (timestamps on/off/renamed/overridden; list-cast shapes) → red, then green.
2. Feature-level assertion through the Resource path (resource exposing `$this->created_at` + a list-cast field).
3. Examples snapshot regen — expected diff: the vanilla `Flight` model schema gains `created_at`/`updated_at` (vanilla has no `Booking` model component); verify nothing else moves.
4. `composer test`, `composer lint`, `vendor/bin/pint --test` green; docs (`docs/auto-derivation.md` cast/precedence section) + CHANGELOG entries for both issues.


Radiergummi:
## Decisions (as implemented)

**#250 — timestamps**

- **Nullability:** timestamp defaults are typed `{type: [string, 'null'], format: date-time}` and never enter `required`. Runtime reality: unsaved models and `NULL` columns carry no value, and the columns are framework-managed rather than app-asserted — claiming non-null would over-promise.
- **Soft deletes:** no `deleted_at` handling. #250 scopes to `created_at`/`updated_at`; `deleted_at` would need its own trait detection and is usually invisible in API output anyway (soft-deleted rows are filtered, the column is commonly hidden). Separate issue if the survey shows demand.
- **Renames/disabling:** read through `getCreatedAtColumn()`/`getUpdatedAtColumn()`, so `CREATED_AT`/`UPDATED_AT` constants are respected and a `null` constant disables that side.
- **Precedence:** the timestamp default sits directly **above the untyped fallback** and below every explicit source — `$casts` entry, `@property` tag, and `$appends` accessor for the same column all win.

**#249 — JSON-cast list disambiguation**

- **Generic-absent default:** untagged columns, map-shaped generics (`array<string, T>`), and single-argument `array<T>` (key type unknown) keep today's conservative `{type: object}`. Only an unambiguous list shape (`list<T>`, `array<int, T>`, `T[]`) flips to `array`.
- **`object` cast:** stays `{type: object}` unconditionally, even when the tag is list-shaped — the cast dictates the serialized PHP value (`stdClass`).
- **Items:** typed only when the element is a scalar keyword; `list<Carbon>` emits a bare `{type: array}`. One level only, no deep generic descent — the Tier-0 boundary from the issue.
- **Nullable tags:** a leading nullable (`?list<string>`) is unwrapped to find the list shape, but nullability is **not** propagated onto the property — the cast path has never been nullable-aware, and widening that is out of scope here.
- `TypeNodeResolver` gains exactly one method (`listValueType()`); no other parsing machinery changed.

**Plan deviation:** the verification step said the vanilla `Flight`/`Booking` schemas gain the timestamp columns — vanilla has no `Booking` model component, so the snapshot diff is `Flight` only (verified byte-exact via the `snapshot` group; no other example moved). PR body corrected.

**README:** the hero snippet mirrors the regenerated `examples/vanilla` output (docs convention: spec snippets grounded in real example output), hence the two added lines there.

Radiergummi:
## Review: empirical test report (review at 8566e71, delta vs `feat/12-resource-toarray-inference` @ af95f95)

**Verdict: APPROVE.**

### Gates (run locally, PHP 8.4)

- `composer test`: **2200 passed** (5487 assertions). One parallel run flaked with exit 2 before passing twice — sequential run green on first try; smells like the known fixture race (#222), not this PR.
- `composer lint` (PHPStan level 8): **no errors**.
- `vendor/bin/pint --test`: **passed**.
- `composer examples` regen: **byte-clean** — no snapshot drift; the committed `examples/vanilla` diff (Flight gains `created_at`/`updated_at`, nothing else) is exactly reproducible. Checked why no other example moved: no example emits a *direct model* component besides vanilla's `Flight` (the Resource/Data/Fractal examples build field-driven component schemas), so the single-file delta is legitimate.

### Code review

The implementation matches the decisions comment exactly. `listValueType()` is genuinely additive — one new public method on `TypeNodeResolver`, reusing `unwrapNullable()`, no existing call-site touched. The precedence move in `propertyFor()`/`schemaFor()` (hoisting `$tag` above the cast branch) changes no behaviour for non-JSON casts: the tag is only *consulted* inside `castToProperty()` for `array`/`json`/`collection`. `required` stays tag-driven (non-nullable `@property` only), so default-typed timestamps can never enter `required` — consistent with how every other inferred model property handles required. Never-required + nullable is the right call for framework-managed columns.

Beyond the shipped tests, I probed the edges empirically (throwaway fixtures, removed):

- `list<list<string>>` → bare `{type: array}` (no recursion) ✓
- `non-empty-list<string>` → `{type: object}` (conservative; acceptable Tier-0 boundary, noted below) ✓
- `array<int|string, string>` → `{type: object}` ✓
- `Carbon[]` → bare `{type: array}` ✓
- `$appends` accessor for `created_at` (returning `int`) **wins** over the timestamp default; `updated_at` still gets the default ✓ — full precedence chain (enum cast → cast → `@property` → appends accessor → timestamp default → untyped) verified
- relation `$ref` path: building `Article` registers `Author`, whose component gains nullable date-time timestamps ✓
- `CREATED_AT` rename + `UPDATED_AT = null`: `propertyFor('creation_date')` typed, `propertyFor('created_at')` and `propertyFor('updated_at')` both null ✓

README hero snippet matches the regenerated vanilla output (same two lines, same OAS 3.1 union form); `docs/auto-derivation.md` example is internally coherent (`created_at` required via its explicit non-nullable tag, `updated_at` via the new default).

### Survey (Layer A, all 11 apps, per the #233 procedure)

Baseline `results-pr-247.json` @ af95f95 (= verified merge base). Composer state reset per app before the run; post-run assertion: all 11 vendor symlinks → `review-251`, lock pinned `dev-8566e71`.

- **Aggregate: completely flat** — apiOperations, responseSchemas, requestBodies, componentSchemas, completeness%, lint totals *and* byRule identical across all 11 apps; `genExit 0` everywhere. Exactly the expected profile for schema-quality fixes.
- **Clean-Vito controlled pair:** regenerating clean Vito at the merge base reproduces the archived `clean-247-spec.json` **byte-identically**, so the pair is sound. Clean-247 → clean-251 diff: **78 differing leaf pointers, 78 classified, 0 unexplained**:
  - **38 timestamp properties** across **19 resource components** gain `{type: [string, 'null'], format: date-time}` (e.g. `SiteResource.created_at`: `{}` → typed);
  - **1 list-cast retype**: `SiteResource.aliases` `{type: object}` → `{type: array, items: {type: string}}` — the exact #249 motivating case.
- **Degradation-notice volume** (controlled, log truncated between runs): base af95f95 emits **24** unconstrained-keys notices on clean Vito, **19 mentioning created_at/updated_at**; at 8566e71 it's **22 notices, 0 mentioning timestamps** (two notices vanish entirely — their only unconstrained keys were the timestamps). The dominant noise term from the #247 review is gone.

### Non-blocking notes

- `non-empty-list<T>` / `non-empty-array<int, T>` aren't recognised as list shapes and keep the `object` default — fine for now, cheap future widening of `listValueType()` if the corpus ever shows them on cast columns.
- Filed **#252**: class-form casts (`AsCollection::class`, `AsArrayObject::class`) bypass `castToProperty()` entirely — column documents fully untyped and the `@property` tag (including this PR's list disambiguation) never gets a say. Pre-existing, found while probing; not a #251 defect.
- Codecov's 2 uncovered patch lines are defensive fallthroughs; not worth chasing.

Radiergummi:

**1. [important] `jsonCastDefinition()` emits an items-less `{type: array}` that crashes `openapi:generate` under swagger-php validation.**
`EloquentModelToSchema::jsonCastDefinition()` (`src/Support/Extraction/EloquentModelToSchema.php:574-584`) sets `items` only when the list element resolves to a *scalar keyword*. `@property array<int, Carbon> $slots` + `casts = ['slots' => 'array']` emits `slots: {type: array}` with no `items`; swagger-php's validator then raises `@OA\Items() is required when @OA\Property(property="slots") has type "array"` — the mechanism that makes generate exit non-zero. Probe ran the full pipeline (model + typed-return controller → component → `Analysis::validate()`) and reproduced the error.

This is the third instance of the #265 family, and the shape is **pinned as intended** by the test "emits a bare array when the list element is not a scalar keyword" (`tests/Unit/Support/Extraction/EloquentModelToSchemaTest.php:264`) and by this PR's decision record ("`list<Carbon>` emits a bare `{type: array}`") — a conscious schema shape whose validation-crash consequence went unnoticed. The identical shape in `TransformerTransformReader::castType()` was found blocking in the #261 review and fixed there with `items: new OA\Items([])`; this is the one-line parity. (I've also commented on #265 — the cleanest fix is a single guard at the `SchemaFromArrayDefinition`/emission chokepoint covering the whole family.)

Radiergummi:
Fix landed for the stack-review crash finding: `jsonCastDefinition` now keeps an unconstrained `items` object for non-scalar list elements (`array<int, Carbon>` + `'array'` cast), matching the #261 fix and the chokepoint guard — swagger-php validation accepts the shape on both supported majors. The pinning test now asserts the items object. Commit `2b2413f`.

Issue #243, opened 2026-06-10
Query parameters are concatenated across resolvers without name dedup


Radiergummi:
## Summary

`OperationBuilder::build()` collects query parameters by concatenating every `QueryParameterResolver`'s output (`src/Support/Generator/OperationBuilder.php`, the `$queryParams` loop) — there is no name-level dedup **across** resolvers. Two resolvers emitting the same `name`+`in` produce a duplicate parameter entry, which is invalid OpenAPI (parameter objects must be unique on name+location).

## How it can happen

Pre-existing example: `#[QueryParam('sort')]` (Core) plus `#[AllowedSort]` (QueryBuilder plugin) on the same action — both emit a `sort` parameter. #241 makes the collision class more reachable without explicit user intent: the Core resolver now *infers* parameters from `$request->query(...)` reads and GET inline-validate rules, so e.g. a controller reading `$request->input('filter.name')` next to an `#[AllowedFilter('name')]` attribute yields two `filter[name]` entries.

Within the Core resolver itself precedence is handled (accessor < validate rules < `#[QueryParam]`, pinned by tests in #241); this issue is only about the cross-resolver seam.

## Proposed direction

Either dedup by `name`+`in` in `OperationBuilder` after the resolver loop (last or first resolver wins — needs a deliberate precedence decision, since resolver order is plugin order), or keep concatenation and add a lint rule flagging duplicate parameter names per operation so the collision is at least visible. Worth deciding before more parameter-emitting resolvers land (#15, QueryBuilder chain inference, will widen the surface again).

## References

Observed while implementing #11 (PR #241). Entry point: `src/Support/Generator/OperationBuilder.php` (query-resolver loop and `parameters:` assembly).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #256, opened 2026-06-10
feat(#15): document spatie/laravel-query-builder fluent chains as query parameters

### LINKED ISSUE #15: `spatie/laravel-query-builder` chain detection
## Summary

Read the `QueryBuilder::for(...)->allowedFilters([...])->allowedSorts([...])->allowedIncludes([...])` fluent chain straight from a controller method body and emit the matching `filter[...]`, `sort`, and `include` query parameters — so a Spatie Query Builder API documents itself without any authoring attributes.

## Motivation

`spatie/laravel-query-builder` is the idiomatic way Laravel apps expose filtering, sorting, and includes, and the convention *is* the fluent chain in the method body. Today the bundled QueryBuilder plugin is attribute-gated (`#[AllowedFilter]` / `#[AllowedSort]` / `#[AllowedInclude]` under `src/Plugins/QueryBuilder/Attributes/`): unless a developer re-declares everything on attributes, the chain produces nothing and the generated spec silently omits every query parameter the endpoint actually accepts. Since the generator reads signatures only and never the body (OAPI-017), the most common form of this convention is invisible.

## Proposed approach

Tier 1, scoped to the literal fluent chain rooted at `QueryBuilder::for(...)` on a known class. Reuse the shared bounded-AST body-scan utility introduced by the parent epic (the `src/Plugins/Core` body-scan utility, see #5) to match the chain in the first N statements: walk `->allowedFilters([...])`, `->allowedSorts([...])`, and `->allowedIncludes([...])` calls off the `QueryBuilder::for(...)` root and collect their literal array arguments. Feed the result through the existing `QueryBuilderParameterResolver` (`src/Plugins/QueryBuilder/Resolvers/QueryBuilderParameterResolver.php`) so the chain and the attributes converge on the same output (`filter[...]`, `sort`, `include`). Keep the matcher narrow: anything more indirect — a builder assigned to a variable and mutated, a chain on an unknown root, dynamically computed allow-lists — is refused, falls back to today's attribute-gated behaviour, and records a generation-log note. An explicit attribute always wins.

## Acceptance criteria

- A controller using a literal `QueryBuilder::for(...)` chain yields the corresponding `filter[...]`, `sort`, and `include` query parameters with no attributes present.
- Indirect or dynamic builders degrade gracefully: no params inferred, attributes still honoured, and a log note is emitted.
- Explicit `#[AllowedFilter]` / `#[AllowedSort]` / `#[AllowedInclude]` continue to take precedence over chain inference.
- Add before/after fixture tests covering the three behaviours (literal chain, indirect-builder degradation, attribute precedence) in `tests/Feature/Plugins/QueryBuilder/` (alongside the existing `QueryBuilderParameterTest.php`), with unit coverage for the chain-matching path in `tests/Unit/Plugins/QueryBuilder/QueryBuilderParameterResolverTest.php`.

## References

Parent: #5 (the Tier-1 bounded-AST mechanism, the shared `src/Core` body-scan utility, and the graceful-degradation contract live there). Related: #11 (query params from `$request->*()` accessors — the other body-scan-derived query-parameter source). Gap: OAPI-017 (no controller method-body inference today). Affected areas: `src/Plugins/QueryBuilder`, the `src/Plugins/Core` body-scan utility.

Verify: `vendor/bin/pest tests/Feature/Plugins/QueryBuilder/ tests/Unit/Plugins/QueryBuilder/QueryBuilderParameterResolverTest.php`

Radiergummi:
Closes #15

> **Stack:** #219 ← #225 ← #230 ← #235 ← #241 ← #247 ← #251 ← #253 ← **this** — based on `feat/108-return-expression-resource-type`; merge after #253.

## Implementation plan

Tier-1 bounded scan (epic #5) for the literal `QueryBuilder::for(...)->allowedFilters([...])->allowedSorts([...])->allowedIncludes([...])` chain, feeding the existing `QueryBuilderParameterResolver` so chain and attributes converge on identical `filter[...]` / `sort` / `include` output.

1. **`Plugins\QueryBuilder\Support\QueryBuilderChainReader`** (`@internal`, scoped, memoised per method) — scans the first 10 top-level statements via the shared `MethodBodyScanner` + `StatementNodeFinder`:
   - Matches `allowedFilters` / `allowedSorts` / `allowedIncludes` calls whose receiver spine roots at a `StaticCall` to `for()` on the **resolved FQCN** `Spatie\QueryBuilder\QueryBuilder` — an impostor `QueryBuilder` class never matches. Other chain links (`->defaultSort()`, `->where()`, `->paginate()`, …) are walked through: chain termination is irrelevant to the accepted parameters.
   - Only single-expression chains on the unconditional path (statement expression, assignment right-hand side, or `return`). A builder assigned to a variable and *mutated across statements* is refused (per spec) — degrade + NOTICE.
   - Elements: string literals (via `AstLiteralEvaluator`, so class-constant strings work) and Spatie value-object static constructors where the wire name is the first literal argument. Spatie's variadic form (`allowedSorts('name', 'created_at')`) reads the same way.
   - Partial-keep (the #9 precedent): a non-literal element is dropped, the rest kept, NOTICE emitted. A wholly dynamic allow-list degrades that call. A `QueryBuilder::for` sighting that yields nothing readable emits the found-but-unreadable NOTICE pointing at the attributes as escape hatch.
2. **`QueryBuilderParameterResolver`** — composes per kind: attributes win **per kind** (any `#[AllowedFilter]` present → filters come from attributes only; same for sort/include). Chain fills only attribute-less kinds; the body scan runs only when at least one kind has no attribute (Tier-0-miss discipline). Output is built through the same `filterParameter()` / `listParameter()` paths as attributes.
3. **Tests** — unit: chain matching per kind, value-object shapes, partial-keep, dynamic-array degrade, impostor-root refusal, variable-mutation refusal, >10-statement miss, precedence granularity. Feature fixtures in `tests/Feature/Plugins/QueryBuilder/`: literal chain → params; indirect builder → degrade + attributes honoured; attribute precedence.
4. **Example** — one idiomatic chain-only route in `examples/query-builder` + snapshot regeneration.
5. **Docs** — `docs/plugins.md` QueryBuilder section, `docs/auto-derivation.md` query-parameters row, `CHANGELOG.md`.

## Pinned decisions

- **Value-object matrix:** whitelisted static constructors per VO class — `AllowedFilter::exact/partial/beginsWith/endsWith/belongsTo/scope/callback/trashed/custom/operator/groupOr/groupAnd`, `AllowedSort::field/custom/callback`, `AllowedInclude::relationship/count/exists/min/max/sum/avg/callback/custom`. All take the public wire name first; the constructor kind affects SQL semantics, not the parameter name. `trashed()` without arguments defaults to `trashed` (Spatie's default). Internal names (second argument) are server-side detail and ignored — matching the attribute vocabulary, which has no internal-name concept.
- **Precedence granularity:** per kind, not per param — if any attribute of a kind is present, that kind comes from attributes only. Mixing one source per name inside a kind would make the document's provenance unreadable.
- **`->defaultSort()`:** not read. Neither `#[AllowedSort]` nor the resolver models a sort default, so reading it would create chain-only vocabulary the attribute path cannot converge on. Candidate follow-up if the attribute ever grows a default.
- **Variable-chain boundary:** the two-statement form `$q = QueryBuilder::for(X)->allowedFilters([...]); return $q->paginate();` works (the chain is one expression); `$q = QueryBuilder::for(X); $q->allowedFilters([...]);` is refused per the issue — no variable tracking (Tier-2).
- **Config naming:** the existing resolver hardcodes `filter[…]`/`sort`/`include` and does not read Spatie's `query-builder.parameters` rename config; the chain path funnels through the same emit methods, so behaviour stays consistent. Honouring the rename config would be a separate change covering both sources.


Radiergummi:
## Implementation decisions (as committed in c4966d0)

Two implementer sessions crashed mid-flight (infra); the orchestrator verified and finished the work inline. Decisions extracted from the committed code:

1. **Value-object support matrix** — string literals plus the full set of Spatie static constructors whose first parameter is the public wire name: `AllowedFilter::` `exact/partial/beginsWith/endsWith/belongsTo/scope/callback/trashed/custom/operator/groupOr/groupAnd`; `AllowedSort::field/custom/callback`; `AllowedInclude::relationship/count/exists/min/max/sum/avg/callback/custom`. The constructor kind changes SQL semantics, never the parameter name, so the name-first convention is safe across the whole matrix. Internal-name second arguments are ignored (the wire name is what's documented).
2. **Precedence granularity: per kind.** Any `#[AllowedFilter]` attribute present → filters come from attributes only; the chain fills only attribute-less kinds. The body scan is skipped entirely when all three kinds are attributed.
3. **Degradation** — non-literal array element: dropped, rest kept (+ NOTICE naming the element). Dynamic array, impostor `QueryBuilder` root, builder mutated across statements, chain inside a conditional context: that source degrades to nothing + NOTICE pointing at the attributes as the escape hatch. Memoised per `Class::method` so repeated resolution (route verbs + lint) parses once.
4. **`defaultSort()`** — not read: the existing attribute resolver does not model sort defaults, and the chain/attribute outputs must stay convergent. Noted as future work if `#[AllowedSort]` ever grows a default.
5. **Spatie config naming** — the resolver already derives `filter[...]`/`sort`/`include` names the same way for attributes; chain output funnels through the same builders, so renamed parameter config behaves identically for both sources.
6. **PHPStan** — one scoped, commented ignore: Spatie types `allowedFilters()` as variadic `AllowedFilter|string` but flattens a single array argument at runtime; the array form is the documented idiom the reader targets, so fixtures must use it.

Gates on c4966d0: `composer test` 2248 passed (5620 assertions); PHPStan L8 no errors; Pint passed; snapshot group 9 passed — the only example delta is the new chain-only `GET /api/bookings` route (filter[seat], filter[passenger_name], sort enum from allowedSorts, include) in the query-builder flavor.

Radiergummi:
## Review: empirical test report (head `c4966d0`, delta vs `feat/108-return-expression-resource-type`)

**Verdict: APPROVE** — one rebase prerequisite, one follow-up issue filed (#257), no changes requested to the #256 delta itself.

### Gates (fresh worktree at `c4966d0`, PHP 8.4)

- `composer test`: **2248 passed (5620 assertions)** on re-run. First run had one parallel flake in `ExampleFileLoaderTest` — that is #222's fixture-copy race, fixed on `main` (9680cd2) but not in this stack's base; passes in isolation and on re-run. Not a #256 problem.
- `composer lint` (PHPStan L8): no errors.
- `vendor/bin/pint --test`: passed.

### Findings (numbered, severity-tagged)

1. **[INFO — resolved by rebase, but verify] Duplicate `(name,in)` pairs with the #11 accessor scan at this stack base.** Probed the #243 lesson: a method with both `QueryBuilder::for(...)->allowedSorts([...])` and `$request->query('sort')` produces **duplicate `sort|query` / `include|query` parameters** at `c4966d0` — because the #243 dedup fix (4884ddb) is **not an ancestor** of this stack (forked at 04194eb). Verified by trial-merging `origin/main` into the head: the dedup applies and the QueryBuilder plugin's richer enum'd `sort`/`include` correctly replace Core's plain accessor-derived ones (later-resolver-wins). No code change needed here; just don't merge the stack without rebasing onto current `main`.
2. **[MEDIUM — stack maintenance, not #256] Rebase semantic conflict ahead.** The same trial merge surfaces 24 test failures: `main` changed `JsonSchemaFromType::__construct` to two parameters, while the stack's `tests/Support/SchemaFromResourceFactory.php:65` passes one (`ArgumentCountError`). Textual conflicts are docs-only (CHANGELOG, auto-derivation, linting). Mechanical fix at rebase time — flagging so it lands deliberately, not as a surprise.
3. **[LOW — follow-up filed: #257] Fluent modifier calls on value-object constructors drop the element.** `AllowedFilter::exact('healthy')->nullable()` is a `MethodCall` wrapping the constructor, so `elementName()` refuses it: dropped + partial-keep NOTICE (contract-compliant). It's a real corpus idiom — SpeedtestTracker's `ResultsController` loses `filter[healthy]` to exactly this. All Spatie modifiers (`nullable/default/ignore/delimiter`, `AllowedSort::defaultDirection`) return `static` and never rename, so unwrapping the spine (like `rootsAtBuilder()` already does) closes it cheaply. Filed as #257.
4. **[NIT] Partial-keep notice can fire for an attribute-covered kind.** The scan runs whenever ≥1 kind is unattributed, and `unreadableCalls` isn't filtered by which kinds the attributes cover — `#[AllowedFilter]` + a chain whose `allowedFilters` has a dynamic element (plus a readable `allowedSorts`) logs "allowedFilters element(s) … dropped" even though filters come from attributes and nothing was lost. Log-noise only; fine to leave.
5. **[OK] PHPStan ignore is properly scoped.** `paths: tests/` only, regex pinned to `Spatie\QueryBuilder\QueryBuilder<…>::allowed(Filters|Sorts|Includes)()` variadic-vs-array — it cannot mask a package-code bug.
6. **[OK] Quality.** `@internal` + `#[Scoped]` on reader and resolver (matches the codebase's 80-class scoped convention — memoisation is per-run, Octane-safe), `// region` folding, verbose naming, docs (plugins.md QueryBuilder section + auto-derivation query-parameters row) accurate against observed behaviour, CHANGELOG entry present. The value-object whitelist matches Spatie 7.3.0's **complete** static-constructor surface (verified against vendor source; every constructor is name-first; `trashed()` default correct).

### Empirical probes (throwaway fixtures, removed after)

All confirmed at `c4966d0`: literal chain in return / assignment RHS; variadic form; value-object constructors including **named arguments** (`operator(name: 'start_at', internalName: 'created_at', …)` → `filter[start_at]`, internal name ignored); split-statement builder refused + NOTICE; chain inside `if` **and inside a closure** degraded; impostor `QueryBuilder` never matches; >10 statements not scanned; `QueryBuilder::for()` root argument never analysed (a dynamic relation/builder argument doesn't impede the chain); unlisted/fictional constructor dropped with the rest kept; per-kind precedence (attribute filters + chain sorts compose correctly); fully-attributed action provably skips the scan (a dynamic chain under full attribution emits zero notices). **Convergence:** chain-derived and attribute-derived `filter[...]`/`sort`/`include` parameters are byte-identical structures for the same allow-list.

### Survey (Layer A, #233 procedure, sole runner)

Full 11-app serialized run (`results-pr-256.json`, per-app composer-state repair + post-run `readlink` verification — all 11 linked to the review-15 worktree). **Zero deltas vs `results-pr-253.json` on every metric, lint count, and crash field, for all 11 apps.**

Honest framing: the QueryBuilderPlugin is **commented out in every corpus app's published config** (and in the package default), so the chain scan cannot fire in Layer A — no corpus movement was *possible*; the run is a pure regression gate, and it's clean. Grep-first census: Pelican (11 controllers), SpeedtestTracker (2), AdvisingApp + AureusERP (modules/plugins dirs) use genuine Spatie `QueryBuilder::for(`; **Vito's `QueryBuilder` is its own `App\Helpers\QueryBuilder`** — a live demonstration of why the resolved-FQCN impostor guard matters.

**Opt-in spot-check** (plugin temporarily enabled for SpeedtestTracker + Pelican, then config + artifacts restored):
- SpeedtestTracker `GET /api/v1/results`: `filter[ping|download|upload|status|scheduled|start_at|end_at]` + `sort` enum `[ping, download, upload, created_at, updated_at]` — exactly the controller's allow-lists; `filter[healthy]` missing per finding 3 with the partial-keep NOTICE in the app log. `GET /api/v1/stats`: `filter[start_at]`, `filter[end_at]` read through ten `->selectRaw()` links.
- Pelican `GET /api/application/nodes`: `filter[uuid|name|fqdn|daemon_token_id]` + sort enum `[id, uuid, memory, disk, cpu]` — exact match to `NodeController`; the accessor-derived `per_page` composes alongside without collision.

Worth considering for future surveys: enabling the QueryBuilderPlugin in the four Spatie-using corpus apps would let Layer A measure this feature (and #257) for real.

Radiergummi:

**1. [important] Chain-inferred parameters silently replace explicit `#[QueryParam]` attributes and richer validate-derived schemas.**
`QueryBuilderParameterResolver` has no awareness of Core's `#[QueryParam]`, and the cross-resolver dedup (`OperationBuilder.php:127-145`, from #243) is unconditional later-wins on name+in. Probe (GET action carrying `#[QueryParam('sort', type: 'string', description: 'Author-described sort param.')]`, `validate(['filter.name' => 'required|integer|min:5'])`, and a literal `QueryBuilder::for()->allowedFilters(['name'])->allowedSorts(['id'])` chain):

- `sort`: the author's attribute (description + type) is replaced by the chain's canned enum parameter — an explicit attribute loses to a Tier-1 inference.
- `filter[name]`: Core alone emits `{type: integer, minimum: 5}` from the validate rule; with the chain present it degrades to `{type: string}` — a typed inference replaced by a typeless one.

Why this isn't covered by the #243 decision: "later resolver wins — a plugin's **richer** parameter replaces an earlier one" was decided when both colliding sources were explicit attributes. This PR (and #241) changed the population flowing through that dedup — inferred params now collide with explicit ones, and "later" no longer implies "richer". It also violates the stack's own invariant ("explicit authoring always wins"), which every response-side resolver implements via the `PrimaryResponseAuthoringAttribute` guard but the parameter side does not. The plugin even implements the correct rule *internally* (attribute-over-chain per kind) — the cross-plugin equivalent is missing. Minimal fix: skip names already claimed by `#[QueryParam]` on the action, or mark attribute-sourced parameters so the OperationBuilder dedup prefers them (the `EXPLICIT_STATUS_EXTENSION` pattern exists for exactly this).

Radiergummi:
Fix landed for the stack-review precedence finding: the cross-resolver dedup in `OperationBuilder` now keeps the first (attribute-shaped) emission for any name claimed by an explicit `#[QueryParam]` on the action/controller — central on purpose, so every future query resolver inherits 'explicit authoring wins' instead of re-implementing it. Regression test: a chain `allowedSorts` no longer replaces an authored `sort` attribute. The second sub-case (typed validate-derived schema vs typeless chain param, both inferred) is unchanged — adjudicating relative richness between two inferences is follow-up material, not a precedence-contract violation. Commit `49854e2`.

PR #258, opened 2026-06-11
feat(#16): detect constructor $this->middleware() — crash-safe fallback for non-instantiable controllers

### LINKED ISSUE #16: `$this->middleware()` constructor detection
## Summary

Detect auth (and other) middleware that a controller applies imperatively in its `__construct` — e.g. `$this->middleware('auth:sanctum')` — so it counts toward the same signals as route-declared middleware.

## Motivation

A lot of older or convention-light controllers never declare `auth` on the route; they wire it up in the constructor instead. Because the generator only sees route-gathered middleware, those endpoints currently look unauthenticated: no security scheme, no implicit 401/403. That is the same misleading "spec says public, app says protected" failure that the security work is trying to stamp out, just reached by a different path. Closing this is what lets constructor-middleware controllers feed the Sanctum security work (#32) and the implicit 401/403 inference (#29) on equal footing with route middleware.

## Proposed approach

Tier 1 — a narrow constructor-body call-shape whitelist, built on the shared bounded-AST body-scan utility introduced by the parent epic (do not introduce dataflow; see #5 for the mechanism and boundary). That utility does not exist yet — it is a deliverable of #5, so this feature lands on top of it rather than against an existing file.

Scan the controller's `__construct` for `$this->middleware(...)` calls with literal string arguments, honouring the fluent `->only([...])` / `->except([...])` scoping so a middleware that is constructor-scoped to a subset of methods only attaches to the matching operations. Merge the discovered middleware names into the same middleware list that `SecurityExtractor::forRoute()` (in `src/Support/Extraction/SecurityExtractor.php`) consumes today — `forRoute()` already drives both `hasAuthMiddleware()` (security) and scope extraction, so constructor middleware flows into security-scheme derivation and the implicit-error contributor with no change to those consumers. Anything outside the whitelist (dynamic middleware names, non-literal scoping) degrades gracefully to today's behaviour plus a generation-log note, and an explicit `#[Security]` / `#[PublicEndpoint]` attribute always wins.

## Acceptance criteria

- A controller whose constructor calls `$this->middleware('auth:sanctum')` contributes to that operation's per-operation security and implicit 401.
- `->only(...)` / `->except(...)` scoping is respected: middleware attaches only to the in-scope operations.
- Non-whitelisted constructor shapes leave generation unchanged and emit a log note.
- Tests cover the whitelist matcher and the scoping cases. Add the security-merge assertions to `tests/Unit/Support/Extraction/SecurityExtractorTest.php`, and add body-scan/whitelist cases alongside the shared utility's tests under `tests/Unit/`.

## References

Parent: #5 (shared bounded-AST body-scan utility, Tier-1 boundary, graceful-degradation contract). Feeds #32 (Sanctum security) and #29 (implicit 401/403). Affected areas: the `src/Plugins/Core` body-scan utility delivered by #5, `SecurityExtractor` (`src/Support/Extraction/SecurityExtractor.php`), and the middleware error contributor. Verify: `vendor/bin/pest tests/Unit/Support/Extraction/SecurityExtractorTest.php`

Radiergummi:
## Summary

Detect auth (and other) middleware that a controller applies imperatively in its `__construct` — `$this->middleware('auth:sanctum')->only([...])` — so constructor-middleware controllers feed security-scheme derivation and the implicit 401/403 inference on equal footing with route-declared middleware. Closes #16.

**Stack:** #219 ← #225 ← #230 ← #235 ← #241 ← #247 ← #248 ← #251 ← #253 ← #256 ← this (base: `feat/15-querybuilder-chain-detection`).

## Reality check (changes the plan vs. the issue text)

Probed against the Laravel 12 vendor code and Testbench before designing:

- `Route::gatherMiddleware()` **already** resolves classic constructor middleware — `controllerMiddleware()` instantiates the controller through the container and reads its recorded middleware, honouring `only`/`except`. For instantiable controllers there is **no gap today**.
- The `HasMiddleware` static `middleware()` form (Laravel 11+) is also resolved natively (`staticallyProvidedControllerMiddleware()`, no instantiation, `only:`/`except:` honoured).
- The **actual gap**: when the controller cannot be instantiated in the generation context (unbound constructor dependency, throwing side effects — the BookStack-style constructor), `gatherMiddleware()` throws `BindingResolutionException` and **crashes the whole generation run**. Worse, `gatherMiddleware()` poisons its `computedMiddleware` cache on the way down: a second call silently returns `[]`.

So the static constructor scan is the **fallback** for exactly that case; the native runtime path stays primary (a Tier-0 hit — per the `MethodBodyScanner` contract the body scan only runs on a Tier-0 miss).

## Plan

1. **`Support\Routing\ConstructorMiddlewareScanner`** (`@internal`, scoped) — Tier-1 whitelist over the first 10 top-level statements of `__construct`:
   - `$this->middleware(<literal string | array-of-strings>)`, fluent `->only(<literal>)` / `->except(<literal>)`, and the equivalent `['only' => …]` / `['except' => …]` options-array argument.
   - Literals via `AstLiteralEvaluator` (class constants work); scoping matched against the route's action method.
   - `SkipConditionalContexts`: a conditionally-applied `$this->middleware()` is **not** documented (documenting conditional auth as always-required overstates the contract) — it degrades to a generation-log NOTICE, like any non-literal name/scope.
   - Inheritance: `ReflectionClass::getConstructor()` lands on the declaring class, so a base-controller constructor applying middleware is scanned through its own file. Pinned with a parent-constructor fixture.
2. **`Support\Routing\RouteMiddlewareGatherer`** (`@internal`, scoped) — the single middleware-assembly seam: try `$route->gatherMiddleware()`; on `Throwable`, NOTICE + fall back to route-declared middleware merged (deduped) with the constructor scan. Caches per route, immune to the poisoned-cache footgun.
3. **Consumers** — all four current `gatherMiddleware()` readers switch to the gatherer, so the merged list is uniform everywhere: `SecurityExtractor` (security schemes + scopes), `MiddlewareErrorContributor` (implicit 401/403/429), `InclusionEvaluator` (spec `match.middleware`), `WhyCommand` (display). No consumer logic changes. `#[Security]` / `#[PublicEndpoint]` precedence is untouched (checked in `OperationBuilder::resolveSecurity()` before the extractor runs) and pinned by a feature test.
4. **HasMiddleware decision:** in scope as a *pinning test + docs* only — the framework already resolves it without instantiation; no new code needed. If the static `middleware()` itself throws, the gatherer's fallback catches that too.

### Layering

The scanner emulates Laravel's own `ControllerDispatcher::getMiddleware()` contract — framework semantics, not an app convention — and its primary consumer (`SecurityExtractor`) is plugin-agnostic `Support\` infrastructure that must not depend on `Plugins\Core`. So both new classes live in `Support\Routing`, next to `RouteIntrospector`, following the `Support\MethodBody\SingleReturnArrayLiteralFinder` precedent (scan machinery in Support, plugin-specific whitelists with their plugin consumers). A registry seam (`Contracts\Registry` middleware contributor) would be speculative DI for a single implementation.

## Tests

- Unit: scanner whitelist + scoping matrix (string/array forms, fluent + options-array scoping, non-literal name, non-literal scope, conditional context, parent constructor, no constructor, receiver discipline, unknown chain link).
- Unit: gatherer happy-path passthrough, throwing-constructor fallback + dedup, cache stability.
- `SecurityExtractorTest`: constructor middleware drives the requirement; dedup against route-declared.
- Feature end-to-end: non-instantiable constructor-middleware controller → sanctum security + 401 on in-scope operations only; `only`/`except`; degrade NOTICE; `#[PublicEndpoint]` wins; `HasMiddleware` pinning.

## Docs

Security docs page (constructor middleware + HasMiddleware + degrade conditions), auto-derivation row, CHANGELOG.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,25 @@ All notable changes to this project are documented here.
 
 ### Added
 
+- **Constructor `$this->middleware(...)` now survives non-instantiable controllers** (#16). Laravel
+  itself resolves controller middleware when routes are gathered — the static
+  `HasMiddleware::middleware()` form without instantiation, the classic constructor form by
+  instantiating the controller through the container, both honouring `only` / `except`. The gap was
+  the controller the container *cannot* build at generation time (unbound constructor dependency, a
+  constructor that throws outside a real request): `Route::gatherMiddleware()` threw, crashed the
+  whole run, and poisoned the route's middleware cache so a retry silently returned `[]`. All
+  middleware reads now flow through one crash-safe seam (`RouteMiddlewareGatherer`): on an
+  instantiation failure it logs a notice and falls back to the route-declared middleware merged with
+  a bounded static scan of the constructor body (Tier-1, first 10 top-level statements, epic #5) —
+  literal `$this->middleware(...)` names with literal fluent `->only(...)` / `->except(...)` or
+  options-array scoping, matched against the route's action method, deduplicated against
+  route-declared names. Inherited (base-controller) constructors are read through their declaring
+  class. Security schemes, per-operation `security`, the implicit 401/403/429 responses, multi-spec
+  `match.middleware`, and `openapi:why` all see the same merged list; `#[Security]` /
+  `#[PublicEndpoint]` keep winning. Dynamic names, non-literal scoping, and conditionally applied
+  registrations are refused with a generation-log notice (documenting conditional auth as
+  unconditional would overstate the contract). See
+  [Auto-derivation → Controller middleware](docs/auto-derivation.md#controller-middleware).
 - **`spatie/laravel-query-builder` fluent chains now document themselves** (#15). The QueryBuilder
   plugin reads the literal `QueryBuilder::for(...)->allowedFilters([...])->allowedSorts([...])
   ->allowedIncludes([...])` chain straight from the controller method body (Tier-1 bounded scan of
--- a/docs/auto-derivation.md
+++ b/docs/auto-derivation.md
@@ -14,8 +14,8 @@ source.
 | Query parameters | Request-accessor reads in the method body — `$request->query('sort')` / `input('q')` → `string`, `string('name')` / `integer('page')` / `boolean('active')` → their named type (bounded scan; see [Query parameters from the method body](#query-parameters-from-the-method-body)). On GET/HEAD routes, inline `validate()` keys become query parameters (name, schema, `required` from the rules) instead of a request body. With the QueryBuilder plugin enabled, a literal `QueryBuilder::for(...)` chain's `allowedFilters` / `allowedSorts` / `allowedIncludes` become `filter[…]` / `sort` / `include` parameters (see [Plugins → QueryBuilder](plugins.md#querybuilder)). `#[QueryParam]` attributes win for their name; other names compose. |
 | Request body | Spatie Data class on the action (or on a configured payload-indirection object); `FormRequest` is supported natively. Schema is built from PHP types and validation rules. Without a typed payload parameter, inline `validate()` calls and a controller-declared `$rules` property / `rules()` method are read from the method body (bounded scan; see [Request bodies → Inline validation in the controller](request-bodies.md#inline-validation-in-the-controller)). |
 | Response body | Spatie Data class or `DataCollection<…>` return type → component `$ref`. `JsonResource` subclass → component schema: fields are inferred from a single-`return [...]` `toArray()` literal (`$this->field` typed from the wrapped `@mixin`/`@extends` model, nested resources as `$ref`s, `when*` wrappers optional) composed with declared `#[ResourceField]` attributes, which win per field; a passthrough or dynamic `toArray()` falls back to the wrapped model's schema (see [Plugins → ApiResources](plugins.md#apiresources)). A **base** resource return type (`JsonResource`, bare `ResourceCollection`, `AnonymousResourceCollection`) resolves the concrete resource from the method's return expression — `X::collection(…)` / `X::make(…)` / `new X(…)`, `->toResource(…)`, or a `@return …Collection<X>` generic; a collection only claims the paginated envelope when its source visibly ends in a `paginate()`-family call, looking through paginator-preserving links like `withQueryString()` (bounded scan; see [Plugins → ApiResources → Resolving the resource from the return expression](plugins.md#resolving-the-resource-from-the-return-expression)). Eloquent `Model` subclass → component schema built from `$casts`, `@property`/`@property-read` annotations, typed `$appends` accessors, and `$hidden`/`$visible`. See [Eloquent model response schemas](#eloquent-model-response-schemas). Without a schema-bearing return type, a literal `response()->json([...])` in the method body is read instead (bounded scan; see [Inline JSON responses](#inline-json-responses)). |
-| Security | `auth:*` / `scope:*` / `scopes:*` (and Sanctum's `abilities:*` / `ability:*`) middleware → a per-operation `security` requirement against the derived scheme(s): Passport's OAuth2 flows, a `sanctum` http/bearer scheme when any route uses `auth:sanctum`, or `openapi.security_schemes`. Sanctum's all-of `abilities:a,b` lists both as scopes on one requirement; its any-of `ability:a,b` emits one OR-alternative requirement per ability. Map project-specific guard middleware to a declared scheme via `openapi.security_middleware_map`. When the route is authed but no scheme is derivable, `security` is omitted (not `[]`, which means *public*) and `operation.security-missing` flags it. |
-| Error responses | `@throws ExceptionClass` → status codes; `abort(403)` / `abort_if(…, 404, 'msg')` / `abort_unless(…, 403, 'msg')` in the method body → that status, with a literal message as the response description (bounded scan of the first 10 statements; class-constant statuses such as `abort(Response::HTTP_FORBIDDEN, …)` resolve too, a genuinely non-literal status is skipped with a generation-log note); a route-model-bound parameter (`show(Post $post)`) → 404; a `FormRequest` parameter → 422; `auth`/`scope`/`can`/`throttle` middleware → 401 / 403 / 403 / 429. An explicit `#[Response]` for the same status always wins. |
+| Security | `auth:*` / `scope:*` / `scopes:*` (and Sanctum's `abilities:*` / `ability:*`) middleware — route-declared or controller-applied, including constructor `$this->middleware(...)` and the static `HasMiddleware` form (see [Controller middleware](#controller-middleware)) — → a per-operation `security` requirement against the derived scheme(s): Passport's OAuth2 flows, a `sanctum` http/bearer scheme when any route uses `auth:sanctum`, or `openapi.security_schemes`. Sanctum's all-of `abilities:a,b` lists both as scopes on one requirement; its any-of `ability:a,b` emits one OR-alternative requirement per ability. Map project-specific guard middleware to a declared scheme via `openapi.security_middleware_map`. When the route is authed but no scheme is derivable, `security` is omitted (not `[]`, which means *public*) and `operation.security-missing` flags it. |
+| Error responses | `@throws ExceptionClass` → status codes; `abort(403)` / `abort_if(…, 404, 'msg')` / `abort_unless(…, 403, 'msg')` in the method body → that status, with a literal message as the response description (bounded scan of the first 10 statements; class-constant statuses such as `abort(Response::HTTP_FORBIDDEN, …)` resolve too, a genuinely non-literal status is skipped with a generation-log note); a route-model-bound parameter (`show(Post $post)`) → 404; a `FormRequest` parameter → 422; `auth`/`scope`/`can`/`throttle` middleware (route-declared or controller-applied, see [Controller middleware](#controller-middleware)) → 401 / 403 / 403 / 429. An explicit `#[Response]` for the same status always wins. |
 | Validation constraints | `Data::rules()` and Spatie validation attributes → `maxLength`, `minLength`, `pattern`, `enum`, `format`, `minimum`/`maximum`, `minItems`/`maxItems`. |
 
 Use an authoring attribute when convention can't produce what you need. See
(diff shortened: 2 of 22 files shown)


Radiergummi:
## Decision record

**1. Issue premise was stale — feature reframed as a crash-safe fallback.** Probed in Testbench against Laravel 12 before designing: `Route::gatherMiddleware()` *already* returns classic constructor middleware (`only`/`except` honoured) by instantiating the controller, and resolves `HasMiddleware::middleware()` statically without instantiation. So for instantiable controllers there was no gap — endpoints did not "look unauthenticated". The real-world failure mode is the controller the container cannot build at generation time: `gatherMiddleware()` throws `BindingResolutionException`, which **crashed the entire run** (no fault isolation in `PathsStage`/`OperationBuilder`/`MiddlewareErrorContributor`), and poisons `Route::$computedMiddleware` so a retry silently returns `[]`. The static constructor scan therefore runs **only on that Tier-0 miss** — when the runtime gather throws — which also satisfies the `MethodBodyScanner` contract (scan only on a Tier-0 miss). All acceptance criteria of #16 are met; they are exercised through the non-instantiable fixture.

**2. HasMiddleware static `middleware()`: in scope as pinning test + docs only.** The framework resolves it natively; new code would duplicate Laravel. Pinned by `tests/Feature/ConstructorMiddlewareTest.php` ("documents HasMiddleware static middleware without any constructor scan") and documented in the new [Controller middleware] docs section. If the static method itself throws, the gatherer's catch covers it (fallback finds no constructor registrations, route-declared middleware survives). No follow-up issue needed.

**3. Merge seam: `Support\Routing\RouteMiddlewareGatherer`, adopted by all four readers.** Route middleware was read raw in `SecurityExtractor`, `MiddlewareErrorContributor`, `InclusionEvaluator` (spec `match.middleware`), and `WhyCommand`. Mutating the live `Route` objects was rejected (runtime mutation of shared router state, unacceptable under Octane); an `ActionDescriptor` property was rejected because `SecurityExtractor::forRoute()`/`sanctumInUse()` operate on `Route`s, not descriptors. All four consumers now inject the gatherer, so security, implicit 401/403/429, spec matching, and diagnostics see one uniform, deduplicated (`Router::uniqueMiddleware`) list. Layering: the scanner emulates Laravel's own `ControllerDispatcher::getMiddleware()` contract — framework semantics, not an app convention — and its primary consumer is plugin-agnostic `Support\` infrastructure, so both classes live in `Support\Routing` (the `SingleReturnArrayLiteralFinder` precedent); a `Contracts\Registry` contributor seam would be speculative DI for a single implementation.

**4. Conditional policy: `SkipConditionalContexts` + NOTICE.** A `$this->middleware()` inside an `if` is refused and logged — documenting conditional auth as unconditional overstates the contract, and on the happy path the runtime gather (which executed the real constructor) remains the accurate source. Understating is surfaced rather than hidden: the notice names the controller and points at `#[Security]`.

**5. Inheritance: handled by reflection, pinned by fixture.** `ReflectionClass::getConstructor()` reflects the declaring class, so the scanner reads the base controller's file (BookStack-style base constructors). Pinned by `ConstructorMiddlewareBaseController`/`ChildController` fixtures at both unit and feature level.

**6. Attribute precedence: untouched, pinned.** `OperationBuilder::resolveSecurity()` checks `#[PublicEndpoint]`/`#[Security]` before consulting the extractor; constructor-derived middleware changes only what the extractor sees. Pinned by the `#[PublicEndpoint]`-wins feature test. Note (existing behaviour, also for route-declared middleware): the implicit 401 from `MiddlewareErrorContributor` follows the middleware list regardless of `#[PublicEndpoint]` — the attribute clears `security` only. Out of scope here.

**7. Examples: no new route.** An instantiable example controller would exercise only Laravel's native path (zero observable diff attributable to this feature), and a deliberately non-instantiable controller is unnatural noise in a showcase app. The feature is exercised end-to-end by the fixture-backed feature tests instead; snapshots are untouched.

**8. Scoping notices fire once per controller class**, not per route, to keep many-route controllers from flooding the generation log.

Radiergummi:
## Review: empirical test report (PR #258, delta vs `feat/15-querybuilder-chain-detection`)

**Verdict: APPROVE.** Gates green at head `149ffeb` (2271 tests / 5673 assertions, PHPStan level 8 clean, Pint clean). The reframing is empirically correct on every claim — details below. No blocking findings; two pre-existing problems filed as #259 and #260.

### Position on the reframing

The deviation from the issue text is **exactly right**, and I verified each claim with throwaway Testbench probes on the merge base (`feat/15` head, Laravel v13.9.0), not just by reading the PR's own fixtures:

1. **Stale premise confirmed.** An *instantiable* controller with `$this->middleware('auth:sanctum')->only('index')` already produced the sanctum requirement + implicit 401 on `index` only (and nothing on the other action) **on the base branch** — the native `Route::gatherMiddleware()` → `controllerMiddleware()` path handles it, `only`/`except` included. Issue #16's "endpoints look unauthenticated" did not reproduce. (My first probe accidentally "reproduced" the issue with a non-PSR-4-autoloadable fixture — `method_exists()` on an unloadable class silently returns false and the native path degrades to `[]`. Worth knowing as a failure mode, but not an app-code-shaped one.)
2. **Crash confirmed.** With an unbound constructor dependency, base-branch generation dies with `BindingResolutionException`; the stack is `SecurityExtractor::expandedMiddlewareFor → Route::gatherMiddleware → controllerMiddleware → getController → Container::make` — one bad controller kills the entire run, exactly as the decision record says.
3. **Poisoning confirmed, with one nuance.** `gatherMiddleware()` assigns `$this->computedMiddleware = []` *before* the merge that throws, so a direct retry on the same `Route` object silently returns `[]` (probed). Nuance: per-route pipeline work runs on **compiled-collection clones** (`RouteIntrospector::discover()` → `toCompiledRouteCollection()`), while `SecurityExtractor::sanctumInUse()` iterates the **live** router routes — so both populations get gathered, and the gatherer's per-object cache + once-per-controller notice handle the duplication correctly.

On the PR head, my independent probes (instantiable / unbuildable / unbuildable-with-dynamic-name-only) all behaved as documented: native path untouched and silent, fallback recovers scoped middleware with the `could not be instantiated` notice, dynamic-only registration claims nothing and emits the `no statically readable name or scope` notice. `#[PublicEndpoint]`-wins and HasMiddleware pinning verified via the shipped feature tests. Probes were scratch-only and are cleaned up; both worktrees are pristine and the suite re-ran green after cleanup.

### Findings

1. **[LOW] `RouteMiddlewareGatherer` caches by `spl_object_id`** (`src/Support/Routing/RouteMiddlewareGatherer.php:73`). The compiled-clone routes are materialized per `discover()` call; PHP recycles object ids aggressively after GC, so in a process that generates several specs in one scoped lifetime (multi-spec `openapi:generate`), a freed clone's id can be reused by a different `Route` and serve a stale middleware list. In practice descriptors keep their routes alive through the run, so I could not construct a failing case — but `WeakMap<Route, array>` is a drop-in replacement that deletes the hazard class entirely (and the comment "Route objects live in the router for the whole generation run" is not true for the compiled clones the pipeline actually passes in). Suggestion, not a blocker.
2. **[LOW/observation] The caught throw still leaves the *live* route poisoned.** When `sanctumInUse()` gathers a live router route and the controller can't build, Laravel's own pre-assignment leaves `computedMiddleware = []` on the shared `Route` object after the gatherer catches. Under in-process generation (Octane, HTTP-triggered docs), a later real dispatch of that route calls `gatherMiddleware()` → `[]` → skips all middleware before failing on controller instantiation. Net behavior for a *deterministically* unbuildable controller is unchanged (the request 500s either way), and the mechanics are Laravel's, pre-existing — but the PR is what makes the process survive into that state, so consider restoring `computedMiddleware` to `null` in the catch (closure-bind), or noting it. Author's call; fine to defer.
3. **[NIT] Degenerate chain `->only('a')->only('b')`:** the scanner applies chain links outermost-first, so the *innermost* (`'a'`) wins, while Laravel's runtime applies the *last* call (`'b'`). Pathological idiom, zero corpus hits; mentioning only for the record.
4. **[NOTE] Acceptance-criteria placement:** issue #16 asked for the security-merge assertions in `SecurityExtractorTest`; they landed in `RouteMiddlewareGathererTest` (merge + dedup + cache stability) and the end-to-end feature test instead, with `SecurityExtractorTest` getting only mechanical DI updates. Functionally covered — the extractor's middleware source is now the gatherer, so the gatherer tests *are* the merge tests.
5. **[PRE-EXISTING → #259]** The `#[PublicEndpoint]`-clears-security-but-401-stays quirk: confirmed pre-existing on the merge base with plain route-declared middleware (`security: []` + a documented 401 on the same operation). Decision-record item 6 is accurate; filed as #259 since a declared-public operation documenting Unauthenticated is spec-misleading regardless of which PR introduced it.
6. **[PRE-EXISTING → #260]** Two lint-side readers (`OperationNode::hasAuthMiddleware()` → `operation.security-missing`, and `publicendpoint.contradicts-middleware`) still read `Route::middleware()` directly, so controller middleware — native *or* fallback — is invisible to lint while the generated document sees it. Not a #258 regression (they never used `gatherMiddleware()`), but the new seam makes the fix a two-line switch; filed as #260.

Layering checked: `Support\Routing` imports nothing from `Plugins\*`; both new classes `@internal` + `#[Scoped]` consistent with the codebase; no registry seam invented for a single implementation (consistent with the no-speculative-DI rule); docs (auto-derivation Security/Error rows + new Controller-middleware section) and CHANGELOG accurate. No remaining direct `gatherMiddleware()` reads anywhere in `src/`.

### Survey (Layer A, #233 procedure, serialized)

Baseline `results-pr-256.json` (= #256 head) vs this PR's head, all 11 apps re-run with vendor pinned to the review worktree; saved as `results-pr-258.json`.

**Prediction first:** grep found constructor `$this->middleware(` in BookStack (23 files) and InvoiceNinja (12 files), none elsewhere; since *every* app already generated with `gen_exit=0` on the baseline — and pre-PR a single unbuildable constructor-middleware controller would have crashed the run — all those controllers must be instantiable, i.e. already documented by the native path. Predicted **zero delta everywhere**. 

**Actual: exactly that.**

| App | gen | lint | ops | api-ops | completeness | Δ vs pr-256 |
|---|---|---|---|---|---|---|
| AdvisingApp | 0 | 1 | 532 | 31 | 96.8 | identical |
| AureusERP | 0 | 1 | 160 | 90 | 91.1 | identical |
| Bagisto | 0 | 1 | 491 | 41 | — | identical |
| BookStack | 0 | 1 | 335 | 86 | 94.2 | identical |
| Coolify | 0 | 1 | 375 | 154 | — | identical |
| InvoiceNinja | 0 | 1 | 694 | 522 | — | identical |
| Koel | 0 | 1 | 294 | 170 | — | identical |
| Lychee | 0 | 1 | 273 | 205 | — | identical |
| Pelican | 0 | 1 | 260 | 158 | — | identical |
| SpeedtestTracker | 0 | 1 | 42 | 8 | — | identical |
| Vito | 0 | 1 | 456 | 144 | — | identical |

"Identical" is the full metrics object (paths/operations/api-ops/response schemas/request bodies/component schemas/completeness/lint findings by rule/crash block), not just the headline counts. Zero fallback notices in any `generate.log` — no corpus app even enters the new code path. So the honest framing: **this PR's user-visible payoff on the current corpus is zero; its value is the crash-safety seam** (fault isolation for a class of controller the corpus happens not to contain, BookStack-style apps with side-effecting constructors being the real-world target) **plus the uniform middleware list**. The corpus result also retroactively validates the reframing: had the issue's premise been true, BookStack's 23 constructor-middleware controllers would have shown a security/401 delta — they show none because the framework already covered them.

One flake note, not attributable to this PR: the very first post-`composer install` suite run failed 4 tests; two clean re-runs (including with a cleared Pest cache) and the final post-cleanup run were all green — consistent with the known fixture-copy race.

Radiergummi:

**1. [nit] Changelog-speak in user docs.** `docs/auto-derivation.md:426-428` ("Controller middleware"): "Previously that crashed the run; now the generator logs a notice and falls back to a **bounded static scan**…" — the package is unpublished pre-1.0; there is no "previously" for users. The CHANGELOG #16 entry already tells the history; the docs section should just describe the behaviour.

Issue #254, opened 2026-06-10
survey: metrics.php counts a contentless 2xx as a "response schema", so adding a real-but-empty schema *decreases* responseSchemas


Radiergummi:
Found while empirically reviewing PR #253 (Koel `responseSchemas` 170 → 169 with **zero** substantive ops lost).

`metrics.php` sets `hasResponse = true` when a 2xx response has **no content at all** ("explicit no-content 2xx") and counts it into `responseSchemas`. Consequence: when an op moves from a bare `200 {description: OK}` to a 2xx **with** a JSON schema that is *not yet substantive* (PR #253's body-resolved `SongResource` whose `toArray()` is dynamic → empty component), the op drops **out** of `responseSchemas` — the metric goes down although the spec strictly gained information (concrete item class + correct `{data: [...]}` envelope, with `resource.fields-undeclared` now carrying the signal).

Observed instance: Koel `POST /api/songs/by-ids` (`FetchSongsByIdsController` → `SongResource::collection(...)`, `SongResource::toArray()` not statically readable).

The metric conflates two questions:
- "does the op document *any* 2xx outcome?" (where bare-200 counting is fair), and
- "does the op have a substantive 2xx schema?" (what delta reviews actually read it as — and what the #106 sizing used).

Suggested fix: split into `responseSchemas` (substantive-only, no bare-200 credit) and, if wanted, a separate `documentedResponses` counter; or at minimum stop counting contentless 2xx into `responseSchemas`. Self-verifying delta tables depend on this not flip-flopping.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Radiergummi:
Already resolved in merged PR #264 (2026-06-11). `tools/survey/metrics.php` now gates `responseSchemas` on `survey_substantive()` — a bare `200 {description: OK}` (no `content`) and an empty-schema 2xx are excluded, while a content-bearing 2xx (incl. a `{data: [...]}` envelope referencing an otherwise-empty component) counts. Covered by `tests/Tools/SurveyMetricsTest.php`.

Note: the shipped fix also split out a separate `documentedResponses` counter (any documented 2xx, incl. 204s). That's the 'substantive-only + new counter' variant rather than the 'no new counter' option — but since it's already merged and the core flip-flop bug is fixed, closing as resolved. If you'd prefer to drop `documentedResponses`, that's a separate follow-up.

PR #261, opened 2026-06-11
feat(#13): infer Fractal transformer schemas from the transform() array literal

### LINKED ISSUE #13: Fractal transformer `transform()` inference
## Summary

Read the array literal a Fractal transformer returns from its `transform()` method and bind it to the endpoint's response, so Fractal-based APIs document their output shape instead of emitting nothing.

## Motivation

A transformer expresses its output entirely inside `transform()`, which the generator never reads — and the existing `fractal.response-unbound` lint rule only catches the rare case where a controller injects a `League\Fractal\Manager` parameter. The two dominant invocation styles — the `fractal()` helper and the `Spatie\Fractalistic\Fractal` facade — are used in the method body and never inject a `Manager`, so they slip past the rule entirely (`FractalResponseUnbound` deliberately ships at level 2 for exactly this reason). The result is that Fractal endpoints silently produce undocumented responses. This was observed in the surveyed apps Invoice Ninja and Pelican; the work also closes the `fractal.response-unbound` blind spot it leaves behind.

## Proposed approach

Tier 1 for reading the literal; the binding is the part that must stay narrow.

- Use the shared bounded body-scan utility (see parent #5) to read the array literal returned by a transformer's `transform()` and map its keys to response properties. This extends the Fractal plugin's existing schema derivation in `src/Plugins/Fractal/Support/SchemaFromTransformer.php` / `src/Plugins/Fractal/Resolvers/TransformerRefSchemaResolver.php`.
- Detect the convention that ties a transformer to an endpoint by matching `fractal()` helper, `Spatie\Fractalistic\Fractal` facade, and injected-`Manager` call shapes in the endpoint body, then bind the resolved transformer to the operation's response via `FractalResponseResolver` / `FractalEnvelopeFactory`.
- Resolving *which* transformer applies is where Tier-2 dataflow would creep in. Keep the binding to unambiguous, single-transformer shapes; refuse ambiguous cases, fall back to today's behaviour with a generation-log note, and let the `#[FractalResponse]` attribute (`src/Plugins/Fractal/Attributes/FractalResponse.php`) always win.

## Acceptance criteria

- A transformer whose `transform()` returns an array literal, bound via the `fractal()` helper, the facade, or an injected `Manager`, yields a response schema in the generated document.
- An ambiguous binding degrades gracefully: no schema is forced, a log note is emitted, and `#[FractalResponse]` remains the explicit hatch — once declared, the envelope emits regardless of call style.
- Add coverage under `tests/Unit/` for the Fractal plugin (the body-scan-to-schema mapping and the bind/refuse decision), plus a Fractal flavor in `tests/Feature/ExamplesTest.php` exercising snapshot byte-match, OpenAPI 3.1 validation, and lint-clean output.

## References

Parent: #5 (shared bounded-AST body-scan utility, the Tier-1 boundary, and graceful-degradation contract). Related: #12 (Resource `toArray()`), #14 (`response()->json()` literal). Affected areas: `src/Plugins/Fractal`, plus the `src/Plugins/Core` body-scan utility from #5. Verify: `vendor/bin/pest tests/Feature/ExamplesTest.php`

Radiergummi:
## What & why

Closes #13 (epic #5 — the last Tier-1 body-scan child). A Fractal transformer expresses its output entirely inside `transform()`, which the generator never reads — Fractal endpoints document only what `#[TransformerField]` attributes restate by hand. This PR reads the single `return [...]` literal of `transform()` (the `SingleReturnArrayLiteralFinder` primitive was extracted in #12 precisely for this) and composes the inferred fields into the transformer's schema, mirroring #12's reader/VO/notice conventions.

**Stack:** #219 ← … ← #256 ← #258 ← this (base: `feat/16-constructor-middleware`).

## Scope — part 1: transform() literal inference

Bounded case (Tier-1, epic #5 contract): the transformer's `transform()` body — within the first 10 top-level statements — contains exactly one top-level `return` whose expression is an array literal. The literal's **string keys** become schema properties; each **value** resolves best-effort, refusing per value:

| Value shape | Resolution |
|---|---|
| `$model->field` where `$model` is `transform()`'s typed parameter (an Eloquent `Model` — Tier-0) | the model's property schema via `EloquentModelToSchema::propertyFor()` (casts, `@property`, appends) |
| `(int)` / `(float)` / `(string)` / `(bool)` / `(array)` cast of anything | the cast's JSON type — the dominant idiom in the motivating corpus (InvoiceNinja casts nearly every field) |
| literal scalar / literal array | typed via the shared `SchemaDefinitionFromLiteral` |
| anything else (method calls, ternaries, …) | property kept with an unconstrained `{}` schema — dropping a response property is silently wrong (#235/#12 precedent) — listed in **one** summarising note per transformer |

Precedence & fallback (the #12 conventions exactly):

- `#[TransformerField]` / `#[TransformerInclude]` win **per field**; inferred fields they don't cover follow in literal order. All inferred fields are required (no `when()` conditionals exist in `transform()`).
- A dynamic `transform()` body (early returns, `return $data`, no array literal) degrades to today's attribute-only behaviour + a generation-log note. A class without `transform()` stays silent (nothing to read).
- `TransformerRefSchemaResolver::canResolve` widens: a class is a transformer when it carries `#[TransformerField]` **or** extends `League\Fractal\TransformerAbstract` (FQCN string — no hard dependency) with a readable `transform()`.
- Envelopes (`{data}`, includes, pagination meta per serializer) unchanged — this fills the **item** schema only. Component registration via the established `ComponentSchemaRegistry::buildOnce()` path.
- `fractal.fields-undeclared` goes quiet when inference yields fields (its "the response schema is empty" message would be false) — same quieting as `resource.fields-undeclared` in #12.

## Scope — part 2: the `$entity_transformer` + `itemResponse()`/`listResponse()` idiom

Recommended by the #258 review + exhaustiveness analysis (~300 corpus operations, InvoiceNinja's BaseController convention). A new `PrimaryResponseResolver` in the Fractal plugin matches, in the scanned method body, a top-level `return $this->itemResponse(...)` or `return $this->listResponse(...)` **and** resolves the controller's `protected $entity_transformer` **property default** (Tier-0, `ReflectionProperty::getDefaultValue()`) to a concrete `TransformerAbstract` class-string. Then: transformer schema per part 1; `itemResponse` → single envelope, `listResponse` → collection envelope (DataArray serializer — the honest default).

Refusals (degrade to today's behaviour):

- the scanned method **reassigns** `$entity_transformer` → refuse with note (the default is no longer the honest answer);
- no usable property default while the call shape matched → note;
- the call shape doesn't appear → silent (most apps don't use this convention);
- any `PrimaryResponseAuthoringAttribute` (`#[FractalResponse]`, `#[ResponseResource]`) → defer; explicit authoring always wins.

The method names stay a **literal whitelist** — no configurable convention knobs (speculative). Same-PR decision (vs the reviewer's second-PR lean) and the spec-widening are recorded as a PR comment + issue #13 comment.

## File structure

| File | Action |
|---|---|
| `src/Plugins/Fractal/Support/TransformerTransformReader.php` | create — the bounded `transform()` reader (memoised, registry-free) |
| `src/Plugins/Fractal/Support/InferredTransformerField.php` | create — one inferred field (name, property, unconstrained flag) |
| `src/Plugins/Fractal/Support/SchemaFromTransformer.php` | modify — compose attribute + inferred fields; notes |
| `src/Plugins/Fractal/Resolvers/TransformerRefSchemaResolver.php` | modify — widen `canResolve` |
| `src/Plugins/Fractal/Resolvers/EntityTransformerResponseResolver.php` | create — part 2 |
| `src/Plugins/Fractal/FractalPlugin.php` | modify — register the new resolver |
| `src/Plugins/Fractal/Lint/Rules/FractalFieldsUndeclared.php` | modify — quiet when inference yields fields |
| tests: `tests/Unit/Plugins/Fractal/TransformerTransformReaderTest.php`, `SchemaFromTransformerTest.php`, `tests/Feature/Plugins/Fractal/…` (+ fixtures) | create/modify |
| `examples/fractal/` — attribute-free `BookingTransformer` + endpoint, snapshot regenerated | modify |
| `docs/plugins.md` (Fractal), `docs/auto-derivation.md` rows, `CHANGELOG.md` | modify |

## Implementation plan (TDD, one commit per step)

1. **`TransformerTransformReader`** — unit tests: literal fields; `$model->field` via the typed parameter; casts; nested literals; unresolvable values → unconstrained; dynamic body → null; no `transform()` → null; dynamic key/spread → null.
2. **`SchemaFromTransformer` composition** — attribute-wins per field, inferred extras in literal order, one summarising note, dynamic-body note; `canResolve` widening; `fractal.fields-undeclared` quieting; feature tests.
3. **`EntityTransformerResponseResolver`** (part 2) — feature tests: itemResponse/listResponse fixtures, reassignment degrade, missing-default note, attribute-wins.
4. **Examples + docs + CHANGELOG** — attribute-free `BookingTransformer` in `examples/fractal`, snapshot regenerated (diff must be exactly this feature).

Verify: `composer test && composer lint && vendor/bin/pint --test`.


Radiergummi:
## Scope decisions (recorded per the issue-#13 spec)

**1. Part 2 ships in this PR, not a second stacked PR.** The #258 review leaned toward a separate PR for the `$entity_transformer` + `itemResponse()`/`listResponse()` idiom. After building part 1, the shared core (the `TransformerTransformReader` + `SchemaFromTransformer` composition) left part 2 at one resolver class (~290 lines incl. docblocks) plus fixtures — small enough that splitting would cost more in stack overhead than it buys in review focus. It lands as its own commit (`9c6314a`) so it reviews in isolation and reverts cleanly if the idiom is unwanted.

**2. The `fractal()` helper / `Fractalistic` facade / injected-`Manager` *binding* detection from the issue's proposed approach is deferred.** This PR binds transformers two ways: the existing `#[FractalResponse]` attribute (now composing inferred fields) and the `$entity_transformer` property-default convention. Detecting *which transformer applies* from a `fractal($invoice, new InvoiceTransformer())` call inside the body is a distinct whitelist with its own ambiguity surface (transformer argument shapes, `->respond()` vs `->toArray()` terminals, serializer switches mid-chain) — split out as a follow-up so this PR stays the transform()-reading core. Until then those endpoints keep today's behaviour, and `#[FractalResponse]` remains the hatch the issue describes.

**3. Cast values are typed by the cast.** Not in the issue's value table, but `(float) $invoice->amount` / `(string) $invoice->number` is the dominant idiom in the motivating corpus (InvoiceNinja casts nearly every transform() field) — without it, most fields would degrade to unconstrained `{}`. A cast node is a single unambiguous AST shape and states the runtime JSON type regardless of what it wraps, so the cast wins even over model metadata (`decimal:2` casts serialize as strings, but `(float)` makes the wire type a number).

**4. Part 2 refuses field-less transformers.** A matched `itemResponse()` whose transformer has neither attributes nor a readable literal degrades with a note instead of binding an envelope around an empty `{}` item — claiming authority over the response while documenting nothing would be worse than today's behaviour. The attribute path (`#[FractalResponse]`) keeps binding regardless, since that authoring is explicit.

Radiergummi:
## Review — empirical test report (PR #261)

Reviewed the `origin/feat/16-constructor-middleware..origin/feat/13-fractal-transform-inference` delta only (stack base #258 approved). Gates on the PR head (`e3dbc0c`, PHP 8.4.16): **PHPStan clean, Pint clean, Pest 2298 passed** — the suite ran 8× total; run 1 (fresh `composer install`) had a single failure that never reproduced (runs 2–8 green, including one with a cleared Pest cache). Matches the flake noted during implementation; identity lost to output truncation, not actionable yet.

### Findings

1. **[blocking]** The `(array)` cast maps to `type: array` with **no `items`** (`TransformerTransformReader::castType()`), and swagger-php's `Analysis::validate()` requires `@OA\Items()` whenever a property has type `array` — so `openapi:generate` under default flags **crashes** (`ErrorException: @OA\Items() is required when @OA\Property(property="client_registration_fields") has type "array"`). This is not an edge case in the motivating corpus: **InvoiceNinja has 11 `(array)` casts across its transformers** and generation with the plugin enabled dies on the first one (`CompanyTransformer`). Reproduced in-package with a minimal fixture (transformer returning `'fields' => (array) $article->subtitle`, `artisan openapi:generate` → exit 1, verified on swagger-php 6.1.2). Neither the unit fixtures (`InferredArticleTransformer` covers int/float/bool but not `(array)`) nor the examples snapshot exercise the `(array)` arm, which is why the validated-snapshot path never caught it. Fix: emit an unconstrained `items` alongside (`{ type: array, items: {} }`) or refuse `(array)` to unconstrained; add an `(array)` cast to a fixture that passes through `openapi:generate`'s validation. Related: the same items-less-array shape pre-exists in `SchemaDefinitionFromLiteral` (list with unreadable first element) on the #235/#12 paths — filed as #265; worth fixing both in one shared place.
2. **[low]** A `$entity_transformer` reassignment *outside* the scanned method (constructor, or a called helper) silently documents the property default. Empirically harmless in InvoiceNinja — no constructor reassigns, and the helper-reassigning controllers (`QuoteController::action` → `performAction()`) return through the helper, which doesn't match the whitelist, so they degrade silently rather than wrongly — but the "default wins when the reassignment is invisible" behaviour isn't pinned by any test. Suggest one pinning test so the decision is deliberate and survives refactors.
3. **[nit]** The part-2 refusal note for a field-less transformer says "has no readable transform() literal" even when the literal is readable but **empty** (`return [];` — `hasDocumentableFields` correctly refuses). "declares no #[TransformerField] and yields no documentable fields" would be accurate for both.
4. **[nit]** An unreadable value *inside a nested literal* (`'totals' => ['amount' => (float) …]` — casts only resolve at the top level of values) degrades to an unconstrained inner property correctly, but isn't listed in the summarising notice, which only covers top-level keys. Harmless; mentioning for completeness.

Verified clean: attribute-wins at both levels (per-field `#[TransformerField]` and `#[FractalResponse]` over the body scan); the `canResolve` widening only fires on explicit class-string references (ref resolvers are never consulted spontaneously), so helper transformers can't self-bind; resolver interplay with #235's `InlineJsonResponseResolver` is correct in both directions (a conditional `response()->json()` guard before a top-level `itemResponse()` yields the Fractal envelope — probed empirically); in-method reassignment refuses with a precise notice including conditional contexts; single vs collection envelope per called method; the snapshot diff is exactly the feature; #263 is well-formed; docs/CHANGELOG thorough.

### Survey Layer A (serialized per #233; baseline `results-pr-258.json` = merge base `149ffeb`, verified)

**Corpus (Fractal disabled — its published-config default): zero deltas across all 11 apps** on every metric (paths, operations, apiOperations, responseSchemas, requestBodies, componentSchemas, completeness, lint totals, crash fields). `results-pr-261.json`.

**InvoiceNinja opt-in spot-check** (FractalPlugin temporarily enabled, regenerated, config + artifacts restored after):

- Generation **crashes** under default flags (finding 1). Measured the rest with `--no-validate`:
- **255 of 522 API operations gained a substantive 2xx** (contentless non-204 200s: 419 → 164); 0 lost, 0 status-set changes; 35 transformer component schemas. The exhaustiveness analysis predicted ~300 ops reachable via the `$entity_transformer` idiom: **295 matched** = 255 resolved + **40 honest refusals with notices** (9 in-method reassignments — `TokenController::*`, `ProjectController::bulk`, … — and 31 ops on the four transformers whose `transform()` is the `$data = [...]; if (…) {…} return $data;` shape: `InvoiceTransformer`, `RecurringInvoiceTransformer`, `RecurringExpenseTransformer`, `CompanyLedgerTransformer`). The remaining ~124 contentless ops never match the whitelist (helper indirection like `performAction()`, returns beyond the 10-statement window, non-`itemResponse` shapes).
- Spot-checks against source: `GET /api/v1/clients/{client}` → single `{data: $ref ClientTransformer}`, 48 properties, `(float) balance` → number, `(int) created_at` → integer, `(bool) is_deleted` → boolean, `encodePrimaryKey(...)`/`?:`-elvis values honestly unconstrained `{}`; `GET /api/v1/products` → collection `{data: [$ref]}`, `(float) $product->cost ?: 0` correctly refused (the elvis wraps the cast); negatives: `/api/v1/tokens` (reassign) and `/api/v1/invoices` (unreadable transformer) stay contentless with precise notices — **no wrong schema anywhere**.
- Note the completeness scoreboard shows **0.0 delta** for this gain — the #254 contentless-2xx artifact (it counts a bare `description: OK` as complete). The per-op specdiff above is the honest measure.

### Verdict

**REQUEST CHANGES** — solely finding 1: the headline feature crashes `openapi:generate` on its motivating app with default flags, and the fix is small and well-localised. Everything else — the reader, the composition, part 2's refusal discipline, the degradation notices — is exactly what the spec ordered and measured beautifully against the corpus: this PR takes InvoiceNinja from 20% to 69% substantive-2xx coverage once the cast fix lands.

Radiergummi:
All four findings addressed in `fa9ce14` / `f8d5c39` / `72f3bf1`; gates green on the new head (Pest 2302 passed, PHPStan clean, Pint clean, snapshot group green).

**1. [blocking] `(array)` cast → items-less `type: array`** — fixed in `fa9ce14`. `TransformerTransformReader::resolveValue()` now emits `items: {}` alongside the `array` type (unconstrained items — array-of-anything is the honest claim for an `(array)` cast). TDD per the report: a new `(array)` arm in the `InferredArticleTransformer` fixture with a unit test asserting `items` is set, plus an `extras => (array) $booking->seat` field on the examples `BookingTransformer`, so the `(array)` shape now passes through `openapi:generate` + `Analysis::validate()` in `ExamplesTest` on every run. Verified on **both majors**: swagger-php 6.1.2 locally (snapshot regenerated, validation + lint green) and the 5.8 CI path replicated exactly (pinned `zircote/swagger-php:^5.8` → resolved 5.8.3, full suite minus the `snapshot` group green, and `php examples/generate.php fractal` succeeds under default flags emitting `items: { }`). No shared helper emerged — the fix is one inline `OA\Items([])` — so nothing to add on #265's twin hazard in `SchemaDefinitionFromLiteral`.

**2. [low] helper-hidden `$entity_transformer` reassignment** — pinned in `f8d5c39`. New fixture (`ArticleEntityController::helperReassigned()` reassigning via a private `swapTransformer()` helper) + feature test *"documents the property default when the reassignment hides in a called helper — invisible to the bounded scan"*. No notice: detecting the hidden reassignment would require following calls (Tier-2 dataflow), and a blanket notice on every helper call would be noise; the assumption is instead named explicitly in the resolver's degradation-contract docblock and in `docs/plugins.md` (with the `#[FractalResponse]` escape hatch called out).

**3. [nit] refusal wording** — fixed in `f8d5c39`. The note now reads *"declares no #[TransformerField] and yields no documentable fields"*, accurate for both the unreadable body and the readable-but-empty literal. The empty-literal branch is now exercised by a new `EmptyTransformer` (`return [];`) fixture + feature test asserting both the refusal and the notice text.

**4. [nit] nested unreadable values missing from the notice** — fixed in `72f3bf1`. `InferredTransformerField` now carries `unconstrainedPaths` (a list of key paths) instead of a boolean; the reader walks nested literal definitions and reports inner refusals by path (`flags.rating` in the fixture — a cast below the top level), and `SchemaFromTransformer`'s summarising notice lists them alongside the top-level keys.

Snapshot delta: `examples/fractal/openapi.yaml` +4 lines (`extras` in `required` + `extras: {type: array, items: {}}` on `BookingTransformer`) — exactly the new fixture field.

Radiergummi:
## Re-review — fixes verified (head `72f3bf1`)

All four findings from the [previous review](https://github.com/Radiergummi/laravel-openapi/pull/261#issuecomment-4680049240) verified against the new head, each empirically.

### Per-finding verification

1. **[blocking] `(array)` cast → items-less array — fixed and confirmed on the motivating app.** Re-ran the original reproduction: InvoiceNinja, state-verified per #233 (vendor symlink → the review worktree at `72f3bf1`, `composer.json` intact), `FractalPlugin` enabled, `openapi:generate` under **default flags (validation on)** → **exit 0**. The 255-op gain holds exactly: substantive 2xx **70 → 325** (contentless 2xx 591 → 336), and the structural spec diff against my `e3dbc0c` `--no-validate` probe artifact is **precisely the fix**: 3 added `items: {}` leaves on the `(array)`-cast properties of resolved transformers (`CompanyTransformer.client_registration_fields`, `RecurringQuoteTransformer.recurring_dates`, `SchedulerTransformer.parameters`) — zero removals, zero other changes beyond the 92 nondeterministic `operationId` nonces. (#254 caveat stands: the completeness scoreboard won't show this gain; the per-op counts above are the honest measure.) The TDD chain is right: unit assertion on non-default `items`, plus the `extras` field on the examples `BookingTransformer` so the `(array)` arm passes through `Analysis::validate()` on every `ExamplesTest` run.
2. **[low] helper-hidden reassignment — pinned; acceptable without a notice.** The new `helperReassigned()` fixture + feature test pin "default wins when the reassignment is invisible". I agree with the no-notice call: detecting it means following calls (Tier-2), and a blanket notice on every helper call is noise. The assumption is named where it should be — the resolver's degradation-contract docblock and `docs/plugins.md`, with `#[FractalResponse]` as the escape hatch — and the pinning test makes the boundary deliberate.
3. **[nit] refusal wording — fixed**, and the previously untested readable-but-empty branch is now exercised: `EmptyTransformer` (`return [];`) fixture + feature test asserting both the refusal and the "yields no documentable fields" note.
4. **[nit] nested unreadable values — fixed properly**, not minimally: `InferredTransformerField` now carries `unconstrainedPaths`, the reader walks nested literal definitions, and the notice names inner refusals by key path (`flags.rating` asserted at both the reader and `SchemaFromTransformer` level).

### Gates (PHP 8.4.16, swagger-php 6.1.2)

- Pest: **2302 passed** (5744 assertions), three runs, no flake this time.
- PHPStan: clean. Pint: clean. Snapshot group: 9 passed.
- Fractal example regenerates byte-identically under default flags; the snapshot delta vs `e3dbc0c` is exactly the new `extras` field (+4 lines).
- **5.8 path independently replicated** (their approach — git-archive + pin `^5.8` in a scratch dir — is sound: it mirrors the CI 5.8 job without touching the worktree lock): resolved 5.8.3, suite minus `snapshot` **2293 passed**, fractal example generates `items: {}` under default flags. One first-run-only "1 warning" that never reproduced — same fresh-install transient family as noted last time. CI fully green on the head, 5.8 job included.

### Residual, non-blocking (#265)

Confirmed empirically that the items-less-array twin I filed as #265 is reachable on this PR's path too: an empty-literal value (`'x' => []`) in `transform()` flows through `SchemaDefinitionFromLiteral` → `['type' => 'array']` with no items → the same `@OA\Items() is required` crash under default flags (probed on the example app, reverted). It does **not** trip InvoiceNinja today — the default-flags run above succeeded — and the shared fix belongs in `SchemaDefinitionFromLiteral` per #265's scoping, which will cover this third consumer automatically (and subsume the small gap that `unreadableNestedPaths` can't name a bare nested `type: array` in the notice). Worth a cross-reference on #265.

### Workspace

InvoiceNinja config restored (Fractal disabled), disabled-state artifacts regenerated: every metric byte-matches `results-pr-261.json` (paths 529, operations 694, apiOperations 522, responseSchemas 519, requestBodies 255, componentSchemas 375, completeness 88.3, lint 194, crash fields identical), and the regenerated spec is identical to the disabled probe baseline modulo `operationId` nonces. Zero-delta state confirmed; all probes cleaned.

### Verdict

**APPROVE.** The blocker is fixed at the right layer with the honest schema (`items: {}`), proven on the app that exposed it, on both swagger-php majors; the three smaller findings got tests, not just patches. Ready for @Radiergummi's merge review — and with the #219 → #261 stack this completes the **last child of Epic #5**: all nine Tier-1 body-scan idioms are implemented and in review.

Radiergummi:
Independent stack review — findings attributable to this PR:

**1. [minor] `docs/linting.md:403` catalog row for `fractal.fields-undeclared` is stale against this PR's rule change.** The row still reads "A transformer bound via #[FractalResponse] declares no #[TransformerField] attributes." — but the rule now also covers `$entity_transformer` bindings and stays quiet when the `transform()` literal yields inferred fields (as `docs/plugins.md:464-465` correctly documents). The catalog is hand-maintained per CLAUDE.md; this stack updated the analogous `resource.*` rows but missed this one.

**2. [minor — shared with #235] The "Tier-0 resolvers stay authoritative regardless of chain order" claim in `EntityTransformerResponseResolver`'s docblock doesn't hold for this plugin's own convention path** — see the detailed note on PR #235 (an untyped method with a non-returned straight-line `json([literal])` assignment shadows the `$entity_transformer` convention).

Issue #157, opened 2026-06-06
Companion composite GitHub Action: post documentation-coverage PR comment + check run


Radiergummi:
## Motivation

Follow-up to the documentation-coverage work (#152, #153). A **companion GitHub Action** makes sense for exactly one job: the GitHub-API glue a plain `run:` step can't do — **not** for running the linter.

This is explicitly gated on #152/#153: until those define a coverage report format, there is nothing for the action to wrap.

## Why composite, not Docker

The linter introspects the **consumer's own routes/classes via reflection** — it only runs inside their already-set-up PHP environment (their PHP version, extensions, `composer install`, app autoloader, env). A Docker/standalone action can't provide that and would fight the consumer's setup. This is why first-party PHP tools (PHPStan, Pint, Pest, Larastan) don't ship their own actions.

So this must be a **composite** action that assumes PHP + `vendor/` are already present in the job (the consumer runs `setup-php` + `composer install` first, per the recipe in the docs issue), and only adds the reporting glue.

## What it does (and only this)

- Post a **sticky PR comment** with the coverage % + per-area table (from #152's report).
- Open a **check run / commit status** for "documentation (patch) coverage" so it can be made a required gate.
- Resolve the correct base ref for `--diff` from the PR event (`github.event.pull_request.base.sha`).

What it explicitly does **not** do: set up PHP, run `composer install`, or "run the linter" — those are one-line `run:` steps already covered by the docs recipe. Inline PR annotations also need no action: `GithubFormatter` already emits `::warning file=…,line=…::` workflow commands via `--format=github`.

## Scope & distribution

- Reporting wrapper only; interoperates with the existing CLI output (#153's Cobertura/LCOV can alternatively feed Codecov directly, in which case this action is redundant for those users — document the trade-off).
- Distribution: a composite `action.yml`, likely in a separate `laravel-openapi-action` repo for Marketplace tagging discipline. Note the added release surface — only worth it once #152/#153 have shipped and a report format is stable.

## Dependencies

Blocked on #152 and #153. Do not start before a report format exists. Milestone no earlier than v1.2.

Related: #158 (integration recipes docs).

Radiergummi:
**Decision (backlog triage): declined in favor of a documented recipe — see #385.**

[...]

PR #266, opened 2026-06-11
feat(#153): coverage report formats (multi-target lint output)

### LINKED ISSUE #153: Emit documentation coverage as Cobertura/LCOV for Codecov/Sonar (patch coverage for free)
## Motivation

Builds on the native documentation-coverage gate (#152). Where the gate is the cheap 80%, this is the shiny 20%: emit coverage in an **established code-coverage format** so existing tooling (Codecov, Coveralls, SonarQube) consumes it with zero new infra on our side.

The clever part: if we emit a **Cobertura/LCOV** report keyed to source lines (controller method, FormRequest rule, Data property), Codecov's **patch coverage** gives "documentation coverage of the lines this PR changed" *for free* — we build no diff logic and no PR-comment bot, and devs get the gutter UI they already know. Codecov **flags** let it ride as a separate stream so it doesn't conflate with real code coverage.

## Why it's feasible now

Findings already carry source location — `FindingLocation` has `file`, `line`, `routeName`, `routeMethod`, `routeUri`, `jsonPointer` (`src/Lint/FindingLocation.php`). That's the prerequisite for a line-keyed report.

## Proposed design

- `openapi:lint --format=cobertura` (and/or `--coverage-output=<path>` emitting Cobertura XML; LCOV as a stretch).
- Map each documentable unit to a representative source line via `FindingLocation.file`/`line`; covered/uncovered per the gate's definition.
- Document using Codecov **flags** (separate stream) so it never mixes with code coverage %.

## Footguns to design around

- **Line attribution is lossy.** Some spec facts have no single source line (response inferred from route-model binding, error response inferred from middleware). Best-effort only — likely *exclude* inference-only facts from the line-keyed report and keep them in the native JSON summary. Coverage gutters pointing at the wrong line erode trust.
- **Graded → binary flattening.** The format is per-line binary; doc quality is graded. Acceptable for gating, but we lose the severity nuance the `Finding` carries — hence keep the rich detail in the native JSON report (#152), not here.
- **Diff scoping keys on the controller file** (see #154): a gap whose only fix lives in an untouched FormRequest/Data class is pulled in via reachability. Arguably correct for coverage, but means "patch coverage" is not strictly your-diff-only — document this.
- **Shifting denominator** across generator versions (as in #152) — pin the generator version in report metadata.

## Scope boundaries

- We **interoperate with** coverage tooling; we don't rebuild it. No PR-comment bot, no threshold UI — those belong to Codecov/Sonar.
- Spec/CI-side only.

Radiergummi:
Closes #153.

## What & why

Emit documentation coverage in established code-coverage formats (Cobertura first) so Codecov/Coveralls/Sonar consume it with no new infra on our side, and — the point of this PR's shape — let a **single lint run** produce both the human stream (GitHub annotations) *and* the machine report, instead of running the command twice (wasted CI minutes).

## Design decision (deviates from the issue's literal CLI)

The issue proposed `--format=cobertura` and/or `--coverage-output=<path>`. Since the package is unreleased (no back-compat constraint), I collapsed output selection into **one repeatable flag** rather than carrying two concepts (`--format` for stdout + a separate report flag) — the shape golangci-lint/RuboCop/Psalm converge on:

```bash
php artisan openapi:lint                                            # auto-detect → stdout (unchanged)
php artisan openapi:lint --format=github                           # one stream → stdout
php artisan openapi:lint --format=github --format=cobertura:coverage.xml   # one run, two outputs
```

- `--format=<fmt>` is **repeatable**; each is `<format>[:<target>]`.
- `<target>` ∈ `stdout` (default when omitted) | `stderr` | a file path. `stdout`/`stderr` are reserved keywords.
- Split on the **first** `:` only (Windows paths survive: `cobertura:C:\out\cov.xml`).
- **Rule:** at most one target per console stream — two formats both writing stdout errors out with a clear hint.
- Any format works as a file target (`json:lint.json`); `cobertura` (and later `lcov`) are just enum cases that happen to be more useful in a file.

## Plan

- [x] Target grammar: `--format` → array option; parse `[(LinterOutputFormat, OutputTarget)]`; same-stream collision guard.
- [x] `LinterOutputFormat::Cobertura`; reuse the `Formatter` interface, writing to a `StreamOutput` for file targets (no formatter signature change).
- [x] Thread per-operation `{file, line, covered}` into the coverage model: widen `CoverageCalculator` input with each op's `OperationNode::file()/line()`; add an **in-memory** `perOperation` list on `CoverageSummary` that is *excluded* from `toArray()`/`jsonSerialize()` (JSON report + gate output stay byte-identical).
- [x] `CoberturaFormatter`: group `perOperation` by file → `<class>`/`<line number hits=0|1>`; `line-rate` from the aggregate; generator version pinned in root. **Exclude** operations with null file/line (inference-only/unattributable facts — per the issue's lossy-attribution footgun).
- [x] Same-line collision (multi-verb route on one method): line is uncovered if *any* mapped op is uncovered (conservative — surfaces the gap).
- [x] Tests mirroring `tests/Unit/Lint/Formatters/`; `docs/linting.md` format section; `CHANGELOG.md [Unreleased]`.

## Out of scope (follow-ups)

- LCOV (issue calls it a stretch) — trivial once the report mechanism lands.
- #157 composite GitHub Action (separate repo, blocked-then-separate).


Radiergummi:
Implemented and green (2022 tests pass, PHPStan clean, Pint clean).

- [x] Target grammar: `--format` → repeatable array option; `FormatTargetParser` parses `<format>[:<target>]` into `FormatTarget`s (first-colon split, same-destination guard). `OutputChannel`/`OutputTarget` value objects.
- [x] `LinterOutputFormat::Cobertura`; dispatch loop in `LintCommand` (one `Formatter` + `OutputInterface` per target; stdout via the OutputStyle for test capture, stderr via the console error output, file via `StreamOutput`).
- [x] Per-operation `{file,line,covered}` threaded from `OperationNode::file()/line()` through `LintRunner` into `CoverageCalculator`; held on `CoverageSummary::$perOperation` **in memory only** (excluded from `toArray()`/`jsonSerialize()` — JSON report + coverage gate output byte-identical).
- [x] `CoberturaFormatter`: `<class>`/`<line hits=0|1>` grouped by file, `line-rate` from valid/covered, generator version stamped; null file/line excluded; same-line collision → uncovered if any op is.
- [x] Tests: `FormatTargetParserTest`, `CoberturaFormatterTest`, `CoverageCalculatorTest` (perOperation), `LintReportTargetsTest` (end-to-end multi-target).
- [x] `docs/linting.md` "Output targets" section; `CHANGELOG.md [Unreleased]`.

### Decisions while implementing (deviations / notes)
- **Unified single `--format` flag** rather than the issue's `--format=cobertura` / `--coverage-output`: pre-release, no back-compat, so collapsing output selection into one repeatable flag (golangci/RuboCop/Psalm shape) avoids carrying two concepts. `stdout`/`stderr` are reserved target keywords; pathless ⇒ stdout.
- **Paths emitted verbatim** (no `base_path()` relativization): Codecov's path-fixing handles absolute→repo-relative, and relativization is awkward under Testbench. Flagging in case repo-relative output is preferred — easy follow-up in the location collector.
- **`--list --format=cobertura`** degrades to the JSON rule catalog (coverage has no catalog representation) — keeps the `RuleCatalogRenderer` match exhaustive under PHPStan.
- **LCOV** deferred (issue's "stretch") — trivial follow-up now the report mechanism exists. **#157** (composite Action, separate repo) stays separate.

PR #272, opened 2026-06-12
feat(#127): array-shape PHPDoc → object schema (general TypeNodeToSchema)

### LINKED ISSUE #127: Tier-0: resolve array-shape PHPDoc (`array{foo: string, bar: int}`) into an object schema
## Summary

Array-shape PHPDoc — `array{foo: string, bar: int}`, `array{data: array{id: int}}`, optional keys (`array{name?: string}`), and the list form `list<array{...}>` — is **not resolved into an object schema**. The shape is silently dropped to a bare `array`/untyped object with no properties.

## Evidence (where it falls through)

- `phpstan/phpdoc-parser` (already a dependency) parses `array{...}` into an `ArrayShapeNode` — so the information reaches us intact.
- `src/Support/Types/TypeNodeResolver.php` only handles `IdentifierTypeNode`, `GenericTypeNode`, `NullableTypeNode`, `UnionTypeNode`. There is **no** `ArrayShapeNode` / `ArrayTypeNode` handling anywhere in `src/` (verified by grep — every `array{` hit is the codebase's own docblocks, not resolver logic).
- Consequence: `@return array{foo: string, bar: int}` → emitted as an opaque `array`, not `{type: object, properties: {foo: string, bar: integer}}`.

## Proposed behaviour (Tier 0)

Resolve `ArrayShapeNode` into an object schema:
- each shape item key → a property; its value `TypeNode` → resolved recursively (reuse the existing identifier/generic/nullable/union paths);
- optional item (`name?: …`) → property not in `required`;
- nested `array{...}` → nested object; `list<array{...}>`/`array{...}[]` → `array` with `items` = the object;
- string-keyed `array<string, T>` (no shape) stays `additionalProperties: T` (already the conservative behaviour).

## Why it matters

Pure reflection/PHPDoc, no body parsing — the cheapest tier. It's the natural typed-source companion to #12 (toArray reader): a Resource that documents `toArray(): array{...}`, or any controller method returning `array{...}`, becomes a fully-typed object schema for free. Surfaced during the Vito full-spec attribute proof, where every response object had to be hand-restated as `#[ResourceField]` partly because typed array shapes carry no weight.

Related: #12, #28 (nested rule → nested schema, the validation-rule analogue), #108.

Radiergummi:
Closes #127

## Problem

Array-shape PHPDoc — `array{foo: string, bar: int}`, nested shapes, optional keys (`array{name?: string}`), and list forms (`list<array{…}>` / `array{…}[]`) — was silently dropped to a bare/untyped array. The single place a `TypeNode` became an `OA\Schema` was `EloquentModelToSchema::schemaFromTagType()`, which was **private and model-specific** and explicitly did not descend into array/generic nodes.

## Approach

A **general `TypeNode` → `OA\Schema` resolver** — the missing front-end companion to the existing `JsonSchemaFromType` (symfony `Type` → schema). Array shapes are a `TypeNode`-level concept (`ArrayShapeNode`) that symfony's `Type` model cannot represent, so this is the correct and only home for them.

### New: `src/Support/Types/TypeNodeToSchema.php`
Owns the **structural** layer, recursively:
- `array{k: T, …}` → `{type: object, properties, required}`; optional key (`k?:`) → omitted from `required`
- nested `array{…}` → nested object; `list<array{…}>` / `array{…}[]` (and scalar lists like `string[]`) → `{type: array, items}` (items always emitted, even for unreadable elements — swagger-php rejects `type: array` without `items`)
- `array<string, T>` → `additionalProperties: T`; `array<int, T>` → list
- scalars / nullable (`?T`, `T|null` → OAS 3.1 idiom) reuse the existing `TypeNodeResolver` paths
- a resolved **leaf class FQCN** → delegated to a consumer-supplied `callable(string $fqcn): ?OA\Schema`. The one real per-consumer divergence: a Model leaf becomes a `$ref` minted by `EloquentModelToSchema::build()`, a plain class goes through `JsonSchemaFromType`. Keeps the resolver dependency-light (only `TypeNodeResolver`) and free of any Eloquent coupling.

### Wire-in
`EloquentModelToSchema::propertyFromTag()` now delegates to `TypeNodeToSchema`, supplying its Model→`$ref` strategy. `@property array{…}` on a model becomes a fully-typed object schema.

## Scope / conflict avoidance
Does **not** touch `OperationBuilder` or the primary-response resolvers — **no overlap with the epic #5 stack** (#12 `toArray()` reader, #108 return-expression resolver, #14 `response()->json()` literal). Those land the `@return array{…}` → response path; this PR hands them a ready-made resolver to call. Pure Tier-0: reflection/PHPDoc only.

## Verification — all green
- `tests/Unit/Support/Types/TypeNodeToSchemaTest.php` — 17 cases (shape / optional / nested / list / suffix-form / scalar-list / additionalProperties / scalar / class-leaf / quoted+int keys / unreadable-value / nullable idiom / null fallbacks)
- `tests/Unit/Support/Extraction/EloquentModelToSchemaTest.php` — added an end-to-end model `@property array{…}` case (fixture `ShapedArticle`)
- `composer test` (2067 passed) · `vendor/bin/pint --test` (passed) · `composer lint` (no errors)
- `docs/auto-derivation.md` (property-sources note) + `CHANGELOG.md` [Unreleased]


src/Support/Generator/SchemaFieldCopier.php
```diff
@@ -0,0 +1,40 @@
...
+    private function __construct() {}
+
+    public static function copy(OA\Schema $source, OA\Schema $target): void
+    {
+        foreach (get_object_vars($source) as $field => $value) {
+            if (in_array($field, ['property', 'schema'], strict: true) || $field[0] === '_') {
+                continue;
+            }
+
+            if ($value !== Generator::UNDEFINED) {
```
Radiergummi: shouldn't we use our `is_undefined` function here?

src/Support/Types/TypeNodeToSchema.php
```diff
@@ -0,0 +1,228 @@
...
+    private const array ARRAY_GENERICS = ['array', 'list', 'non-empty-array', 'non-empty-list', 'iterable'];
+
+    public function __construct(
+        private TypeNodeResolver $typeNodeResolver,
+    ) {}
+
+    public static function create(): self
+    {
+        return new self(TypeNodeResolver::create());
+    }
```
Radiergummi: Couldn't we just use the normal constructor here, with a default for the resolver?

```php
public function __construct(private TypeNodeResolver $typeNodeResolver = new TypeNodeResolver())
```

src/Support/Types/TypeNodeToSchema.php
```diff
@@ -0,0 +1,228 @@
...
+
+    /**
+     * Resolves a type node to a schema, or null when the node is a shape this resolver does not
+     * model (so the caller falls through to its own fallback). Nullability is applied internally —
+     * `?T` / `T|null` is wrapped via the OAS 3.1 idiom ({@see NullableSchema}) — so callers must
+     * not wrap again.
+     *
+     * @param callable(string): ?OA\Schema $classSchema resolves a leaf FQCN to a schema
+     */
+    public function resolve(TypeNode $node, Reflector $context, callable $classSchema): ?OA\Schema
```
Radiergummi, line 64: This method could use a `return match (true) { ... }` expression for greater clarity

src/Support/Generator/SchemaFieldCopier.php
Radiergummi: Done — switched to `is_undefined($value)` (folded into the single skip condition). ebc8f2a→latest.

src/Support/Types/TypeNodeToSchema.php
Radiergummi: Done — dropped `create()` and gave the constructor a default: `__construct(private TypeNodeResolver $typeNodeResolver = new TypeNodeResolver())`. That needed `TypeNodeResolver` to be no-arg constructable, so I added promoted defaults there too (`create()` stays as a thin `new self()` alias for the service-provider binding). Tests now use `new TypeNodeToSchema()`.

src/Support/Types/TypeNodeToSchema.php
Radiergummi: Done — `resolve()` is now a `return match (true) { … }`, with the multi-statement nullable handling extracted to a private `resolveNullable()` so the match stays a clean one-arm-per-node-kind dispatch.

Radiergummi:
Addressed the three review comments (replied inline) and closed the Codecov gaps on the type-inference infra:

**Review comments**
- `SchemaFieldCopier` now uses `is_undefined()`.
- `TypeNodeToSchema` drops `create()` for a constructor default; `resolve()` is a `match (true)`.

**Coverage** (full suite)
- `TypeNodeToSchema` 94.0% → **98.4%**
- `TypeNodeResolver` 81.1% → **96.5%** — added direct tests for the public `unwrapNullable` / `isNullable` / `resolveClassName` union + scalar branches (previously only `genericValueClass` / `throwsClasses` were exercised)
- `EloquentModelToSchema` patch path 33% → covered (non-model-class `@property` via `JsonSchemaFromType`, unresolvable `@property` → empty fallback)
- `SchemaFieldCopier` 100%

Per the "sensible public-API coverage, not 100%" goal, the handful of lines left uncovered are degenerate/defensive `return null` branches (a 3-argument `array<a,b,c>` generic, the context-retry catch, a nullable-`@throws` flatten) — not worth contrived tests. Remaining `EloquentModelToSchema` misses are pre-existing cast/accessor paths unrelated to this PR.

`composer test` (2078) · `vendor/bin/pint --test` · `composer lint` all green.

PR #275, opened 2026-06-12
fix(#252): recognise class-form object casts so they type the column and let @property have a say

### LINKED ISSUE #252: Class-form object casts (AsCollection::class, AsArrayObject::class) bypass the JSON-cast mapping — column documents untyped and the @property tag never gets a say
Found while empirically reviewing PR #251 (#249's list-disambiguation half); the gap predates it — `castToProperty()` has never recognised class-form casts.

## Problem

Modern Laravel spells the JSON casts as castable class-strings:

```php
protected function casts(): array
{
    return ['tags' => AsCollection::class];   // or AsArrayObject::class, AsEncryptedCollection::class, …
}
```

`Model::getCasts()` then reports the FQCN (`Illuminate\Database\Eloquent\Casts\AsCollection`) as the cast string. In `EloquentModelToSchema::castToProperty()` the normalisation (`strtolower`, strip after `:`) leaves that FQCN unmatched by the `match`, so the method returns null and the property falls back to **fully untyped** (`{}`).

Two compounding effects (verified empirically on the #251 head, 8566e71):

- The column doesn't even get the conservative `{type: object}` the string-form `array`/`json`/`collection` casts produce.
- Because `$casts` outranks the `@property` tag in `propertyFor()`/`schemaFor()`, a precise docblock generic (`@property list<string> $tags`) is bypassed entirely — including #249's new list disambiguation, which only fires for the string-form casts.

## Proposal (Tier 0)

Recognise the framework's castable classes in `castToProperty()` (e.g. `is_a($cast, AsCollection::class, true)` family, or a small FQCN map):

- `AsCollection`, `AsEncryptedCollection` → same treatment as the `collection` string cast (incl. the #249 `@property`-tag list disambiguation);
- `AsArrayObject`, `AsEncryptedArrayObject` → same as `array`;
- `AsStringable` → `string`.

Custom `CastsAttributes` implementations stay at the untyped fallback (their value shape is genuinely unknowable at Tier 0 — `@property` tag / attribute territory), but should arguably fall through to the `@property` tag instead of swallowing it, same mechanism #249 used.

## Impact

Any model using the class-form casts (the documented modern style for `casts()` method declarations) loses both the JSON-object default and any docblock precision on those columns.

Radiergummi:
Closes #252

## Problem

Modern Laravel spells the JSON casts as castable class-strings in the `casts()` method:

```php
protected function casts(): array
{
    return ['tags' => AsCollection::class];   // or AsArrayObject::class, AsEncryptedCollection::class, …
}
```

`Model::getCasts()` reports the FQCN (`Illuminate\Database\Eloquent\Casts\AsCollection`). In `EloquentModelToSchema::castToProperty()` the normalisation (`strtolower`, strip after `:`) leaves that FQCN unmatched by the keyword `match`, so the method returns null and the column falls back to **fully untyped** (`{}`). Two compounding effects:

- The column doesn't even get the conservative `{type: object}` the string-form `array`/`json`/`collection` casts produce.
- Because `$casts` outranks the `@property` tag, a precise docblock generic (`@property list<string> $tags`) is bypassed entirely — including #249's list disambiguation.

## Plan

1. Fixtures + unit tests (`EloquentModelToSchemaTest`, `EloquentModelPropertyLookupTest`): a model whose `casts()` uses `AsCollection::class` / `AsArrayObject::class` / `AsEncryptedCollection::class` / `AsStringable::class`, plus a custom `CastsAttributes` with a `@property` tag. → red before fix
2. **A — recognise the framework castables** in `castToProperty()` (before the keyword normalisation): `AsCollection` / `AsEncryptedCollection` / `AsArrayObject` / `AsEncryptedArrayObject` → the JSON object/list shape (`jsonCastDefinition()`, inheriting #249's `@property` list disambiguation); `AsStringable` → `string`. Parameterised forms (`AsCollection::of(...)` → `…AsCollection:Foo`) are handled by stripping at `:`.
3. **B — stop swallowing the `@property` tag.** When a cast is unrecognised (a custom `CastsAttributes`, unknowable at Tier 0), fall through to the `@property` tag before the untyped fallback at both call sites (`propertyFor()`, `schemaFor()`) — the same precedence #249 established. A cast column is still always at least an untyped property.
4. `composer test && composer lint && vendor/bin/pint --test`

## Scope / boundary

`AsEnumCollection` / `AsEnumArrayObject` (which carry an enum-class parameter) are out of scope — they stay null and fall through to the `@property` tag via (B). Custom `CastsAttributes` value shapes remain genuinely unknowable at Tier 0 (attribute / `@property` territory).


src/Support/Extraction/EloquentModelToSchema.php
```diff
@@ -519,10 +548,21 @@ private function propertyFromSchema(string $name, OA\Schema $schema): OA\Propert
      */
     private function castToProperty(string $name, string $cast, ?TypeNode $declaredType = null): ?OA\Property
     {
-        // Normalise: take the part before `:` (e.g. `decimal:2` → `decimal`) and lowercase.
-        $normalised = strtolower(
-            str_contains($cast, ':') ? explode(':', $cast, 2)[0] : $cast,
-        );
+        // The part before `:` — a parameterised cast (`decimal:2`, `AsCollection:Foo`) or the
+        // bare keyword / class-string.
+        $castHead = str_contains($cast, ':') ? explode(':', $cast, 2)[0] : $cast;
```
Radiergummi: could just use strtok here?

Radiergummi: Done — `$castHead = strtok($cast, ':') ?: $cast;`. The `?: $cast` keeps the type a non-false `string` for PHPStan (strtok returns `false` only on an empty/all-delimiter input, which a cast value never is). Equivalent for cast strings, which never start with `:`.

PR #277, opened 2026-06-12
fix(#260): lint reads gathered (controller-aware) middleware, matching the generator

### LINKED ISSUE #260: Lint reads only route-declared middleware — controller middleware (HasMiddleware / constructor) is invisible to `operation.security-missing` and `publicendpoint.contradicts-middleware`
## Summary

After #258, all four generator-side middleware consumers read through the crash-safe `Support\Routing\RouteMiddlewareGatherer` seam, so security schemes, implicit error responses, spec matching, and `openapi:why` all see controller-applied middleware (native `HasMiddleware` / constructor resolution, plus the static-scan fallback).

The **lint subsystem** still reads `Route::middleware()` (route-declared only) in two places:

- `src/Lint/Tree/OperationNode.php` — `hasAuthMiddleware()`, consumed by `operation.security-missing`
- `src/Lint/Rules/PublicEndpointContradictsMw.php` — `findConflictingMiddleware()`

So generator and linter can disagree about the same route:

1. A controller that applies `auth:*` middleware in its constructor or via static `HasMiddleware::middleware()` is authed in the generated document, but `operation.security-missing` treats the route as unauthenticated — when no scheme is derivable, the missing-security finding that #76 introduced **silently does not fire** for exactly these routes.
2. `publicendpoint.contradicts-middleware` misses the contradiction when the conflicting auth middleware is controller-wired rather than route-declared.

Pre-existing (these readers never used `gatherMiddleware()`, so they were equally blind before #258) — but the one-uniform-list seam now makes the divergence pointed: switching both readers to `RouteMiddlewareGatherer::middlewareFor()` is a two-line change per site and inherits the crash-safety for free. Note `hasAuthMiddleware()`'s closure-middleware handling: the gathered list can contain non-strings, so the `str_starts_with` callbacks need an `is_string` guard after the switch.

Found while reviewing #258.

Radiergummi:
Closes #260

## Problem

Since #258, all generator-side middleware consumers read through the crash-safe `RouteMiddlewareGatherer` seam, so they see controller-applied middleware (constructor `$this->middleware(...)`, the static `HasMiddleware` form, plus the static-scan fallback). The **lint subsystem** still read `Route::middleware()` (route-declared only) in two places, so generator and linter could disagree about the same route:

1. `operation.security-missing` (via `OperationNode::hasAuthMiddleware()`) treated a controller-authed route as unauthenticated — when no scheme is derivable, the finding **silently did not fire**.
2. `publicendpoint.contradicts-middleware` (`PublicEndpointContradictsMw::findConflictingMiddleware()`) missed the contradiction when the conflicting auth middleware was controller-wired.

## Fix

Both readers now source the gathered middleware via `RouteMiddlewareGatherer::middlewareFor()`, filtering to strings (the gathered list can contain closure middleware). Both rules are container-resolved, so the gatherer is injected via constructor — the natural DI seam.

`OperationNode::hasAuthMiddleware()` had exactly one caller (`OperationSecurityMissing`); a plain value object can't reach the scoped gatherer, so the detection moves into the rule and the orphaned VO method is removed. Route-declared middleware is unaffected: the gathered list is a superset, so existing route-declared cases still resolve identically — the change only *adds* the controller-applied middleware the linter was blind to.

## Plan

1. Add a `HasMiddleware` lint fixture controller (`auth:sanctum`, no route-declared middleware) with a plain and a `#[PublicEndpoint]` method. Add unit cases to both rule tests: controller-applied auth → `operation.security-missing` fires; `#[PublicEndpoint]` + controller-applied auth → `publicendpoint.contradicts-middleware` fires. → red before fix
2. Inject `RouteMiddlewareGatherer` into `OperationSecurityMissing` and `PublicEndpointContradictsMw`; read the gathered list (string-filtered). Remove `OperationNode::hasAuthMiddleware()`.
3. Update the two rule unit tests to construct the rule with the gatherer (`app(RouteMiddlewareGatherer::class)`).
4. `composer test && composer lint && vendor/bin/pint --test`.


src/Lint/Rules/PublicEndpointContradictsMiddleware.php
Radiergummi, line 1: I dislike the abbreviation; even if the name is long, spell out "Middleware" consistently

Radiergummi: Done — renamed the class to `PublicEndpointContradictsMiddleware` (file, registration in `BaselineRegistration`, the `OperationSecurityMissing` cross-reference, and the test). The rule ID string `publicendpoint.contradicts-middleware` is unchanged, so configs/suppressions are unaffected.

PR #313, opened 2026-06-12
feat(#292): `query-builder.filter-duplicate` lint rule for duplicate `#[AllowedFilter]` names

### LINKED ISSUE #292: Duplicate #[AllowedFilter] with same name silently last-wins (no lint rule)
## Problem

`#[AllowedFilter]` is repeatable (`src/Plugins/QueryBuilder/Attributes/AllowedFilter.php:27`, `IS_REPEATABLE`), and `QueryBuilderParameterResolver` emits one parameter per attribute instance:

```php
// src/Plugins/QueryBuilder/Resolvers/QueryBuilderParameterResolver.php:73
$parameters[] = $this->filterParameter($attribute->newInstance());
```

When two `#[AllowedFilter]` attributes declare the **same** `$name` with different schemas (valid PHP — the attribute enforces no uniqueness), two `filter[name]` parameters are produced. `OperationBuilder`'s name+`in` dedup then silently keeps the last one and drops the first, with no diagnostic.

```php
#[AllowedFilter('status', type: 'string')]    // silently dropped
#[AllowedFilter('status', type: 'integer')]   // wins
public function index(): JsonResponse { … }
```

The existing `querybuilder.filter-type-missing` rule inspects every `AllowedFilter` instance, but nothing flags a duplicate `name`.

## Proposed fix

Add a lint rule (e.g. `querybuilder.filter-duplicate`) that flags two `#[AllowedFilter]` instances on the same action declaring the same wire name, so the silent last-wins collision is surfaced. Mirrors the per-instance inspection `querybuilder.filter-type-missing` already does.

Flagged during a holistic code review.

Radiergummi:
## Summary
`#[AllowedFilter]` is repeatable and `QueryBuilderParameterResolver` emits one parameter per instance. When two instances declare the **same** wire name with different schemas, `OperationBuilder`'s name+`in` dedup silently keeps the last and drops the first — no diagnostic:
```php
#[AllowedFilter('status', type: 'string')]    // silently dropped
#[AllowedFilter('status', type: 'integer')]   // wins
```

Per the #292 decision, this adds a lint rule `query-builder.filter-duplicate` (severity 3, matching the sibling `query-builder.filter-type-missing`) that flags two or more `#[AllowedFilter]` on the same action sharing a name — surfacing the silent last-wins collision. (Note: the actual rule-ID convention in the codebase is hyphenated `query-builder.*`, not the `querybuilder.*` in the issue text.)

It groups instances by name and emits **one** finding per duplicated name (regardless of how many extra instances).

## Changes
- `src/Plugins/QueryBuilder/Lint/Rules/QueryBuilderFilterDuplicate.php` (new) + registration in `QueryBuilderPlugin`.
- `tests/Unit/Plugins/QueryBuilder/Lint/QueryBuilderFilterDuplicateTest.php` — 2 same-name → 1 finding; 3 same-name → 1 finding; distinct names → 0; single → 0.
- `docs/linting.md` — catalog row (between the hand-maintained `lint-rule-catalog` markers).
- `CHANGELOG.md` — `[Unreleased] > Added`.

## Verification
- `composer test` → 2446 passed · `vendor/bin/pint --test` → passed · `composer lint` → no errors

Closes #292


Radiergummi (commented):
shouldn't this be accompanied by a PhpStan rule too?

Radiergummi:
Yes — added in 1ff36f2. `QueryBuilderFilterDuplicateRule` (`openapi.allowedFilter.duplicate`), registered in `extension.neon` against `FunctionLike` (the attribute targets methods/functions only). It's the natural fit here: the lint rule reads the attributes straight off the action via reflection — it never touches the generated spec — so the collision is a pure source-code mistake PHPStan can catch in the IDE before any spec exists. This mirrors the existing pairing pattern (`DuplicateResponseStatusRule` / `DuplicateResponseHeaderRule` + their runtime counterparts).

Two notes:
- It lives under `src/Plugins/QueryBuilder/PhpStan/Rules/` rather than the general `src/PhpStan/Rules/`, since it's plugin-specific (per CLAUDE.md). First plugin-scoped PHPStan rule — registration is still in the one global `extension.neon` (unavoidable; the rule is inert unless code actually uses `#[AllowedFilter]`). Happy to move it to `src/PhpStan/Rules/` if you'd rather keep all static rules colocated.
- `AttributeHelpers` deliberately resolves named args only, but `#[AllowedFilter('status')]` passes the name positionally (the documented idiom), so the rule falls back to the first unnamed argument. Without that it would miss essentially all real usage.

Docs catalog row + CHANGELOG updated; the lint rule's intentional-misuse test fixtures get a scoped `ignoreErrors` entry.

PR #316, opened 2026-06-13
fix(#294): inspect nullable `oneOf` / `anyOf` fields, flag uninspected unions

### LINKED ISSUE #294: Lint field tree skips oneOf/anyOf branches — under-reports on 3.1 nullable shapes
## Problem

`SpecTreeBuilder::buildFields()` resolves `allOf` composition (following `$ref`s into the component index) but **intentionally leaves `oneOf` / `anyOf` unresolved**:

```
// src/Lint/Tree/SpecTreeBuilder.php:511
// `oneOf` / `anyOf` are intentionally left [unresolved] …
```

The consequence: a schema written as `oneOf: [{$ref: A}, {$ref: B}]` (or `anyOf`) produces **zero field nodes**, so field-level rules (`field.description-missing`, `schema.enum-type-mismatch`, etc.) silently emit nothing for those properties — the linter under-reports.

This is most relevant for OpenAPI 3.1 nullable types, which the library itself now emits as `oneOf: [{type: …}, {type: 'null'}]` (see `NullableSchema`, and the `schema.nullable-via-deprecated-keyword` rule that pushes authors toward the 3.1 form). So newer specs increasingly use exactly the shape the field walker skips, and nothing tells the user their fields went uninspected.

## Proposed direction

Decide how the lint tree should treat `oneOf`/`anyOf`:
- inspect the common nullable shape (`oneOf` of a concrete schema + `{type: 'null'}`) by unwrapping to the non-null branch, and/or
- walk each branch and union/flag the field set, and/or
- at minimum emit a diagnostic ("fields under oneOf/anyOf were not inspected") so the gap isn't silent.

Filed as a design call — the builder comment marks the current omission deliberate, so the right resolution needs a decision rather than a drive-by change.

Flagged during a holistic code review.

Radiergummi:
Closes #294.

## Problem

`SpecTreeBuilder::buildFields()` resolved `allOf` but left `oneOf`/`anyOf` unresolved, so a property written as `oneOf: [{$ref:A},{$ref:B}]` produced **zero field nodes** — field-level rules (`field.description-missing`, `schema.enum-type-mismatch`, …) silently emitted nothing. This increasingly bit real specs because the library emits 3.1 nullable types as `oneOf: [{type:…}, {type:'null'}]` (`NullableSchema`) — exactly the skipped shape — with no signal that fields went uninspected.

## Fix (maintainer-approved direction)

1. **Unwrap the nullable shape.** Exactly one non-null branch + one-or-more pure `{type:'null'}` branches → unwrap to the concrete branch, build its fields normally, force `nullable`. Field/schema rules now inspect them. New `SchemaAccessor::classifyComposition()` helper.
2. **Flag genuine unions.** 2+ non-null alternatives → set a marker on `FieldNode` and emit a finding via a new lint rule **`schema.composite-fields-uninspected`** (level 3) so the gap is non-silent. Branch-union semantics are explicitly out of scope.
3. `allOf` resolution untouched.

## Seam

The tree builder doesn't emit findings (rules do, during the walk), so: builder records uninspected-union on the `FieldNode`; a new `FieldRule` reads the marker and emits the finding — suppressible and severity-overridable like any rule. Registered in `BaselineRegistration`; catalog row added to `docs/linting.md`; live in `openapi:lint --list`.

## Tests

New `SpecTreeBuilderCompositeFieldsTest` (4 cases): nullable-unwrap fires `field.description-missing`; genuine `anyOf` emits exactly one finding at level 3; nullable not flagged; empty `oneOf` doesn't crash.

## Gates

`composer test` 2446 passed · `vendor/bin/pint --test` passed · `composer lint` no errors. (Suite +4 over the `main` baseline of 2442 — exactly the new tests; reviewer verified no existing test was removed or weakened.)

Reviewed by a code-review subagent (PASS).

## Follow-up noted

A *top-level* component/response/request schema that is itself a bare multi-alternative `oneOf`/`anyOf` (not a property) still yields no nodes and no diagnostic — narrower than #294's property scope; will file separately.


src/Lint/Tree/FieldNode.php
```diff
@@ -35,6 +35,12 @@ public function __construct(
         public readonly array $examples,
         public readonly ?string $ref,
         public readonly ?OA\Property $raw,
+        /**
+         * True when this field's schema is a `oneOf` / `anyOf` of two or more genuine (non-null)
+         * alternatives, which the tree builder leaves uninspected. The standard nullable shape
+         * (one concrete branch plus `{type: 'null'}`) is unwrapped instead and does not set this.
+         */
+        public readonly bool $uninspectedCompositeBranches = false,
```
Radiergummi: I don't like this - we're leaking requirements for a single rule into the node state here.

src/Lint/Tree/SpecTreeBuilder.php
Radiergummi, line 1: You're rambling in the code comments again, repeating yourself over and over, across this PR. Trim it down.

src/Lint/Tree/FieldNode.php
Radiergummi: Removed the flag. `SchemaCompositeFieldsUninspected` now derives the condition itself from `$field->raw` via `SchemaAccessor::classifyComposition()` — same pattern `SchemaNullableViaDeprecatedKeyword` already uses to read a raw-schema condition off the node. The domain node carries no rule-specific state anymore; the builder still uses `classifyComposition()` only for the structural unwrap (branch + nullable). c123eac

src/Lint/Tree/SpecTreeBuilder.php
Radiergummi: Trimmed. The nullable-vs-union explanation now lives in one place — `SchemaAccessor::classifyComposition()`'s docblock. The `buildFields()` docblock, the inline loop comment, the rule docblock, and the `level()` comment were all repeating it; they now point at or just summarize it. c123eac

Issue #236, opened 2026-06-10
Read `->noContent()` and `->setStatusCode()` in the response body scan


Radiergummi:
## Summary

Complete the Tier-1 status-argument branch that #19 folded into #14: recover the literal success status from the two remaining whitelisted response-chain shapes the #14 scan does not read —

- `return response()->noContent();` → `204`, empty body
- `return response()->json($x)->setStatusCode(201);` / `->setStatusCode(Response::HTTP_CREATED)` → that code

## Context

PR #235 shipped the `response()->json(...)` literal scan, including the literal/class-constant **`status` argument** (`json($x, 201)`, `json(data: [...], status: 201)`). The `->noContent()` and `->setStatusCode()` chain forms from #19's folded plan were deliberately left out to keep that PR's matcher narrow.

Also belongs here (needs a resolver→builder signal, not just matching): **precedence against the #25 resource convention.** `OperationBuilder::applyConventionStatus()` currently layers the convention status (`store`→201, `destroy`→204) over any resolver-produced response, so an explicit literal status that disagrees with the convention loses. #19's stated precedence is the inverse — explicit literal (Tier-1) beats convention default (Tier-0) — and applies equally to the `json($x, 200)`-in-a-`store` case from #14.

## References

Parent: #5. Siblings: #14 (PR #235 — entry point `src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php`), #19 (closed; Tier-0 half shipped via #25, Tier-1 folded into #14, this issue carries the remainder), #227 (class-constant statuses resolve for free via `AstLiteralEvaluator`).

PR #330, opened 2026-06-14
feat: read explicit status argument on response()->noContent($status)

### LINKED ISSUE #328: Inline response scan: read an explicit status argument on response()->noContent($status)
## Context

Surfaced during #236 (PR #327). The new `response()->noContent()` matcher in `InlineJsonResponseResolver` hard-codes `204` and **intentionally ignores `noContent()`'s status argument** — #236 scoped the shape strictly to 204 to keep the matcher narrow.

So a caller that explicitly overrides the status:

```php
return response()->noContent(200);   // documented as 204, not 200
```

is mis-documented. `Illuminate\Routing\ResponseFactory::noContent(int $status = 204, array $headers = [])` accepts a status, and an author writing `noContent(202)` (e.g. an accepted-but-still-processing delete) means it.

## Proposed approach

Read the first positional / `status:` named argument of `noContent()` through the same `AstLiteralEvaluator` + `ensureSuccessStatus` path the `->setStatusCode()` / `json(..., status:)` branches already use: a literal 2xx becomes the documented status (still body-less, since `noContent()` strips the body), absent → 204, a non-2xx or non-literal degrades per the existing contract. Reuse the existing #240 explicit-status→convention precedence signal (already wired for `noContent()`).

Tier-1 — reuses the existing bounded body-scan machinery; no dataflow.

## Entry point

`src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php` (`findNoContentCall` / the no-content branch of `responseFromCall`).

Radiergummi:
Closes #328

## Tier

**Tier-1** — extends the existing bounded body-scan in `InlineJsonResponseResolver`. It reuses
the `AstLiteralEvaluator` + `ensureSuccessStatus` path the `json(..., status:)` and
`->setStatusCode()` branches already use. No variable tracking, no dataflow. Built at the lowest
tier that captures the idiom.

## Problem

The `noContent()` matcher (added in #236 / PR #327) hard-codes `204` and intentionally ignores
the call's status argument:

```php
return response()->noContent(200);   // documented as 204, not 200
```

`Illuminate\Routing\ResponseFactory::noContent(int $status = 204, array $headers = [])` accepts a
status, and an author writing `noContent(202)` means it. We should read it.

## Approach

The only behaviour to change is the no-content branch in
`InlineJsonResponseResolver::resolvePrimaryResponse()` (currently lines 138-143), which builds a
fixed `204`. Replace the fixed status with a small helper that reads the first positional /
`status:` named argument of the matched `noContent()` call:

- **Absent argument** → `204` (the helper default), explicit-status marker = `false` (still the
  helper default, the resource convention may override it — same rule the `json()` default-200
  branch follows at line 453). The response is `204 No Content`, body-less.
- **Literal / class-constant 2xx** → that status, explicit-status marker = `true` (the author
  wrote it; it must beat the resource-action convention per #240). The response stays **body-less**
  regardless of the status — `noContent()` strips the body, so e.g. `noContent(200)` documents a
  `200` response with no `content`.
- **Non-literal** (`noContent($request->integer('status'))`) → degrade: skip + note, exactly the
  contract `resolveStatus()` / `statusFromSetStatusCode()` already enforce ("no statically
  readable status code").
- **Non-2xx literal** (`noContent(404)`) → degrade via `ensureSuccessStatus()` + its existing
  non-2xx note. An error status must not claim the primary response.

`ensureSuccessStatus()` already returns the status for 2xx and `null` (with a note) otherwise, so
the success-status guard is reused as-is.

### Reusing the helpers

`argument()` resolves by position-or-name but is currently `private` and keyed to the json()
signature constants — it is signature-agnostic, so it can be called directly with
`('status', 0)`. `AstLiteralEvaluator::evaluate()` + `is_int` + `ensureSuccessStatus()` mirror the
`statusFromSetStatusCode()` shape. The new helper is essentially `statusFromSetStatusCode()`'s
structure with a `204` default instead of a degrade-when-absent, returning `int|null` (status, or
`null`/degrade) plus a flag for whether the status was author-written.

Concretely, add a private `statusFromNoContent(MethodCall $call, ReflectionMethod $method): int|false`
that returns the resolved 2xx status (`204` when absent), or `false` to degrade (non-literal or
non-2xx, both already noted by the reused evaluator/`ensureSuccessStatus`). The branch then:

```
$status = $this->statusFromNoContent($noContentCall, $method);
if ($status === false) { return null; }
$explicit = $this->argument($noContentCall->getArgs(), 'status', 0) !== null;
return $this->markStatusExplicit(
    new OA\Response(['response' => (string) $status, 'description' => HttpFoundationResponse::$statusTexts[$status] ?? sprintf('HTTP %d', $status)]),
    $explicit,
);
```

(Argument-spread/`unpack` guard: `noContent(...$args)` is non-literal anyway — the evaluator on a
non-literal-readable arg degrades; an explicit `unpack` check is unnecessary for a single scalar
status arg, matching `statusFromSetStatusCode()` which already guards `unpack`. Mirror that
`unpack` guard for consistency.)

## Files

- **Modify** `src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php`
  - Rework the no-content branch in `resolvePrimaryResponse()` to read the status.
  - Add private `statusFromNoContent()` (modeled on `statusFromSetStatusCode()`).
  - Update the two class-level docblock sentences that assert `noContent()` is "matched directly
    as a 204 ... its (defaulted) status argument is not read" (lines ~68 and ~240-243) to reflect
    that the explicit status is now read.
- **Modify** `tests/Fixtures/InlineJsonFixtureController.php` — add fixtures:
  - `noContentExplicitStatus(): SymfonyResponse` → `return response()->noContent(200);`
  - `noContentNamedStatus(): SymfonyResponse` → `return response()->noContent(status: 202);`
  - `noContentDynamicStatus(Request $request): SymfonyResponse` →
    `return response()->noContent($request->integer('status'));`
  - `noContentNonSuccess(): SymfonyResponse` → `return response()->noContent(404);`
  - (keep the existing `noContent()` → defaults-to-204 fixture as the absent-arg case.)
- **Modify** `tests/Unit/Core/Inference/InlineJsonResponseResolverTest.php` — add the proving
  tests (below). The existing `noContent()`→204 test stays green (regression guard for the default).

## Proving tests (TDD — write first, watch them fail)

In `InlineJsonResponseResolverTest.php`, alongside the existing no-content test:

1. **explicit 2xx is read and is body-less**: `noContentExplicitStatus` →
   `$response->response === '200'`, `description === 'OK'`, serialized has no `content` key,
   logger empty. (This is the headline failing test — currently emits `204`.)
2. **named `status:` argument**: `noContentNamedStatus` → `'202'`, `'Accepted'`, body-less.
3. **non-literal status degrades with a note**: `noContentDynamicStatus` → `null`, one note
   containing "status".
4. **non-2xx literal degrades**: `noContentNonSuccess` → `null`, one note containing "404".
5. **absent argument still defaults to 204** (regression): the existing
   `'documents a 204 from ->noContent() without a body schema'` test already covers this; keep it.

Optionally extend `OperationBuilderConventionStatusTest.php` only if the explicit-status →
convention precedence needs a no-content case; the json() path already covers the #240 marker, and
the unit `markStatusExplicit` flag assertion is sufficient — **skip** unless review asks.

## Verify

```
vendor/bin/pest tests/Unit/Core/Inference/InlineJsonResponseResolverTest.php
composer test            # full suite
vendor/bin/pint --test   # style
composer lint            # PHPStan level 8
```

## Bookkeeping

- **CHANGELOG.md** `[Unreleased]` → `Changed` (or `Added`): one line —
  "Inline response scan: `response()->noContent($status)` now documents an explicit literal 2xx
  status (and `status:` named arg); absent → 204, non-literal/non-2xx degrade per the body-scan
  contract."
- **docs/** — the inline response-scan behaviour is documented in the responses page (the same
  page that covers `response()->json()` body scanning). Add one sentence to the `noContent()`
  note that an explicit literal 2xx status is honored (still body-less). Locate via the existing
  `noContent` mention; if no observable doc covers `noContent()` yet, add the sentence where the
  inline json status argument is described.

## Controversy / merge gate

**Low.** Pure extension of an existing Core-plugin resolver; no `Contracts/**`, no stage-order
change, no new config key, no public-signature change. Reuses established helpers and the existing
degradation + #240 explicit-status contract. Should be an auto-merge candidate once CI is green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,7 @@ All notable changes to this project are documented here.
 - Pipeline order is expressed as a single ordered sequence in `BaselineRegistration::assemble()`; resolver fault isolation is centralized in `ResolverFaultBoundary`; pipeline classes are scoped singletons (Octane-safe).
 - Restructured namespaces to separate the public extension surface (`Contracts\`) from internal infrastructure (`Support\`).
 - `PaginatorKind` recognises Spatie's `PaginatedDataCollection`; `DataResponseResolver` handles union return types as `oneOf`.
+- The inline `response()->noContent($status)` body-scan reads an explicit literal/named 2xx status (body-less); a non-literal or non-2xx status degrades. (#328)
 
 ### Fixed
 - Lint now surfaces top-level and field-level bare `oneOf` / `anyOf` schemas across components, responses, and request bodies. (#294, #318)
@@ -49,7 +50,6 @@ All notable changes to this project are documented here.
 ### Documentation
 - New [Migrating from L5-Swagger](docs/migrating-from-l5-swagger.md) guide and [CI integration](docs/ci.md) guide. (#123, #158)
 - Documentation restructure: the monolithic `docs/usage.md` is split into focused pages, with an accuracy sweep across the set.
-- The inline `response()->noContent($status)` body-scan now reads an explicit literal/named 2xx status argument (a body-less `202` etc.); a non-literal or non-2xx status degrades. (#328)
 
 ## [0.1.0] - 2026-05-18
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,7 @@ All notable changes to this project are documented here.
 ### Documentation
 - New [Migrating from L5-Swagger](docs/migrating-from-l5-swagger.md) guide and [CI integration](docs/ci.md) guide. (#123, #158)
 - Documentation restructure: the monolithic `docs/usage.md` is split into focused pages, with an accuracy sweep across the set.
+- The inline `response()->noContent($status)` body-scan now reads an explicit literal/named 2xx status argument (a body-less `202` etc.); a non-literal or non-2xx status degrades. (#328)
 
 ## [0.1.0] - 2026-05-18
 
--- a/docs/auto-derivation.md
+++ b/docs/auto-derivation.md
@@ -286,7 +286,10 @@ A status you wrote in the call wins over the resource-action
 With no status argument the response documents as `200` and the convention
 still applies, so a conventional `store` with no explicit status is promoted to
 `201`. A literal `204` documents as `204 No Content` without a body schema — the
-runtime strips the body — as does a bare `response()->noContent()`. A chained
+runtime strips the body — as does a bare `response()->noContent()`. A
+`response()->noContent($status)` is read at its status argument
+(`noContent(202)` / `noContent(status: 202)` documents `202`, still body-less);
+a non-literal or non-2xx status degrades. A chained
 `->setStatusCode(<literal>)` — `response()->json([...])->setStatusCode(201)`,
 class constants such as `Response::HTTP_CREATED` included — overrides the status
 and likewise wins over the convention. When several calls match, a **returned**
--- a/src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php
+++ b/src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php
@@ -320,12 +320,16 @@ final readonly class InlineJsonResponseResolver implements PrimaryResponseResolv
         return !($namespacedName instanceof Name && function_exists($namespacedName->toString()));
     }
 
-    private function note(ReflectionMethod $method, string $reason): void
-    {
+    private function note(
+        ReflectionMethod $method,
+        string $reason,
+        string $callExpression = 'response()->json()',
+    ): void {
         $this->logger->notice(
             sprintf(
-                'response()->json() call in %s::%s %s; no response inferred. '
+                '%s call in %s::%s %s; no response inferred. '
                 . 'Annotate the action with #[Response] to document it.',
+                $callExpression,
                 $method->getDeclaringClass()->getName(),
                 $method->getName(),
                 $reason,
@@ -438,13 +442,14 @@ final readonly class InlineJsonResponseResolver implements PrimaryResponseResolv
             }
 
             if (is_int($status)) {
-                return $this->ensureSuccessStatus($status, $method) ?? false;
+                return $this->ensureSuccessStatus($status, $method, 'response()->noContent()') ?? false;
             }
         }
 
         $this->note(
             $method,
             'has no statically readable status code, so the body must not be documented under a guessed status',
+            'response()->noContent()',
         );
 
         return false;
@@ -588,8 +593,11 @@ final readonly class InlineJsonResponseResolver implements PrimaryResponseResolv
      * or a `->setStatusCode(403)`) is an error response, and taking it as primary would evict the
      * operation's success response.
      */
-    private function ensureSuccessStatus(int $status, ReflectionMethod $method): ?int
-    {
+    private function ensureSuccessStatus(
+        int $status,
+        ReflectionMethod $method,
+        string $callExpression = 'response()->json()',
+    ): ?int {
         if ($status < 200 || $status > 299) {
             $this->note(
                 $method,
@@ -597,6 +605,7 @@ final readonly class InlineJsonResponseResolver implements PrimaryResponseResolv
                     'has a literal non-2xx status (%d) — an error response must not claim the primary response',
                     $status,
                 ),
+                $callExpression,
             );
 
             return null;
--- a/src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php
+++ b/src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php
@@ -65,7 +65,8 @@ use function strtolower;
  * status may claim the primary response: a straight-line non-2xx literal (the guarded-success +
  * terminal-error-fallback idiom) degrades with a note rather than evicting the operation's
  * success response. A 204 documents without content — the runtime strips the body. A
- * `response()->noContent()` is matched directly as a 204 with no body. A chained
+ * `response()->noContent(<status>)` is matched as a body-less response at its status argument
+ * (204 when absent, the literal 2xx otherwise); a non-literal or non-2xx status degrades. A chained
  * `->setStatusCode(<literal|class-constant>)` overrides the status (and beats the resource-action
  * convention); a non-literal `->setStatusCode()` or a body-mutating `->setData()` degrades the
  * call. Header/cookie chains stay matched.
@@ -136,8 +137,20 @@ final readonly class InlineJsonResponseResolver implements PrimaryResponseResolv
         $noContentCall = $this->findNoContentCall($statements);
 
         if ($noContentCall instanceof MethodCall) {
+            $status = $this->statusFromNoContent($noContentCall, $method);
+
+            if ($status === false) {
+                return null;
+            }
+
+            // Writing noContent() at all is an affirmative choice of a body-less success response,
+            // so it beats the resource-action convention regardless of whether a status argument
+            // is present (absent → 204; literal 2xx → that status). Only the value varies.
             return $this->markStatusExplicit(
-                new OA\Response(['response' => '204', 'description' => 'No Content']),
+                new OA\Response([
+                    'response' => (string) $status,
+                    'description' => HttpFoundationResponse::$statusTexts[$status] ?? sprintf('HTTP %d', $status),
+                ]),
                 true,
             );
         }
@@ -238,7 +251,7 @@ final readonly class InlineJsonResponseResolver implements PrimaryResponseResolv
     /**
      * The matched `response()->noContent()` call, preferring a *returned* call over one only
      * assigned to a variable, mirroring {@see self::findJsonCall}. `noContent()` always documents
-     * a 204 with no body, so its (defaulted) status argument is not read.
+     * a body-less response; {@see self::statusFromNoContent()} reads its status argument.
      *
      * @param list<Stmt> $statements
      */
@@ -398,6 +411,40 @@ final readonly class InlineJsonResponseResolver implements PrimaryResponseResolv
             $method,
             'is chained into ->setStatusCode() with no statically readable status code, '
             . 'so the body must not be documented under a guessed status',
+        );
+
+        return false;
+    }
+
+    /**
+     * The status of a matched `response()->noContent(<status>)`: 204 when the argument is absent
+     * (the helper default), the literal/class-constant 2xx status when present, or false (degrade
+     * with a note) when the argument is non-literal or a non-2xx literal — the same contract
+     * {@see self::statusFromSetStatusCode()} follows. The response stays body-less either way.
+     */
+    private function statusFromNoContent(MethodCall $call, ReflectionMethod $method): int|false
+    {
+        $statusArgument = $this->argument($call->getArgs(), 'status', 0);
+
+        if ($statusArgument === null) {
+            return 204;
+        }
+
+        if (!$statusArgument->unpack) {
+            try {
+                $status = AstLiteralEvaluator::evaluate($statusArgument->value);
+            } catch (NonLiteralValueException) {
+                $status = null;
+            }
+
+            if (is_int($status)) {
+                return $this->ensureSuccessStatus($status, $method) ?? false;
+            }
+        }
+
+        $this->note(
+            $method,
+            'has no statically readable status code, so the body must not be documented under a guessed status',
         );
 
         return false;
--- a/tests/Fixtures/InlineJsonFixtureController.php
+++ b/tests/Fixtures/InlineJsonFixtureController.php
@@ -68,6 +68,26 @@ class InlineJsonFixtureController extends Controller
         return response()->noContent();
     }
 
+    public function noContentExplicitStatus(): SymfonyResponse
+    {
+        return response()->noContent(200);
+    }
+
+    public function noContentNamedStatus(): SymfonyResponse
+    {
+        return response()->noContent(status: 202);
+    }
+
+    public function noContentDynamicStatus(Request $request): SymfonyResponse
+    {
+        return response()->noContent($request->integer('status'));
+    }
+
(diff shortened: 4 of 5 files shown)


Radiergummi:

Radiergummi:
🤖 **[lead]** plan looks tight (Tier-1, surgical, reuses existing machinery). → plan-review.

Radiergummi:
🤖 **[reviewer]** plan-review: **changes needed** — one correctness bug in the explicit-status marker rule for the absent-argument case.

**Tier / scope / reuse: all sound.** Genuinely Tier-1 (extends the existing bounded body-scan in `InlineJsonResponseResolver`, no dataflow). The helpers it claims to reuse exist with the described shapes: `argument()` is signature-agnostic (`InlineJsonResponseResolver.php:487`), `statusFromSetStatusCode()` is a faithful template (`:380`), `ensureSuccessStatus()` returns `int|null` with a note (`:544`). Fixture names do not collide — the existing `noContentStatus` (`tests/Fixtures/InlineJsonFixtureController.php:151`) is a `json(..., 204)` case, not a `noContent()` case. Proving tests 1–4 are well-specified.

**Required change — the absent-argument case must keep `explicit = true`, not flip to `false`.**

The plan says: absent argument → 204, explicit-status marker = `false` ("the resource convention may override it — same rule the `json()` default-200 branch follows at line 453").

That breaks an existing, intentional contract. The current no-content branch (`:138-143`) marks `explicit = true` unconditionally, and that is load-bearing:

- `tests/Unit/Support/Generator/OperationBuilderConventionStatusTest.php:64` — "keeps a `->noContent()` 204 over the store convention (201)" — asserts a **bare** `noContent()` in a POST `store` action produces `204`, not the convention's `201`.
- The fixture comment spells out the intent: `tests/Fixtures/ResourceConventionLiteralController.php:36` — "POST store → convention says 201, but `->noContent()` must win with 204."

Under the plan, `explicit = false` for the absent-arg case lets the resource convention relabel a bare `noContent()` to `201` — breaking that test and silently regressing real behavior.

The `json()` analogy is false: `json([])`'s defaulted 200 is the helper's invisible default the author did not choose, but **writing `noContent()` at all is an affirmative choice of a body-less 204** — which is exactly why the current code marks it explicit.

**Fix:** the explicit marker for `noContent()` must be `true` whenever the matcher fires, regardless of whether the status argument is present (absent → 204 explicit; literal 2xx → that status, explicit). Only the *value* of the status changes with the argument; the explicit-wins-over-convention contract does not. Drop the "absent → explicit `false`" rule and the line-453 analogy from the Approach section.

**Also:** the plan's "regression test stays green" bullet names the wrong test (the json()→204 unit test). The test actually at risk is `OperationBuilderConventionStatusTest.php:64` — add it to the proving/regression list and assert it stays `204`.

Everything else (CHANGELOG `[Unreleased]`, docs at `docs/auto-derivation.md:288-289`) is correctly identified. Fix the marker rule and this is an auto-merge candidate.

Radiergummi:
🤖 **[lead]** plan-review found a real contract bug (explicit-status marker on bare noContent must stay true — 204 beats convention 201). Folding the correction into implementation → coding.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed ba26a84. Ready for review.

- Reworked the no-content branch in `InlineJsonResponseResolver::resolvePrimaryResponse()` to read the status argument via new private `statusFromNoContent()` (modeled on `statusFromSetStatusCode()`): absent → 204, literal/named 2xx → that status (still body-less), non-literal or non-2xx → degrade with a note (reusing `ensureSuccessStatus()`).
- **Per reviewer-1's plan-review correction:** the explicit-status marker stays `true` whenever the noContent() matcher fires (only the status *value* varies with the argument). The at-risk convention test `OperationBuilderConventionStatusTest.php:64` (bare `noContent()` 204 beats store convention 201) stays green.
- Added 4 fixtures + 4 unit tests covering literal/named/dynamic/non-2xx status; updated class + `findNoContentCall` docblocks, `docs/auto-derivation.md`, and a `CHANGELOG.md [Unreleased]` line.
- Remote CI green across the full matrix incl. the swagger-php 5.8 job.

Radiergummi:
🤖 **[lead]** implementation green on the full CI matrix (incl. swagger-php 5.8), diff 129/+4 across 5 files (non-controversial). → code review (two independent reviewers).

Radiergummi:
🤖 **[reviewer]** review round 1 — 1 finding (minor, bookkeeping):

The implementation is correct and my plan-review correction holds. Verified directly:

- **Marker stays `true` whenever the matcher fires** (`InlineJsonResponseResolver.php:139-153`) — absent → 204 explicit, literal 2xx → that status explicit. The protected contract test `OperationBuilderConventionStatusTest.php:64` ("keeps a `->noContent()` 204 over the store convention (201)") is green, so a bare `noContent()` still beats the resource convention.
- **`statusFromNoContent()` (`:419-450`)** faithfully mirrors `statusFromSetStatusCode()`: `unpack` guard, `AstLiteralEvaluator::evaluate` + `is_int`, then `ensureSuccessStatus($status) ?? false`. Edge paths trace correctly — non-2xx literal (404) degrades via `ensureSuccessStatus`'s own note (`?? false`), non-literal falls through to the generic note. No risk of a valid 2xx status being coerced to `false` (2xx is never falsy).
- **Body-less guarantee** holds for the non-204 explicit case too: the no-content branch never attaches `content`, and test 1 asserts `noContentExplicitStatus` → `200` with no serialized `content` key.
- **Tests** cover all four new paths (literal, named, non-literal degrade, non-2xx degrade) plus the absent-arg 204 regression; each would fail without the change. Conventions clean (no abbreviations, concise concrete comment at `:144-146`, no strict-types/`@internal` regressions). Docs prose at `docs/auto-derivation.md:286-292` is accurate. No scope creep.

**Finding 1 — `CHANGELOG.md:52`: entry is in the wrong section.** The new bullet sits under `### Documentation`, but it describes a behavior change to the inference engine ("the inline `response()->noContent($status)` body-scan **now reads** an explicit ... status argument"), not a documentation change. The plan called for `### Changed` (or `Added`), and there is an existing `### Changed` section at `:29`. Please move it there.

(Style note, non-blocking: given #325's "headline entries only" curation — individual tier-1 sub-features are folded into the single headline at `:8` — consider whether this warrants a standalone bullet at all, or a fold-in. Your call; moving it to `### Changed` resolves the finding either way.)

Fix the section placement and this is an approve from me.

Radiergummi:
🤖 **[reviewer-2]** review round 1: **approved** with two non-blocking nits.

Reviewed adversarially against commit ba26a84 (independent of reviewer-1).

**Correctness — all four edge cases hold:**
- Absent arg → `204`, marker `explicit=true` (unconditional in the branch), so a bare `noContent()` still beats the resource-action `201` convention. Verified at the integration level: `OperationBuilderConventionStatusTest > keeps a ->noContent() 204 over the store convention (201)` stays green.
- Literal 2xx → that status, **body-less** (no `content` key), marker `true` — beats the convention per #240.
- Non-literal (`noContent($request->integer(...))`) → degrades to `null` with a note.
- Non-2xx literal (`noContent(404)`) → degrades via the reused `ensureSuccessStatus()` + its 404 note.

`statusFromNoContent()` faithfully mirrors `statusFromSetStatusCode()` (the `unpack` guard, `AstLiteralEvaluator::evaluate()`, `is_int`, `ensureSuccessStatus(...) ?? false`). `argument('status', 0)` is correctly used signature-agnostically. Class-constant 2xx is covered transitively by the shared evaluator (already tested on the json path).

**Tests — TDD verified:** reverting only the resolver to `main` makes all four new tests fail (`200`/`202`/non-literal/`404`) while the existing `noContent()`→204 regression test stays green. With the change, 37/37 pass.

**Gates:** `vendor/bin/pint --test` clean; `composer lint` (PHPStan L8) clean; resolver + convention suites green.

**Findings (non-blocking, fix-on-merge if convenient):**

1. **CHANGELOG bucket** — `CHANGELOG.md`: the entry sits under `### Documentation`, but its text ("the body-scan **now reads** an explicit literal/named 2xx status") describes a **behaviour change**, not a docs change. The sibling Tier-1 body-inference work lives under `### Added`/`### Changed`. Move this line to `### Changed`.

2. **`note()` prefix wording** — `InlineJsonResponseResolver.php` `note()` hard-codes the prefix `response()->json() call in %s::%s ...`. The new degradation paths route `noContent($dynamic)` / `noContent(404)` through `note()`, so an author gets a generation-log line mislabeling their `noContent()` call as a `json()` call. On `main` noContent never degraded, so this is newly reachable. Cosmetic (the `#[Response]` remedy still applies), but the message is now inaccurate for the noContent shape. Consider generalizing the prefix or passing the helper name.

Neither blocks merge. Approving.

Radiergummi:
🤖 **[coder]** addressed round 1: (1) moved the CHANGELOG `[Unreleased]` entry from `### Documentation` to `### Changed` as a concise one-liner; (2) `note()` now takes an optional call-expression label (default `response()->json()`), and the noContent degradation paths pass `response()->noContent()` so the log prefix reads accurately — threaded through `ensureSuccessStatus()` for the non-2xx case too. Strengthened the two noContent degradation tests to assert the prefix. `composer check` green (Pint, PHPStan, 2546 tests); remote CI green across the full matrix incl. swagger-php 5.8. Pushed e3ec92d.

Radiergummi:

Round-1 finding resolved and verified:

- **CHANGELOG** entry moved out of `### Documentation` into `### Changed` (`CHANGELOG.md:36`), removed from the old location. Correctly categorized as a behavior change now.
- **Bonus accuracy fix (not scope creep):** `note()` and `ensureSuccessStatus()` take an optional `$callExpression` label (default `response()->json()`, preserving existing behavior), and the noContent degradation paths pass `response()->noContent()` so the generation log no longer misleadingly says `response()->json()` for a noContent() call. Backward-compatible default, threaded through the non-2xx note correctly.
- **Tests strengthened:** both noContent degradation tests now assert the `response()->noContent()` prefix (`InlineJsonResponseResolverTest.php:234,248`) — these would fail under the old hard-coded label, so the corrected log is properly proven.

Ran the resolver + convention suites locally: 39 passed (127 assertions), including the `->noContent()` 204-over-convention contract test. CI green across the matrix. Correctness, edge cases, tests, conventions, boundaries (Core-plugin only, no Contracts/Support/host-app changes), and bookkeeping all hold. Approve.

Radiergummi:
🤖 **[reviewer-2]** review round 2: **approved, no concerns.**

Both round-1 nits resolved on e3ec92d with no regression:

1. **CHANGELOG** — entry moved from `### Documentation` to `### Changed` and reworded to drop "now". ✅
2. **`note()` prefix** — `note()` and `ensureSuccessStatus()` gained an optional `$callExpression` (default `response()->json()`, so every existing json-path call site is byte-for-byte unchanged); both noContent degradation paths thread `response()->noContent()` through. The two degradation tests now assert that prefix, so the threading is proven. Surgical, no signature break. ✅

Verified on e3ec92d (worktree HEAD, clean tree): 37/37 pass (131 assertions, +2 from the strengthened prefix asserts), `vendor/bin/pint --test` clean, `composer lint` (PHPStan L8) clean. Final verdict: **approve.**

Radiergummi:
🤖 **[lead]** two independent reviewer approvals on e3ec92d + full CI matrix green. → survey Layer A gate (area:responses is generation-affecting).

Radiergummi:

Gate: survey Layer A (deterministic corpus). area:responses → generation-affecting, gate applies.

- **Baseline:** regenerated on current `main` `3eca466` (the cached baseline at `2a5d718` was stale — `main` advanced 25+ commits since, several generation-affecting). All 11 apps booted, `generateExit=0`.
- **Candidate:** PR branch `feat/328-inline-response-scan-read-an-explicit-st` @ `e3ec92d`; all 11 apps installed that exact library commit, same pinned app SHAs.
- **Result:** every deterministic metric (paths, operations, apiOperations, responseSchemas, documentedResponses, requestBodies, componentSchemas) is **identical** baseline↔candidate for all 11 apps; no new generation crash.

Why a 0-delta pass is correct here: #328 only changes a spec when an app calls `response()->noContent(<status>)` with an explicit non-204 2xx (status changes) or a non-literal/non-2xx status (degrades to no inferred response). No corpus app has an **application-level** `noContent(<status>)` call — the only two non-bare calls in the trees (`noContent(401)`, a ternary) live in vendor framework code (`Illuminate/.../Exceptions/Handler.php`) the body-scanner never reaches. Expected and observed: no spec moves.

| metric | corpus baseline = candidate |
|--------|------------------------------|
| crashes | 0 → 0 |
| regressions | 0 |
| improvements | 0 |

Radiergummi:
🤖 **[lead]** **SUPERVISED HOLD — clean auto-merge candidate, not merging.**

This is the autonomous-team's supervised first run (`stop before the first merge` directive), so I'm escalating instead of auto-merging. **Nothing is wrong with this PR** — it cleared every gate:

- ✅ Plan-review caught and corrected a real contract bug (bare `noContent()` must keep explicit-status = true so 204 beats the resource convention's 201).
- ✅ Two independent reviewer approvals on `e3ec92d` (reviewer-1, reviewer-2).
- ✅ Full remote CI matrix green incl. the swagger-php 5.8 job, snapshot, quality.
- ✅ Survey Layer A: 0 regressions across all 11 corpus apps (baseline refreshed on current `main`).
- ✅ Non-controversial: no Contracts / stage-order / config-key / public-signature change; diff 145/-9 across 5 files (sits right at the ~150-line heuristic).

**To merge:** `gh pr merge 330 --squash --delete-branch` (or merge via the UI). In unsupervised mode the team would have squash-merged this automatically.

PR #332, opened 2026-06-14
QueryBuilder chain reader: unwrap fluent modifier calls on value-object constructors

### LINKED ISSUE #257: QueryBuilder chain reader: unwrap fluent modifier calls on value-object constructors (->nullable(), ->default(), ->ignore(), ->delimiter())
## Summary

The #15 chain reader (`QueryBuilderChainReader::elementName()`) matches an allow-list element only when the element expression **is** a whitelisted static constructor (`StaticCall`). Spatie's fluent modifiers wrap that constructor in a `MethodCall` — `AllowedFilter::exact('healthy')->nullable()` — so the element is treated as non-literal and dropped (partial-keep + NOTICE).

All of Spatie's instance modifiers return `static` and never change the public wire name: `AllowedFilter::nullable()`, `default()`, `ignore()`, `delimiter()` (and `AllowedSort::defaultDirection()`). Unwrapping the method-call spine down to the underlying static constructor — exactly like the chain-root walk in `rootsAtBuilder()` — would read these without any dataflow.

## Evidence (real corpus app)

SpeedtestTracker `App\Http\Controllers\Api\V1\ResultsController::list`:

```php
->allowedFilters([
    AllowedFilter::operator('ping', FilterOperator::DYNAMIC),
    AllowedFilter::exact('healthy')->nullable(),   // <- dropped today
    AllowedFilter::exact('status'),
    ...
])
```

With the QueryBuilderPlugin enabled (verified during the PR #256 review), the generated operation documents `filter[ping]`, `filter[status]`, … but **not** `filter[healthy]`; the partial-keep NOTICE fires as designed. Degradation is graceful and contract-compliant — this issue is about closing the gap, not a bug.

## Sketch

In `elementName()`, before the `StaticCall` check, walk `MethodCall->var` until a non-`MethodCall` node is reached; if that root is a whitelisted constructor of the matching kind, read its name argument as today. No modifier whitelist needed — a modifier cannot rename the parameter (the name is fixed in the constructor).

Found while reviewing PR #256 (issue #15).

Radiergummi:
## Plan

Closes #257.

### Problem

`QueryBuilderChainReader::elementName()` reads an allow-list element only when the element expression **is** a whitelisted value-object static constructor (`StaticCall`). Spatie's fluent instance modifiers wrap that constructor in a `MethodCall`:

```php
->allowedFilters([
    AllowedFilter::exact('healthy')->nullable(),  // MethodCall(var: StaticCall, name: nullable)
])
```

so the element is seen as non-`StaticCall`, falls through to the literal evaluator, fails, and is dropped (partial-keep + the `request.query.parameters` degrade NOTICE). The wire name is fully determined by the underlying constructor's first argument; the modifier (`->nullable()`, `->default()`, `->ignore()`, `->delimiter()`, `AllowedSort::defaultDirection()`) only changes server-side semantics and can never rename the parameter.

### Tier

**Tier 1, no widening.** This is a pure AST-shape fix inside the existing bounded chain reader — unwrapping a method-call spine to its root expression, exactly the walk `rootsAtBuilder()` already performs on the chain receiver (`QueryBuilderChainReader.php:227`). No method-name whitelist for the modifiers is needed (the name is fixed in the constructor, so any modifier is safe to walk through), no variable tracking, no new statement budget. It stays well inside the Tier-1 envelope.

### Change

One file: `src/Plugins/QueryBuilder/Support/QueryBuilderChainReader.php`.

In `elementName()` (currently `:291`), before the `instanceof StaticCall` check, unwrap any `MethodCall` spine down to its root expression — mirroring `rootsAtBuilder()`:

```php
private function elementName(Expr $element, string $kind): ?string
{
    // Spatie's instance modifiers (->nullable(), ->default(), ->ignore(), ->delimiter(),
    // AllowedSort::...->defaultDirection()) wrap the value-object constructor but never
    // change the wire name — walk the method-call spine to the underlying constructor.
    $root = $element;

    while ($root instanceof MethodCall) {
        $root = $root->var;
    }

    if ($root instanceof StaticCall) {
        return $this->valueObjectName($root, $kind);
    }

    // A bare literal element is read directly (no spine), so evaluate the original node.
    try {
        $value = AstLiteralEvaluator::evaluate($element);
    } catch (NonLiteralValueException) {
        return null;
    }

    return is_string($value) ? $this->normaliseName($value, $kind) : null;
}
```

Note the literal path keeps evaluating `$element` (the original), not `$root`: a string literal is never a `MethodCall`, so `$root === $element` there; using `$root` for the static-call branch and `$element` for the literal branch keeps each branch reading the node it expects.

`valueObjectName()` already enforces the FQCN + constructor whitelist, so a non-Spatie method chain whose root happens to be some other `StaticCall` still returns `null` and is dropped — no false positives.

Feature-level, matching the existing pattern in `tests/Feature/Plugins/QueryBuilder/QueryBuilderChainDetectionTest.php` (controller fixture → `generateSpec()` → assert `parameters[].name`). Add fixtures + cases:

1. **Single modifier on a filter constructor** — `AllowedFilter::exact('healthy')->nullable()` yields `filter[healthy]`. (The headline regression from the issue's SpeedtestTracker evidence.)
2. **Chained modifiers** — `AllowedFilter::exact('q')->default('x')->ignore(null)` yields `filter[q]` (spine of length > 1 walked fully).
3. **Sort modifier** — `AllowedSort::field('created_at')->defaultDirection(...)` yields the `created_at` sort, with the `-` normalisation still applied.
4. **Mixed allow-list** — a list combining a plain literal, a bare constructor, and a modifier-wrapped constructor keeps **all three** (no partial-drop NOTICE) — guards the "literal branch still reads `$element`" invariant.
5. **Negative / degrade preserved** — a `MethodCall` rooted at a non-Spatie static call (or a dynamic expression) is still dropped and still triggers the partial-keep behaviour; the existing degrade test for a genuinely non-literal element must stay green.

I'll also confirm the existing `QueryBuilderChainDetectionTest` and `QueryBuilderParameterTest` suites stay green (the bare-constructor and literal paths are unchanged).

### Docs + changelog

- **`docs/plugins.md`** (QueryBuilder value-object paragraph, ~`:260`): the text currently lists the value-object constructors as readable but is silent on instance modifiers. Add a sentence: Spatie's fluent modifiers on a value-object element (`->nullable()`, `->default()`, `->ignore()`, `->delimiter()`, `->defaultDirection()`) are walked through — they change server-side behaviour, not the wire name.
- **`CHANGELOG.md` `[Unreleased]`**: #15's chain-reader bullet already exists. Append a short "Changed"/"Fixed"-style note that chain reading now sees value-object constructors wrapped in fluent modifiers (so `AllowedFilter::exact('x')->nullable()` is documented, not dropped), referencing #257.
- **No lint-rule change**, no config key, no `src/Contracts/**` touch, no stage-order change, no public-signature change.

### Controversy assessment

Low. Single internal `@internal` file, one private method, no public surface, no config, no contract change. The walk is identical in spirit to the already-shipped `rootsAtBuilder()` spine walk. Safe auto-merge candidate once tests + Pint + PHPStan are green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -50,6 +50,7 @@ All notable changes to this project are documented here.
 ### Documentation
 - New [Migrating from L5-Swagger](docs/migrating-from-l5-swagger.md) guide and [CI integration](docs/ci.md) guide. (#123, #158)
 - Documentation restructure: the monolithic `docs/usage.md` is split into focused pages, with an accuracy sweep across the set.
+- The query-builder chain reader sees value-object constructors wrapped in fluent modifiers, so `AllowedFilter::exact('healthy')->nullable()` is read as `filter[healthy]` instead of dropped. (#257)
 
 ## [0.1.0] - 2026-05-18
 
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -260,7 +260,11 @@ are read from string literals (class-constant strings included) and from
 Spatie's value-object static constructors (`AllowedFilter::exact(…)`,
 `AllowedSort::field(…)`, `AllowedInclude::relationship(…)`, …) — the first
 argument is the public wire name; internal names and `->defaultSort(…)` are
-server-side detail and not modelled. Both the array and the variadic form
+server-side detail and not modelled. Fluent modifiers on a value-object element
+(`->nullable()`, `->default()`, `->ignore()`, `->delimiter()`,
+`->defaultDirection()`) are walked through — they change server-side behaviour,
+not the wire name, so `AllowedFilter::exact('healthy')->nullable()` still reads
+as `filter[healthy]`. Both the array and the variadic form
 (`allowedSorts('name', 'created_at')`) work, and so does a chain assigned to
 a variable in one expression. A chain-derived filter gets a `string` schema —
 use `#[AllowedFilter]` where a filter needs a type, format, or description.
--- a/src/Plugins/QueryBuilder/Support/QueryBuilderChainReader.php
+++ b/src/Plugins/QueryBuilder/Support/QueryBuilderChainReader.php
@@ -47,9 +47,11 @@ use function ltrim;
  * Allow-list elements resolve from string literals (via {@see AstLiteralEvaluator}, so
  * class-constant strings work too) and from Spatie's value-object static constructors
  * (`AllowedFilter::exact('status')`, `AllowedSort::field('created_at')`, …), whose first
- * argument is always the public wire name. Spatie's variadic form
- * (`allowedSorts('name', 'created_at')`) reads the same way. A non-literal element is dropped
- * and the remaining ones kept; the call is reported as partially unreadable.
+ * argument is always the public wire name. Fluent instance modifiers on such a constructor
+ * (`->nullable()`, `->default()`, `->ignore()`, `->delimiter()`, `->defaultDirection()`) are
+ * walked through — they change server-side semantics, never the wire name. Spatie's variadic
+ * form (`allowedSorts('name', 'created_at')`) reads the same way. A non-literal element is
+ * dropped and the remaining ones kept; the call is reported as partially unreadable.
  *
  * @internal
  */
@@ -290,10 +292,21 @@ final class QueryBuilderChainReader
      */
     private function elementName(Expr $element, string $kind): ?string
     {
-        if ($element instanceof StaticCall) {
-            return $this->valueObjectName($element, $kind);
+        // Spatie's instance modifiers (->nullable(), ->default(), ->ignore(), ->delimiter(),
+        // AllowedSort::...->defaultDirection()) wrap the value-object constructor but never
+        // change the wire name — walk the method-call spine to the underlying constructor,
+        // exactly as the chain receiver is walked in rootsAtBuilder().
+        $root = $element;
+
+        while ($root instanceof MethodCall) {
+            $root = $root->var;
         }
 
+        if ($root instanceof StaticCall) {
+            return $this->valueObjectName($root, $kind);
+        }
+
+        // A bare literal element carries no spine, so evaluate the original node directly.
         try {
             $value = AstLiteralEvaluator::evaluate($element);
         } catch (NonLiteralValueException) {
--- a/tests/Unit/Plugins/QueryBuilder/QueryBuilderParameterResolverTest.php
+++ b/tests/Unit/Plugins/QueryBuilder/QueryBuilderParameterResolverTest.php
@@ -20,6 +20,7 @@ use Radiergummi\OpenApi\Tests\Support\ActionDescriptorFactory;
 use Spatie\QueryBuilder\AllowedFilter as SpatieAllowedFilter;
 use Spatie\QueryBuilder\AllowedInclude as SpatieAllowedInclude;
 use Spatie\QueryBuilder\AllowedSort as SpatieAllowedSort;
+use Spatie\QueryBuilder\Enums\SortDirection;
 use Spatie\QueryBuilder\QueryBuilder;
 
 use function array_find;
@@ -150,6 +151,35 @@ class QbChainFixtureController
     {
         return QueryBuilder::for(Author::class)
             ->allowedSorts('name', 'created_at')
+            ->get();
+    }
+
+    public function modifiedValueObjectChain(): mixed
+    {
+        return QueryBuilder::for(Author::class)
+            ->allowedFilters([
+                SpatieAllowedFilter::exact('healthy')->nullable(),
+                SpatieAllowedFilter::exact('q')->default('x')->ignore(null),
+            ])
+            ->allowedSorts([SpatieAllowedSort::field('-created_at')->defaultDirection(SortDirection::Ascending)])
+            ->get();
+    }
+
+    public function mixedModifierAllowListChain(): mixed
+    {
+        return QueryBuilder::for(Author::class)
+            ->allowedFilters([
+                'status',
+                SpatieAllowedFilter::exact('origin'),
+                SpatieAllowedFilter::exact('healthy')->nullable(),
+            ])
+            ->get();
+    }
+
+    public function modifiedImpostorChain(): mixed
+    {
+        return QueryBuilder::for(Author::class)
+            ->allowedFilters(['status', QbImpostorQueryBuilder::for('x')->get()])
             ->get();
     }
 
@@ -351,6 +381,27 @@ it('reads the variadic allow-list form', function (): void {
         ->and($sort->schema->items->enum)->toBe(['name', 'created_at']);
 });
 
+it('reads through fluent modifiers wrapping value-object constructors', function (): void {
+    $descriptor = chainDescriptor(QbChainFixtureController::class, 'modifiedValueObjectChain');
+    $params = makeQueryBuilderParameterResolver()->resolveQueryParameters($descriptor);
+
+    $sort = array_find($params, static fn(OA\Parameter $p): bool => $p->name === 'sort');
+
+    expect(parameterNames($params))->toContain('filter[healthy]')
+        ->and(parameterNames($params))->toContain('filter[q]')
+        ->and($sort)->not->toBeNull()
+        ->and($sort->schema->items->enum)->toBe(['created_at']);
+});
+
+it('keeps literal, bare and modifier-wrapped elements together without a partial drop', function (): void {
+    $logger = recordingLogger();
+    $descriptor = chainDescriptor(QbChainFixtureController::class, 'mixedModifierAllowListChain');
+    $params = makeQueryBuilderParameterResolver($logger)->resolveQueryParameters($descriptor);
+
+    expect(parameterNames($params))->toBe(['filter[status]', 'filter[origin]', 'filter[healthy]'])
+        ->and($logger->records)->toBe([]);
+});
+
 // endregion
 
 // region Graceful degradation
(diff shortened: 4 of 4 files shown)


Radiergummi:

Radiergummi:
🤖 **[lead]** planning done → plan-review. Tier-1 single-file chain-reader unwrap; low controversy.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved**, no changes required.

Verified against the code:
- `elementName()` is at `QueryBuilderChainReader.php:291`, the `StaticCall` check at `:293` — plan matches.
- The proposed `MethodCall`-spine unwrap mirrors the shipped `rootsAtBuilder()` walk (`:227-236`) exactly. `MethodCall` is already imported (`:12`), so no import churn.
- Tier-1 classification is correct: pure AST-shape unwrap, no modifier whitelist, no variable tracking, no statement-budget change. The name-invariance argument (a fluent modifier cannot rename the parameter, so walking through any modifier is safe) is sound, and `valueObjectName()` still enforces the FQCN + constructor whitelist on the root, so non-Spatie chains stay dropped — no false positives.
- The `$root` (static-call branch) vs `$element` (literal branch) distinction is a correct, deliberate detail — a string literal is never a `MethodCall`, so `$root === $element` there.

Bookkeeping all named and present:
- Proving tests: 5 cases (single modifier, chained modifiers, sort `->defaultDirection()` with `-` normalisation, mixed allow-list guarding the literal-branch invariant, negative/degrade preserved) extend `tests/Feature/Plugins/QueryBuilder/QueryBuilderChainDetectionTest.php`; the existing degrade test (`:131`) is the green-must-stay anchor.
- Docs: `docs/plugins.md` value-object paragraph (actual span 258-266; plan says `~:260` — minor) is genuinely silent on instance modifiers; a one-sentence addition is the right surgical edit. No `docs/README.md` index change needed.
- CHANGELOG: `[Unreleased]` #15 bullet exists (line 4); appending a fluent-modifier note is correct. No lint rule, no config key, no `Contracts`/stage-order/public-signature change.

Scope is surgical — one private method in one `@internal` file. Sound auto-merge candidate once tests + Pint + PHPStan are green.

Radiergummi:
🤖 **[lead]** plan-review approved (reviewer-1) → coding. TDD: failing tests first, then unwrap fix.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed 2fac5d8. Ready for review.

- **Fix:** `QueryBuilderChainReader::elementName()` now walks the `MethodCall` spine to its root before the `StaticCall` check (mirroring `rootsAtBuilder()`), so fluent modifiers on a value-object constructor are read through. `valueObjectName()` still enforces the FQCN + constructor whitelist on the root, so a modifier-wrapped non-Spatie static call is still dropped.
- **Tests** (TDD, all 5 plan cases live in `QueryBuilderParameterResolverTest`): single modifier → `filter[healthy]`; chained modifiers → `filter[q]`; `AllowedSort::field('-created_at')->defaultDirection(...)` → `created_at` sort with `-` still stripped; mixed literal + bare + modifier-wrapped list keeps all three with **no** partial-drop notice; modifier-wrapped non-Spatie root dropped while the readable element is kept (partial-keep notice preserved).
- **Deviation (minor):** the negative-case fixture uses a *mixed* list (`['status', QbImpostorQueryBuilder::for('x')->get()]`) rather than a single dropped element. A lone dropped element yields the *full* "could not be read statically" degrade message, not the partial-keep one; the mixed list exercises the partial-keep path the plan named and proves the `$element` (not `$root`) literal-branch invariant.
- **Bookkeeping:** `docs/plugins.md` value-object paragraph + `CHANGELOG.md [Unreleased]`.
- Remote CI green across the full matrix (PHP 8.4/8.5 × Laravel 12/13, swagger-php 5.8, snapshot, quality, coverage).

Radiergummi:
🤖 **[lead]** coding done; CI green full matrix (2fac5d8) → code-review. Two independent approvals required for auto-merge.

Radiergummi:

Adversarial pass on 2fac5d8:

- **Correctness:** the `MethodCall`-spine unwrap in `elementName()` (`QueryBuilderChainReader.php:295-309`) is tight. `valueObjectName($root)` re-enforces the FQCN + constructor whitelist on the unwrapped root, so a wrong-class root drops — no false positives. The literal branch correctly evaluates `$element` (not `$root`); a literal has no spine, so the distinction is load-bearing and right. `NullsafeMethodCall` (`?->`) is not unwrapped, but Spatie never emits it and it degrades gracefully — not a gap.
- **TDD verified independently:** reverting only the production file (head restored byte-identical afterward) makes `reads through fluent modifiers wrapping value-object constructors` and `keeps literal, bare and modifier-wrapped elements together without a partial drop` fail with the expected assertions (missing `filter[healthy]`; array mismatch). Genuine red-green.
- **Documented deviation confirmed sound:** the negative test `drops a modifier-wrapped non-Spatie root` passes on old code too — exactly because a *lone* dropped element yields the full-degrade message, not partial-keep. The coder switched to a mixed list to exercise the partial-keep NOTICE path the plan named; the existing `degrades a dynamic allow-list` test still covers full-degrade, so nothing is lost. Good call.
- **Test coverage of the public surface:** sort `-` normalisation (`AllowedSort::field(-created_at)->defaultDirection(...)` → `created_at`), chained modifiers (`->default()->ignore()`), single modifier, mixed list, and the no-false-positive negative are all covered. 39/39 QB tests green.
- **Conventions/bookkeeping:** Pint clean, PHPStan level-8 clean on the touched file, strict-types intact, comments concise (no issue numbers in code), modern PHP. CHANGELOG `[Unreleased]` entry + `docs/plugins.md` paragraph both accurate. No new lint rule, no Contracts/Core boundary touch, no host-app mutation, no public-signature change.

Approving as one of the two independent reviewers.

Radiergummi:
🤖 **[reviewer-2]** review round 1: **approved, no concerns.**

Second independent review (not deferring to reviewer-1). Verified locally against `2fac5d8`.

**Correctness.** The `MethodCall`-spine unwrap in `elementName()` (`QueryBuilderChainReader.php:295-309`) is right and mirrors `rootsAtBuilder()` (`:227-236`). Traced every branch:
- spine walked to a Spatie `StaticCall` → `valueObjectName($root, $kind)`; the FQCN + constructor whitelist is still enforced on the *root*, so wrong-class/wrong-kind/off-whitelist roots stay `null` — no false positives.
- spine walked to a non-Spatie `StaticCall` (`QbImpostorQueryBuilder::for(...)->get()`) → rejected by `valueObjectName`, dropped.
- spine walked to a non-`StaticCall` non-literal (e.g. `$var->nullable()`) → falls to the literal evaluator on the **original** `$element`, throws `NonLiteralValueException` → `null`. The deliberate `$root` (static branch) vs `$element` (literal branch) split is correct: a string literal is never a `MethodCall`, so `$root === $element` there.
- sort `-` normalisation still applied via `valueObjectName → normaliseName`.

**Tests / TDD.** Confirmed the two proving tests fail on `origin/main` and pass on the branch (reverted the source, reran: *reads through fluent modifiers* and *keeps literal, bare and modifier-wrapped together* both ⨯ → ✓). Full QB suite green (39 passed, 117 assertions). Pint + PHPStan (level 8) clean on both changed files. Codecov: 100% on the modified lines.

**Deviation (mixed negative fixture) — reasoning holds.** A lone dropped element yields the full "could not be read statically" degrade message, not the partial-keep one; the mixed `['status', QbImpostorQueryBuilder::for('x')->get()]` list correctly exercises the partial-keep path the plan named and asserts `['filter[status]']` kept + an `allowedFilters` notice. One honest caveat: this negative test passes **with or without** the source change (an impostor `MethodCall` was always non-`StaticCall` → dropped), so it is a false-positive *regression guard*, not a *proving* test — valuable to keep, just not load-bearing for the fix. The two VO-modifier tests carry the proof. Not a blocker.

**Test-gap hunt.** Walked the changed branches and the public surface (chain-derived param names). Covered: no-spine static call, spine length 1, spine length >1, sort normalisation, mixed literal+bare+modifier (literal-branch invariant), non-Spatie modifier root drop. The modifier-wrapped cross-kind case (e.g. `AllowedSort::field('x')->nullable()` inside `allowedFilters`) routes identically to the existing `crossKindValueObjectChain` test plus the spine walk — functionally redundant, not worth a dedicated case. No material gap.

**Conventions / boundaries / docs.** strict-types intact; no abbreviations; comments concise/concrete with no issue numbers in code; change confined to one `@internal` plugin file + its test; no `Support`/`Contracts`/Core leak, no host-app mutation, no public-signature/config/lint-rule change. `docs/plugins.md` value-object paragraph now documents the walked-through modifiers with a concrete `filter[healthy]` example and the honest "server-side behaviour, not wire name" framing; `CHANGELOG.md [Unreleased]` entry present. No `docs/README.md` index change needed.

Sound auto-merge candidate.

Radiergummi:
🤖 **[lead]** two independent approvals (reviewer-1 + reviewer-2), red-green verified → survey. Generation-affecting (area:plugins/params), Layer A required. Docs handled inline, no sub-phase.

Radiergummi:

Baseline regenerated on current \`main\` (\`7b2253c\`) — the cached baseline was at \`3eca466\`, two commits stale (incl. #330, responses). Candidate = PR branch \`2fac5d8\` (contains current main).

**Deterministic corpus (default config — QB plugin off):** all 11 apps **identical** across every metric (paths/operations/apiOperations/responseSchemas/documentedResponses/requestBodies/maxRequestProperties/componentSchemas/completeness); no new \`generate\` crashes. Expected: the QueryBuilder plugin is commented out in the published config, and `metrics.php` does not count parameters, so the fix is invisible to the headline corpus.

**Targeted QB demonstration (SpeedtestTracker, QB plugin enabled, scratch):**
| op | baseline (main) | candidate (#332) |
|----|-----------------|------------------|
| `GET /api/v1/results` | 7 filters + sort, **`filter[healthy]` dropped** | adds **`filter[healthy]`** |
| `GET /api/v1/stats` | `filter[start_at]`, `filter[end_at]` | unchanged |

`AllowedFilter::exact('healthy')->nullable()` (the modifier-wrapped element from issue #257) is now unwrapped and read; only `filter[healthy]` is added — nothing removed or renamed. Equal-or-improved, no regressions. App tree restored, vendor link back on main, library never modified.

Clear to merge.

PR #337, opened 2026-06-14
feat(#333): finish multipart/form-data file-upload request bodies

Radiergummi:
Closes #333

## Summary

Issue #333 asks us to *generate* a `multipart/form-data` request body when a request DTO carries
an `UploadedFile` field, and to retire/soften the `multipart.file-without-multipart` lint rule
once generation exists.

**Investigation finding: generation already ships, on both paths.** The issue's "current state"
is stale (it predates the resolvers wiring this up). Verified directly:

- **SpatieData** — `Plugins/SpatieData/Resolvers/DataClassRequestSchemaResolver.php:54-59` already
  returns `ResolvedSchema(mediaType: MediaType::MultipartFormData)` when
  `SchemaFromDataClass::hasFileProperties()` is true (transitive, cached, nested-aware).
  `SchemaFromDataClass::resolveObjectSchema()` (`:581`) emits `{type: string, format: binary}` for
  an `UploadedFile` property.
- **Core FormRequest** — `Plugins/Core/Resolvers/FormRequestRequestSchemaResolver.php:55-60` does
  the same via `SchemaFromFormRequest::hasFileFields()`; `ValidationRulesToSchema::applyFile()`
  (`:439`) maps `file|image|mimes|mimetypes` (and the `File`/`Dimensions` rule objects) to
  `format: binary` + `isFile`.
- `RequestBodyExtractor::buildRequestBody()` keys `content` by the resolved `MediaType`, so the
  multipart content-type string is already emitted.

So the **feature exists**. This issue is "finishing started work + closing the gaps," exactly as
the issue's own Notes section hedges. Scope is therefore narrow and surgical.

**Tier:** Tier-0. Pure reflection/signature reads (typed `UploadedFile` property; `file`/`image`
validation-rule strings). No body parsing, no dataflow. Nothing here climbs to Tier-1/2.

## What is actually missing

1. **SpatieData has no end-to-end test** that the Data-class path emits a `multipart/form-data`
   body with a binary field. `tests/Feature/Plugins/SpatieData/DataClassRequestSchemaResolverTest.php`
   and `SchemaFromDataClassTest.php` have **zero** file/multipart coverage; the only SpatieData
   file coverage is the *lint-rule* test (`hasFileProperties`), which never asserts the generated
   body. The Core path is well covered (`FormRequestSchemaTest.php:74-118` — file/image/mimes/
   dimensions + multipart + the plain-string-stays-string negative).
[...]

PR #338, opened 2026-06-14
feat(#334): emit additionalProperties for string-keyed Data map properties

### LINKED ISSUE #334: Emit `additionalProperties` for string-keyed array/collection types (map inference)
Spun off from #200 (parent), candidate "`additionalProperties`". The decision: this is an **inference gap, not an attribute** — build the Tier-0 inference; defer any attribute to the narrow residue.

## Current state

`additionalProperties` is emitted **nowhere** in generated output — a grep over `src/` finds it only in internal PHPDoc (`array<string, …>` annotations on the library's own code), never in a produced schema.

## Gap

A string-keyed array or collection is a map/dictionary and has a faithful OpenAPI shape, but the generator drops the value type:

```php
/** @var array<string, Money> */    // → should be {type: object, additionalProperties: {$ref: Money}}
/** @var array<string, string> */   // → should be {type: object, additionalProperties: {type: string}}
/** @var array<string, mixed> */    // → should be {type: object, additionalProperties: true}
```

`symfony/type-info` already carries the key and value types, so this is a **Tier-0 read** — the type system expresses it; the schema mapper just doesn't act on it. Treating it as an authoring attribute would be the smell (re-stating an inferable type).

## Proposal (Tier-0)

- In the schema mappers that resolve a property/return type, detect a string-keyed `array`/`Collection` value type and emit `{type: object, additionalProperties: <value schema>}`:
  - named value class → `additionalProperties: {$ref}`
  - scalar value → `additionalProperties: {type: …}`
  - `mixed`/untyped value → `additionalProperties: true`
- Locus: `src/Plugins/SpatieData/Support/SchemaFromDataClass.php` and the ApiResources field path; check whether a shared `Support/Extraction` helper is the right home so both reuse it.

## Attribute fallback (deferred, not this issue)

Only the residue where the value type cannot be named in PHP (e.g. a constrained `array<string, mixed>` shape) would want an attribute knob — and that belongs on the field-attribute surface (see the explicit-schema escape-hatch issue), not a new attribute. Do **not** build it until a concrete case demands it.

## Tests

- `array<string, Money>` → `additionalProperties: {$ref}`.
- `array<string, string>` → `additionalProperties: {type: string}`.
- `array<string, mixed>` / untyped value → `additionalProperties: true`.
- `list<X>` / int-keyed array → still a plain array (negative: not a map).
- Nested map (`array<string, array<string, X>>`) → nested `additionalProperties`.

## Bookkeeping

`docs/` schema page + `CHANGELOG.md [Unreleased]`.

Radiergummi:
Closes #334

## Summary

Emit OpenAPI `additionalProperties` for **string-keyed** array/collection types (maps/dictionaries)
read off a Spatie Data class's typed properties. `array<string, Money>` should become
`{type: object, additionalProperties: {$ref: Money}}` instead of being dropped to `{type: array}`.

**Tier:** Tier-0. `symfony/type-info` already carries the key type
(`CollectionType::getCollectionKeyType()`); the schema mapper just doesn't read it. No body
parsing, no dataflow — a pure signature/type read. Treating it as an attribute would be the smell
the issue calls out.

## Investigation finding — the gap is narrower than the issue states

The issue says `additionalProperties` is "emitted **nowhere**." That is **not accurate** — map
inference already works on the other type-reading paths. Verified:

- **PHPDoc generics** — `Support/Types/TypeNodeToSchema::fromGeneric()` (`:205-216`) already detects
  a string key and emits `{type: object, additionalProperties: <value>}`; int keys → `{type: array,
  items}`. `isStringKey()` accepts `string` / `array-key` / `non-empty-string` (`:232-238`). Covered
  by `tests/Unit/Support/Types/TypeNodeToSchemaTest.php:111-116`.
- **Eloquent model `@property` + JSON casts** — `docs/auto-derivation.md:151-152` documents
  `array<string, T>` → `additionalProperties` for model columns.
- FormRequest / inline-`validate()` wildcard rules emit `OA\AdditionalProperties` too
  (`SchemaFromFormRequest.php:198`, `InlineValidationRequestSchemaResolver.php:162`).

The **one genuine gap** is the **SpatieData typed-property path**:
`src/Plugins/SpatieData/Support/SchemaFromDataClass.php::resolveCollectionSchema()` (`:530-568`)
reads `getCollectionValueType()` but **never reads `getCollectionKeyType()`** — so `array<string,X>`
and `array<int,X>` both collapse to `{type: array, items}`. That single method is the locus.

**ApiResources is out of scope** (and the issue's "check the ApiResources field path" resolves to:
not applicable). That plugin is attribute-driven (`#[ResourceField]`); it has **no** symfony
type-info / `CollectionType` read at all (`SchemaFromResource::buildProperty()` `:270-299` derives
`items` from the attribute's declared `items` class-string, not from a reflected map type). There is
nothing to change there for type-derived maps.

## Shared-helper decision (the issue asks)

The issue asks whether a shared `Support/Extraction` helper is the right home so "both reuse it."
Answer: **no — keep it surgical in `resolveCollectionSchema()`.** Rationale (CLAUDE.md simplicity-
first, no speculative abstraction):

- There is no second consumer. ApiResources doesn't read types; the PHPDoc path
  (`TypeNodeToSchema`) already has its own correct map branch and operates on a different input
  (`GenericTypeNode`, not `CollectionType`) — extracting a shared helper would have to unify two
  different type representations for zero present benefit.
- The change is ~one branch inside one method. A new `Support/` class for it would be the
  over-abstraction the guidelines warn against.

So: add a key-type check at the top of `resolveCollectionSchema()`. If the key is **not** a string
key, fall through to the existing list/array logic unchanged.

### 1. Tests first (`tests/Feature/Plugins/SpatieData/SchemaFromDataClassTest.php` + a fixture)

Add a fixture Data class (e.g. `tests/Fixtures/MapPropertiesFixtureData.php`) with properties
exercising each branch, and assert the generated component schema:

- `array<string, Money>` (Money = a Data class) → `{type: object, additionalProperties: {$ref: Money}}`.
- `array<string, string>` → `{type: object, additionalProperties: {type: string}}`.
- `array<string, mixed>` / untyped value → `{type: object, additionalProperties: true}` (the
  permissive-map case — value type opaque but the key proves it's a map).
- **Negatives (must stay arrays):** `list<Money>` and `array<int, Money>` → `{type: array, items:
  {$ref}}` (int key / list is not a map); a bare untyped `array` property → stays the existing
  opaque-array shape (key is `array-key` = int|string union, **not** a pure string key → not a map).
- **Nested map:** `array<string, array<string, Money>>` → outer `additionalProperties` whose value
  is itself `{type: object, additionalProperties: {$ref: Money}}`.

Edge to decide in code (note in PR): a bare `array` with no `@var` resolves its key to `array-key`
(int|string), and `non-empty-string` keys exist too. **Map only when the key is a pure string
identifier** (`string` / `non-empty-string`) — mirror `TypeNodeToSchema::isStringKey()` semantics
but **exclude `array-key`** here, because an untyped `array`'s `array-key` is the "I don't know"
case, not an asserted map. (PHPDoc's `isStringKey` includes `array-key` because there the generic
was *written* with two args; in the reflected-type path a bare `array` also surfaces `array-key`, so
including it would wrongly turn every untyped array into a map. Verify the exact key Type that
type-info produces for `array<string,X>` vs bare `array` during implementation and lock it with the
negative test.)

### 2. Implementation — `SchemaFromDataClass::resolveCollectionSchema()`

At the method top, read `$type->getCollectionKeyType()`. When it is a string key (per the rule
above) **and** the collection is not a list (`!$type->isList()`):

- value is a `Data` class → `{type: object, additionalProperties: {$ref}}`
- value is a concrete scalar `BuiltinType` (not mixed/null) → `{type: object, additionalProperties:
  <schemaFromType->fromType(value)>}`
- value is mixed/opaque → `{type: object, additionalProperties: true}` (bool `true`, which
  swagger-php's `OA\Schema::$additionalProperties` accepts — it is typed `bool|AdditionalProperties`,
  confirmed in `vendor/zircote/swagger-php/.../JsonSchemaTrait.php:126`).

Otherwise fall through to the **unchanged** existing array/list branches. Reuse the existing value
resolution (Data → `build()` + `$ref`; scalar → `schemaFromType->fromType()`) so the value-schema
logic isn't duplicated between the array and map branches — factor the value-schema resolution into
a small private helper *within this class* if it reads cleaner, but no cross-class extraction.

swagger-php note to verify on the 5.8 CI path: a nested `OA\Schema` as `additionalProperties`
serializes/validates on 5.8 (the envelope strategies already emit `OA\AdditionalProperties`, and the
PHPDoc path emits a nested `OA\Schema` — so the shape is proven). `additionalProperties: true`
(bool) likewise. Confirm both serialize byte-stably for the snapshot group.

### 3. Bookkeeping

- `docs/auto-derivation.md` — the Spatie Data response-body / property mapping should state that a
  string-keyed `array<string, T>` property becomes `additionalProperties` (the model/PHPDoc paths
  are already documented at `:151-152`; add the Data-class property case so all three type-reading
  paths read consistently).
- `CHANGELOG.md [Unreleased]` — under **Added**: string-keyed array/collection Data properties now
  emit `additionalProperties` (map inference) instead of a bare array.

## Files

- *Code:* `src/Plugins/SpatieData/Support/SchemaFromDataClass.php` (`resolveCollectionSchema`, ~one
  branch + optional private value-schema helper).
- *Tests:* `tests/Feature/Plugins/SpatieData/SchemaFromDataClassTest.php` (+ new fixture
  `tests/Fixtures/MapPropertiesFixtureData.php`, possibly a `Money`-like nested Data fixture if none
  reusable exists).
- *Docs/bookkeeping:* `docs/auto-derivation.md`, `CHANGELOG.md`.

No changes to `src/Contracts/**`, no config key, no stage-order change, no public-signature change.

## Verification

- `composer test` green (new map cases + negatives).
- `vendor/bin/pint --test` clean; `composer lint` (PHPStan L8) clean.
- Snapshot/byte-exact group still green on **both** swagger-php 5.8 and 6.x (the nested-schema
  `additionalProperties` shape is the 5.x-sensitive bit — see the project's swagger-php 5.x/6.x
  validation note).

## Controversial / escalation flags

Low-risk, surgical. The only judgement call is the **`array-key` exclusion** (a bare untyped `array`
must NOT become a map) — locked by a negative test, flagged here so review confirms the intended
semantics. No public surface touched.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -25,6 +25,7 @@ All notable changes to this project are documented here.
 - `SpecStage`, `ErrorResponseContributor`, and `ResourceTargetLocator` plugin extension points; observability events for generation and linting.
 - Default route filters for first-party operational packages (Horizon, Pulse, Nova, Telescope, Ignition, Passport, and the library's own routes via `SkipSelfRoutes`).
 - `openapi:diff:config` command reporting drift between the published config and the package default.
+- String-keyed array/collection properties on Spatie `Data` classes (`array<string, T>`) emit `additionalProperties` (a map) instead of a bare array. (#334)
 
 ### Changed
 - `spatie/laravel-data` is now a soft dependency (moved from `require` to `require-dev`); Fractal and query-builder packages are opt-in.
--- a/docs/request-bodies.md
+++ b/docs/request-bodies.md
@@ -18,6 +18,7 @@ validation.
 | Request-side support | Yes, always on | Yes, via the SpatieData plugin |
 | Response-side support | No. Return a typed resource (`JsonResource`) or use `#[ResponseResource]`. | Yes. Return a `Data`, `DataCollection<…>`, or paginated variant. |
 | Nested objects | Yes, via dotted/wildcard rule keys (`address.city`, `items.*.name`) → inline nested object/array schemas | Yes. Nested `Data` classes become nested component schemas via `$ref`. |
+| Maps / dictionaries | Inline rules describe lists, not string-keyed maps | A string-keyed property (`@var array<string, T>`) → `{type: object, additionalProperties: …}`; an int-keyed/`list<T>`/bare `array` stays a plain array |
 | Enums | `Rule::in([...])` → inline `enum`; `Rule::enum(Status::class)` (full case set) → `$ref` to a [shared enum component](auto-derivation.md#shared-enum-components) | Native `BackedEnum` property → `$ref` to the shared enum component; `Rule::in([...])` still works |
 | Field-level enrichment | `#[RequestField]` on a `PARAM_*` class-constant whose value matches the field name | `#[RequestField]` on the promoted constructor parameter or property |
 | Transformations / computed properties | No. The generator reads signatures only — method bodies are not read. | The `Data` class can carry `Optional`-typed and computed properties. PATCH semantics are inferred from `Optional\|…\|null`. |
--- a/src/Plugins/SpatieData/Support/SchemaFromDataClass.php
+++ b/src/Plugins/SpatieData/Support/SchemaFromDataClass.php
@@ -158,7 +158,10 @@ final class SchemaFromDataClass implements FilePropertyChecker
             }
 
             try {
-                $type = $this->typeResolver->resolve($rawType, $this->typeContextFor($context->reflection));
+                // Resolve the property itself (not its ReflectionNamedType) so the docblock-aware
+                // resolver reads `@var` generics — `array<string, T>` key/value types collapse to a
+                // bare `array` when only the native type is resolved.
+                $type = $this->typeResolver->resolve($context->reflection, $this->typeContextFor($context->reflection));
             } catch (UnsupportedException) {
                 $properties[] = new OA\Property([
                     'property' => $context->wireName,
@@ -531,6 +534,16 @@ final class SchemaFromDataClass implements FilePropertyChecker
     {
         $valueType = $type->getCollectionValueType();
 
+        // A pure string key denotes a map (`array<string, T>`) → `additionalProperties`. A bare
+        // `array` and `array<array-key, T>` both surface an int|string union key, which is the
+        // "unknown" case, not an asserted map — those fall through to the list/array branches.
+        if (!$type->isList() && $this->isStringKey($type->getCollectionKeyType())) {
+            return new OA\Schema([
+                'type' => 'object',
+                'additionalProperties' => $this->mapValueSchema($valueType, $propertyName),
+            ]);
+        }
+
         if (
             $valueType instanceof ObjectType
             && is_a($valueType->getClassName(), Data::class, allow_string: true)
@@ -565,6 +578,41 @@ final class SchemaFromDataClass implements FilePropertyChecker
                 $propertyName,
             ),
         ]);
+    }
+
+    /**
+     * Schema for a map's value. Reuses the full property-schema resolution (Data → `$ref`, scalar,
+     * nullable/union → `oneOf`, nested collection), so a map value gets the same fidelity as any
+     * other property. An opaque `mixed` value has no useful schema, so it degrades to a permissive
+     * `additionalProperties: true` rather than an empty schema.
+     *
+     * @return OA\Schema|true permissive `additionalProperties` for an opaque `mixed` value
+     *
+     * @throws ReflectionException
+     * @throws RuntimeException
+     * @throws UnsupportedException
+     */
+    private function mapValueSchema(Type $valueType, string $propertyName): OA\Schema|bool
+    {
+        if (
+            $valueType instanceof BuiltinType
+            && $valueType->isIdentifiedBy(TypeIdentifier::MIXED)
+        ) {
+            return true;
+        }
+
+        return $this->resolvePropertySchema($valueType, $propertyName);
+    }
+
+    /**
+     * Whether the collection key is a pure `string` identifier (a map key). The reflected-type path
+     * surfaces a bare `array` / `array<array-key, T>` as an int|string {@see UnionType}, which is
+     * deliberately not a map; `non-empty-string` normalizes to a `string` {@see BuiltinType}.
+     */
+    private function isStringKey(Type $keyType): bool
+    {
+        return $keyType instanceof BuiltinType
+            && $keyType->isIdentifiedBy(TypeIdentifier::STRING);
     }
 
     /**
--- a/src/Support/Extraction/FieldDescriptor.php
+++ b/src/Support/Extraction/FieldDescriptor.php
@@ -266,8 +266,12 @@ final class FieldDescriptor
             }
         }
 
-        // Nested array element (from wildcard validation keys).
-        if ($this->items !== null) {
+        // Nested array element (from wildcard validation keys). Skip when the target is already a
+        // map (`type: object` with `additionalProperties`): a string-keyed array resolves to a map,
+        // and a rule pass that read it as a plain `array` must not graft a stray `items` onto it.
+        $targetIsMap = $target->type === 'object' && is_defined($target->additionalProperties);
+
+        if ($this->items !== null && !$targetIsMap) {
             if ($overwrite || is_undefined($target->items)) {
                 $childItems = new OA\Items([]);
                 $this->items->applyTo($childItems);
--- a/tests/Feature/Plugins/SpatieData/SchemaFromDataClassTest.php
+++ b/tests/Feature/Plugins/SpatieData/SchemaFromDataClassTest.php
@@ -12,6 +12,7 @@ use Radiergummi\OpenApi\Tests\Fixtures\Alpha\SelfRefData as AlphaSelfRefData;
 use Radiergummi\OpenApi\Tests\Fixtures\Beta\SelfRefData as BetaSelfRefData;
 use Radiergummi\OpenApi\Tests\Fixtures\ConditionalResponseFieldFixtureData;
 use Radiergummi\OpenApi\Tests\Fixtures\MapInputNameFixtureData;
+use Radiergummi\OpenApi\Tests\Fixtures\MapPropertiesFixtureData;
 use Radiergummi\OpenApi\Tests\Fixtures\PropertyFixtureData;
 use Radiergummi\OpenApi\Tests\Fixtures\SchemaTitleDescriptionFixtureData;
 
@@ -202,6 +203,14 @@ class ConditionalResponseFieldController extends Controller
     }
 }
 
+class MapPropertiesController extends Controller
+{
+    public function store(MapPropertiesFixtureData $data): JsonResponse
+    {
+        return new JsonResponse();
+    }
+}
+
 it('keeps a conditional field in properties but removes it from required[]', function (): void {
     Route::get('/spatie-data/conditional', [ConditionalResponseFieldController::class, 'show']);
 
@@ -222,6 +231,94 @@ it('does not remove non-conditional fields from required[]', function (): void {
 
     expect($required)->toContain('id')
         ->and($required)->toContain('alwaysRequired');
+});
+
+// endregion
+
+// region #334: string-keyed array properties → additionalProperties (map inference)
+
+it('emits additionalProperties with a $ref for a string-keyed Data-class map', function (): void {
+    Route::post('/spatie-data/maps', [MapPropertiesController::class, 'store']);
+
+    $props = generateSpec()['components']['schemas']['MapPropertiesFixtureData']['properties'];
+
+    expect($props['addressMap']['type'])->toBe('object')
+        ->and($props['addressMap'])->not->toHaveKey('items')
+        ->and($props['addressMap']['additionalProperties']['$ref'])
+        ->toBe('#/components/schemas/AddressFixtureData');
+});
+
+it('emits additionalProperties with a scalar value schema for a string-keyed scalar map', function (): void {
+    Route::post('/spatie-data/maps', [MapPropertiesController::class, 'store']);
+
+    $props = generateSpec()['components']['schemas']['MapPropertiesFixtureData']['properties'];
+
+    expect($props['scalarMap']['type'])->toBe('object')
+        ->and($props['scalarMap']['additionalProperties']['type'])->toBe('string');
+});
+
+it('emits permissive additionalProperties for a string-keyed map with an opaque value', function (): void {
+    Route::post('/spatie-data/maps', [MapPropertiesController::class, 'store']);
+
+    $props = generateSpec()['components']['schemas']['MapPropertiesFixtureData']['properties'];
+
+    expect($props['opaqueMap']['type'])->toBe('object')
+        ->and($props['opaqueMap']['additionalProperties'])->toBeTrue();
+});
+
+it('emits a nested map for array<string, array<string, Data>>', function (): void {
+    Route::post('/spatie-data/maps', [MapPropertiesController::class, 'store']);
+
+    $props = generateSpec()['components']['schemas']['MapPropertiesFixtureData']['properties'];
+
+    $outer = $props['nestedMap'];
+    $inner = $outer['additionalProperties'];
+
+    expect($outer['type'])->toBe('object')
+        ->and($inner['type'])->toBe('object')
+        ->and($inner['additionalProperties']['$ref'])
+        ->toBe('#/components/schemas/AddressFixtureData');
+});
+
+it('preserves a nullable map value as oneOf [$ref, null] (full fidelity, not degraded to true)', function (): void {
+    Route::post('/spatie-data/maps', [MapPropertiesController::class, 'store']);
+
+    $props = generateSpec()['components']['schemas']['MapPropertiesFixtureData']['properties'];
+
+    $value = $props['nullableValueMap']['additionalProperties'];
+    $refs  = array_column($value['oneOf'], '$ref');
+    $types = array_column($value['oneOf'], 'type');
+
+    expect($props['nullableValueMap']['type'])->toBe('object')
+        ->and($refs)->toContain('#/components/schemas/AddressFixtureData')
+        ->and($types)->toContain('null');
+});
+
+it('preserves a union map value as oneOf of the member schemas', function (): void {
+    Route::post('/spatie-data/maps', [MapPropertiesController::class, 'store']);
+
+    $props = generateSpec()['components']['schemas']['MapPropertiesFixtureData']['properties'];
+
+    $types = array_column($props['unionValueMap']['additionalProperties']['oneOf'], 'type');
+
+    expect($props['unionValueMap']['type'])->toBe('object')
+        ->and($types)->toContain('integer')
+        ->and($types)->toContain('string');
+});
+
+it('keeps int-keyed, list and bare-array properties as plain arrays (not maps)', function (): void {
+    Route::post('/spatie-data/maps', [MapPropertiesController::class, 'store']);
+
+    $props = generateSpec()['components']['schemas']['MapPropertiesFixtureData']['properties'];
+
+    expect($props['addressList']['type'])->toBe('array')
+        ->and($props['addressList'])->not->toHaveKey('additionalProperties')
+        ->and($props['addressList']['items']['$ref'])->toBe('#/components/schemas/AddressFixtureData')
+        ->and($props['indexedList']['type'])->toBe('array')
+        ->and($props['indexedList'])->not->toHaveKey('additionalProperties')
+        ->and($props['indexedList']['items']['$ref'])->toBe('#/components/schemas/AddressFixtureData')
+        ->and($props['bareArray']['type'])->toBe('array')
+        ->and($props['bareArray'])->not->toHaveKey('additionalProperties');
 });
 
 // endregion
--- a/tests/Fixtures/MapPropertiesFixtureData.php
+++ b/tests/Fixtures/MapPropertiesFixtureData.php
@@ -1,0 +1,41 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Tests\Fixtures;
+
+use Spatie\LaravelData\Data;
+
+/**
+ * Exercises map (string-keyed array) inference on the SpatieData typed-property path.
+ * String keys become `additionalProperties`; int keys, lists and bare arrays stay plain arrays.
+ */
+final class MapPropertiesFixtureData extends Data
+{
+    public function __construct(
+        /** @var array<string, AddressFixtureData> */
+        public array $addressMap,
+
+        /** @var array<string, string> */
+        public array $scalarMap,
+
+        /** @var array<string, mixed> */
+        public array $opaqueMap,
+
+        /** @var array<string, array<string, AddressFixtureData>> */
+        public array $nestedMap,
+
+        /** @var array<string, ?AddressFixtureData> */
+        public array $nullableValueMap,
+
+        /** @var array<string, int|string> */
+        public array $unionValueMap,
+
+        /** @var list<AddressFixtureData> */
+        public array $addressList,
+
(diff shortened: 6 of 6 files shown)


Radiergummi:

Scope note for review: the issue says `additionalProperties` is emitted "nowhere," but map inference already works on the PHPDoc-generic path (`TypeNodeToSchema::fromGeneric`) and the Eloquent model/cast path. The **one real gap** is the SpatieData typed-property path — `SchemaFromDataClass::resolveCollectionSchema()` reads the collection *value* type but never the *key* type, so `array<string,X>` collapses to a plain array. Fix is one branch in that method.

Two decisions recorded in the plan: (1) **no shared helper** — ApiResources is attribute-driven with no type-info read, so there is no second consumer to justify extraction (kept surgical per simplicity-first); (2) a bare untyped `array` (key = `array-key`, int|string) must **not** become a map — only a pure string key does. That `array-key` exclusion is the one semantic judgement call; locked by a negative test.

Radiergummi:
🤖 **[lead]** plan posted (PR #338) → plan-review. Planner scope finding: map inference already works on PHPDoc-generic + Eloquent-cast paths; the single real gap is SpatieData's SchemaFromDataClass::resolveCollectionSchema() never reading the collection key type, so array<string,X> collapses to a plain array. One-branch fix. Two decisions to validate: (1) no shared helper (ApiResources has no type-info read → no 2nd consumer); (2) bare untyped array (array-key) must NOT become a map — only a pure string key, locked by a negative test.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved** — sound, surgical, Tier-0. No plan changes; one coding-time refinement for the key-detection mechanism (the plan already flagged it as verify-during-implementation, so this just settles it).

Verified the load-bearing claims against the code, and **empirically probed type-info** to ground the `array-key` decision:

- **Gap locus confirmed.** `SchemaFromDataClass::resolveCollectionSchema()` (`:530-568`) reads only `getCollectionValueType()`; it never reads the key, so `array<string,X>` and `array<int,X>` both collapse to `{type: array, items}`. Single-method fix. Correct.
- **Other paths already work.** `TypeNodeToSchema::fromGeneric()` (`:205-216`) emits the map shape for PHPDoc generics. Confirmed.
- **No shared helper — agreed.** The two paths do not share a representation: PHPDoc operates on `GenericTypeNode` *string keywords*; this path operates on `CollectionType` *`Type` objects*. They do not even encode the "unknown key" case the same way (see below). Extracting a shared helper would unify two different type models for zero present benefit — speculative abstraction. Keep it in `resolveCollectionSchema()`.
- **ApiResources out of scope — agreed.** Its only type read is `ResourceClassLocator::isCollectionType()` (a `ReflectionType` collection-vs-not check for locating the resource class), not a `CollectionType` field-map read. Nothing to change there.

**Coding-time refinement (settles the plan's open `array-key` question).** I ran the actual `StringTypeResolver` the library uses:

```
array<string, int>           => key=string        isList=false
array<non-empty-string, int> => key=string        isList=false   (normalized to string)
array<int, int>              => key=int            isList=false
list<int>                    => key=int            isList=true
array                        => key=int|string (UNION)  isList=false
array<array-key, int>        => key=int|string (UNION)  isList=false
```

So in **this** (reflected `CollectionType`) path the map test must be: **key is a pure `string` BuiltinType** (e.g. `isIdentifiedBy(TypeIdentifier::STRING)` and not a union). It must NOT mirror `TypeNodeToSchema::isStringKey()` literally — that matches the keyword string `"array-key"`, but here a bare `array` and `array<array-key,X>` both surface an `int|string` **UnionType**, never an `array-key` token. The plan's *intent* (only a pure string key is a map; bare array stays an array) is exactly right and the negative test locks it; only the "exclude the `array-key` keyword" mechanism description is imprecise for this code path. `non-empty-string` normalizes to `string`, so it is correctly a map (matches the issue).

**Test coverage adequate:** positive `$ref` map + scalar map + permissive `additionalProperties: true`; negatives list / int-key / bare-array; nested map. Reusable nested Data fixtures exist (`AddressFixtureData`, `NestedParentData`) for the `$ref` case — a new `Money` fixture may be unnecessary.

No contract/config/stage/public-surface change. Snapshot-on-5.8 verification correctly called out. Clear to proceed to coding.

Radiergummi:
🤖 **[lead]** plan-review approved → coding. Coding-time refinement from reviewer-1 (settles the plan's verify-during-impl flag): the key test must assert a **pure `string` BuiltinType**, NOT mirror `TypeNodeToSchema::isStringKey()` literally — in the reflected CollectionType path a bare array surfaces an `int|string` UnionType (never an `array-key` token), so copying isStringKey() would be wrong. Plan intent + negative test correct. Snapshot-on-5.8 called out.

Radiergummi:
🤖 **[coder]** deviation: the plan assumed `resolveCollectionSchema` already read the collection **value** generic and only the **key** was missing. Empirically, neither was reaching the method — the call site (`buildPropertyContexts` loop) resolved `$context->reflection->getType()` (a `ReflectionNamedType`), which drops all PHPDoc generics; only resolving the `ReflectionProperty` itself runs the docblock-aware resolver. So `array<string|int, T>` collapsed to a bare `array` regardless of key.

Two necessary changes beyond the planned one branch:

1. **Call site** (`SchemaFromDataClass` :161): resolve `$context->reflection` (the property) instead of its `getType()`, so `@var` generics surface. This also fixes value generics (`array<int, ScalarType>` now yields typed `items` where it previously went opaque) — a latent improvement consistent with the issue intent. Full suite (2554) stays green.
2. **Rules-merge guard** (`FieldDescriptor::applyTo` :270): Spatie infers a plain `array` rule for the map property; the descriptor carried an `items` child that `applyTo(overwrite: false)` grafted onto the resolved `type: object` map (stray `items: {}`). Guarded the items-merge to skip when the target is already a map (`type: object` + `additionalProperties`). Plugin-agnostic, lives in `Support/`.

Key detection is exactly the reviewer-confirmed rule: pure `string` `BuiltinType` (`isIdentifiedBy(STRING)`), `!isList()`. `non-empty-string` normalizes to `string` (map); bare `array` / `array<array-key, T>` are int|string unions (not maps) — locked by the negative test. Verified the full suite + the `additionalProperties: true`/`$ref` shapes on **swagger-php 5.8** (snapshot excluded, matching the 5.x CI job) and 6.1.2.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed c620f8b. Remote CI fully green (all matrix incl. swagger-php 5.8, Snapshot, quality, PHPStan). Ready for review.

Radiergummi:
🤖 **[lead]** coding done (sha c620f8b), remote CI green incl. swagger-php 5.8 → review. Needs TWO independent reviewer approvals. Note two coder deviations to scrutinize: (1) call-site fix at SchemaFromDataClass:161 (resolve ReflectionProperty so @var generics surface — broadened from 'key only' to also fix value generics); (2) a guard in src/Support/ FieldDescriptor::applyTo:270 (skip grafting stray items onto a resolved map). Both have a 'deviation' PR comment with rationale. Generation-affecting → survey Layer A required before merge.

Radiergummi:
🤖 **[reviewer-2]** review round 1 — 1 finding (non-blocking; raising for a decision):

Independent review against PR ref `c620f8b` (probed the actual `StringTypeResolver` rather than trusting the summaries). The core change and **both deviations are sound** — see verification below — but I found one consumer-visible fidelity gap in the new map-value resolver.

**1. `SchemaFromDataClass::mapValueSchema()` silently degrades nullable / union map values to `additionalProperties: true`.**
`mapValueSchema` (`src/Plugins/SpatieData/Support/SchemaFromDataClass.php`) is a narrower re-implementation of the sibling `resolvePropertySchema` (:497): it only routes `CollectionType`, a `Data` `ObjectType`, and a concrete `BuiltinType`, falling back to `return true` for everything else. Probed:
- `array<string, ?Foo>` → value is a `NullableType` → falls to `true` (loses `Foo`). `resolvePropertySchema` would `NullableSchema::wrap` it.
- `array<string, A|B>` → value is a `UnionType` → falls to `true` (loses the `oneOf`). `resolvePropertySchema` would emit `oneOf`.

Not a correctness bug (`true` is valid, conservative `additionalProperties`), but it is a fidelity gap on the public map surface with **no test**, and it duplicates resolution logic that already exists one method up. Recommend either routing the value through `resolvePropertySchema($valueType, $propertyName)` and reserving the `true` fallback for genuinely opaque values (`mixed`/no value type) — simpler *and* higher fidelity — or, if the narrow handling is intentional, a one-line comment + a test locking `array<string, ?Foo>` → `additionalProperties: true` so the degradation is deliberate, not incidental. Suggested test inputs: a `@var array<string, ?AddressFixtureData>` property asserts `additionalProperties` carries the `$ref` (or, if kept as-is, `=== true`).

---

**Verification of the load-bearing claims (all confirmed):**

- **Deviation 1 — call-site :161 (`resolve($context->reflection)` not `getType()`).** Correct: Symfony `TypeResolver::resolve()` is docblock-aware only when handed the `ReflectionProperty`; the `ReflectionNamedType` path drops every `@var` generic. The `$rawType === null` guard above is untouched, so untyped properties still degrade. The widening is real (all `@var` generics now surface, incl. value generics) but **safe**: the byte-exact Snapshot CI job passed and **no snapshot fixture changed** in this PR — i.e. no existing shape’s output shifted. The value-generic side-effect is covered (`indexedList: array<int, AddressFixtureData>` → `items.$ref`).
- **Deviation 2 — `FieldDescriptor::applyTo` :270 guard.** Correct layer and precise. `$targetIsMap = type === "object" && is_defined(additionalProperties)` trips *only* for a resolved map; plain-array (`type: array`) and composed-object (no `additionalProperties`) targets are unaffected, so non-map paths keep grafting `items`. `is_defined`/`is_undefined` are already imported and used throughout the method. The graft-stray-items bug is a general type-pass-then-rule-pass merge concern, not Spatie-specific → `Support/` is right.
- **Key detection.** Probed: `array<string,T>`, `non-empty-string`, `lowercase-string`, `class-string` → pure `string` BuiltinType → map; `array<int,T>` / `list<T>` → int key → array; bare `array` / `array<array-key,T>` → `int|string` UnionType → **not** a map (a `UnionType` is not a `BuiltinType`, so `isStringKey` correctly returns false). Negatives test-locked. The `!$type->isList()` pre-check is redundant for a string key but harmless/defensive.
- **Bookkeeping/docs.** `CHANGELOG.md [Unreleased] / Added` present; `docs/request-bodies.md` comparison table row added and accurate (string-key → map, int/list/bare → array). No new public method/attribute/config/lint rule, so no further docs surface. Index unaffected.
- **CI:** full matrix + swagger-php 5.8 + Snapshot + quality (PHPStan L8) + Pint + codecov patch/project all green.

Conventions clean (strict-types, PHP 8.4, no abbreviations, region markers, concise comments). Tier-0, surgical.

Verdict: **approve once finding #1 is addressed or consciously waived with a locking test.** Everything else is merge-ready.

Radiergummi:
🤖 **[lead]** review round 1 → back to coding. reviewer-2 confirmed both deviations sound, but raised 1 finding (approve-once-addressed): mapValueSchema() degrades nullable (array<string,?Foo>) and union (array<string,A|B>) map values to additionalProperties:true, losing the type, untested. Not a correctness bug but a fidelity gap + duplication smell. Coder to either route the value through the sibling resolvePropertySchema (simpler + higher fidelity, preferred if clean) OR keep narrow with a comment + a test locking the degradation as deliberate.

Radiergummi:
🤖 **[coder]** addressed round 1: took the preferred path — `mapValueSchema()` now routes the map value through the existing `resolvePropertySchema()` (removing the scalar/Data re-implementation duplication), so a map value gets full property fidelity: `array<string, ?Foo>` → `additionalProperties: {oneOf: [$ref, {type: null}]}` and `array<string, A|B>` → `{oneOf: […]}` instead of degrading to `true`. The single deliberate degradation that remains is an opaque `mixed` value (`array<string, mixed>`) → `additionalProperties: true`, since `mixed` has no useful schema — commented as intentional and locked by the existing `opaqueMap` test. Added two tests (nullable-value map, union-value map). Verified the new `oneOf`-under-`additionalProperties` shapes on swagger-php 5.8 (snapshot excluded) and 6.1.2.

Rebased onto current origin/main (clean; #340 absorbed) and squashed to one commit. composer check + remote CI fully green. Pushed 36f7123.

Radiergummi:
🤖 **[lead]** round 1 addressed (sha 36f7123, rebased onto current origin/main, #340 absorbed) → back to review. Coder took the preferred path: mapValueSchema now delegates to resolvePropertySchema — nullable→oneOf[$ref,null], union→oneOf[...], only opaque mixed→additionalProperties:true (commented intentional + opaqueMap test). 2 new tests. Needs reviewer-2 re-review (finding closed?) + reviewer-1 independent 1st pass = two approvals, then survey.

Radiergummi:
🤖 **[reviewer-2]** review round 2: approved, no concerns.

Verified round-1 finding #1 is genuinely closed at `36f7123`:

- `mapValueSchema()` now returns `true` **only** for a `mixed` value (`isIdentifiedBy(MIXED)`), commented as the intentional opaque-value degradation; everything else delegates to `resolvePropertySchema()`. So a map value gets the same fidelity as any other property — nullable wraps, union → `oneOf`, nested collection recurses, Data → `$ref`, scalar → its schema. No bespoke re-implementation remains.
- The two requested value tests are present and assert the upgraded behaviour: `nullableValueMap` (`array<string, ?AddressFixtureData>`) → `additionalProperties.oneOf` = `[$ref, null]`; `unionValueMap` (`array<string, int\|string>`) → `additionalProperties.oneOf` of member schemas. The `mixed` degradation stays locked by `opaqueMap` → `additionalProperties === true`. 7 map tests total.
- **No unwanted behaviour pulled in.** The delegation lives entirely inside the new string-key branch of `resolveCollectionSchema()`; the existing list / int-key / bare-array branches are untouched, so no existing shape changed — confirmed by **zero snapshot files in the diff** and the green Snapshot CI job.
- Fixture `MapPropertiesFixtureData` carries the two new annotated properties; key detection and the deviations remain as verified in round 1.

CI green at `36f7123`: full matrix + swagger-php 5.8 + Snapshot + quality (PHPStan L8) + Pint + codecov.

Approved — one of the two required approvals.

Radiergummi:
🤖 **[reviewer]** review round 2: approved, no concerns. This is the second of two approvals.

Independent **code** review of `36f7123` (I did the plan-review; verified the diff, the round-1 fidelity fix, both deviations, and ran the suite + snapshot group myself — did not defer to reviewer-2).

**Round-1 fidelity finding is genuinely closed.** `mapValueSchema()` returns `true` only for a `mixed` value (`isIdentifiedBy(TypeIdentifier::MIXED)`), commented as the intentional opaque degradation; everything else delegates to `resolvePropertySchema()`, which returns `OA\Schema` **unconditionally** (`:497-521`) — so the `OA\Schema|bool` union is sound and a map value now gets identical fidelity to any other property: nullable → `NullableSchema::wrap`, union → `oneOf`, collection → recurse (nested maps), object → `$ref`, scalar → `fromType`. No bespoke re-implementation remains. (Aside: a bare `null` value type would now emit `{type: null}` rather than `true` — strictly *more* faithful, no regression.)

**Tests — ran them, non-vacuous: 18 passed (59 assertions)** in `SchemaFromDataClassTest`, all 7 map cases included. The fidelity pair is real: `nullableValueMap` → `additionalProperties.oneOf` contains `$ref` + `null`; `unionValueMap` → `oneOf` of `integer`+`string`; `opaqueMap` → `additionalProperties === true`; `nestedMap` recurses; the three negatives (`addressList` list / `indexedList` int-key / `bareArray`) stay `type: array` with **no** `additionalProperties`.

**Deviation 1 (call-site `:161`, resolve the `ReflectionProperty` not its named type) — verified SAFE myself.** This is the highest-risk change (it broadens `@var`-generic resolution across the whole class). I ran the **snapshot group → 9 passed (18 assertions)** and confirmed **zero snapshot files in the diff** (`gh pr diff --name-only`) — so no existing byte-exact shape shifted. The value-generic side-effect is itself locked (`indexedList: array<int, AddressFixtureData>` → `items.$ref`). The `$rawType === null` untyped-degrade guard above is untouched.

**Deviation 2 (`FieldDescriptor::applyTo:270` map guard) — verified precise.** `is_defined`/`is_undefined` are real `src/functions.php` helpers. `$targetIsMap = type === "object" && is_defined(additionalProperties)` trips only for a resolved map; I ran the `items`-grafting consumers (`FormRequestSchemaTest` + `InlineValidationSchemaTest`) → **18 passed (74 assertions)**, so plain-array and composed-object targets still graft `items` correctly. Correctly in `Support/` — it is a general type-pass-then-rule-pass merge concern, not Spatie-specific.

**Key detection + negatives test-locked.** `isStringKey()` requires a `string` `BuiltinType`; a bare `array` / `array<array-key, T>` surfaces an `int|string` `UnionType` (not a `BuiltinType`) so it correctly returns false. `non-empty-string` normalizes to `string` → map. Matches the empirical type-info behaviour I probed at plan time.

**Conventions / boundaries / bookkeeping** — strict-types, PHP 8.4, no abbreviations, `// region` markers, concise concrete comments; no `Contracts`/config/stage-order touched; `CHANGELOG.md [Unreleased] / Added` present; `docs/request-bodies.md` comparison-table row added and accurate. Tier-0, surgical.

One trivial non-finding: `mapValueSchema()`'s docblock `@return OA\Schema|true` vs the `OA\Schema|bool` signature — harmless, PHPStan L8 green; not worth a change.

No findings. **Approved** — #338 now has both approvals.

Radiergummi:
🤖 **[lead]** two independent approvals in (reviewer-2 rounds 1+2, reviewer-1 ran suite 18/59 + snapshot group 9 + grafting consumers 18/74; deviations verified safe) → survey. Generation-affecting (area:plugins map inference). After survey: rebase onto current origin/main (now incl. #337) + CI re-confirm before merge.

Radiergummi:
🤖 **[surveyor]** Layer A pass — 0 regressions across 11 apps (1 app improved).

Generation-affecting PR (`area:plugins`), full corpus run. Fresh baseline regenerated on current `main` (`57a2fbf`, post-#337-merge); candidate installed PR HEAD `36f7123` into all 11 apps (manifest-verified).

**Deterministic metrics — no regression on any app; one improvement:**

| app | paths | operations | requestBodies | componentSchemas (base → cand) | gen_exit |
|-----|-------|------------|---------------|--------------------------------|----------|
| Lychee | 213 | 273 | 92 | **273 → 280 (+7)** | 0 |

paths/operations/requestBodies identical everywhere; the only metric delta is Lychee componentSchemas **+7 (an increase)**, and no metric dropped on any app.

**Content-level spec diff** (normalized JSON, baseline vs candidate):
- **7 apps byte-identical** (AdvisingApp, Bagisto, BookStack, Coolify, Koel, Pelican, Vito).
- **AureusERP, InvoiceNinja, SpeedtestTracker** differ **only** in non-deterministic identifiers (Livewire per-install asset-route hash; random `api.generated_`/`client.generated_` operationIds) — zero feature-related lines.
- **Lychee** carries the feature's footprint, and it is **purely additive**: a string-keyed map property (`EmbedPhotoResource.exif`) now resolves to `{type: object, additionalProperties: {type: [string, null]}}` instead of collapsing to a bare `array`, and reading the `@var` generic value type pulled 7 previously-unreferenced Data classes (`EmbedPhotoResource`, `ThumbResource`, `ThumbAlbumResource`, `OrderItemResource`, `ImportDirectoryResource`, `UserMemberGroupResource`, `DayCount`) into the component pool as `$ref`s.

**Regression guard — no schema lost detail:** I diffed every baseline component schema's property set against the candidate — **0 schemas removed, 0 properties dropped**. The `FieldDescriptor` map guard ("don't graft a stray `items` onto an object map") removed no observable output. The change only adds map detail where a Data property is typed `array<string, T>`.

Clear to merge.

Issue #344, opened 2026-06-14
Explicit-schema escape hatch — Sub-item B: class-level `#[Schema]` body replacement


Radiergummi:
Spun off from #336 (parent) as **Sub-item B** of the explicit-schema escape hatch. The first slice — Sub-item A (`x-*` vendor-extension passthrough) — is in PR #343; the issue's proposed sequencing is **A (→ #343) → C → B**, so this lands after A (and likely after C).

## What

A class-level `#[Schema([...])]` attribute that says "**skip inference, use this literal schema as the component body**." Nothing today does this: `#[SchemaName]` only *labels* the component key (read by `ComponentSchemaRegistry`); it does not replace the inferred body. So this is the part of #336 that genuinely earns a new symbol — a healthy authoring attribute supplying something the code cannot express.

## Why it's its own issue (heavier than A)

Per the #343 plan, B is materially larger than the `x-*` slice:

- **New attribute** `src/Attributes/Schema.php` (`TARGET_CLASS`).
- **Inference-skip branch in all three plugin assembly loci** — each has a `private buildSchema()` that produces the component `OA\Schema`:
  - `src/Plugins/SpatieData/Support/SchemaFromDataClass.php::buildSchema()`
  - `src/Plugins/ApiResources/Support/SchemaFromResource.php::buildSchema()`
  - `src/Plugins/Core/Support/SchemaFromFormRequest.php::buildSchema()`
  An early branch returns the literal schema and skips the multi-pass inference.
- **Honesty guardrail (issue #336 Guardrail 2)** — a new lint rule `schema.raw-keyword-unsupported` that rejects/flags keywords `swagger-php`'s `OA\Schema` cannot serialize, rather than dropping them silently. Mirrors the `OverridesUnknownField` rule shape (a `Rule` + visitor, stable id, fixHint) and gets a `docs/linting.md` catalog row.

## Serializer boundary (do not over-promise)

Scope the accepted keywords to what `OA\Schema` actually models. **Modeled** (safe): `const`, `oneOf`/`anyOf`/`allOf`, `not`, and the standard validation keywords (`type`, `enum`, `minimum`/`maximum`, `pattern`, `minItems`/`maxItems`, `properties`, `required`, `additionalProperties`, …). **NOT modeled** (the #140 wall — must be rejected, not silently dropped): `if`/`then`/`else`, `dependentRequired`, `dependentSchemas`. Truly arbitrary keywords are deferred to the IR ADR (#189).

[...]

PR #346, opened 2026-06-14
feat(#345): field-level `additionalProperties:` override

### LINKED ISSUE #345: Explicit-schema escape hatch — Sub-item C: field-level `schema:` / `additionalProperties:` override
Spun off from #336 (parent) as **Sub-item C** of the explicit-schema escape hatch. The first slice — Sub-item A (`x-*` vendor-extension passthrough) — is in PR #343; the issue's proposed sequencing is **A (→ #343) → C → B**, so this lands after A and before B.

## What

Two field-level explicit-override knobs on the existing `FieldAttribute` base (used by `PathParam`/`QueryParam`/`RequestField`/`ResponseField`/`ResourceField`/`TransformerField`) — **not** a new attribute, since the base already holds the full JSON-Schema field surface:

- **`additionalProperties:`** — an explicit map-value override on a field.
- **`schema:`** — a merge knob to override/augment the field's resolved schema.

These thread through the same shared chokepoint A uses: `FieldAttribute` → `SchemaDescriptor` → `SchemaDescriptor::applyTo($property)`.

## Why it's its own issue

- **`additionalProperties:` overlaps #334.** #334 covers the *inferable* map case (string-keyed `array<string, T>` → `additionalProperties` from the type). This issue covers the *explicit override* — when the author wants to declare/override the map-value schema directly. The two must be designed to coexist (explicit override wins over inference); sequencing this after #334 lands keeps the inference baseline settled first.
- **`schema:` needs its own merge-semantics design.** "Merge knob" is underspecified: replace the resolved schema, deep-merge into it, or compose (allOf)? Pick one and justify it — this is the design work that makes C distinct from the trivial `x-*` of A. Bound the accepted keywords to what `OA\Schema` serializes (same #140 wall as Sub-item B; reuse B's `schema.raw-keyword-unsupported` lint rule if it exists by then).

## Serializer boundary

Same as B — only keywords `swagger-php`'s `OA\Schema` models are safe (`const`, `oneOf`/`anyOf`/`allOf`, `not`, standard validation keywords); `if`/`then`/`else`/`dependent*` are the #140 wall and must be rejected/linted, not dropped. Truly arbitrary keywords are deferred to the IR ADR (#189).

## Bookkeeping

`docs/attributes.md`, `CHANGELOG.md [Unreleased]`, and a `docs/linting.md` row if a new/extended lint rule is added.

## Related

- Parent: #336. First slice: #343 (Sub-item A). Sibling: Sub-item B (class-level `#[Schema]`).
- #334 (map inference — `additionalProperties` from string-keyed types; the inference half this override complements), #140 (unmodeled swagger-php keywords), #189 (IR ADR).

Radiergummi:
Closes #345

## Recommendation up front: ship `additionalProperties:` now; do NOT build the free-form `schema:` knob

Issue #345 proposes two field-level knobs on `FieldAttribute`: `additionalProperties:` and a `schema:` merge knob. After reading the merged Sub-item A pattern and #334's inference, I recommend implementing **`additionalProperties:` only**, and **dropping `schema:`** (not merely deferring it), with the reasoning recorded for the maintainer.

**Why drop `schema:`** (per the inference philosophy — an attribute is healthy only when it supplies something the code cannot express):

- `FieldAttribute` *already* exposes every JSON-Schema keyword `OA\Schema` models as a typed, individually-overridable constructor param (`type`, `format`, `minimum`/`maximum`, `pattern`, `minItems`, `enum`, `nullable`, …). And `SchemaDescriptor::applyTo()` is **already a shallow overlay** — it writes each non-null field onto the inferred `OA\Property`, so the author already wins per-keyword. There is no "merge-semantics gap" for the modeled keywords.
- The *only* thing a free-form `schema: array` adds beyond the typed params is **unmodeled/composed keywords** (`if`/`then`/`else`, `dependent*`, exotic compositions) — which are exactly the #140 wall + IR (#189) territory, and the class-level literal-body case is **Sub-item B's** job (#344). A field-level free-form `schema:` would (a) re-introduce the keyword-validation/lint burden #345 itself flags, (b) duplicate B's surface at field scope, and (c) be a second way to set keywords the typed params already cover — the *smell* the philosophy warns against.
- So `schema:` is either redundant (modeled keywords) or blocked (unmodeled). Building it now would add public surface that mostly re-states existing capability. **Drop it**; if a concrete unmodeled-keyword case appears, it rides the IR work, not a field knob.

This is a deliberate scope decision (narrower than the issue's literal two-knob framing) and will be posted as a PR comment for maintainer confirmation.

## Tier & boundary

`additionalProperties:` is **Tier-2 → attribute** (an explicit override the code can't always express) and respects the `openapi.overrides` boundary unchanged: the attribute is co-located with the class and travels with refactors; `overrides` is document-level for code you can't annotate. No Contracts change, no config key, no stage reorder.

## The merged A pattern this builds on (verified)

Sub-item A (#343, merged) established the exact chokepoint:

- `FieldAttribute` base constructor carries the knob (e.g. `public ?array $x`), forwarded into `descriptor()` → `SchemaDescriptor`.
- `SchemaDescriptor::applyTo(OA\Property)` (`:152`) overlays `toOpenApi()` fields, then calls `applyVendorExtensions()`; `SchemaDescriptor::toSchema()` (`:72`) is the parallel path for callers that build a `Schema` directly (parameter resolvers), also ending in `applyVendorExtensions()`.
- The **`UriParametersExtractor` caveat**: path params build a `Schema` via `toSchema()`/a direct call rather than `applyTo()`, so any new descriptor field that must reach path params has to be honored in **both** `applyTo()` and `toSchema()` (A added `applyVendorExtensions()` to both for exactly this reason).
- A exposed the knob on the four core concrete attributes: `ResponseField`, `RequestField`, `QueryParam`, `PathParam` (each forwards `x: $x` to `parent::__construct`).

## How `additionalProperties:` interacts with #334 (design question 2 — resolved)

#334 (merged) sets `additionalProperties` on the property schema **during inference**
(`SchemaFromDataClass::resolveCollectionSchema()` → `{type: object, additionalProperties: …}` for a string-keyed map). The attribute overlay runs **after** inference: `applyValidationRules()` → then `applyPropertyAttribute()` (`SchemaFromDataClass.php:212/807`) calls `descriptor()->applyTo($property)`. So **the explicit override naturally wins** by running last — no special arbitration needed, identical to how every other field keyword already overrides inference. The plan just has to make `applyTo()`/`toSchema()` write `additionalProperties` when the descriptor carries it. **Explicit author value wins over #334's inferred map value — spelled out, and locked by a test.**

`swagger-php` types `OA\Schema::$additionalProperties` as `bool|AdditionalProperties`
(`JsonSchemaTrait.php:126`) — so the descriptor carries either a `bool` (`true`/`false`) or a nested schema. **Scope the attribute arg to what's expressible in an attribute literal:** `bool` (the common `additionalProperties: true|false` case) plus an optional value-type as a `class-string`/primitive-type-string that we wrap into an `OA\AdditionalProperties([...])` (mirroring how `items:` already takes a type string in `FieldAttribute`/`itemsSchema()`). Final arg shape to confirm in review — recommend starting with **`bool|string|null`**: `true`/`false` for the boolean form, a primitive type string (`'string'`/`'integer'`) or class-string for a typed map value; `null` = not set (inference stands).

### 1. Tests first

`tests/Unit/Support/Generator/SchemaDescriptorTest.php` + a feature test/fixture:

- **Unit:** `SchemaDescriptor(additionalProperties: true)` → `applyTo($property)` sets `$property->additionalProperties === true`; same via `toSchema()`. `false` → `false`. A typed   value (`'string'`) → `$property->additionalProperties` is an `OA\AdditionalProperties`/`OA\Schema` with `type: string`.
- **Feature — override beats inference (#334 interaction):** a Data property typed `array<string, Foo>` (which #334 infers to `additionalProperties: {$ref: Foo}`) carrying `#[ResponseField(additionalProperties: false)]` → emitted `additionalProperties: false` (author wins). And `#[ResponseField(additionalProperties: 'string')]` → `additionalProperties: {type: string}`.
- **Feature — no inference present:** a plain `object`-typed field with `#[ResponseField(type: 'object', additionalProperties: true)]` → `additionalProperties: true`.
- **PathParam path (UriParametersExtractor caveat):** `#PathParam(additionalProperties: …)]` reaches the emitted parameter schema via `toSchema()` (proves both apply paths honor the field). (Edge: path params are scalar — this mainly guards that `toSchema()` doesn't drop the field; keep the assertion minimal.)
- **Negative / default:** no `additionalProperties:` arg → property unchanged; #334 inference (when present) survives untouched.

### 2. Thread `additionalProperties` through the chokepoint

- `src/Support/Generator/SchemaDescriptor.php` — add `public bool|string|null $additionalProperties = null` to the constructor. Add a private `additionalPropertiesValue(): bool|OA\Schema|null` helper (bool passes through; a type/class-string wraps into `OA\AdditionalProperties`/an `OA\Schema` with `type`/`$ref`, mirroring `itemsSchema()`). In **both** `applyTo()` and `toSchema()`, after the existing field application, set `$target->additionalProperties` when the helper returns non-null. (Two apply paths because of the `UriParametersExtractor` caveat.) Do **not** put it in `toOpenApi()` — like `items`/`nullable`/`x`, it's a structural field set directly.
- `src/Attributes/FieldAttribute.php` — add `public bool|string|null $additionalProperties = null` to the base constructor; forward into `descriptor()`.

### 3. Expose `additionalProperties:` on the core field attributes

Add the `additionalProperties:` param (forwarding to `parent::__construct`) on the same four core attributes A used: `ResponseField`, `RequestField`, `QueryParam`, `PathParam`. (PathParam's value is nearly always scalar, but exposing it keeps the surface symmetric with A and costs one line; note in PR if review prefers omitting it there.)

Plugin-owned subclasses (`ResourceField`/`TransformerField`/`AllowedFilter`) inherit base support but don't expose the param — same trivial follow-up line as A left for them.

### 4. Bookkeeping

- `docs/attributes.md` — document `additionalProperties:` on the field family, with the override-beats-inference note (links to #334's map inference) and the `openapi.overrides` boundary. State that free-form `schema:` is intentionally **not** offered (typed params + the IR/Sub-item B cover that ground).
- `CHANGELOG.md [Unreleased]` — under **Added**: field-level `additionalProperties:` override on field attributes.
- **No `docs/linting.md` row** — dropping `schema:` means no `schema.raw-keyword-unsupported` rule is needed for C (the bool/typed `additionalProperties` value can't carry unmodeled keywords). That lint rule lands with Sub-item B (#344) where free-form keyword arrays actually appear.

### Files

- *Modified:* `src/Support/Generator/SchemaDescriptor.php`, `src/Attributes/FieldAttribute.php`, `src/Attributes/ResponseField.php`, `src/Attributes/RequestField.php`, `src/Attributes/QueryParam.php`, `src/Attributes/PathParam.php`
- *Tests:* `tests/Unit/Support/Generator/SchemaDescriptorTest.php`, a feature test + fixture
- *Docs:* `docs/attributes.md`, `CHANGELOG.md`

No `src/Contracts/**`, no config key, no stage-order change.

## Verification

- `composer test` green; `vendor/bin/pint --test` clean; `composer lint` (PHPStan L8) clean.
- Emitted `additionalProperties` (bool + nested `OA\AdditionalProperties`) validates on swagger-php **5.8 and 6.x** — both forms are core swagger-php (the envelope strategies and #334 already emit the nested form), so no #140 risk.

## Controversial / escalation flags (escalation-bound: new public surface, area:core)

1. **Public authoring-surface addition** — a new `additionalProperties:` constructor param on the public `FieldAttribute` base + four public attributes. Human gate; lands ready-but-escalated for Moritz. Expected.
2. **Dropping the `schema:` knob** — narrower than the issue's two-knob framing. Recorded as a deliberate "healthy-attribute" decision (typed params already cover modeled keywords; unmodeled keywords are #189/#344). **Maintainer confirm** whether to proceed `additionalProperties`-only.
3. **Attribute arg shape for `additionalProperties`** (`bool|string|null`, with string = type/ class-string wrapped into a nested schema) — one design choice for review to confirm.

All three are design/surface questions, not implementation risk; no Contracts seam, no stage reorder, no host-app runtime mutation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -28,6 +28,7 @@ All notable changes to this project are documented here.
 - String-keyed array/collection properties on Spatie `Data` classes (`array<string, T>`) emit `additionalProperties` (a map) instead of a bare array. (#334)
 - `#[CookieParam]` authoring attribute documenting an `in: cookie` parameter read off the request at runtime. (#335)
 - `x:` vendor-extension passthrough on the field attributes (`#[ResponseField]`, `#[RequestField]`, `#[QueryParam]`, `#[PathParam]`, `#[CookieParam]`): attach `x-*` specification extensions to a field's schema, co-located with the field (the field-level analogue of `openapi.overrides`). (#336)
+- `additionalProperties:` override on the field attributes (`#[ResponseField]`, `#[RequestField]`, `#[QueryParam]`, `#[PathParam]`): set a field's map behaviour to a bool or a typed value schema; applied last, it wins over inferred `array<string, T>` map values. (#345)
 
 ### Changed
 - `spatie/laravel-data` is now a soft dependency (moved from `require` to `require-dev`); Fractal and query-builder packages are opt-in.
--- a/docs/attributes.md
+++ b/docs/attributes.md
@@ -55,7 +55,7 @@ target. The wrong scope is caught by `field.attribute-wrong-scope`.
 |---|---|---|---|
 | `RequestField` | property, parameter, class-constant | Request-body input fields | Place on a Spatie Data class property / promoted constructor parameter, or on a `FormRequest` field constant. Supports `writeOnly` and `default`. No `readOnly` (a request field is never read-only). |
 | `ResponseField` | class-constant, property | Response output fields | Place on a response class field constant or property. Supports `readOnly` and `conditional`. `conditional: true` keeps the field in `properties` but removes it from `required`. Use for conditionally-present fields. |
-| `PathParam` | parameter | URI path parameters | Place on a controller action parameter for a route-bound model or scalar segment. Only `description`, `example`, `format`, `pattern`, and `x` apply (type is inferred from the binding). |
+| `PathParam` | parameter | URI path parameters | Place on a controller action parameter for a route-bound model or scalar segment. Only `description`, `example`, `format`, `pattern`, `x`, and `additionalProperties` apply (type is inferred from the binding). |
 | `QueryParam` | class, method | Ad-hoc query parameters | See operation-level table above. |
 | `CookieParam` | class, method | Cookie parameters | See operation-level table above. |
 
@@ -66,9 +66,9 @@ All five subclasses share the same JSON Schema field surface inherited from
 type), `default`, `nullable`, `enum`, `minimum` / `maximum`,
 `exclusiveMinimum` / `exclusiveMaximum`, `multipleOf`, `minLength` /
 `maxLength`, `pattern`, `minItems` / `maxItems`, `uniqueItems`, `readOnly`,
-`writeOnly`, `x` (where applicable to the scope — e.g. `RequestField` has no
-`readOnly`, `PathParam` only `description` / `example` / `format` / `pattern` /
-`x`).
+`writeOnly`, `x`, `additionalProperties` (where applicable to the scope — e.g.
+`RequestField` has no `readOnly`, `PathParam` only `description` / `example` /
+`format` / `pattern` / `x` / `additionalProperties`).
 
 ### Vendor extensions (`x-*`)
 
@@ -90,6 +90,28 @@ This is the field-level analogue of the `openapi.overrides` config key: use the
 attribute when the extension is **co-located with the field and should travel
 with refactors**; use `openapi.overrides` for **document-level** `x-*` targeted
 at code you can't or won't annotate (third-party routes, generated controllers).
+
+### `additionalProperties` override
+
+The `additionalProperties:` argument overrides the schema's open/closed map
+behaviour. Pass a bool, or a type string wrapped into a nested value schema
+(mirroring `items:`):
+
+```php
+#[ResponseField(additionalProperties: false)]    // forbid extra keys
+#[ResponseField(additionalProperties: 'string')] // values are strings → {type: string}
+```
+
+It **wins over inference**: a property typed `array<string, Foo>` is otherwise
+inferred to `additionalProperties: {$ref: Foo}` (the [map inference](auto-derivation.md)),
+and the attribute — applied last — replaces that inferred value. Leave the
+argument off (`null`) to keep the inferred map untouched; `additionalProperties:
+false` is distinct from unset and emits a closed map.
+
+A free-form `schema:` knob is intentionally **not** offered: every keyword
+`OA\Schema` models as a scalar is already an individually-overridable field
+argument, and composition keywords (`oneOf`/`allOf`/…) belong to the class-level
+schema-body escape hatch, not a field knob.
 
 ### Enum from a backed-enum class-string
 
--- a/src/Attributes/FieldAttribute.php
+++ b/src/Attributes/FieldAttribute.php
@@ -39,38 +39,45 @@ abstract readonly class FieldAttribute
      * @param null|non-empty-string                                                        $description
      * @param null|class-string|OpenApiPrimitiveType                                       $type
      * @param null|non-empty-string                                                        $format
-     * @param null|array<int, BackedEnum|int|string>|class-string<BackedEnum>|FieldDefault $enum        A list of
-     *                                                                                                  allowed values,
-     *                                                                                                  or a
-     *                                                                                                  backed-enum
-     *                                                                                                  class-string
-     *                                                                                                  resolved to its
-     *                                                                                                  cases.
+     * @param null|array<int, BackedEnum|int|string>|class-string<BackedEnum>|FieldDefault $enum                 A list of
+     *                                                                                                           allowed values,
+     *                                                                                                           or a
+     *                                                                                                           backed-enum
+     *                                                                                                           class-string
+     *                                                                                                           resolved to its
+     *                                                                                                           cases.
      * @param null|int<0, max>                                                             $minLength
      * @param null|int<0, max>                                                             $maxLength
      * @param null|non-empty-string                                                        $pattern
      * @param null|int<0, max>                                                             $minItems
      * @param null|int<0, max>                                                             $maxItems
-     * @param bool                                                                         $conditional When true, the
-     *                                                                                                  field is kept
-     *                                                                                                  in
-     *                                                                                                  `properties`
-     *                                                                                                  but removed
-     *                                                                                                  from `required`
-     *                                                                                                  — used by
-     *                                                                                                  response fields
-     *                                                                                                  emitted via
-     *                                                                                                  `$this->when()`
-     *                                                                                                  /
-     *                                                                                                  `$this->whenLoaded()`.
-     * @param null|array<string, mixed>                                                    $x           Vendor extensions
-     *                                                                                                  (`x-*`); keys are
-     *                                                                                                  written with the
-     *                                                                                                  `x-` prefix and
-     *                                                                                                  emitted verbatim,
-     *                                                                                                  matching the
-     *                                                                                                  `openapi.overrides`
-     *                                                                                                  contract.
+     * @param bool                                                                         $conditional          When true, the
+     *                                                                                                           field is kept
+     *                                                                                                           in
+     *                                                                                                           `properties`
+     *                                                                                                           but removed
+     *                                                                                                           from `required`
+     *                                                                                                           — used by
+     *                                                                                                           response fields
+     *                                                                                                           emitted via
+     *                                                                                                           `$this->when()`
+     *                                                                                                           /
+     *                                                                                                           `$this->whenLoaded()`.
+     * @param null|array<string, mixed>                                                    $x                    Vendor extensions
+     *                                                                                                           (`x-*`); keys are
+     *                                                                                                           written with the
+     *                                                                                                           `x-` prefix and
+     *                                                                                                           emitted verbatim,
+     *                                                                                                           matching the
+     *                                                                                                           `openapi.overrides`
+     *                                                                                                           contract.
+     * @param null|bool|string                                                             $additionalProperties Map-value
+     *                                                                                                           override: `true`/
+     *                                                                                                           `false`, or a type
+     *                                                                                                           string wrapped into
+     *                                                                                                           a nested value
+     *                                                                                                           schema. Wins over
+     *                                                                                                           inferred map values.
      *
      * @throws InvalidArgumentException When `$enum` is a string that is not a backed-enum class-string.
      */
@@ -99,6 +106,7 @@ abstract readonly class FieldAttribute
         public ?bool $writeOnly = null,
         public bool $conditional = false,
         public ?array $x = null,
+        public bool|string|null $additionalProperties = null,
     ) {
         // A backed-enum class-string is resolved to its cases here, so every downstream reader
         // (descriptor, lint rules) sees a uniform value list rather than a bare class name.
@@ -191,6 +199,7 @@ abstract readonly class FieldAttribute
             readOnly: $this->readOnly,
             writeOnly: $this->writeOnly,
             x: $this->x,
+            additionalProperties: $this->additionalProperties,
         );
     }
 }
--- a/src/Attributes/PathParam.php
+++ b/src/Attributes/PathParam.php
@@ -28,7 +28,8 @@ final readonly class PathParam extends FieldAttribute
      * @param null|non-empty-string     $description
      * @param null|non-empty-string     $format
      * @param null|non-empty-string     $pattern
-     * @param null|array<string, mixed> $x           Vendor extensions (`x-*`).
+     * @param null|array<string, mixed> $x                    Vendor extensions (`x-*`).
+     * @param null|bool|string          $additionalProperties Map-value override.
      */
     public function __construct(
         ?string $description = null,
@@ -36,6 +37,7 @@ final readonly class PathParam extends FieldAttribute
         ?string $format = null,
         ?string $pattern = null,
         ?array $x = null,
+        bool|string|null $additionalProperties = null,
     ) {
         parent::__construct(
             description: $description,
@@ -43,6 +45,7 @@ final readonly class PathParam extends FieldAttribute
             format: $format,
             pattern: $pattern,
             x: $x,
+            additionalProperties: $additionalProperties,
         );
     }
 }
--- a/src/Attributes/QueryParam.php
+++ b/src/Attributes/QueryParam.php
@@ -27,16 +27,17 @@ final readonly class QueryParam extends FieldAttribute
      * @param null|non-empty-string                                                        $description
      * @param null|OpenApiPrimitiveType                                                    $type
      * @param null|non-empty-string                                                        $format
-     * @param null|array<int, BackedEnum|int|string>|class-string<BackedEnum>|FieldDefault $enum        Allowed values,
-     *                                                                                                  or a backed-enum
-     *                                                                                                  class-string; renders as a
-     *                                                                                                  dropdown.
+     * @param null|array<int, BackedEnum|int|string>|class-string<BackedEnum>|FieldDefault $enum                 Allowed values,
+     *                                                                                                           or a backed-enum
+     *                                                                                                           class-string; renders as a
+     *                                                                                                           dropdown.
      * @param null|int<0, max>                                                             $minLength
      * @param null|int<0, max>                                                             $maxLength
      * @param null|non-empty-string                                                        $pattern
      * @param null|int<0, max>                                                             $minItems
      * @param null|int<0, max>                                                             $maxItems
-     * @param null|array<string, mixed>                                                    $x           Vendor extensions (`x-*`).
+     * @param null|array<string, mixed>                                                    $x                    Vendor extensions (`x-*`).
+     * @param null|bool|string                                                             $additionalProperties Map-value override.
      */
     public function __construct(
         public string $name,
@@ -63,6 +64,7 @@ final readonly class QueryParam extends FieldAttribute
         ?int $maxItems = null,
         ?bool $uniqueItems = null,
         ?array $x = null,
+        bool|string|null $additionalProperties = null,
     ) {
         parent::__construct(
             title: $title,
@@ -86,6 +88,7 @@ final readonly class QueryParam extends FieldAttribute
             maxItems: $maxItems,
             uniqueItems: $uniqueItems,
             x: $x,
+            additionalProperties: $additionalProperties,
         );
     }
 }
--- a/src/Attributes/RequestField.php
+++ b/src/Attributes/RequestField.php
@@ -32,10 +32,10 @@ use Radiergummi\OpenApi\Support\Attributes\FieldDefault;
 final readonly class RequestField extends FieldAttribute
 {
     /**
-     * @param null|non-empty-string                                                        $name        Field name; required
-     *                                                                                                  on a method, derived
-     *                                                                                                  from the target
-     *                                                                                                  otherwise
+     * @param null|non-empty-string                                                        $name                 Field name; required
+     *                                                                                                           on a method, derived
+     *                                                                                                           from the target
+     *                                                                                                           otherwise
      * @param null|non-empty-string                                                        $title
      * @param null|non-empty-string                                                        $description
      * @param null|OpenApiPrimitiveType                                                    $type
@@ -46,7 +46,8 @@ final readonly class RequestField extends FieldAttribute
      * @param null|non-empty-string                                                        $pattern
      * @param null|int<0, max>                                                             $minItems
      * @param null|int<0, max>                                                             $maxItems
-     * @param null|array<string, mixed>                                                    $x           Vendor extensions (`x-*`).
+     * @param null|array<string, mixed>                                                    $x                    Vendor extensions (`x-*`).
+     * @param null|bool|string                                                             $additionalProperties Map-value override.
      */
     public function __construct(
         public ?string $name = null,
@@ -73,6 +74,7 @@ final readonly class RequestField extends FieldAttribute
         ?bool $uniqueItems = null,
         ?bool $writeOnly = null,
         ?array $x = null,
+        bool|string|null $additionalProperties = null,
     ) {
         parent::__construct(
             title: $title,
@@ -97,6 +99,7 @@ final readonly class RequestField extends FieldAttribute
             uniqueItems: $uniqueItems,
             writeOnly: $writeOnly,
             x: $x,
+            additionalProperties: $additionalProperties,
         );
     }
 }
--- a/src/Attributes/ResponseField.php
+++ b/src/Attributes/ResponseField.php
@@ -37,7 +37,8 @@ final readonly class ResponseField extends FieldAttribute
      * @param null|non-empty-string                                                        $pattern
      * @param null|int<0, max>                                                             $minItems
      * @param null|int<0, max>                                                             $maxItems
-     * @param null|array<string, mixed>                                                    $x           Vendor extensions (`x-*`).
+     * @param null|array<string, mixed>                                                    $x                    Vendor extensions (`x-*`).
+     * @param null|bool|string                                                             $additionalProperties Map-value override.
      */
     public function __construct(
         ?string $title = null,
@@ -62,6 +63,7 @@ final readonly class ResponseField extends FieldAttribute
         ?bool $readOnly = null,
         bool $conditional = false,
         ?array $x = null,
+        bool|string|null $additionalProperties = null,
     ) {
         parent::__construct(
             title: $title,
@@ -86,6 +88,7 @@ final readonly class ResponseField extends FieldAttribute
             readOnly: $readOnly,
             conditional: $conditional,
             x: $x,
+            additionalProperties: $additionalProperties,
         );
     }
 }
--- a/src/Support/Extraction/UriParametersExtractor.php
+++ b/src/Support/Extraction/UriParametersExtractor.php
@@ -72,6 +72,7 @@ final readonly class UriParametersExtractor
             $schema->example = $fieldDescriptor->example;
         }
 
+        $fieldDescriptor?->applyAdditionalProperties($schema);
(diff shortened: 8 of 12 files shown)


Radiergummi:
🤖 **[planner]** plan posted — see PR body. Ready for plan-review. **Escalation-bound** (new public surface, area:core) — lands ready-but-escalated for Moritz regardless.

Resolves the two design questions I flagged when filing #345:

1. **schema: merge semantics → recommend dropping the schema: knob entirely.** FieldAttribute already exposes every OA\Schema-modeled keyword as a typed, individually-overridable param, and SchemaDescriptor::applyTo is already a shallow explicit-wins overlay — so there is no merge gap for modeled keywords. The only thing free-form schema: would add is unmodeled/composed keywords, which are the #140/#189 wall and Sub-item B's (#344) job. A field-level schema: would re-introduce the keyword-lint burden and duplicate B at field scope — the "smell" (re-stating expressible capability). So ship additionalProperties: only.

2. **additionalProperties: vs #334 inference → explicit wins, naturally.** #334 sets additionalProperties during inference; applyPropertyAttribute runs descriptor()->applyTo AFTER inference, so the override wins by running last — identical to every other field keyword. No special arbitration. Locked by a test.

Built on the merged A chokepoint (applyTo + toSchema both end in applyVendorExtensions; the UriParametersExtractor path-param caveat means additionalProperties must be honored in both apply paths). Exposed on the same four core attributes A used.

Maintainer-confirm: (a) additionalProperties-only vs. also building schema:; (b) the arg shape bool|string|null (string = type/class-string wrapped into a nested schema, mirroring items:).

Radiergummi:
🤖 **[lead]** plan posted (PR #346) → plan-review. **Lead endorses the planner's scope cut: additionalProperties-ONLY, drop the schema: knob.** Rationale is sound and matches the inference philosophy: FieldAttribute already exposes every OA\Schema-modeled keyword as a typed explicit-wins param (applyTo is a shallow overlay) — a free-form schema: only adds unmodeled/composed keywords, which is the #140/#189 wall and Sub-item B's (#344) job; a field-level schema: would duplicate B at field scope + re-introduce the keyword-lint burden (re-stating-expressible-capability smell). So C = one additionalProperties: knob threaded exactly like A threaded x: (same 4 core attrs, SchemaDescriptor::applyTo+toSchema chokepoint, UriParametersExtractor path-param caveat), no new lint rule (that lands with B). additionalProperties: vs #334 = explicit wins naturally (override runs after inference; lock with a test). Reviewer: validate the no-merge-gap claim + the arg-shape decision (bool|string|null, string=type/class-string wrapped like items:). Escalation-bound regardless.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved** — the `additionalProperties:`-only scope is sound and both load-bearing claims hold against the code. One sharpened justification for the `schema:` drop, and one implementation hazard to flag for coding. No plan changes required.

**Claim 2 (`additionalProperties:` vs #334) — verified exactly.** `SchemaFromDataClass::buildSchema()` is explicitly three passes and the code even comments it: Pass 1 inference (`:184`, incl. `additionalProperties` from `resolveCollectionSchema`), Pass 2 validation rules (`:201`), **Pass 3 (`:203` comment: "apply scoped field attribute overrides last so authoring annotations win") → `applyPropertyAttribute` (`:212`) → `descriptor()->applyTo()` (`:818`)**. So an explicit `additionalProperties` overlays the #334-inferred value by running last — no arbitration needed, identical to every other field keyword. ✓

> **Implementation hazard to lock with the test (the one place this can go backwards):** the override only wins if `applyTo()` writes `$property->additionalProperties` **unconditionally** when the descriptor carries it. The #334 value is *already on the property* from Pass 1, so a "skip-if-already-defined" guard would make **inference** win, inverting the intent. Note that the sibling validation-rule path (`applyValidationRules` → `applyTo($prop, overwrite: false)` at `:789`) uses `overwrite: false` — the coder must NOT copy that here. The override-beats-inference feature test (`array<string, Foo>` + `#[ResponseField(additionalProperties: false)]` → emits `false`) is what proves the write is unconditional; keep it.

**Claim 1 (drop `schema:` — no merge gap) — holds, with a sharper justification than the plan gives.** `applyTo()` (`:152-154`) overlays each non-null `toOpenApi()` field onto the inferred property — a shallow, per-keyword, explicit-wins overlay; `toOpenApi()` enumerates every modeled **scalar** keyword as an individually-overridable param. So for those, there is genuinely no merge gap. **But the plan's "redundant or blocked" dichotomy is slightly imprecise:** `FieldAttribute` does **not** expose the *composition* keywords `OA\Schema` models — `oneOf`/`anyOf`/`allOf`/`not`/`const`/`discriminator` (grep confirms none are typed params). A free-form `schema:` *could* set those, and they are serializer-safe — so `schema:` is not purely redundant. The reason to still drop it is cleaner: **field-level composition is Sub-item B's design space** (B is the literal-schema-body case, bounded to exactly that modeled-composition keyword set + its `schema.raw-keyword-unsupported` lint). Shipping `schema:` in C would pre-empt B's bounded-keyword + lint design at field scope — the plan's own argument (b)/(c). So: **drop `schema:` confirmed**, but because composition belongs to B (not because `schema:` adds nothing modeled). Either way no Tier-2 escalation and no field-level need is stranded — a concrete field-composition case is B/#189, not a C knob.

**Arg shape `bool|string|null` — right surface.** Mirrors the existing `items:` type-string→nested-schema pattern (`itemsSchema()` `:132`), so it's a proven idiom, not a new one. Default **must** stay `null` (not `false`) so "unset" (inference stands) is distinguishable from explicit `additionalProperties: false` (forbid extra props) — the plan has this right.

**UriParametersExtractor wiring — yes, needed in both `applyTo()` and `toSchema()`** (A established this; path params build via `toSchema()` and bypass `applyTo()`). Realistically `additionalProperties` on a scalar path param is near-meaningless, so that test is purely a "`toSchema()` doesn't drop the field" guard and `PathParam` exposure is pure symmetry — the plan flags this for review, which is the correct call (omitting it on `PathParam` would also be defensible).

**No scope creep toward B.** No `#[Schema]`, no literal-schema body, no lint rule, no `docs/linting.md` row — correct (the bool/typed `additionalProperties` value can't carry unmodeled keywords, so C needs no keyword-lint; that lands with B/#344). Clean extension of the merged A chokepoint. Escalation-bound for Moritz (area:core, new public surface), as flagged. Ready to code.

Radiergummi:
🤖 **[lead]** plan-review approved → coding. Scope confirmed: additionalProperties-only, threaded like A (4 core attrs, SchemaDescriptor applyTo+toSchema, UriParametersExtractor path-param drop-guard). **CRITICAL IMPLEMENTATION HAZARD (reviewer-1):** the override only beats #334's inference if applyTo writes additionalProperties UNCONDITIONALLY (overwrite). #334 already set the value in Pass 1; a skip-if-defined guard would INVERT it (inference would win). Do NOT copy the sibling validation path's applyTo(..., overwrite: false) at :789. Lock with a test: array<string,Foo> + additionalProperties: false → emits false. Arg shape bool|string|null (string=type/class-string wrapped like items:); default stays null, not false. No new lint rule (that's B/#344).

Radiergummi:
🤖 **[coder]** scope confirmed + critical hazard handled.

**Scope: additionalProperties-only, schema: dropped** — implemented exactly as planned/approved. No `schema:` knob, no `#[Schema]`, no lint rule, no `docs/linting.md` row (those belong to Sub-item B / #344).

**Critical hazard (override must beat #334 inference) — handled and locked.** `applyTo()`/`toSchema()` write `additionalProperties` **unconditionally** via a dedicated `applyAdditionalProperties()` (NOT the sibling validation path's `overwrite: false`). #334 sets the map value in Pass 1; the attribute overlay runs in Pass 3 (last), so the explicit override wins. Proven by a feature test asserting the **serialized spec**: `array<string, AddressFixtureData>` + `#[ResponseField(additionalProperties: false)]` → emits `additionalProperties: false` (inferred `{$ref}` replaced); `additionalProperties: 'string'` → `{type: string}`.

**Arg shape:** `bool|string|null`, default `null` (unset = inference stands; distinct from explicit `false`). A string wraps into `OA\AdditionalProperties(['type' => …])`, mirroring `items:`. Both forms are core swagger-php (#334 and the envelope strategies already emit the nested form), so no #140 risk — will confirm on the 5.8 CI job.

**Deviation (minor, consistent with A):** like A's `x:`, the `additionalProperties` override had to be wired into `UriParametersExtractor` too — path params build their schema via a direct `toSchema()`/type path that bypasses `applyTo()`, so I made `applyAdditionalProperties()` public and call it there. On a scalar path param this is near-meaningless; the test is a "toSchema() doesn't drop the field" guard, and PathParam exposure is pure symmetry with A (omitting it there would also be defensible).

Plugin-owned subclasses (`ResourceField`/`TransformerField`/`AllowedFilter`) inherit base support but don't expose the param — same trivial follow-up A left.

Radiergummi:
🤖 **[coder]** implemented; local gates green (Pint clean, PHPStan L8 no errors, 2588 tests pass); pushed a4e111a. Ready for review — escalation-bound for Moritz (new public authoring surface, area:core).

Files: `SchemaDescriptor` (`additionalProperties` param + `additionalPropertiesValue()` helper + public `applyAdditionalProperties()`, applied in both `applyTo` and `toSchema`), `FieldAttribute` (base param + `descriptor()` forward), the four core attributes, `UriParametersExtractor` (PathParam drop-guard). Tests: `tests/Unit/Support/Generator/SchemaDescriptorTest.php` (bool/string/toSchema/no-touch, serialized-output), `tests/Feature/AdditionalPropertiesOverrideTest.php` + fixture (override-beats-inference for false and type-string, PathParam guard). Docs: `docs/attributes.md` additionalProperties section + boundary; `CHANGELOG.md` Added.

Radiergummi:
🤖 **[lead]** coding done (sha a4e111a), remote CI green incl. swagger-php 5.8 (confirms nested OA\AdditionalProperties serializes on 5.8) → review. TWO independent approvals needed. Hazard handled: applyAdditionalProperties writes UNCONDITIONALLY (no overwrite:false guard); override-beats-inference locked by serialized test (array<string,Foo> + additionalProperties:false → emits false). Arg bool|string|null, default null. Minor deviation: applyAdditionalProperties made public for the PathParam drop-guard (consistent with A's applyVendorExtensions). Scope strictly additionalProperties-only (no #[Schema]/lint/linting.md — B/#344). Escalation-bound (area:core/public surface).

Radiergummi:
🤖 **[reviewer]** review round 1: approved, no concerns. First of two approvals. (Escalation-bound for Moritz — new public authoring surface, area:core.)

Independent **code** review of `a4e111a` (I did the plan-review and flagged the precedence hazard; verified the diff, ran the tests, and proved the override-beats-inference test is non-vacuous).

**The precedence hazard I flagged at plan time is CLOSED and regression-locked.** `applyAdditionalProperties()` writes `$target->additionalProperties = $value` **unconditionally** when the descriptor carries one — no `overwrite: false` skip-guard — and the docblock spells out exactly why ("an inferred map value … is already on the property … a skip-if-defined guard would invert that"). It did **not** copy the `applyValidationRules` → `applyTo(…, overwrite: false)` pattern. Confirmed against `SchemaFromDataClass`’s 3-pass order (inference Pass 1 → validation Pass 2 → attribute overlay Pass 3, last), so the explicit value wins by running last.

**The override test is non-vacuous — verified.** The fixture's `closedMap`/`retypedMap` are `array<string, AddressFixtureData>`, the *exact* shape #334's own test asserts infers `additionalProperties.$ref = '#/components/schemas/AddressFixtureData'` (`SchemaFromDataClassTest:247`). The override test asserts the emitted spec instead shows `false` / `{type: string}` — which inference for that type never produces, so the only path to that output is the override **replacing** the inferred `{$ref}`. That proves replacement, not mere addition. Ran the suite: **14 passed (20 assertions)**, incl. the unit `false`/`true`/type-string-wrap/`toSchema()` cases and the no-touch default (`$property->additionalProperties` stays `Generator::UNDEFINED` — locks unset ≠ explicit `false`).

**Arg shape `bool|string|null`, default `null` — right surface.** `additionalPropertiesValue()` passes a bool through and wraps a type-string into `OA\AdditionalProperties(['type' => …])`, mirroring `itemsSchema()` — a proven idiom. Threaded through **both** `applyTo()` and `toSchema()`, plus the `UriParametersExtractor` PathParam drop-guard (the path-param test confirms `toSchema()` doesn't drop the field; the assertion is honestly commented as near-meaningless on a scalar param).

**Deviation (public `applyAdditionalProperties`) — sound, parity with A.** `SchemaDescriptor` is already `@internal` (`:26`), so widening this method to `public` changes nothing in the documented surface (the public API is the attributes). It's the minimal seam for the `UriParametersExtractor` bypass — identical to A's `applyVendorExtensions`. No public-surface addition beyond the attribute constructor params themselves.

**Scope held `additionalProperties`-only.** File list: only the four core attributes + `FieldAttribute` base + `SchemaDescriptor` + `UriParametersExtractor` + tests/docs. **No `src/Attributes/Schema.php`, no lint rule, no `docs/linting.md` row** — correct (a bool/typed value can't carry unmodeled keywords, so C needs no keyword-lint; that lands with B/#344). No B leakage.

**Conventions / boundaries / bookkeeping** — strict-types, `// region` markers, `match(true)` for the value resolution, comments explain the *why* (the unconditional-write invariant); `Support/` changes are plugin-agnostic infra (no Core/plugin breach, no config key, no `Contracts`/stage change); host app untouched. `CHANGELOG.md [Unreleased] / Added` accurate; `docs/attributes.md` documents the override, the override-beats-inference note, and the honest "no `schema:` knob" rationale (composition belongs to the class-level escape hatch).

No findings. **Approved** (first of two) — escalation-bound for Moritz after the second approval.

Radiergummi:
🤖 **[reviewer-2]** review round 1: approved, no concerns (1 minor doc-accuracy nit, non-blocking). (Escalation-bound — recording a thorough verdict for Moritz.)

Reviewed independently at `a4e111a`; verified the precedence chain and the swagger-php serialization against the actual code, not the summaries.

**The load-bearing precedence is correct — override wins over #334 inference.** Traced the SpatieData path:
- **Pass 1** sets the inferred `additionalProperties: {$ref: Foo}` from #334 map inference on `array<string, Foo>`.
- **Pass 2** (`:201`) merges validation rules.
- **Pass 3** (`:203` → `applyPropertyAttribute` `:818`) calls `$attribute->descriptor()->applyTo($property)` — i.e. **`SchemaDescriptor::applyTo`**, which now calls `applyAdditionalProperties`, writing `$target->additionalProperties = $value` **unconditionally**. Running last + overwriting is exactly what makes the override win.
- The coder correctly did **not** copy the sibling `FieldDescriptor::applyTo(overwrite: false)` at `:789` — that is a *different class* (the Support validation-rule merge), and copying its skip-if-defined precedence would have inverted the result. The new method’s docblock calls this out explicitly.

**Proving tests assert serialized output and genuinely exercise both passes.** The fixture properties are real `array<string, AddressFixtureData>` (so #334 *does* infer `{$ref}` in Pass 1), and the feature tests assert on `generateSpec()`: `closedMap.additionalProperties === false` and `retypedMap.additionalProperties === {type: string}` — proving the attribute overwrote the inferred `$ref`, not a vacuous pass. Unit tests cover false / true / type-string-wrap / standalone `toSchema()` / **no-attr no-touch** (`=== Generator::UNDEFINED`).

**swagger-php 5.8 nested form — safe.** The code builds a proper `OA\AdditionalProperties(["type" => …])` *model object* (not a raw nested array), then assigns it post-construction. Probed locally: both `additionalProperties: false` and the assigned nested `{type: string}` serialize correctly, and the green 5.8 matrix job confirms the model form round-trips there (the raw-array shape would be the 5.8-risky one; this isn’t that).

**Scope — C only, no B leak.** `bool|string|null` (default `null`); the `schema:` knob is deliberately omitted and the omission is documented in both the code and `docs/attributes.md` ("a free-form `schema:` knob is intentionally not offered"). No composition keywords. `CookieParam` is correctly **excluded** (a cookie value is a string scalar — a map override is meaningless there), unlike `x:` in #343 which applied to all five; the four-attribute scope is internally consistent across CHANGELOG, docs row, and shared-surface list.

**Boundaries clean.** Touches only `src/Attributes/` (the new author API) + `src/Support/Generator|Extraction/`. **No `Plugins/`, `Contracts/`, `config/`, stage-order, or lint.** `SchemaDescriptor` stays `@internal`, so the new public `applyAdditionalProperties` is public-within-internal — the same justified pattern as #343’s `applyVendorExtensions` (path params bypass `applyTo()`).

**Minor non-blocking nit:** the `SchemaDescriptor::$additionalProperties` docblock says "a type/**class-string** wrapped into a nested value schema", but the code does `new OA\AdditionalProperties(["type" => $value])` verbatim — a class-string would emit an invalid `{type: "App\\Foo"}` (no `$ref` resolution). The public attribute contract only promises a *type string* (`"string"`), and the docs example only shows that, so this is a doc overreach in the internal docblock, not a code bug. Suggest dropping "class-string" from that one `@param` line. (Could also be read as a future hook — fine to leave, but the wording currently implies behaviour the code doesn’t do.)

CHANGELOG `[Unreleased]/Added` present; docs gain a dedicated "`additionalProperties` override" section with the precedence + closed-map semantics and the explicit no-`schema:` rationale. Conventions clean (strict-types, PHP 8.4, `match`, no abbreviations, concise comments).

CI green at `a4e111a`: full matrix + swagger-php 5.8 + Snapshot + quality (PHPStan L8) + Pint + codecov.

Approved. Recommend Moritz merge once the second approval + survey Layer A are in; the doc-line nit is optional and can ride a follow-up.

Radiergummi:
🤖 **[lead]** two independent approvals in (reviewer-1 + reviewer-2, both verified the unconditional-write precedence over #334's Pass-1 inference + non-vacuous serialized override test + 5.8 uses a proper OA\AdditionalProperties model object, not the 5x-risky raw nested array) → survey. reviewer-2 noted CookieParam correctly EXCLUDED here (cookie value is a string scalar; map override meaningless) — unlike x: in A. Generation-affecting (area:core); running Layer A to complete the escalation package. One trivial non-blocking docblock nit (drop 'class-string' from the @param) being fixed in the pre-escalation rebase — no generation impact, so this survey holds.

Radiergummi:
🤖 **[surveyor]** Layer A pass — 0 regressions across 11 apps. (Informational run; merge is human-gated.)

Generation-affecting PR (`area:core`), full corpus run. Fresh baseline regenerated on current `origin/main` (`82971b5`, post-#343-merge). Candidate manifest-verified across all 11 apps.

**Note on the candidate sha:** the run installed `c1559b6` — the worktree advanced from `a4e111a` to `c1559b6` mid-run as the pre-announced docblock-only fix landed (`docs(attributes): match additionalProperties docblock to behaviour`, a 4-line `@param` wording tweak in `SchemaDescriptor`, zero code/behaviour change). So this result measures the **final** sha, and is equally valid for `a4e111a`.

**Deterministic metrics — identical across the corpus** (paths / operations / apiOperations / responseSchemas / documentedResponses / requestBodies / componentSchemas). No metric changed on any app; `generateExit=0` everywhere.

| app | paths | operations | requestBodies | componentSchemas | gen_exit |
|-----|-------|------------|---------------|------------------|----------|
| Lychee | 213 | 273 | 92 | 280 | 0 |

(all columns equal to baseline)

**Content-level spec diff** (normalized JSON, baseline vs candidate): 8/11 apps byte-identical; AureusERP, SpeedtestTracker, InvoiceNinja differ **only** in non-deterministic Livewire per-install asset-route hashes and random `api.generated_`/`client.generated_` operationIds.

**`additionalProperties` is unchanged on every app** (count identical baseline↔candidate, zero `additionalProperties` diff lines) — exactly as expected: the new `bool|string|null $additionalProperties` override is opt-in, and `applyAdditionalProperties()` writes nothing when the parameter is `null` (every path through `SchemaDescriptor::toSchema()`, `applyTo()`, and `UriParametersExtractor` is gated on the author passing it). No corpus app uses it, so the footprint is invisible — same opt-in shape as #339/#343.

Radiergummi:
🤖 **[lead]** ESCALATED → needs-human (Moritz to merge). Same reason as Sub-item A: new public author-facing API surface (the `additionalProperties:` parameter on the 4 core field attributes) in area:core — outside the auto-merge subset.

**Ready to merge — full package (sha c1559b6, CLEAN/up-to-date):**
- **Scope:** Sub-item C, de-scoped to **additionalProperties: ONLY** (the `schema:` knob was deliberately dropped — field-level composition is Sub-item B's design space (#344); shipping schema: in C would pre-empt B + re-introduce the keyword-lint burden). No #[Schema], no lint rule, no linting.md row.
- **Precedence (the load-bearing point):** explicit override beats #334's map inference. #334 sets additionalProperties in Pass 1; the attribute overlay runs Pass 3 (last) and `applyAdditionalProperties` writes UNCONDITIONALLY — so explicit wins. Deliberately did NOT copy the sibling validation path's `applyTo(overwrite:false)` at :789 (that would invert it). Locked by a non-vacuous serialized-output test: `array<string,AddressFixtureData>` (the exact shape #334 infers as {$ref}) + `additionalProperties: false` → emits `false`.
- **Arg shape:** bool|string|null, default null (unset ≠ explicit false); string wraps into an OA\AdditionalProperties MODEL object (not the 5.8-risky raw nested array), mirroring items:.
- **Two independent approvals** (reviewer-1 + reviewer-2, both ran the suite + traced the 3-pass order + verified 5.8 serialization). CookieParam correctly EXCLUDED (cookie value is a string scalar; map override meaningless) — note this differs from A's x: which hit all 5 attrs.
- **Survey Layer A:** 0 regressions / 11 apps (informational; opt-in footprint invisible).
- **CI:** full matrix green incl. swagger-php 5.8 + Snapshot.
- **Boundaries:** only src/Attributes + src/Support; no Contracts/config/stage-order/lint.

PR #348, opened 2026-06-14
feat(#344): class-level #[Schema] body replacement (Sub-item B)

### LINKED ISSUE #344: Explicit-schema escape hatch — Sub-item B: class-level `#[Schema]` body replacement

Radiergummi:
Closes #344

## Scope: ship the full body-replacement mechanism; defer only the *secondary* conflict-lint

B genuinely bundles six things: (1) the `#[Schema]` attribute, (2) a shared literal-schema
builder + keyword validation, (3) inference-skip wiring in **three** plugin `buildSchema()` loci,
(4) the `schema.raw-keyword-unsupported` lint guardrail, (5) the class-`#[Schema]`-vs-field-attribute
**precedence**, (6) a *second* lint rule warning about ignored field attributes
(`schema.class-attribute-conflicts-with-field-attributes`).

The **minimal honest first slice = (1)–(5)**. That is the complete, usable feature: the attribute
replaces the inferred body, keywords are validated honestly, and the precedence is *implemented*
(class `#[Schema]` wins — the early return skips all field-attribute reads, so there is no silent
half-merge). The only deferred piece is **(6)**, the *cosmetic* lint that warns "you put
`#[ResponseField]` on a class that also has `#[Schema]`; the field attributes are ignored." That's a
nicety, not correctness — the behaviour is already correct and documented without it. Deferring it
keeps this PR to one reviewable mechanism + one lint rule. **(6) → follow-up issue** (file it after).

Rationale for not slicing thinner: the three inference-skips can't be split (a `#[Schema]` that only
works on Spatie Data but silently keeps inferring for resources/form-requests would be the dishonest
capability the parent issue warns against). And the keyword guardrail (4) must ship *with* the
attribute — emitting `#[Schema(['if' => …])]` that silently drops `if` is exactly the trap #336
Guardrail 2 exists to prevent. So (1)–(5) are one coherent, indivisible mechanism.

## Tier & boundary

Tier-2 → attribute. `#[Schema]` supplies a literal body the inference genuinely cannot derive — a
healthy attribute. Boundary vs. `openapi.overrides` (per #336 Guardrail 1): `#[Schema]` is
**co-located with the class, replaces its component body, travels with refactors**; `overrides` is
**document-level, operation-targeted** (it can't even address a component schema body). Different
locus, no overlap. Boundary vs. SwaggerPhp harvester (#17): the harvester ingests hand-authored
`#[OA\Schema]` annotations wholesale; `#[Schema]` is a *first-party* terse knob scoped to the
validated keyword set — complementary, not a replacement (note in docs).

## Seam placement (the cross-cutting concern the lead flagged)

The inference-skip must be **consistent across the three plugins without leaking plugin-specifics
into `Support/`**. Solution: one plugin-agnostic helper in `Support/` that does only reflection +
array→`OA\Schema` + keyword validation — no plugin types touched.

- **New:** `src/Support/Generator/ExplicitClassSchema.php` (or similar) — `@internal`. Static/simple
  methods: `read(ReflectionClass): ?Schema` (returns the parsed `#[Schema]` attribute instance or
  null) and `toSchema(Schema $attr): OA\Schema` (builds the literal `OA\Schema`, applying the
  validated keyword set). This is pure infra (reflection + swagger-php), so `Support/` is correct
  per the seam rules — it does **not** depend on any plugin.
- Each plugin's `private buildSchema()` gains an **early-return branch at the very top**, mirroring
  the existing `#[Discriminator]` early-return precedent already in
  `SchemaFromDataClass::buildSchema()` (`:121-140`):
  ```
  if (($explicit = ExplicitClassSchema::read($reflection)) !== null) {
      return ExplicitClassSchema::toSchema($explicit);
  }
  ```
  Loci (all confirmed same shape — reflect class, then build; all wrapped by
  `registry->buildOnce()`):
  - `src/Plugins/SpatieData/Support/SchemaFromDataClass.php::buildSchema()` (`:115`)
  - `src/Plugins/ApiResources/Support/SchemaFromResource.php::buildSchema()` (`:163`)
  - `src/Plugins/Core/Support/SchemaFromFormRequest.php::buildSchema()` (`:88`)
  Each plugin keeps its own reflection/registry glue; the shared helper carries the
  plugin-agnostic literal-schema logic. No plugin-specifics in `Support/`.

The `#[Schema]` attribute itself lives in `src/Attributes/Schema.php` (the general package — it's
not plugin-specific; all three plugins honor it).

## The keyword boundary (verified against swagger-php)

`OA\Schema` (via `Schema.php` + `JsonSchemaTrait.php`) **models** (safe to accept): `type`, `enum`,
`const`, `not`, `allOf`/`anyOf`/`oneOf`, `properties`, `required`, `additionalProperties`,
`patternProperties`, `propertyNames`, `contains`, `minimum`/`maximum`/`exclusiveMinimum`/
`exclusiveMaximum`, `multipleOf`, `minLength`/`maxLength`, `pattern`, `minItems`/`maxItems`,
`uniqueItems`, `default`, `title`, `description`, `format`, `dependencies`, plus `x-*` (via `$x`).
It does **NOT model** (the #140 wall — must reject, not drop): `if`, `then`, `else`,
`dependentRequired`, `dependentSchemas`. Maintain the accepted set as a `const` allowlist on the
helper (single source of truth, also consumed by the lint rule).

## The `schema.raw-keyword-unsupported` lint rule

- **id:** `schema.raw-keyword-unsupported`; **severity:** level 1 (error) — emitting a `#[Schema]`
  with a keyword the serializer drops produces a silently-wrong spec, the dishonest-capability case
  the guardrail targets. (Confirm 1 vs 3 in review; I lean 1 because it's silent data loss.)
- **What it flags:** a `#[Schema]` whose keyword array contains any key outside the helper's
  accepted allowlist. **fixHint:** "swagger-php cannot serialize `<keyword>`; remove it or express
  the shape with a supported keyword (see docs). Arbitrary keywords await the IR (#189)."
- **Shape:** mirror `OverridesUnknownField` (a `Rule` reading the parsed attribute; pick the visitor
  that reaches component schemas / class attributes — likely a class/operation visitor that can
  resolve the payload class, same traversal `MultipartFileWithoutMultipart` uses). Registered in
  `BaselineRegistration` (it's plugin-agnostic config-level infra — the rule keys off the attribute,
  not a plugin) **or** per the existing lint-registration convention; confirm during impl which
  registration path keeps it plugin-agnostic.
- **docs/linting.md** — add the catalog row between the `lint-rule-catalog` markers (hand-maintained
  per CLAUDE.md; `--list` is the live source).
- **Degrade-at-build:** independently of the lint rule, `ExplicitClassSchema::toSchema()` should
  **drop unsupported keys and log** (matching the library's "degrade gracefully + log" philosophy,
  e.g. the `EloquentModelToSchema`/`RequestBodyExtractor` precedent) so a non-linted run still
  produces a valid document rather than a swagger-php hard-fail. Lint surfaces it; the builder
  survives it.

## Conflict precedence (design question — RESOLVED)

**Class-level `#[Schema]` wins; field attributes on that class are ignored.** This falls out of the
early-return: `buildSchema()` returns the literal schema before any property/field-attribute pass
runs, so there is no merge and no ambiguity. Documented as the contract. The **warning** that field
attributes were ignored (`schema.class-attribute-conflicts-with-field-attributes`) is the deferred
secondary lint (follow-up issue) — the *behaviour* is defined and correct now; only the advisory
nudge is deferred.

### 1. Tests first

- **Unit (`tests/Unit/Support/Generator/ExplicitClassSchemaTest.php`):** `read()` returns the
  attribute when present, null otherwise; `toSchema(['type'=>'object','required'=>['a'],
  'properties'=>[...]])` → correct `OA\Schema`; an `oneOf`/`const`/`not` keyword survives; an
  unsupported `if` key is dropped + logged (assert it's absent from output).
- **Feature, per plugin (the indivisible three):** a Spatie Data class, a JsonResource, and a
  FormRequest each carrying `#[Schema([...])]` → the component schema is the literal body, and
  **inference is skipped** (e.g. a property the class would otherwise infer is absent because the
  literal body omits it). One feature test per locus to prove all three honor it.
- **Precedence:** a class with both `#[Schema([...])]` and `#[ResponseField]`/`#[ResourceField]` on
  members → the literal `#[Schema]` wins, field attributes have no effect.
- **Composition keywords:** `#[Schema(['oneOf' => [...]])]` round-trips (guards the #336 "don't
  silently drop oneOf/anyOf/allOf" honesty).
- **Lint rule:** `#[Schema(['if' => [...]])]` → `schema.raw-keyword-unsupported` finding with the
  offending keyword; a fully-supported `#[Schema]` → no finding.
- **Negative:** no `#[Schema]` → inference proceeds exactly as today (regression guard for all three
  plugins).

### 2–4. Implement

- `src/Attributes/Schema.php` — `#[Attribute(Attribute::TARGET_CLASS)]`, `final readonly`,
  constructor `(array $schema)` (PHPDoc the accepted shape; reference the allowlist). Docblock states
  "replaces the inferred component body; keywords bounded to what swagger-php serializes."
- `src/Support/Generator/ExplicitClassSchema.php` — the helper + accepted-keyword `const` allowlist
  + degrade-and-log.
- The three early-return branches (above).
- The lint rule + `BaselineRegistration`/lint registration + `docs/linting.md` row.

### 5. Bookkeeping

- `docs/attributes.md` (+ `docs/README.md` index if a section is added) — document `#[Schema]`, the
  accepted-keyword boundary, precedence over field attributes, and the boundary vs. `overrides` /
  the SwaggerPhp harvester.
- `CHANGELOG.md [Unreleased]` — **Added:** class-level `#[Schema]` body replacement + the
  `schema.raw-keyword-unsupported` lint rule.

### Files

- *New:* `src/Attributes/Schema.php`, `src/Support/Generator/ExplicitClassSchema.php`,
  the lint rule under `src/Lint/Rules/`, tests + fixtures.
- *Modified:* the three plugin `buildSchema()` loci, `BaselineRegistration` (or the lint registry),
  `docs/attributes.md`, `docs/linting.md`, `docs/README.md` (maybe), `CHANGELOG.md`.

No `src/Contracts/**` change, no config key, no stage-order change, no host-app runtime mutation.

## Deferred follow-up (file after this PR)

`schema.class-attribute-conflicts-with-field-attributes` — a level-3 advisory that field attributes
on a `#[Schema]`-bearing class are dead. Behaviour is already correct (they're ignored); this is the
nudge that tells the author. Own small issue.

## Controversial / escalation flags (escalation-bound: new public attribute, area:core + area:plugins)

1. **New public authoring symbol** `#[Schema]` + a new lint rule → human gate. Lands
   ready-but-escalated for Moritz. Expected — the lead pre-flagged it.
2. **Cross-cutting three-plugin change** — touches all three extraction backends. Mitigated by the
   shared `Support/` helper (one logic locus; plugins only add an early return) and the per-plugin
   feature tests. Still, it's the broadest blast radius of the batch → call out for careful review.
3. **Accepted-keyword allowlist + lint severity (1 vs 3)** — one judgement call; recommend level 1
   (silent data loss). Confirm in review.
4. **Naming:** `#[Schema]` is a common word and collides conceptually with swagger-php's
   `OA\Schema`. Confirm the maintainer wants this exact name (the issue specifies it) vs. e.g.
   `#[RawSchema]`/`#[SchemaBody]` to avoid import confusion. Flag, don't decide.

All are surface/design questions for the human gate, not implementation risk.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,7 @@ All notable changes to this project are documented here.
 - `#[CookieParam]` authoring attribute documenting an `in: cookie` parameter read off the request at runtime. (#335)
 - `x:` vendor-extension passthrough on the field attributes (`#[ResponseField]`, `#[RequestField]`, `#[QueryParam]`, `#[PathParam]`, `#[CookieParam]`): attach `x-*` specification extensions to a field's schema, co-located with the field (the field-level analogue of `openapi.overrides`). (#336)
 - `additionalProperties:` override on the field attributes (`#[ResponseField]`, `#[RequestField]`, `#[QueryParam]`, `#[PathParam]`): set a field's map behaviour to a bool or a typed value schema; applied last, it wins over inferred `array<string, T>` map values. (#345)
+- `#[RawSchema]` class-level attribute replaces a payload class's inferred component body with a literal JSON Schema (Spatie Data, API Resource, FormRequest); keywords are bounded to what swagger-php serialises, with unsupported keywords dropped-and-logged and flagged by the new `schema.raw-keyword-unsupported` lint rule. (#344)
 
 ### Changed
 - `spatie/laravel-data` is now a soft dependency (moved from `require` to `require-dev`); Fractal and query-builder packages are opt-in.
--- a/docs/attributes.md
+++ b/docs/attributes.md
@@ -188,6 +188,7 @@ Resource, a `FormRequest`, an Eloquent model).
 | Attribute | Target | Purpose |
 |---|---|---|
 | `SchemaName` | class | Override the component key the class maps to in `#/components/schemas/{key}`. |
+| `RawSchema` | class | Replace the inferred component body with a literal JSON Schema. Skips all convention inference and field attributes on the class. |
 
 By default the key is derived from the class basename (disambiguated with namespace segments, or a
 short hash as a last resort). That key is a public, consumer-facing contract — it becomes the type
@@ -208,6 +209,48 @@ restates what derivation already produces. The name still goes through the `comp
 lint rule, so keep it consistent with the rest of the document (PascalCase by default). Two distinct
 classes declaring the same name is an unresolvable conflict: generation throws
 `DuplicateSchemaNameException`.
+
+### Replace a component body with a literal schema (`#[RawSchema]`)
+
+When the shape a class produces genuinely cannot be derived — a hand-tuned composition, a body the
+runtime assembles dynamically — `#[RawSchema]` lets you supply the component body verbatim. The
+array becomes the component schema; the generator does not read the class's properties, `rules()`,
+`toArray()`, or any field-level attribute on it.
+
+```php
+use Radiergummi\OpenApi\Attributes as OpenApi;
+
+#[OpenApi\RawSchema([
+    'type' => 'object',
+    'required' => ['kind'],
+    'properties' => [
+        'kind' => ['type' => 'string', 'enum' => ['circle', 'square']],
+        'shape' => ['oneOf' => [
+            ['$ref' => '#/components/schemas/Circle'],
+            ['$ref' => '#/components/schemas/Square'],
+        ]],
+    ],
+])]
+final class ShapeData extends Data { … }
+```
+
+**Class-level wins.** Because the literal body replaces inference outright, any `#[ResponseField]`,
+`#[RequestField]`, etc. on the same class are ignored — there is no merge.
+
+**Bounded keyword set.** Keywords are limited to what swagger-php can serialise: `type`, `enum`,
+`const`, `not`, `oneOf`/`anyOf`/`allOf`, `properties`, `required`, `additionalProperties`,
+`patternProperties`, `propertyNames`, `contains`, `items`, the numeric/string/array constraints,
+`default`, `title`, `description`, `format`, plus any `x-*` extension. Unsupported keywords —
+`if`/`then`/`else`, `dependentRequired`/`dependentSchemas`, and the draft-07 `dependencies` (OpenAPI
+3.1 splits it into the two `dependent*` keywords; accepting only the deprecated spelling would be
+dialect-inconsistent) — are **dropped from the output and logged**, and the
+`schema.raw-keyword-unsupported` lint rule flags them so the loss is never silent.
+
+**Boundaries.** `#[RawSchema]` is co-located with the class and travels with it through refactors;
+use `openapi.overrides` instead for **document-level, operation-targeted** edits (it cannot address a
+component body). It is also distinct from the SwaggerPhp plugin's harvester, which ingests
+hand-authored `#[OA\Schema]` annotations wholesale — `#[RawSchema]` is a terse first-party knob
+scoped to the validated keyword set.
 
 ## Exception-level attribute
 
--- a/docs/linting.md
+++ b/docs/linting.md
@@ -463,6 +463,7 @@ is enabled.
 | `schema.allof-type-conflict` | 1 | allOf members declare conflicting type values. |
 | `schema.enum-empty` | 1 | A schema declares an empty enum (enum: []) and is unsatisfiable. |
 | `schema.nullable-via-deprecated-keyword` | 1 | Schema uses the deprecated OpenAPI 3.0 nullable: true keyword instead of a type array. |
+| `schema.raw-keyword-unsupported` | 1 | A #[RawSchema] uses a keyword swagger-php cannot serialise (if/then/else, dependentRequired/dependentSchemas, dependencies); it is silently dropped from the output. |
 | `security.invalid-scope` | 1 | Operation requires a scope not declared in securitySchemes. |
 | `streaming.no-content-type` | 1 | Streaming operation has no content-type: text/event-stream response. |
 | `throws.transitive-missing` | 1 | An action's handler declares @throws exceptions not redeclared on the controller method. |
--- a/src/Attributes/RawSchema.php
+++ b/src/Attributes/RawSchema.php
@@ -1,0 +1,45 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Attributes;
+
+use Attribute;
+
+/**
+ * Replaces a class's inferred component schema body with a literal JSON Schema. Placed on a
+ * payload class the generator would otherwise introspect (a Spatie Data class, an API Resource,
+ * or a FormRequest), it short-circuits that inference entirely: the array given here becomes the
+ * component body verbatim, and any field-level attributes on the class are ignored.
+ *
+ * ```php
+ * #[OpenApi\RawSchema([
+ *     'type' => 'object',
+ *     'required' => ['kind'],
+ *     'properties' => [
+ *         'kind' => ['type' => 'string', 'enum' => ['a', 'b']],
+ *     ],
+ * ])]
+ * final class WidgetData extends Data { … }
+ * ```
+ *
+ * Keywords are bounded to what swagger-php can serialise (see {@see \Radiergummi\OpenApi\Support\Generator\ExplicitClassSchema::ACCEPTED_KEYWORDS}).
+ * Unsupported keywords (`if`/`then`/`else`, `dependentRequired`/`dependentSchemas`,
+ * `dependencies`) are dropped at build time and flagged by the `schema.raw-keyword-unsupported`
+ * lint rule.
+ *
+ * For document-level, operation-targeted overrides use `openapi.overrides` instead; `#[RawSchema]`
+ * is co-located with the class and travels with it through refactors.
+ *
+ * @api
+ */
+#[Attribute(Attribute::TARGET_CLASS)]
+final readonly class RawSchema
+{
+    /**
+     * @param array<string, mixed> $schema A literal JSON Schema definition.
+     */
+    public function __construct(
+        public array $schema,
+    ) {}
+}
--- a/src/Lint/Rules/SchemaRawKeywordUnsupported.php
+++ b/src/Lint/Rules/SchemaRawKeywordUnsupported.php
@@ -1,0 +1,134 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Lint\Rules;
+
+use Override;
+use Radiergummi\OpenApi\Attributes\RawSchema;
+use Radiergummi\OpenApi\Contracts\Lint\Rule;
+use Radiergummi\OpenApi\Lint\Finding;
+use Radiergummi\OpenApi\Lint\LintContext;
+use Radiergummi\OpenApi\Lint\Tree\OperationNode;
+use Radiergummi\OpenApi\Lint\Visitors\OperationRule as OperationRuleVisitor;
+use Radiergummi\OpenApi\Support\Extraction\PayloadParameterScanner;
+use Radiergummi\OpenApi\Support\Generator\ExplicitClassSchema;
+use ReflectionClass;
+use ReflectionNamedType;
+
+use function class_exists;
+use function implode;
+use function sprintf;
+
+/**
+ * Flags a `#[RawSchema]` whose definition carries a keyword swagger-php cannot serialise
+ * ({@see ExplicitClassSchema::ACCEPTED_KEYWORDS}). Such keywords are silently dropped at build
+ * time, so without this rule a `#[RawSchema(['if' => …])]` would produce a quietly-wrong spec.
+ *
+ * Discovers `#[RawSchema]`-bearing classes by reflecting each operation's request-payload
+ * candidates ({@see PayloadParameterScanner}) and its action return type. The rule keys off the
+ * attribute, not any plugin, so it lives in the baseline rule set.
+ */
+final class SchemaRawKeywordUnsupported implements OperationRuleVisitor, Rule
+{
+    public const string ID = 'schema.raw-keyword-unsupported';
+
+    /**
+     * Classes already inspected this run, so a `#[RawSchema]` referenced by several operations is
+     * reported once. The rule is scoped per generation run, so the set resets with the instance.
+     *
+     * @var array<class-string, true>
+     */
+    private array $seen = [];
+
+    public function __construct(
+        private readonly PayloadParameterScanner $scanner,
+    ) {}
+
+    #[Override]
+    public function id(): string
+    {
+        return self::ID;
+    }
+
+    #[Override]
+    public function level(): int
+    {
+        return 1;
+    }
+
+    #[Override]
+    public function description(): string
+    {
+        return 'A #[RawSchema] uses a keyword swagger-php cannot serialise '
+            . '(if/then/else, dependentRequired/dependentSchemas, dependencies).';
+    }
+
+    /**
+     * @return iterable<Finding>
+     */
+    #[Override]
+    public function checkOperation(OperationNode $operation, LintContext $context): iterable
+    {
+        $method = $operation->descriptor?->method;
+
+        if ($method === null) {
+            return;
+        }
+
+        $candidates = $this->scanner->candidates($method);
+
+        $returnType = $method->getReturnType();
+
+        if ($returnType instanceof ReflectionNamedType && !$returnType->isBuiltin()) {
+            /** @var class-string $returnClass */
+            $returnClass = $returnType->getName();
+            $candidates[] = $returnClass;
+        }
+
+        foreach ($candidates as $className) {
+            yield from $this->checkClass($className);
+        }
+    }
+
+    /**
+     * @param class-string $className
+     *
+     * @return iterable<Finding>
+     */
+    private function checkClass(string $className): iterable
+    {
+        if (isset($this->seen[$className]) || !class_exists($className)) {
+            return;
+        }
+
+        $this->seen[$className] = true;
+
+        $reflection = new ReflectionClass($className);
+        $attributes = $reflection->getAttributes(RawSchema::class);
+
+        if ($attributes === []) {
+            return;
+        }
+
+        $unsupported = ExplicitClassSchema::unsupportedKeywords($attributes[0]->newInstance()->schema);
+
+        if ($unsupported === []) {
+            return;
+        }
+
+        yield new Finding(
+            ruleId: self::ID,
+            level: $this->level(),
+            message: sprintf(
+                '#[RawSchema] on %s uses unsupported keyword(s): %s.',
+                $className,
+                implode(', ', $unsupported),
+            ),
+            fixHint: 'swagger-php cannot serialise these keywords; remove them or express the '
+                . 'shape with a supported keyword (see docs/attributes.md). Arbitrary keywords '
+                . 'await the IR (#189).',
+            context: [Finding::CONTEXT_SOURCE_CLASS => $className],
+        );
+    }
+}
--- a/src/OpenApiServiceProvider.php
+++ b/src/OpenApiServiceProvider.php
@@ -546,6 +546,7 @@ class OpenApiServiceProvider extends ServiceProvider
                     wrappedModelLocator: $app->make(Plugins\ApiResources\Support\WrappedModelLocator::class),
                     modelToSchema: $app->make(Support\Extraction\EloquentModelToSchema::class),
                     logger: $app->make(LoggerInterface::class),
+                    explicitSchema: $app->make(Support\Generator\ExplicitClassSchema::class),
                 );
             },
         );
--- a/src/Plugins/ApiResources/Support/SchemaFromResource.php
+++ b/src/Plugins/ApiResources/Support/SchemaFromResource.php
@@ -16,6 +16,7 @@ use Radiergummi\OpenApi\Plugins\ApiResources\Resolvers\ResourceRefSchemaResolver
 use Radiergummi\OpenApi\Support\Extraction\EloquentModelToSchema;
 use Radiergummi\OpenApi\Support\Extraction\FieldReferenceProperty;
 use Radiergummi\OpenApi\Support\Generator\ComponentSchemaRegistry;
+use Radiergummi\OpenApi\Support\Generator\ExplicitClassSchema;
 use ReflectionClass;
 use ReflectionException;
 
@@ -68,6 +69,7 @@ final class SchemaFromResource
         private readonly WrappedModelLocator $wrappedModelLocator,
         private readonly EloquentModelToSchema $modelToSchema,
         private readonly LoggerInterface $logger,
+        private readonly ExplicitClassSchema $explicitSchema,
     ) {}
 
     /**
@@ -163,6 +165,11 @@ final class SchemaFromResource
     private function buildSchema(string $resourceClass): OA\Schema
     {
         $reflection = new ReflectionClass($resourceClass);
+
+        // #[RawSchema] replaces the inferred body wholesale; field-level attributes are ignored.
+        if (($rawSchema = $this->explicitSchema->read($reflection)) !== null) {
+            return $this->explicitSchema->toSchema($rawSchema, $reflection);
+        }
 
         /** @var list<OA\Property> $properties */
         $properties = [];
--- a/src/Plugins/Core/Support/SchemaFromFormRequest.php
+++ b/src/Plugins/Core/Support/SchemaFromFormRequest.php
@@ -18,6 +18,7 @@ use Radiergummi\OpenApi\Support\Extraction\FakerExampleSynthesiser;
 use Radiergummi\OpenApi\Support\Extraction\FieldDescriptor;
 use Radiergummi\OpenApi\Support\Extraction\ValidationRulesToSchema;
 use Radiergummi\OpenApi\Support\Generator\ComponentSchemaRegistry;
+use Radiergummi\OpenApi\Support\Generator\ExplicitClassSchema;
 use ReflectionAttribute;
 use ReflectionClass;
 use ReflectionClassConstant;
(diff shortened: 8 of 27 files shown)


Radiergummi:
🤖 **[planner]** plan posted — see PR body. Ready for plan-review. **Escalation-bound** (new public attribute, area:core + area:plugins) — lands ready-but-escalated for Moritz; broadest blast radius of the batch.

Scope: ship the full body-replacement mechanism (attribute + shared Support/ literal-schema builder + inference-skip in all three plugin buildSchema() loci + keyword validation + the schema.raw-keyword-unsupported lint). Defer only the *secondary* cosmetic lint (schema.class-attribute-conflicts-with-field-attributes) to a follow-up — behaviour is already correct without it. The three inference-skips and the keyword guardrail cannot be sliced thinner without shipping dishonest capability (a #[Schema] that silently keeps inferring on some backends, or silently drops if/then/else).

Seam: one plugin-agnostic helper in Support/ (reflection + array→OA\Schema + the keyword allowlist), each plugin adds a one-line early-return mirroring the existing #[Discriminator] precedent — no plugin-specifics leak into Support.

Design questions resolved: (1) precedence — class #[Schema] wins, field attributes ignored, falls out of the early return (no merge); (2) keyword boundary verified against swagger-php — accept type/enum/const/not/oneOf|anyOf|allOf/properties/required/additionalProperties/etc., reject if/then/else/dependentRequired/dependentSchemas (the #140 wall), drop-and-log at build + lint-flag.

Maintainer-confirm: (a) the attribute name #[Schema] (collides conceptually with OA\Schema — #[RawSchema]/#[SchemaBody]?); (b) lint severity 1 vs 3 (I lean 1 — silent data loss).

Radiergummi:
🤖 **[lead]** plan posted (PR #348) → plan-review. **Lead endorses the planner's scope:** ship the FULL mechanism — #[Schema]-family attribute + a shared Support/ literal-schema builder (reflection + array→OA\Schema + keyword allowlist) + inference-skip early-return in all three plugin buildSchema() loci (mirroring the #[Discriminator] precedent) + keyword validation + the schema.raw-keyword-unsupported lint. (1)-(5) are one indivisible mechanism (can't slice thinner without shipping dishonest capability). Defer ONLY the cosmetic schema.class-attribute-conflicts-with-field-attributes lint to a follow-up (behaviour is already correct; advisory only). Precedence resolved: class-level wins via the early return, no merge ambiguity. Keyword boundary: accept type/enum/const/not/oneOf|anyOf|allOf/properties/required/additionalProperties; reject if/then/else/dependent* (the #140 wall) — drop-and-log at build + lint-flag. TWO maintainer-confirm items pending Moritz (attribute NAME + lint SEVERITY) — does NOT block plan-review; reviewer validate the mechanism/seam/keyword-boundary/no-Support-leak. Escalation-bound regardless.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved with one required correction** — the scope, seam, precedence, and indivisibility are all sound, the keyword boundary is verified accurate, but the plan's literal-schema builder has a **concrete serialization defect for the composition keywords** that must be fixed (and is the exact 5.8 risk the lead flagged). Not a scope change — a build-mechanism + test correction.

**1. Seam — sound.** Verified the `#[Discriminator]` precedent: `SchemaFromDataClass::buildSchema()` (`:121-138`) already does an early class-attribute reflect-and-return before the property passes — a `#[Schema]` early-return mirrors it exactly. All three loci share the `private buildSchema(string): OA\Schema` shape (`SpatieData:115`, `ApiResources:163`, `FormRequest:88`). A plugin-agnostic `Support/` helper (reflection + array→OA + validation) leaks no plugin types. ✓ Two impl notes: the **FormRequest locus already has an early-return** (no-`rules()` method, `:91`) so the `#[Schema]` branch must sit **above** it; and FormRequest reflects via `class_basename`/`method_exists`, so it must `new ReflectionClass` to feed the helper's `read()`.

**2. Keyword allowlist — verified accurate.** Enumerated `OA\Schema` + `JsonSchemaTrait` properties. **Modeled (accept):** `type`, `enum`, `const`, `not`, `allOf`/`anyOf`/`oneOf`, `properties`, `required`, `additionalProperties`, `patternProperties`, `propertyNames`, `contains`, validation keywords. **NOT modeled (reject — the #140 wall):** `if`, `then`, `else`, `dependentRequired`, `dependentSchemas` — none exist as properties. The plan's accept/reject lists match. ✓ **One nuance to decide:** `OA\Schema` models `dependencies` (the **draft-07** spelling) and it serializes — but OpenAPI 3.1 uses JSON-Schema 2020-12, which **deprecated `dependencies`** in favor of `dependentRequired`/`dependentSchemas` (the rejected ones). So `dependencies` is *serialization-safe but spec-deprecated for 3.1*. The allowlist needs a conscious call: accept it (serializes) or reject it (wrong dialect). Either is defensible; just don't leave it implicit.

**3. Precedence — sound, no ambiguity.** Class `#[Schema]` wins by the early return short-circuiting before any field-attribute pass — identical to how the Discriminator early-return already bypasses property inference. ✓

**4. Indivisibility — planner is right.** The three skips can't split (a `#[Schema]` honored only by SpatieData but silently still inferring for resources/form-requests is the dishonest capability the parent issue forbids), and the keyword guardrail (4) must ship *with* the attribute (emitting `#[Schema(['if'=>…])]` that silently drops `if` is precisely #336 Guardrail 2). (1)–(5) are one mechanism; deferring only the cosmetic (6) conflict-lint is the correct cut. ✓

**REQUIRED CORRECTION — the literal-schema builder must recursively convert composition keywords, and the test must assert `validate()`.** I built `new OA\Schema([...])` from a raw keyword array with nested `properties` and `oneOf` (the plan's `toSchema()` shape). On **swagger-php 6.1.2** it **`validate()` → false** and logs *"`->properties` is an object literal, use nested `@OA\Property()`"* and *"`->oneOf` is a array, expecting `[OpenApi\Annotations\Schema]`"*. So dropping a raw array onto `OA\Schema` is invalid for the composition/object keywords on **both** 5.8 and 6.x — and those keywords (`oneOf`/`properties`) are the entire point of `#[Schema]`.

- Good news: `src/Support/Generator/SchemaFromArrayDefinition.php` **already exists** for exactly this and recursively converts `properties`→`OA\Property` and `items`→`OA\Items` (its docblock even cites the "5.x rejects a raw array under properties" bug). **Reuse it.**
- Gap: `SchemaFromArrayDefinition::apply()` (`:45-75`) special-cases **only** `properties` and `items`; every other key (incl. `oneOf`/`anyOf`/`allOf`/`not`) falls to `$node->{$key} = $value` — the raw array that fails `validate()`. So the helper must be **extended to recursively convert `oneOf`/`anyOf`/`allOf` (→ `OA\Schema[]`) and `not` (→ `OA\Schema`)** before `ExplicitClassSchema` can rely on it (cleanest: extend `SchemaFromArrayDefinition` once — existing callers benefit; alternatively handle it in the new helper).
- **Test consequence:** the unit test must assert **`$schema->validate() === true`** and/or round-trip through `generateSpec()` + the snapshot path — **not** inspect the output array. My probe shows the JSON output *looks* correct while `validate()` is false, so an array-shape assertion would pass green over a broken spec (and the 5.8 CI job — or any consumer `validate()` — would then fail). The "composition keywords round-trip" feature test is the one that proves this; make it assert validity, not shape.

**Scope held B = (1)–(5).** No drift into C; (6) correctly deferred. Naming (`#[Schema]` vs `#[RawSchema]`) and lint severity (1 vs 3) are the maintainer's calls, as flagged — I won't block, though I agree level 1 fits silent data loss.

Fix the builder to convert composition keywords + assert `validate()` in the test, and this is ready (escalation-bound for Moritz regardless).

Radiergummi:
Regarding the two maintainer decisions:
1. The attribute should be named `#[RawSchema]` for clarity.
2. Lint severity `1` makes sense.

Radiergummi:
🤖 **[lead]** plan-review **approved** → coding. Coder, here is the consolidated brief (plan in PR body is endorsed; these are the deltas/decisions on top of it):

**Moritz's two maintainer decisions (binding):**
1. Attribute name is **`#[RawSchema]`** — NOT `#[Schema]`. Name every symbol/file/test/docs accordingly (`src/Attributes/RawSchema.php`).
2. Lint severity is **1** (error) for `schema.raw-keyword-unsupported`.

**Reviewer's REQUIRED correction (must land — it's the 5.8 serialization defect):**
- A raw keyword array dropped onto `new OA\Schema([...])` fails `validate()` for the composition/object keywords (`oneOf`/`anyOf`/`allOf`/`not`, nested `properties`) on **both** 5.8 and 6.x — and those are the whole point of the attribute.
- `src/Support/Generator/SchemaFromArrayDefinition.php` already recursively converts `properties`→`OA\Property` and `items`→`OA\Items`, but every other key (incl. `oneOf`/`anyOf`/`allOf`/`not`) falls through to `$node->{$key} = $value` (the raw array that fails validation). **Extend `SchemaFromArrayDefinition` once** to recursively convert `oneOf`/`anyOf`/`allOf` (→ `OA\Schema[]`) and `not` (→ `OA\Schema`); existing callers benefit. Build `ExplicitClassSchema`/`RawSchema` on top of it.
- **Test consequence:** the unit + composition-keyword tests must assert **`$schema->validate() === true`** (and/or round-trip through `generateSpec()`/the snapshot path), **NOT** inspect the output array — a shape assertion passes green over a spec the 5.8 job rejects.

**Lead's call on the one open allowlist nuance (`dependencies`):** **reject** it. It's the draft-07 spelling; OpenAPI 3.1 uses JSON-Schema 2020-12 where it's deprecated in favor of `dependentRequired`/`dependentSchemas` — both already rejected. Accepting only the deprecated form would be dialect-inconsistent. Add `dependencies` to the reject set; document the rationale in a code comment + the docs keyword-boundary note. (Flagging for the human gate; Moritz can overrule.)

**Reviewer impl notes (carry these):**
- The FormRequest locus already has an early-return (no-`rules()` method, `SchemaFromFormRequest.php:91`) — the `#[RawSchema]` branch must sit **above** it.
- FormRequest reflects via `class_basename`/`method_exists`, so it must `new ReflectionClass(...)` to feed the helper's `read()`.
- All three loci share `private buildSchema(string): OA\Schema` (`SpatieData:115`, `ApiResources:163`, `FormRequest:88`); mirror the existing `#[Discriminator]` early-return precedent in `SchemaFromDataClass::buildSchema()` (:121-138).

**Scope:** B = (1)–(5) per the plan. Defer the cosmetic `schema.class-attribute-conflicts-with-field-attributes` lint to a follow-up issue (file it after, do NOT fold in). No `src/Contracts/**`, no config key, no stage-order change.

**This PR is escalation-bound regardless** (new public attribute + new lint rule + area:core/area:plugins) — drive it to green + ready, it lands for Moritz's final merge.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed `1f44dab` (rebased onto current `main`, CHANGELOG conflict resolved keep-both with #345). Ready for review.

**Scope shipped: B = (1)–(5).** Deferred cosmetic lint filed as #351.

**What landed (per the consolidated brief):**
- `src/Attributes/RawSchema.php` — `#[Attribute(TARGET_CLASS)]`, `@api`. (Named `#[RawSchema]`, not `#[Schema]`.)
- `src/Support/Generator/ExplicitClassSchema.php` (`@internal`, `#[Scoped]`) — `read()` / `toSchema()` + the single `ACCEPTED_KEYWORDS` allowlist; drops-and-logs unsupported keywords; static `unsupportedKeywords()` reused by the lint rule.
- **Extended `SchemaFromArrayDefinition`** to recursively convert `oneOf`/`anyOf`/`allOf` (→ `OA\Schema[]`) and `not` (→ `OA\Schema`). This is the 5.8 serialization fix — a raw array there fails `validate()`. Existing callers benefit.
- Early-return in all three loci, mirroring the `#[Discriminator]` precedent: `SchemaFromDataClass:buildSchema()`, `SchemaFromResource:buildSchema()`, `SchemaFromFormRequest:buildSchema()` (above the no-`rules()` early return, via `new ReflectionClass(...)`).
- `src/Lint/Rules/SchemaRawKeywordUnsupported.php` — id `schema.raw-keyword-unsupported`, **level 1**; registered in `BaselineRegistration`; appears in `openapi:lint --list`. `docs/linting.md` catalog row added.
- `dependencies` keyword **rejected** (draft-07 spelling; 3.1 uses `dependent*`) — rationale in a code comment + `docs/attributes.md`.

**Tests** assert `$schema->validate() === true` and round-trip through `generateSpec()`, not output-array shapes. New: `ExplicitClassSchemaTest`, `SchemaFromArrayDefinitionTest` (composition keywords), `SchemaRawKeywordUnsupportedTest`, `RawSchemaAttributeTest` (one feature test per plugin + precedence + unsupported-keyword degrade).

**Remote CI green on `1f44dab`** — full matrix incl. **PHP 8.5** and the **swagger-php 5.8** leg + the byte-exact Snapshot job all pass. PR is **MERGEABLE** (BLOCKED = awaiting review).

This PR is escalation-bound (new public attribute + new lint rule). Lands ready for Moritz's final merge.

Radiergummi:
🤖 **[lead]** coding done (1f44dab, full CI green incl. swagger-php 5.8 + Snapshot; rebased onto current main) → code-review. Reviewer: escalation-bound (new public `#[RawSchema]` + new lint rule level 1) so it lands for Moritz's final merge regardless — but give it a full adversarial pass. Focus: (1) `SchemaFromArrayDefinition` extension correctly recurses `oneOf`/`anyOf`/`allOf`(→Schema[])/`not`(→Schema) with no regression to existing callers; (2) tests assert `validate()===true` (not array-shape) and snapshot/3.1 paths green; (3) keyword allowlist + `dependencies` rejection + degrade-and-log; (4) the three plugin early-returns sit above the FormRequest no-rules() early-return; (5) docs (attributes.md keyword boundary + linting.md catalog row) + CHANGELOG. Cosmetic conflict-lint correctly deferred to #351.

Radiergummi:
🤖 **[reviewer]** code-review round 1: **approved with 1 non-blocking hardening finding.** Escalation-bound (new public `#[RawSchema]` + level-1 lint) — lands for Moritz's final merge regardless; this pass found no merge blocker.

Verified all five focus areas, plus an adversarial sweep of the keyword-conversion boundary:

**1. `SchemaFromArrayDefinition` recursion — correct, no regression.** `oneOf`/`anyOf`/`allOf` → `OA\Schema[]` via `array_map(build)`, `not` → nested `OA\Schema`. I reproduced on the vendored **6.1.2** that a raw `oneOf` array fails `validate()` (`@OA\Schema()->oneOf is a "array", expecting a "[OpenApi\Annotations\Schema]"`) — so the fix is genuinely necessary and lands exactly where needed. The four existing callers (`InlineJsonResponseResolver:530`, `OperationBuilder:599`, `ResourceToArrayReader:327`, `TransformerTransformReader:223`) previously left these keywords as raw arrays; the new branches only convert them — a strict improvement, and the Snapshot job staying green confirms no byte-shift on existing fixtures.

**2. Tests assert `validate()`, not shape.** `SchemaFromArrayDefinitionTest` (`...->validate())->toBeTrue()` on the oneOf/not cases), `ExplicitClassSchemaTest` (validate() on object/oneOf/const/degrade), and `RawSchemaAttributeTest` round-trips through `generateSpec()`. The "shape assertion passes green over a 5.8-rejected spec" trap the plan-review flagged is avoided. ✓

**3. Allowlist + `dependencies` rejection + degrade-and-log.** `ACCEPTED_KEYWORDS` matches the verified swagger-php-modeled set; `dependencies` correctly absent with the dialect rationale in both the `ExplicitClassSchema` docblock and `docs/attributes.md`. `toSchema()` filters + `logger->warning`s dropped keywords; `unsupportedKeywords()` is pure and shared with the lint rule (single source of truth). ✓

**4. Three early-returns positioned correctly.** All at the top of `buildSchema()` mirroring the `#[Discriminator]` precedent. The FormRequest branch (`SchemaFromFormRequest.php:94-99`) sits **above** the no-`rules()` early-return and correctly calls `setHasFileFields($class, false)` to match that path's invariant, reflecting via `new ReflectionClass`. ✓

**5. Bookkeeping.** `docs/attributes.md` (keyword boundary, class-wins precedence, `overrides`/harvester boundaries), `docs/linting.md` catalog row, `CHANGELOG.md [Unreleased]` all present and accurate. Lint rule registered in `BaselineRegistration` (plugin-agnostic, keys off the attribute via `PayloadParameterScanner` + return type), level 1, `$seen` dedup so a shared payload is reported once. ✓

---

**Finding 1 (non-blocking — test-suite/keyword-boundary hardening).**
`ExplicitClassSchema::ACCEPTED_KEYWORDS` admits several **nested-schema** keywords that `SchemaFromArrayDefinition` does **not** recursively convert: `additionalProperties` (when given a schema, not a bool), `patternProperties`, `propertyNames`, `contains`, and `discriminator`. On the vendored **6.1.2** these all `validate()` true as raw arrays (they are not typed `[OpenApi\Annotations\Schema]` in `Schema::$_nested`, so the type-check that rejects raw `oneOf` does not fire) — so there is **no current bug**. But:
- The plan-review established that **5.8 diverges** from 6.x on raw-array-vs-`OA`-object handling, and **no shipped test exercises these five keywords** (the tests cover `type/required/properties/oneOf/const/not/if`). So whether `#[RawSchema(['additionalProperties' => ['type' => 'string']])]` or `['patternProperties' => [...]]` validates on the 5.8 leg is **unverified**.
- Concrete tests to add (each asserting `$schema->validate() === true`, ideally also run on the 5.8 path): `additionalProperties` as a value-schema; `patternProperties`; `propertyNames`; `discriminator` with a `mapping`. If any fails on 5.8, extend `SchemaFromArrayDefinition::apply()` to convert that keyword (same pattern as `not`); if all pass, the test pins the boundary against future swagger-php drift.

Why non-blocking: no proven defect, full CI (incl. 5.8 + Snapshot) is green on `1f44dab`, and the feature is escalation-bound to Moritz. I'd recommend filing this as a small follow-up (sibling to the already-deferred #351) rather than holding the PR.

No other concerns. Mechanism, seam, precedence, and the Support/-no-plugin-leak boundary are all clean.

Radiergummi:
🤖 **[lead]** all gates clear → **needs-human** (ready, NOT merged). Controversial by policy (new public `#[RawSchema]` attribute + new lint rule `schema.raw-keyword-unsupported` level 1 + area:core/area:plugins) — Moritz's call to merge.

**State:** tip `1f44dab`; remote CI green on the full matrix (PHP 8.4/8.5 × Laravel 12/13 + swagger-php 5.8 + Snapshot). Plan-review approved (prior session). Code-review **approved, no merge blocker** (one non-blocking hardening finding → filed as #352). Two maintainer decisions already honored: name `#[RawSchema]`, lint severity 1; `dependencies` rejected (3.1-dialect consistency — overrule here if you'd rather accept the serializable draft-07 form).

**Survey skipped (deliberate):** the attribute is opt-in and no consumer app uses it, so a consumer survey exercises none of the new path; the `SchemaFromArrayDefinition` extension's only blast radius (4 existing callers) is covered by the green in-repo Snapshot.

**Follow-ups filed, not folded in:** #351 (cosmetic conflict-lint), #352 (5.8 validate() coverage for nested-schema keywords). Worktree left in place. Ready for your merge.

PR #349, opened 2026-06-15
feat(#168): implicit 404 from findOrFail/firstOrFail body scan (Tier-1)

Radiergummi:
Closes #168

### Tier

**Tier 1** — bounded AST whitelist on the first N top-level statements, no variable
tracking, graceful degrade. The sibling `AbortErrorContributor`
(`src/Plugins/Core/ErrorContributors/AbortErrorContributor.php`) is the exact model to follow.

### Scope correction (important — read before implementing)

The issue lists **two** new signals (inline `validate()` → 422 and `findOrFail`/`firstOrFail` → 404).
**The 422 half is already shipped.** `ValidationErrorContributor`
(`src/Plugins/Core/ErrorContributors/ValidationErrorContributor.php:97-104`) already detects
inline `validate()` via `InlineValidatorRulesReader` and emits the 422 with Laravel's
`{message, errors}` shape. The acceptance criterion "an action calling `$request->validate([...])`
(no FormRequest) emits a 422" already passes — we will add a confirming test, not new code, for it.

**The only genuinely new contributor is `findOrFail`/`firstOrFail` → 404.** Build just that.

### What the new contributor does

A new `FindOrFailErrorContributor` (`ErrorResponseContributor`) scans the action body for the
not-found lookup idioms and, if any is present, emits a single **404** `ErrorDescriptor` keyed off
`ModelNotFoundException` — identical in shape to the binding-derived 404 from
`RouteModelBindingErrorContributor`. Presence is the whole signal; there is no literal to evaluate
(unlike `abort()`), so the match is simpler than `AbortErrorContributor`.

Matched call shapes (case-insensitive on the method name; no receiver/variable tracking):
- `Model::findOrFail(...)` / `User::findOrFail($id)` — `PhpParser\Node\Expr\StaticCall`
- `$query->findOrFail(...)`, `Model::where(...)->firstOrFail()`, `$this->repo->firstOrFail()` —
  `PhpParser\Node\Expr\MethodCall`
- method names: `findOrFail`, `firstOrFail`, **`findOrFail`/`firstOrFail` only** — `sole()` and
  `firstOr()` are deliberately out of scope (they throw different / no exceptions).

Conditional-context policy: use `ConditionalContextPolicy::IncludeConditionalContexts` (same as
`AbortErrorContributor`) — a `findOrFail` inside an `if` branch is still a documented 404. Scan the
first `STATEMENT_LIMIT = 10` top-level statements via `MethodBodyScanner::firstStatements()` +
[...]

Issue #168, opened 2026-06-06
Implicit errors from method body: inline `validate()` → 422, `findOrFail` / `firstOrFail` → 404 (Tier-1)


Radiergummi:
## Summary

Extend the implicit-error inference with the two **Tier-1** (bounded AST whitelist) signals deferred from #29, which closed the Tier-0 half (route-model binding → 404, `can:` → 403, plus the pre-existing `@throws` / `FormRequest` / middleware signals).

## Motivation

#29 covers errors derivable from signatures, route URIs, and middleware names alone. Two common idioms still emit no error response because the signal lives in the **method body**:

- **Inline `validate()`** — actions that validate a bare `Illuminate\Http\Request` (rather than a typed `FormRequest`) throw `ValidationException` on failure but currently produce no 422.
- **`findOrFail()` / `firstOrFail()`** — explicit not-found lookups (no route-model binding) throw `ModelNotFoundException` → 404, but the binding-based detector in #29 does not see them.

Both are broadly-used patterns, so they sit at the economically-sensible end of Tier-1.

## Proposed approach

Tier-1 — parse the action body but match only a small whitelist of well-known call shapes in the first N statements, with no variable tracking across calls. Degrade gracefully and log when the shape isn't matched (per the Tier-1 discipline in `CLAUDE.md`).

- `$request->validate([...])` / `$this->validate(...)` / a top-level `Validator::make(...)->validate()` → **422**, carrying Laravel's `{message, errors}` shape via `ValidationRulesToSchema`, identical to the FormRequest-derived 422 from #29.
- `Model::findOrFail(...)` / `->findOrFail(...)` / `firstOrFail(...)` → **404** keyed off `ModelNotFoundException`, deduped against the binding-derived 404 from #29 (first contributor wins per status).

These would be new `ErrorResponseContributor`s in `src/Plugins/Core/`, gated behind the same
Tier-1 body-scan infrastructure as the rest of epic #5. `abort()` / `abort_if` / `abort_unless` body reads stay in #10.

## Acceptance criteria

- An action calling `$request->validate([...])` (no `FormRequest`) emits a 422 matching the standard error envelope.
- An action calling `findOrFail`/`firstOrFail` (no route-model binding) emits a 404.
- The 404/422 dedupe correctly against #29's binding/FormRequest-derived responses (no duplicate statuses on the operation).
- Unit tests per signal + an end-to-end `generateSpec()` assertion, tagged `uses()->group('openapi')`.
[...]

Issue #18, opened 2026-06-01
Eloquent model → schema from `$casts` / `$hidden` / `$appends`


Radiergummi:
## Summary

Teach the generator to build a response-body schema from an Eloquent model's own metadata when a controller returns that model (or a `JsonResource` wrapping it), so the most common Laravel response shape stops emitting an empty body.

## Motivation

Returning a model directly (`return $user;`, `return User::find($id);`) or a resource around one (`UserResource::make($user)`) is the single most common way Laravel controllers produce a response, yet today the generator reads none of the model's metadata and produces no schema for it. This is the highest-leverage response-fidelity gap in the surveyed apps. The recovered schema already exists as model metadata, so this is pure reflection — not method-body parsing.

## Proposed approach

**Tier 0** — deterministic reflection only. Add a model-metadata reader plus a primary-response resolver under `src/Core/` (reader in `src/Core/Support/`, resolver in `src/Core/Resolvers/`, registered by `src/Core/CorePlugin.php` alongside the existing primary-response resolvers).

Resolve the model class from the action's return type hint, from `Model::find` / `findOrFail` static calls, or from a `JsonResource` wrapping a model. From that class, build the schema by reading:

- `$casts` → property types (mirror the type mapping already done in `src/Core/Support/ValidationRulesToSchema.php` rather than reinventing it),
- `$hidden` / `$visible` → property visibility,
- `$appends` → extra accessor-backed properties,
- `$fillable` / migration columns → the base property set.

Reuse the shared backed-enum component (see #35) for enum casts instead of inlining enum schemas.

## Acceptance criteria

- A controller with a typed model return yields a response schema reflecting `$casts`, visibility, and `$appends`.
- `$hidden` fields are excluded; `$appends` fields are included.
- Model type resolution works for all three sources: return type hint, `find`/`findOrFail` static call, and a resource wrapping a model.
- Add unit tests for the metadata reader covering each source (`$casts`, `$hidden`/`$visible`, `$appends`, `$fillable`) under `tests/Unit/Core/`, and an end-to-end flavor under `tests/Feature/ExamplesTest.php` (snapshot byte-match + OpenAPI 3.1 validate + lint-clean).
- Escape hatch unchanged: `#[ResponseField]` (`src/Attributes/ResponseField.php`) still covers computed/virtual fields not present in `$casts`.

## References

[...]

PR #356, opened 2026-06-15
feat(#97): model-class response schema from find()/findOrFail() return (Tier-1)

### LINKED ISSUE #97: Model-class resolution from `find()` / `findOrFail()` return (Tier-1)
## Summary

Recover the model class for response-schema inference from `return Model::find($id);` / `findOrFail($id)` statements in a controller body, feeding the same `EloquentModelToSchema` reader that #18 builds.

## Context

Spun out of #18 (Eloquent model → response schema). #18 deliberately limits itself to **Tier-0 reflection** sources (return type hint + plain `Collection<Model>` generic). Its original acceptance criteria also listed a `find()`/`findOrFail()` static-call source, but resolving that requires **reading the method body** — Tier-1, not the reflection-only mechanism the parent epic #6 commits to. It is therefore descoped from #18 and tracked here.

See the implementation-ready design comment on #18: https://github.com/Radiergummi/laravel-openapi/issues/18#issuecomment-4613636060

## Scope

- Detect `Model::find(...)` / `findOrFail(...)` as the returned expression via the bounded body-scan utility (Tier-1 whitelist).
- Resolve the static-call class to a `Model` subclass and hand it to the existing `EloquentModelToSchema` reader / `EloquentModelResponseResolver` path.
- Degrade gracefully + log when the shape isn't matched (non-literal, wrapped, conditional).

Radiergummi:
Closes #97

### Tier

**Tier 1** — a bounded AST whitelist on the *returned expression* in the first N statements. No
dataflow. Reuses the merged body-scan infra (`MethodBodyScanner`, `StatementNodeFinder`,
`ConditionalContextPolicy`) and #18's merged Tier-0 model-schema machinery (`EloquentModelToSchema`,
`ComponentSchemaRegistry`). The closest siblings are `InlineJsonResponseResolver` (the other
body-scanning `PrimaryResponseResolver`, #14) and `EloquentModelResponseResolver` (the reflection
model resolver this feeds, #18).

### The gap

`EloquentModelResponseResolver` (`src/Plugins/Core/Resolvers/EloquentModelResponseResolver.php`)
resolves a model success schema **purely from the return type** (`Model` subclass, or
`Collection<Model>`). It misses the extremely common controller shape where the return is **untyped
or `JsonResponse`/`Model`-typed via a body expression**:

```php
public function show($id)            { return User::findOrFail($id); }   // untyped return
public function show($id): ?User     { return User::find($id); }         // ?User — already Tier-0? see guard
```

For the untyped form, no resolver produces a 200 schema today. #97 recovers the model class from the
returned `Model::find()/findOrFail()/firstOrFail()` static call and hands it to the same
`EloquentModelToSchema` reader for the success-response `$ref`.

### What to build — a new body-scanning resolver

**Create** `src/Plugins/Core/Resolvers/FindReturnModelResponseResolver.php` implementing
`PrimaryResponseResolver`. `#[Scoped] final readonly`. Ctor:
`(MethodBodyScanner $scanner, EloquentModelToSchema $modelToSchema, ComponentSchemaRegistry $registry, LoggerInterface $logger)`
(owns a `StatementNodeFinder` like the json/abort siblings).

`resolvePrimaryResponse(ActionDescriptor): ?OA\Response`:
1. `$method = $descriptor->method`; null → return null.
2. **Return-type guard** (mirror `InlineJsonResponseResolver::returnTypeAllowsBodyScan`): only scan
   when the declared return type leaves room — untyped, builtin, or an HTTP-response class. A return
   type that is *itself* a `Model` subclass or `Collection` is already
   `EloquentModelResponseResolver`'s territory (Tier-0) — defer (return null) so the reflection
   resolver, registered earlier, stays authoritative and we don't double-resolve. Union/intersection
   → refuse.
3. Scan first `STATEMENT_LIMIT = 10` statements; find the **returned** `Model::find/findOrFail/firstOrFail`
   static call (see matching below). No match → null (silent — nothing unreadable).
4. Resolve the static-call class to a `Model` subclass; build/register via
   `$this->registry->qualifyKey($this->modelToSchema->build($modelClass))`; wrap a
   `200 application/json` `$ref` response (reuse the exact `jsonResponse()` shape from
   `EloquentModelResponseResolver`).
5. Any `Throwable` from the reader → log + null (mirror #18's "any exception → log + null").

### Call-shape matching (the Tier-1 whitelist)

The **class must be statically resolvable** — that is the whole point. Match a `PhpParser\Node\Expr\StaticCall`:
- `$call->class instanceof Name` (a class reference, NameResolver-resolved to an FQCN), **not** an
  `Expr` (`$class::find()` is Tier-2 — degrade).
- method name (lowercased) in `['find', 'findorfail', 'firstorfail']`.
- the resolved class is `class_exists()` and `is_a($class, Model::class, true)`.

**Returned-expression discipline** (mirror `InlineJsonResponseResolver::findJsonCall`): consider only
`return <expr>;` statements, prefer the first returned match. A `find()` assigned to a variable and
not directly returned is **not** matched (no dataflow) — degrade with a NOTICE only if a `find`-shaped
call exists but isn't a direct return (so the author learns why nothing was inferred); a body with no
such call at all is silent.

**Conditional policy:** use `ConditionalContextPolicy::SkipConditionalContexts` — same reasoning as
the *primary*-slot json scan: a model return inside a conditional branch is not unambiguously the
canonical success response. (Contrast the abort/error scans, which include conditionals because an
error branch is exactly worth documenting. The success slot wants the straight-line return.) A
`find()` found only conditionally → NOTICE, return null.

**Degrade cases → NOTICE + null** (per epic #5; reuse a `note()` helper like the json resolver):
- dynamic class (`$model::find()`), wrapped result (`new UserResource(User::find($id))`,
  `response()->json(User::find($id))` — these belong to other resolvers/attributes),
- `find()` not in a `return`, or only in a conditional branch,
- method name outside the whitelist (`->where(...)->first()`, `User::all()` — not this issue).

**`find` vs `findOrFail` nullability note:** `find()` can return null (no 404), `findOrFail`/`firstOrFail`
throw. The success *schema* is the same model `$ref` either way; nullability of the 200 body is not
modeled here (consistent with `EloquentModelResponseResolver`, which already accepts `?Model` without
emitting a nullable schema). So all three map to the same 200 `$ref` — no branching needed.

### Registration

`src/Plugins/Core/CorePlugin.php`: add `addPrimaryResponseResolver(FindReturnModelResponseResolver::class)`.
**Ordering matters** (first non-null wins in the primary-response chain). Current order:
`Paginator → EloquentModel(reflection) → … → InlineJson(body-scan, last)`. Place the new resolver
**alongside the body-scan tier** — after `EloquentModelResponseResolver` (Tier-0 reflection stays
authoritative) and **before `InlineJsonResponseResolver`** (a `return User::find($id)` is a model
return, not a `response()->json()` literal; the json resolver won't match it anyway, but ordering the
model-return scan first keeps intent clear). Verify the exact slot against the live `register()` body.

### #168 interplay (compose, don't conflict)

`return User::findOrFail($id)` now yields **both**:
- a **404** error response — from `FindOrFailErrorContributor` (#168, PR #349, in flight), via the
  `ErrorResponseContributor`/`ErrorResponseInferenceStage` machinery, and
- a **200 User schema** — from this resolver, via the `PrimaryResponseResolver` chain.

These are **different machineries** that run in different stages and write different status keys —
they compose cleanly, no dedup needed. The plan/PR should add an end-to-end test asserting *both* a
404 and a 200-with-User-`$ref` appear on the same operation for a `findOrFail` return, to lock in the
composition. (If #349 hasn't merged when this is implemented, the 404 half of that assertion may need
to wait — note the dependency; the 200 half stands alone.)

### Tests (TDD — failing test first)

**Create** `tests/Unit/Plugins/Core/Resolvers/FindReturnModelResponseResolverTest.php` (`group('openapi')`),
descriptors via `tests/Support/ActionDescriptorFactory.php` + fixture controllers/models:
- `return User::find($id)` (untyped return) → 200 with `$ref` to the User component.
- `return User::findOrFail($id)` → 200 `$ref`.
- `return User::firstOrFail()` → 200 `$ref`.
- Return type already a `Model` subclass → null (defers to `EloquentModelResponseResolver`).
- Dynamic class `$class::find($id)` → null + NOTICE.
- Wrapped `return new UserResource(User::find($id))` → null + NOTICE.
- `find()` assigned then returned via variable (`$u = User::find($id); return $u;`) → null (no dataflow).
- `find()` only inside an `if` branch → null + NOTICE (SkipConditionalContexts).
- Non-Model class `Foo::find()` (class exists, not a Model) → null.
- Whitelist boundary: `User::all()` / `->where()->first()` → null.
- `method === null` → null.

**Extend E2E** — add a `find()`-returning controller to the **`vanilla`** example flavor through
`tests/Feature/ExamplesTest.php` (snapshot byte-match + 3.1 validation + lint-clean), matching #18's
E2E precedent. Plus a focused `tests/Feature/` test asserting the **#168 composition** (404 + 200 on
the same `findOrFail` operation).

### Observable changes → docs + changelog

- `CHANGELOG.md` `[Unreleased]`: "Infer the success-response model schema from a returned
  `Model::find()`/`findOrFail()`/`firstOrFail()` call in the controller body (Tier-1), feeding the
  existing Eloquent model→schema reader."
- `docs/` response-inference page (the one #18 updated — locate via `docs/README.md`): add the
  `find()`/`findOrFail()` return as a model-schema source; note `#[ResponseField]`/`#[Response]`
  remain the escape hatch and that the scan only matches a directly-returned static call.

### Controversy assessment

**Auto-merge-eligible.** No `Contracts/**` change, no config key, no new attribute, no stage-order
change, no public-signature change. One new Core `PrimaryResponseResolver` + one registration line,
feeding the existing #18 reader — exactly the shape the lead flagged as non-controversial.
Generation-affecting → survey applies. (Soft dependency on #349 only for the 404-half of the
composition E2E assertion; the feature itself is independent.)

Verify: `composer test`, `vendor/bin/pint --test`, `composer lint` (and the `openapi` group / snapshot
group if the example snapshot is touched).


Radiergummi:

New body-scanning `FindReturnModelResponseResolver` (sibling to `InlineJsonResponseResolver`) recovers the model class from a directly-returned `Model::find()/findOrFail()/firstOrFail()` static call and feeds #18's existing `EloquentModelToSchema` reader for the 200 `$ref`. Return-type guard defers to the Tier-0 `EloquentModelResponseResolver`; SkipConditionalContexts (success slot, straight-line return); degrade+NOTICE on dynamic class / wrapped / variable-assigned / conditional. Composes with #168 (`findOrFail` → 404 + 200 on the same op). Auto-merge-eligible — no Contracts/config/attribute/stage-order changes.

Radiergummi:
🤖 **[lead]** plan posted (PR #356) → plan-review. New Core `FindReturnModelResponseResolver` (body-scanning PrimaryResponseResolver, sibling to InlineJsonResponseResolver): matches a **directly-returned** Model::find/findOrFail/firstOrFail static call → resolves Model subclass → feeds #18's EloquentModelToSchema for the 200 $ref. Return-type guard defers to the Tier-0 EloquentModelResponseResolver (reflection stays authoritative; only scans untyped/builtin/HTTP-response returns). SkipConditionalContexts (success slot wants the straight-line return). Degrade+NOTICE on dynamic class / wrapped / variable-assigned / conditional-only. Reviewer focus: (1) the return-type-guard precedence vs Tier-0 resolver is correct and registered in the right order; (2) SkipConditionalContexts is the right policy for the success slot (vs the error scans' Include); (3) degrade cases complete; (4) the #168 interplay — note **#349 is already MERGED**, so the 404+200 compose E2E can assert both halves now (no soft dependency left); (5) Tier-1 lowest-viable, surgical scope. Auto-merge-eligible (no Contracts/config/attribute/stage-order). Survey applies.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved.** Tier-1 is the lowest viable tier, the design is surgical and auto-merge-eligible, and all five focus items check out against source. Minor notes below; none blocking.

**1. Return-type-guard precedence + registration order — correct.** The plan's guard mirrors `InlineJsonResponseResolver::returnTypeAllowsBodyScan` (`:191`), which I verified returns true only for: null return type, builtin, or an `HttpFoundationResponse` subclass; **false** for any other named type (so a `Model`/`Collection`/`?Model` return → guard false → this resolver defers) and for union/intersection (non-`ReflectionNamedType` → false). That is exactly right: a `Model`-subclass or `Collection<Model>` return is a non-builtin named type, so the guard hands it to the Tier-0 `EloquentModelResponseResolver`, which is registered **earlier** (`CorePlugin.php:70`, vs the new resolver slotting between `:70` and `InlineJson :75`). First-non-null-wins in the chain means no double-resolution. The plan's own `?User` example is therefore already Tier-0 (the guard correctly excludes it) — the plan flags this with "already Tier-0? see guard," which is accurate.

**2. `SkipConditionalContexts` is the right policy for the success slot — confirmed.** Mirrors the *primary*-slot json scan (`findJsonCall` uses `SkipConditionalContexts`), and is correctly the opposite of the error scans (`AbortErrorContributor`/`FindOrFailErrorContributor` use `IncludeConditionalContexts`). The reasoning holds: a model return inside an `if` branch is not unambiguously the canonical 200 body, whereas an error branch is exactly the outcome worth documenting. Consistent with the codebase's established split.

**3. Degrade cases — complete.** Dynamic class (`$class::find()` — `$call->class` is an `Expr`, not a `Name`), wrapped (`new UserResource(...)`, `response()->json(...)`), variable-assigned-then-returned (no dataflow), conditional-only, non-whitelist methods (`->where()->first()`, `User::all()`), and non-Model class (`class_exists` + `is_a(Model)` guard). The NOTICE-vs-silent split is well-judged: NOTICE only when a `find`-shaped call exists but isn't a direct return (the author learns why nothing was inferred), silent when no such call exists at all (no noise on unrelated bodies). Returned-expression discipline mirrors `findJsonCall`'s returned-first preference — good reuse of the established pattern.

**4. #168 interplay — #349 is MERGED (confirmed: `487fb07` on `origin/main`, `FindOrFailErrorContributor` present).** So the compose E2E can and **should** assert **both** halves now — a `return User::findOrFail($id)` op carries a 404 (error contributor, `ErrorResponseInferenceStage`) **and** a 200 with the User `$ref` (this resolver, `PrimaryResponseResolver` chain). Different machineries, different status keys, no dedup. The plan's "soft dependency / 404 half may need to wait" hedge is now moot — **drop the hedge and assert both halves** in the focused composition test. (Update the controversy line too: no remaining dependency on #349.)

**5. Tier-1 lowest-viable + surgical — yes.** Reuses `MethodBodyScanner`/`StatementNodeFinder`/`ConditionalContextPolicy` + #18's `EloquentModelToSchema` (verified `build(string): string`, then `registry->qualifyKey(...)` — exactly the `EloquentModelResponseResolver::refKey` pattern). One new Core resolver + one registration line. No `Contracts/**`, no config key, no attribute, no stage-order change. Auto-merge-eligible. ✓

**Minor (coder notes, non-blocking):**
- The plan says "reuse the exact `jsonResponse()` shape from `EloquentModelResponseResolver`" — that method is **private** there, so it must be **replicated** (3 lines: `200`/`OK`/`application/json` via `MediaType::Json->schema($schema)`), not shared. Don't extract a helper for this; a private copy matches the sibling's own pattern and keeps scope surgical.
- `EloquentModelToSchema::build()` already does the `buildOnce`/cycle-guard internally and returns the key string — the new resolver just calls `build()` then `qualifyKey()`, same as `refKey()`. No need to call `buildOnce` directly (the plan's step 4 wording "build/register via qualifyKey(build(...))" is right; just don't add a redundant `buildOnce` wrapper).
- E2E: per #18's precedent, the `vanilla` example flavor + `ExamplesTest` snapshot path is the right home; ensure the new fixture's snapshot is regenerated (byte-match + 3.1 validate + lint-clean).

Verdict: **approved, auto-merge-eligible.** Proceed to coding; assert both 404+200 halves of the #168 composition (no longer gated).

Radiergummi:
🤖 **[lead]** plan-review **approved, auto-merge-eligible** → coding. Brief (plan + reviewer-1's coder notes):
- New Core `FindReturnModelResponseResolver` (body-scanning PrimaryResponseResolver, sibling to InlineJsonResponseResolver): match a directly-returned `Model::find()/findOrFail()/firstOrFail()` static call → resolve Model subclass → feed #18's `EloquentModelToSchema` for the 200 $ref.
- **Return-type guard** mirrors `InlineJsonResponseResolver::returnTypeAllowsBodyScan` (:191) — fire only on null/builtin/HttpFoundationResponse returns; Model/Collection/?Model/unions defer to the Tier-0 `EloquentModelResponseResolver`. **Register between CorePlugin:70 (EloquentModelResponseResolver) and :75 (InlineJson)** so first-non-null-wins gives no double-resolution.
- `SkipConditionalContexts` (success slot wants the straight-line return).
- Degrade+NOTICE: dynamic class / wrapped / variable-assigned (no dataflow) / conditional-only / non-whitelist / non-Model.
- **#349 is MERGED** — drop the soft-dependency hedge: the E2E asserts BOTH the 404 (#168) and the 200 model schema; update the controversy line (no remaining dependency).
- Coder notes: `jsonResponse()` is private in the sibling → **replicate the 3 lines, don't extract**; `build()` already calls buildOnce → no wrapper; E2E via a vanilla flavor + ExamplesTest snapshot (per #18 precedent).
- Tier-1; auto-merge-eligible (no Contracts/config/attribute/stage-order). Survey applies.

Radiergummi:
🤖 **[coder]** implemented; local gates green (`composer check`: 2632 passed, PHPStan clean, Pint clean); rebased onto current main (0435bfd, clean); pushed adf1edc. Remote CI all green incl. swagger-php 5.8, PHP 8.5, snapshot, quality, coverage.

- New `FindReturnModelResponseResolver` matches a **directly-returned** `Model::find()/findOrFail()/firstOrFail()` static call → resolves the Model subclass → feeds #18's `EloquentModelToSchema` for the 200 $ref. Return-type guard mirrors `InlineJsonResponseResolver::returnTypeAllowsBodyScan`; registered between `EloquentModelResponseResolver` and `InlineJsonResponseResolver`. `SkipConditionalContexts`.
- **Directly-returned discipline:** the lookup must be the returned expression itself — a wrapped result (`new UserResource(User::find())`, `response()->json(User::find())`), a variable-assigned lookup, a dynamic class (`$class::find()`), or a conditional-only call all degrade with a NOTICE; non-Model / non-whitelist stay silent.
- **#168 composition:** E2E (vanilla `/find/flights/{flight}` + a focused feature test) asserts **both** the 404 and the 200 model $ref on the same `findOrFail` op. `jsonResponse()` replicated (3 lines), not extracted.

Ready for review.

Radiergummi:
🤖 **[lead]** coding done (adf1edc; full CI green incl. swagger-php 5.8 + Snapshot; clean rebase on 0435bfd) → code-review. **Auto-merge candidate → TWO independent reviewer approvals required.** Focus: (1) the **directly-returned discipline** — the lookup must BE the returned expression, not nested/wrapped (try to break it: `return [User::find()]`, `return response()->json(User::find())`, ternary, assigned-then-returned); (2) return-type guard defers correctly to the Tier-0 EloquentModelResponseResolver (registered between :70 and :75, first-non-null-wins, no double-resolution); (3) degrade split (NOTICE on dynamic/wrapped/variable/conditional, silent on non-Model/non-whitelist); (4) the #168 composition E2E asserts BOTH 404+200 with no collision (different status keys); (5) jsonResponse replicated not extracted; docs (auto-derivation.md limitation→positive) + CHANGELOG; test gaps. Responsible coder: coder-3.

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer-1, independent A): **approved.** One non-blocking test-gap finding.

Full adversarial pass on `adf1edc`. The directly-returned discipline holds against every break-case, the guard/ordering/degrade logic is correct, and the #168 composition asserts both halves.

**Directly-returned discipline — tried to break it, couldn't.** `directlyReturnedLookupCall` iterates only top-level `Return_` statements and matches only when `$statement->expr instanceof StaticCall` (the call *is* the whole returned value). I verified each evasion degrades correctly:
- **Wrapped** `return new UserResource(User::find())` / `return response()->json(User::find())` → expr is `New_`/`MethodCall`, not matched → NOTICE (tests `wrapped` + a `New_` path covered by the wrapped/notice assertion).
- **Variable-assigned** `$u = User::find(); return $u;` → return expr is a `Variable` → not matched → NOTICE (test `assignedThenReturned`).
- **Conditional-only** `if (...) { return User::find(); }` → the inner return is nested in `If_`, never a top-level `Return_` → skipped → NOTICE (test `conditionalOnly`).
- **First-class callable** `User::find(...)` → guarded by `!isFirstClassCallable()`.
- **Dynamic class** `$class::find()` → `$call->class` is an `Expr` not `Name` → `resolveModelClass` returns null → NOTICE (test `dynamicClass`).

**Return-type-guard deferral + registration order — correct.** `returnTypeAllowsBodyScan` is a faithful copy of `InlineJsonResponseResolver`'s (true for null/builtin/`HttpFoundationResponse`; false for any other named type incl. Model/Collection/Data, and for union/intersection). A typed `?User` return → guard false → defers to the Tier-0 `EloquentModelResponseResolver` (registered earlier at `CorePlugin.php:70`); the new resolver slots between it and `InlineJsonResponseResolver` (`:75`), with a clear comment. First-non-null-wins → no double-resolution. Test `defers a typed Model return to the reflection resolver` pins it.

**Degrade NOTICE/silent split — precisely tested.** NOTICE fires only when a lookup-shaped call exists but isn't a clean direct return; a body with no lookup is silent. The subtle case — a directly-returned, whitelisted `NonModel::find()` whose class resolves but fails the `is_a(Model)` guard — is **silent** (no notice), and has a dedicated test (`keeps a non-Model class lookup silent`). That's the right call (the author wrote a real `find()` return; we just can't model a non-Model). `noteUnmatchedLookup` uses `IncludeConditionalContexts` so it catches conditional/ternary/wrapped lookups for the advisory.

**#168 composition — both halves asserted (no hedge).** `FindReturnModelResponseTest::composes the findOrFail 200 model schema with the inferred 404` asserts `toHaveKeys(['200','404'])` with the 200 carrying the Article `$ref`. Good isolation: the route param is typed `string $article` (not `Article`), so the 404 comes from `FindOrFailErrorContributor`'s body scan, **not** route-model binding — proving the body-scan composition specifically. Different machineries, different status keys, no collision. ✓

**`jsonResponse()` replicated, not extracted** — exactly as the plan-review advised (the sibling's is private); the docblock even explains why. Scope stayed surgical.

**Docs + CHANGELOG — thorough.** `docs/auto-derivation.md` adds the new source to the response-body summary row, a dedicated subsection (honest about the directly-returned constraint and the degrade cases incl. ternary), the #168 composition note, **and removes the now-stale "Model::find() not read" known-limitation**. CHANGELOG `[Unreleased]` entry present. Vanilla example (`FindFlightController` → snapshot showing 200 `Flight` + 404) is the real-world proof, byte-matched (Snapshot job green).

**Finding 1 (non-blocking — test gap).** The docs explicitly promise a `ternary` lookup degrades (not read), and the resolver handles it correctly — I verified empirically: `return $c ? User::find($id) : null` is not a direct match (`Return_` expr is a `Ternary`), and `noteUnmatchedLookup`'s Include-scan finds the lookup → emits the NOTICE. But the unit suite covers only the `if` form (`conditionalOnly`), not a ternary. Recommend adding one case to `FindReturnFixtureController` — `ternary(string $id, bool $flag): mixed { return $flag ? User::find($id) : null; }` — asserting `null` + one notice, to pin the documented branch. Not a blocker: behavior is correct and CI is green; it's a resilience-of-the-suite addition.

Full CI green on `adf1edc` (all matrix legs incl. swagger-php 5.8, Snapshot, quality/PHPStan, Pint, coverage). Approved (independent A).

Radiergummi:
🤖 **[reviewer]** code-review round 1 (reviewer-2, independent B): **approved, no concerns.**

Full adversarial pass on `adf1edc` (not rubber-stamping reviewer-1's A). Ran a clean worktree: the 2 new suites + ExamplesTest snapshot pass (**39 tests**), the whole `--group openapi` is green (**2363 tests**), Pint clean on the new file. All five focus points verified against source:

**1. Directly-returned discipline — correct, and tighter than a literal reading of the plan.** Rather than `StatementNodeFinder::findFirst` under `SkipConditionalContexts` (which descends into argument lists and would match a *wrapped* call), `directlyReturnedLookupCall()` iterates top-level statements and matches only when `$statement instanceof Return_` **and** `$statement->expr` *is* the `StaticCall` (`FindReturnModelResponseResolver.php:160-176`). I traced the AST for each break-it case the lead listed:
   - `return [User::find($id)]` → `expr` is `Array_` → not matched → NOTICE (the `IncludeConditionalContexts` walk in `noteUnmatchedLookup` finds the nested call). ✓
   - `return response()->json(User::find())` / `return new UserResource(User::find())` → `expr` is `MethodCall`/`New_` → not matched → NOTICE (tested as `wrapped`). ✓
   - `$cond ? User::find() : null` → `expr` is `Ternary` → not matched → NOTICE. ✓
   - `$u = User::find(); return $u;` → return `expr` is `Variable` → not matched → NOTICE (tested as `assignedThenReturned`). ✓
   - `if (...) { return User::find(); }` → the `if` is an `If_` stmt, not a top-level `Return_` → not matched → NOTICE (tested as `conditionalOnly`). ✓

**2. Return-type-guard deferral — correct, no double-resolution.** `returnTypeAllowsBodyScan()` mirrors `InlineJsonResponseResolver`: true only for null/builtin/`HttpFoundationResponse`; a `Model`/`Collection`/`?Model`/union return is a non-builtin named (or non-named) type → false → defer. Registered between `EloquentModelResponseResolver` (`CorePlugin:71`) and `InlineJsonResponseResolver`; first-non-null-wins means the Tier-0 reflection resolver (earlier) stays authoritative. Tested (`typedModelReturn` → null).

**3. Degrade split — precise.** NOTICE on dynamic class / wrapped / variable-assigned / conditional-only; **silent** on non-Model (directly-returned whitelist call whose class resolves but fails the `is_a(Model)` guard — `resolveModelClass` returns null and `$call->class instanceof Name` is true, so no note) and non-whitelist (`User::all()`). The non-Model-silent branch is the subtle one and it has an explicit `records->toBeEmpty()` test. Reader `Throwable` → log + null mirrors #18.

**4. #168 composition — clean, no collision.** Asserted both in the focused feature test (200 `Article` `$ref` + 404 on the same `findOrFail` op) and the byte-exact `examples/vanilla/openapi.yaml` snapshot (`/api/find/flights/{flight}` carries a 200 `Flight` `$ref` *and* a 404 `NotFound`). Different stages/status keys → compose without dedup. #349 is merged so both halves are live.

**5. `jsonResponse()` replicated (3 lines), not extracted** — documented why on the method; keeps scope surgical, as agreed.

Conventions: strict-types, `final readonly`, `#[Scoped]`, no abbreviations, ≤100 cols; `@internal` is consistent with the semantic sibling it feeds (`EloquentModelResponseResolver` carries it too). Docs flip the #97 limitation into a positive "From a returned find()/findOrFail()" section that honestly states the directly-returned discipline, the degrade cases, the 404 composition, and the escape hatch; the stale limitation bullet is removed. CHANGELOG accurate. No `Contracts/**`, config, attribute, or stage-order change.

**Optional, non-blocking:** an explicit `return [User::find($id)]` (array-wrapped) unit case would pin a distinct AST shape (`Array_`) — though it rides the identical `!($expr instanceof StaticCall)` path already covered by `wrapped`, so it's a nice-to-have, not a gap. **Approved.**

Radiergummi:
🤖 **[lead]** **two independent approvals** (reviewer-1 A + reviewer-2 B), both non-blocking test-gap findings → coder-3 adding ternary + array-wrap degrade tests concurrently (test-only, generation-neutral). CI green incl. 5.8 on adf1edc. → **survey Layer A**. Surveyor: rebaseline on current main 0435bfd (cached baseline matches — main hasn't moved since #350), candidate on the #356 branch. Additive change (find/findOrFail/firstOrFail return → 200 model $ref on untyped-return ops). Expect responses/schema counts equal-or-up, 0 regressions, 0 crashes. Report corpus delta. Pass → I merge.

Radiergummi:
🤖 **[coder]** added ternary + array-wrap degrade tests; pushed acf040c. Both return shapes (`return $cond ? Model::find() : null`, `return [Model::find()]`) assert null + 1 NOTICE — closing the documented-but-untested branches both reviewers flagged. Local `composer check` green (2634 passed); remote CI all green incl. swagger-php 5.8, PHP 8.5, snapshot, quality, coverage. No re-review needed.

Radiergummi:

Baseline `main` @ `0435bfd` (reused cached baseline — main unchanged since #350) vs candidate `adf1edc` (rebased onto current main). All 7 deterministic metrics (paths / operations / apiOperations / responseSchemas / documentedResponses / requestBodies / componentSchemas) are **exactly equal** on every app; **0 new crashes** (gen_exit=0 corpus-wide).

**Corpus delta of newly-resolved 200 model schemas: 0** (responseSchemas +0, documentedResponses +0, componentSchemas +0 across all 11 apps).

This is correct conservative behavior, not a silent no-op — verified: the corpus does contain `->find()`/`->findOrFail()` returns (InvoiceNinja 14, Koel 2, Coolify/Vito/Lychee 1 each), but none sit in an **API route action** as a straight-line `return $model->findOrFail($id)`. The InvoiceNinja matches are all in non-route classes (`PaymentDrivers/`, `Livewire/`) and are `Model::withTrashed()->find(...)` query-builder chains — outside the introspected route surface and outside the Tier-1 shape the resolver targets. So the resolver fires on exactly the shape it should; this OSS corpus just doesn't exercise it. Its value is covered by the PR's own `examples/vanilla` + feature tests.

No regression, no crash — clean from the deterministic gate's perspective.

PR #361, opened 2026-06-15
feat(#36): model factory definition() seeds model-schema examples

### LINKED ISSUE #36: Model factories → response/request examples
## Summary

When a model has a Laravel factory, use that factory's `definition()` as the example payload for the model's schema, rather than relying solely on synthesised Faker values.

## Motivation

A model's `factory()->definition()` is a hand-curated, realistic example payload — exactly the kind of data the generated document should show consumers. Today examples come only from `FakerExampleSynthesiser`, which produces plausible-but-generic values seeded per field. Factory definitions are richer and already maintained by the app, so reusing them raises example quality at essentially no authoring cost. This is **Tier 0/1** polish: nice-to-have, not load-bearing.

## Proposed approach

Extend the example source in `src/Support/Extraction/FakerExampleSynthesiser.php` so that, when a model exposes a factory, the factory definition feeds the schema's example alongside or in place of the Faker-synthesised values.

A factory `definition()` is an array literal (Tier 1 — readable by reflection), but it commonly calls `fake()`/Faker at runtime, so a static read of the literal will not resolve those values. The safer route is to instantiate the factory and invoke its definition (Tier 0-ish) so runtime Faker calls resolve to concrete values. Note the determinism constraint inherited from the synthesiser: `FakerExampleSynthesiser` reseeds Faker deterministically per call (from `openapi.examples.faker_seed` + the field name), so factory-sourced values that call Faker must not reintroduce nondeterminism into the byte-exact snapshot output.

## Acceptance criteria

- A model with a factory yields schema examples sourced from the factory definition; a model without a factory still falls back to the existing Faker synthesis path unchanged.
- Add coverage to `tests/Unit/Support/Extraction/FakerExampleSynthesiserTest.php` (the synthesiser's existing unit test), defining a model + factory fixture and asserting the example reflects the factory definition.
- Generated examples remain deterministic so the byte-exact `snapshot` group in `tests/Feature/ExamplesTest.php` stays green.

## References

Parent: #67 — EPIC: Shared components, metadata & examples (shared example/`ComponentSchemaRegistry` context and the `FakerExampleSynthesiser` deterministic-seed mechanism live there).

Verify: `vendor/bin/pest tests/Unit/Support/Extraction/FakerExampleSynthesiserTest.php`

Radiergummi:
Closes #36

### What & why

When an Eloquent model exposes a Laravel factory, source the model schema's per-property
**examples** from `factory()->definition()` — a hand-curated, app-maintained realistic payload —
instead of (or above) the generic Faker synthesis. Tier-0/1 polish: nice-to-have, not load-bearing.

### Key architectural finding (verify-and-rely)

The issue says "extend example sourcing in `FakerExampleSynthesiser`", but on inspection that class
is **per-field** and is wired only into request/Data-class schema builders
(`SchemaFromFormRequest`, `SchemaFromDataClass`, `InlineValidationRequestSchemaResolver`). **Model
schemas do not currently carry examples at all** — `EloquentModelToSchema::schemaFor()` builds
`OA\Property` objects with no `example` set, and never consults the synthesiser.

A factory `definition()` is a **whole-model associative array** (`['name' => fake()->name(), ...]`),
not a per-field synthesis. So the right seam is **not** inside `FakerExampleSynthesiser` (which would
distort its narrow per-field contract). Instead:

- Add a small, plugin-agnostic reader in `src/Support/Extraction/` —
  `ModelFactoryExampleReader` — that, given a `class-string<Model>`, resolves the model's factory,
  invokes `definition()`, and returns `array<string, mixed>` of column → concrete example value (or
  `[]` when the model has no factory / instantiation throws).
- `EloquentModelToSchema::schemaFor()` consults it once per model, then sets `$property->example`
  on each built property whose name has a factory value. This is the single, surgical wiring point.

This keeps `src/Support/` plugin-agnostic (models + factories are vanilla Laravel = Tier 0 metadata),
and leaves `FakerExampleSynthesiser` untouched. I will note this deviation from the issue's literal
"extend FakerExampleSynthesiser" wording as a PR comment (per CLAUDE.md: record plan deviations).

### Tier

**Tier 0-ish** (the issue's framing). `definition()` is an array literal but commonly calls
`fake()` at runtime, so a static AST read won't resolve values — instead we **instantiate the
factory and invoke `definition()`** so Faker calls resolve to concrete values. This is reflection +
runtime invocation of app-owned metadata, not dataflow analysis. No tier escalation.

### Factory resolution mechanism

`Model::factory()` is provided by the `HasFactory` trait, **not** `Model` itself — confirmed via
reflection (`Illuminate\Database\Eloquent\Model` has no `factory()` method; `HasFactory` provides
`newFactory()`/`factory()`). So detection is: the model class has a callable static `factory()`
(`method_exists($modelClass, 'factory')`). The reader calls `$modelClass::factory()->definition()`
inside a `try { } catch (Throwable)` — factory discovery, constructor side effects, or a
`definition()` that touches the DB/config can throw; on any failure return `[]` and **log one
warning** (matching the `EloquentModelToSchema` degrade-and-log convention).

Only **scalar** definition values (string/int/float/bool/null) become examples. A definition value
that is itself a factory, a relationship closure, a `Closure`, or an array/object is skipped — those
aren't representable as a simple scalar property example and would risk recursion. (Guard with an
`is_scalar() || is_null()` filter.)

### Determinism constraint (MUST cover)

`FakerExampleSynthesiser` reseeds Faker deterministically per call from
`(openapi.examples.faker_seed, fieldName)` so the byte-exact `snapshot` group in
`tests/Feature/ExamplesTest.php` stays green. A factory `definition()` calling `fake()` uses the
**global** Faker instance (Laravel's container-bound `Faker\Generator`), whose seed we do not
control per-call. **Two options, plan picks (a):**

(a) **Seed the global Faker before invoking `definition()`.** The reader, before calling
`definition()`, seeds the app's bound `Faker\Generator` via `mt_srand`/`$faker->seed()` from the
configured `openapi.examples.faker_seed` (passed into the reader, mirroring how the synthesiser gets
its seed in `OpenApiServiceProvider`). This makes factory-sourced `fake()` values reproducible across
runs. When `faker_seed` is `null` (config default opt-out), factory examples are **disabled** (would
be non-deterministic) — fall back to no example, same as today.

(b) rejected: post-hoc normalisation — too fragile.

The determinism test is the gating test: generate the same model's schema twice, assert identical
example values; and the existing `snapshot` group must stay byte-identical after a fixture is added.

### Files

**Create:**
- `src/Support/Extraction/ModelFactoryExampleReader.php` — `#[Scoped]`, `@internal`. Constructor
  takes `?int $seed` + `LoggerInterface`. One public method
  `examplesFor(class-string<Model> $modelClass): array<string,mixed>`, memoised per model. Returns
  `[]` when: no `factory()` method, `$seed === null`, or any `Throwable` during resolution.

**Modify:**
- `src/Support/Extraction/EloquentModelToSchema.php` — inject `ModelFactoryExampleReader`; in
  `schemaFor()`, fetch `$examples = $reader->examplesFor($modelClass)` once, and when building each
  property set `$property->example = $examples[$name]` if the key exists and value is scalar/null.
  (Also apply in `propertyFor()` for the Resource-key path? — **No**, keep scope to `schemaFor()`
  per the issue's "model's schema example"; note this in the PR. `propertyFor()` is the
  resource-driven single-property path and out of scope.)
- `src/OpenApiServiceProvider.php` — register `ModelFactoryExampleReader` with the same
  `openapi.examples.faker_seed`-derived seed used for `FakerExampleSynthesiser` (lines ~403-408).
- `config/openapi.php` — no new key needed; reuses `examples.faker_seed` and (optionally) gate on
  `examples.synthesise`. **Decision:** gate the factory path on `examples.synthesise` too, so the
  master switch turns off all auto-examples consistently. State in PR.

**`tests/Unit/Support/Extraction/` — new `ModelFactoryExampleReaderTest.php`:**
- factory present → `examplesFor()` returns the definition's scalar values.
- model without `factory()` → returns `[]`.
- `seed === null` → returns `[]` (determinism opt-out).
- `definition()` that throws → returns `[]` + warning logged (use a fixture factory whose
  `definition()` throws).
- non-scalar definition value (nested array / factory ref) → that key absent from the result.
- **determinism**: two calls on a `fake()`-using factory yield identical values.

**`tests/Unit/Support/Extraction/EloquentModelToSchemaTest.php`** (if exists; else add to the model
schema feature test): model with a factory → built property carries `example` from the factory;
model without a factory → no `example` (unchanged behaviour).

**Fixtures (new):**
- `tests/Fixtures/Models/<X>` model `use`ing `HasFactory` with a `newFactory()` (or a discoverable
  factory) whose `definition()` calls `fake()` (covers the determinism path) and includes one
  non-scalar value (covers the skip filter).
- a factory whose `definition()` throws (covers degrade).

**Snapshot guard:** `tests/Feature/ExamplesTest.php` `snapshot` group MUST stay green. Since adding a
factory to an *existing* snapshot fixture model would change the committed YAML, **prefer adding a
new fixture used only by the unit/feature tests, not by the snapshot flavors** — OR, if a snapshot
flavor model gains a factory, regenerate and commit the snapshot and call it out as an intentional
diff. Coder picks the lower-churn route; default to a dedicated fixture outside the snapshot set.

### Verification

```
vendor/bin/pest tests/Unit/Support/Extraction/ModelFactoryExampleReaderTest.php
vendor/bin/pest tests/Feature/ExamplesTest.php --group snapshot
composer test
vendor/bin/pint --test
composer lint
```

### Docs / CHANGELOG

- `docs/auto-derivation.md` → "Eloquent model response schemas" section (around the worked
  `Article` example, ~line 164): add a short subsection noting that when a model has a factory, its
  `definition()` seeds per-property examples (deterministic via `examples.faker_seed`; disabled when
  the seed is null).
- `CHANGELOG.md [Unreleased]` "Added": model schemas now source per-property examples from the
  model's factory `definition()` when one exists, deterministically seeded.

### Survey gate

**Generation-affecting** (area:core, examples in output) → survey gate WILL run. Adding examples to
model schemas is a real output delta on any dogfood app whose models have factories — the survey
baseline will move. That is **expected and intended**; reviewer should confirm the delta is only
added `example:` keys on model properties, deterministic, and nothing else regressed.

### Controversy

Mild. No `src/Contracts/**` change, no stage-order change, no new config key (reuses
`examples.faker_seed`/`examples.synthesise`), no public-signature change. The deviation from the
issue's literal "extend FakerExampleSynthesiser" wording (new reader instead) is the one judgment
call — documented as a PR comment. The global-Faker seeding for determinism is the subtle part; the
determinism test pins it. Reviewer should sanity-check the snapshot stays byte-identical.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,7 @@ All notable changes to this project are documented here.
 - A `findOrFail()` / `firstOrFail()` lookup in the controller body infers a 404 response, deduped against the route-model-binding 404. (#168)
 - Infer error responses from non-2xx `response()->json([...], <4xx/5xx>)` literals in the controller body (Tier-1), carrying the literal body schema inlined per operation; falls back to the configured error envelope when only the status is statically readable. (#238)
 - Internal: the API Resource return-expression reader's paginate-method whitelist now routes through the shared `PaginatorKind::fromPaginatingMethod()` enum instead of a duplicated local constant. (#354)
+- Eloquent model schemas seed their per-property examples from the model's Laravel factory `definition()` (scalar values only), reseeded deterministically per model from `openapi.examples.faker_seed`; disabled when example synthesis is off or the seed is null. (#36)
 
 ## [0.1.0] - 2026-05-18
 
--- a/docs/auto-derivation.md
+++ b/docs/auto-derivation.md
@@ -219,6 +219,22 @@ Article:
 
 `password` is absent because it is in `$hidden`.
 
+### Examples from model factories
+
+When a model exposes a Laravel factory (the `HasFactory` trait), the generator
+invokes the factory's `definition()` and uses its **scalar** values as the
+per-property `example` in the model's schema. A factory definition is a
+hand-curated, app-maintained realistic payload, so reusing it raises example
+quality at no extra authoring cost. Only scalar (and `null`) values are taken;
+nested arrays, relationship closures, and factory references are skipped.
+
+Factory `definition()` typically calls `fake()`, so the values are reseeded
+deterministically from `openapi.examples.faker_seed` (mixed with the model
+class) before each read — reads are order-independent and stable across runs.
+This path follows the same switches as the rest of example synthesis: it is
+disabled when `openapi.examples.synthesise` is `false` or `faker_seed` is `null`
+(see [Configuration](config.md)).
+
 ### Documenting computed fields
 
 A field in `$appends` without a typed legacy accessor falls back to an
--- a/docs/config.md
+++ b/docs/config.md
@@ -26,7 +26,7 @@ this page is the at-a-glance summary.
 | `security_middleware_map` | Map of custom guard-middleware name → a scheme declared in `security_schemes`. A route carrying a mapped middleware emits that scheme's per-operation requirement (taking precedence over the auto-derived `auth:*`/`scope:*` resolution for that route), so project-specific guards don't make a protected endpoint look public. |
 | `plugins` | Ordered list of `Plugin` class-strings. Ships with `SpatieDataPlugin` and `ApiResourcesPlugin` enabled; `QueryBuilderPlugin` and `FractalPlugin` shipped commented out. |
 | `request_payload_indirection` | Base classes whose constructors are also scanned for Data-class parameters. See [Request bodies → Indirect request payloads](request-bodies.md#indirect-request-payloads). |
-| `examples` | Example synthesis: `synthesise` (default `true`) toggles Faker-generated examples for fields with no authored example; `faker_seed` makes them deterministic. |
+| `examples` | Example synthesis: `synthesise` (default `true`) toggles Faker-generated examples for fields with no authored example and model-factory `definition()` examples for Eloquent model schemas; `faker_seed` makes them deterministic (a `null` seed disables both). |
 | `visibility` | `default` accepts `'public'` (every route documented unless `#[Hide]`) or `'hidden'` (every route excluded unless `#[Expose]`). See [Recipes → Switch between public-default and hidden-default visibility](recipes.md#switch-between-public-default-and-hidden-default-visibility). |
 | `overrides` | Spec-only per-route operation field overrides, keyed by route name or URI glob. See [Operation overrides](#operation-overrides) below. |
 | `routes` | Spec/playground route registration. See below. |
--- a/src/OpenApiServiceProvider.php
+++ b/src/OpenApiServiceProvider.php
@@ -410,6 +410,20 @@ class OpenApiServiceProvider extends ServiceProvider
         );
 
         $this->app->scoped(
+            Support\Extraction\ModelFactoryExampleReader::class,
+            static fn(Container $app): Support\Extraction\ModelFactoryExampleReader
+                => new Support\Extraction\ModelFactoryExampleReader(
+                    // Disabled (null seed) when auto-examples are switched off or no fixed seed is
+                    // configured — factory fake() values would otherwise be non-deterministic.
+                    seed: (bool) (config('openapi.examples.synthesise') ?? true)
+                        && config('openapi.examples.faker_seed') !== null
+                        ? (int) config('openapi.examples.faker_seed')
+                        : null,
+                    logger: $app->make(LoggerInterface::class),
+                ),
+        );
+
+        $this->app->scoped(
             Plugins\Core\Resolvers\DiscriminatedRequestSchemaResolver::class,
             static function (Container $app): Plugins\Core\Resolvers\DiscriminatedRequestSchemaResolver {
                 $registry = $app->make(OpenApiRegistry::class);
--- a/src/Support/Extraction/EloquentModelToSchema.php
+++ b/src/Support/Extraction/EloquentModelToSchema.php
@@ -84,6 +84,7 @@ final class EloquentModelToSchema
         private readonly TypeNodeResolver $typeNodeResolver,
         private readonly DocBlockParser $docBlockParser,
         private readonly LoggerInterface $logger,
+        private readonly ModelFactoryExampleReader $factoryExampleReader,
     ) {}
 
     /**
@@ -589,6 +590,7 @@ final class EloquentModelToSchema
         }
 
         $names = array_values(array_diff($allNames, $hidden));
+        $examples = $this->factoryExampleReader->examplesFor($modelClass);
         $properties = [];
 
         foreach ($names as $name) {
@@ -653,6 +655,16 @@ final class EloquentModelToSchema
             }
 
             $properties[] = new OA\Property(['property' => $name]);
+        }
+
+        // Seed examples from the model's factory definition() (when one exists and is deterministic);
+        // leaves untouched any property the factory does not produce.
+        if ($examples !== []) {
+            foreach ($properties as $property) {
+                if (array_key_exists($property->property, $examples)) {
+                    $property->example = $examples[$property->property];
+                }
+            }
         }
 
         // A property is required when it has a @property/@property-read annotation whose type is
--- a/src/Support/Extraction/ModelFactoryExampleReader.php
+++ b/src/Support/Extraction/ModelFactoryExampleReader.php
@@ -1,0 +1,107 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Support\Extraction;
+
+use Faker\Generator as Faker;
+use Illuminate\Container\Attributes\Scoped;
+use Illuminate\Container\Container;
+use Illuminate\Database\Eloquent\Model;
+use Psr\Log\LoggerInterface;
+use Throwable;
+
+use function array_key_exists;
+use function crc32;
+use function is_scalar;
+use function method_exists;
+
+/**
+ * Reads a model's Laravel factory `definition()` as a source of per-property example values.
+ *
+ * A factory `definition()` is a hand-curated, app-maintained realistic payload — exactly the kind
+ * of data the generated document should show consumers. `Model::factory()` comes from the
+ * `HasFactory` trait (the base `Model` has no `factory()` method), so a model exposes a factory
+ * when `method_exists($modelClass, 'factory')`. The factory is instantiated and its `definition()`
+ * invoked so runtime `fake()` calls resolve to concrete values (Tier 0-ish: reflection + runtime
+ * invocation of app-owned metadata, not dataflow).
+ *
+ * Only scalar (and null) definition values become examples — a value that is itself a factory, a
+ * relationship closure, or an array/object is skipped, as it cannot stand in as a simple property
+ * example.
+ *
+ * Determinism — a factory's `definition()` draws from the container-bound `Faker\Generator`, whose
+ * seed this reader controls. Before **every** `definition()` invocation it reseeds that generator
+ * from `(faker_seed ^ crc32($modelClass))` so reads are order-independent (Faker delegates `seed()`
+ * to the global `mt_srand()`, so any RNG consumer between reads would otherwise drift the state) and
+ * distinct models draw distinct values. When `faker_seed` is null (the config opt-out), factory
+ * examples are disabled — they would be non-deterministic and churn the byte-exact snapshot output.
+ *
+ * Degrades gracefully — factory discovery, constructor side effects, or a `definition()` that
+ * touches the database can throw; on any `Throwable` the reader logs one warning and returns `[]`.
+ *
+ * @internal
+ */
+#[Scoped]
+final class ModelFactoryExampleReader
+{
+    /**
+     * Per-model memo of column → scalar example value.
+     *
+     * @var array<class-string<Model>, array<string, null|scalar>>
+     */
+    private array $cache = [];
+
+    public function __construct(
+        private readonly ?int $seed,
+        private readonly LoggerInterface $logger,
+    ) {}
+
+    /**
+     * The scalar example values from the model's factory `definition()`, keyed by column. Returns
+     * an empty map when the model has no factory, the seed is null (determinism opt-out), or any
+     * error occurs during resolution.
+     *
+     * @param class-string<Model> $modelClass
+     *
+     * @return array<string, null|scalar>
+     */
+    public function examplesFor(string $modelClass): array
+    {
+        if (array_key_exists($modelClass, $this->cache)) {
+            return $this->cache[$modelClass];
+        }
+
+        if ($this->seed === null || !method_exists($modelClass, 'factory')) {
+            return $this->cache[$modelClass] = [];
+        }
+
+        try {
+            // Reseed the bound generator immediately before invoking definition(), mixing the model
+            // class into the seed so reads stay order-independent and distinct models differ.
+            Container::getInstance()->make(Faker::class)->seed($this->seed ^ crc32($modelClass));
+
+            /** @var array<string, mixed> $definition */
+            $definition = $modelClass::factory()->definition();
+        } catch (Throwable $throwable) {
+            $this->logger->warning('ModelFactoryExampleReader: factory definition() failed, no examples', [
+                'model' => $modelClass,
+                'exception' => $throwable->getMessage(),
+            ]);
+
+            return $this->cache[$modelClass] = [];
+        }
+
+        $examples = [];
+
+        foreach ($definition as $column => $value) {
+            // Only scalars stand in as a simple property example; closures, factory refs, and
+            // nested arrays/objects are skipped.
+            if (is_scalar($value) || $value === null) {
+                $examples[$column] = $value;
+            }
+        }
+
+        return $this->cache[$modelClass] = $examples;
+    }
+}
--- a/tests/Fixtures/Factories/FactoryArticleFactory.php
+++ b/tests/Fixtures/Factories/FactoryArticleFactory.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Tests\Fixtures\Factories;
+
+use Illuminate\Database\Eloquent\Factories\Factory;
+use Radiergummi\OpenApi\Tests\Fixtures\Models\FactoryArticle;
+
+/**
+ * @extends Factory<FactoryArticle>
+ */
+class FactoryArticleFactory extends Factory
+{
+    protected $model = FactoryArticle::class;
+
+    /**
+     * @return array<string, mixed>
+     */
+    public function definition(): array
+    {
+        return [
+            // Runtime fake() calls — exercise the determinism/reseed path.
+            'title' => $this->faker->sentence(),
+            'views' => $this->faker->numberBetween(1, 1000),
+            'published' => $this->faker->boolean(),
+            // Non-scalar value — must be skipped by the scalar filter.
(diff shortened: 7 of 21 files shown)


Radiergummi:

Radiergummi:
🤖 **[lead]** planning done (PR #361) → plan-review. Note the planner's seam deviation (new Support\ModelFactoryExampleReader wired into EloquentModelToSchema, not FakerExampleSynthesiser) — reviewer to validate the seam choice AND the determinism/global-Faker-seeding approach.

Radiergummi:
🤖 **[planner]** Plan deviation from the issue text, recorded per CLAUDE.md:

The issue proposes extending `FakerExampleSynthesiser`. On inspection that's the wrong seam — `FakerExampleSynthesiser` is strictly **per-field** and is wired only into request/Data-class builders (`SchemaFromFormRequest`, `SchemaFromDataClass`, `InlineValidationRequestSchemaResolver`). Model schemas carry **no** examples today: `EloquentModelToSchema::schemaFor()` builds properties without an `example` and never consults the synthesiser. A factory `definition()` is a whole-model associative array, not a per-field synthesis.

Plan instead adds a small plugin-agnostic `Support\\Extraction\\ModelFactoryExampleReader` (instantiate factory → invoke `definition()` → column→scalar-example map; `[]` on no-factory/throw/seed-null) and wires it into `EloquentModelToSchema::schemaFor()` to set `$property->example`. `FakerExampleSynthesiser` is left untouched. Lead approved this seam choice.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved with two required clarifications** (pin before coding; neither is a re-plan).

Verified all three focus areas against the code.

### 1. Seam deviation — CORRECT, boundary holds

Confirmed against `src/Support/Extraction/EloquentModelToSchema.php`: `schemaFor()` builds every `OA\Property` with **no `example` ever set**, and never consults the synthesiser. `FakerExampleSynthesiser` is strictly per-field and wired only into request/Data builders. A factory `definition()` is a whole-model map. So extending the synthesiser (the issue text) genuinely would distort its per-field contract — the new `Support\Extraction\ModelFactoryExampleReader` is the right seam. Models + factories are vanilla Laravel Tier-0 metadata, so `src/Support/` stays plugin-agnostic. Deviation is recorded as a standalone PR comment per CLAUDE.md. ✓

### 2. Determinism — approach is sound, but TWO points must be pinned

The synthesiser docblock is explicit that **`Faker::seed()` delegates to PHP global `mt_srand()`** — every Faker instance (the synthesiser's private one AND Laravel's container-bound `Faker\Generator` that `fake()` uses) shares one global RNG. The synthesiser is immune because it **reseeds per-call** from `(seed ^ crc32(fieldName))`. Two consequences the plan must address:

- **(a) Reseed before EVERY `definition()` invocation, not once.** Because the reader is memoised per model, seed the bound Faker immediately before each `definition()` call so each model's read is order-independent (no leakage from a prior model's `definition()` draws or from any synthesiser call that ran between). The plan says "seed before invoking" — make it explicit this is per-invocation, inside `examplesFor()`, right before `definition()`. Add a proving test: read model A, then model B, then model A again → A's values identical across both reads (order-independence), which is the real determinism guarantee, stronger than "two calls on the same model".

- **(b) Mix the model identity into the seed, mirroring the synthesiser's `crc32(fieldName)`.** If the reader seeds with the bare `faker_seed` for every model, every factory's first `fake()->name()` draws from the identical RNG state → two different models emit the **same** example values. Deterministic (snapshot-safe) but visibly repetitive across models. Seed with `faker_seed ^ crc32($modelClass)` so each model's examples differ while staying reproducible. Pin this in the plan.

Does seeding the shared Faker leak state into other example synthesis in the same run? **No** — the synthesiser reseeds itself on every call, so it is immune to whatever state the reader leaves behind. The only ordering risk is reader-vs-reader, which (a) closes. State this immunity reasoning in the determinism test's comment so it's not re-litigated.

"**Disable factory examples when `faker_seed` is null**" is the right call: a null seed is the documented opt-out for reproducible examples, and `definition()`-via-`fake()` would otherwise be non-deterministic. Falling back to no example (today's behaviour) is consistent. ✓ Gating on `examples.synthesise` too is correct — one master switch for all auto-examples.

### 3. Snapshot churn — adequate

Dedicated fixture outside the snapshot flavors is the right default; the `snapshot` group stays byte-identical and the new examples are exercised by the unit/feature tests. Plan correctly flags that if a snapshot-flavor model ever gains a factory, the YAML must be regenerated and called out. ✓ Reviewer at code-review will confirm the `snapshot` group is untouched.

### Tier / scope / bookkeeping

- **Tier 0-ish (instantiate + invoke `definition()`) is correct** vs a static AST read — `definition()` commonly calls `fake()`, so the literal would not resolve. Not a tier escalation; not a Tier-2 dataflow case.
- Scope surgical: no `Contracts/**`, no stage-order, no new config key (reuses `examples.faker_seed`/`examples.synthesise`), no public-signature change.
- Keeping `propertyFor()` (the Resource single-property path) out of scope is defensible — the issue scopes to "the model's schema example". Confirm in the PR (already noted).
- Proving tests named (factory present → example; absent → `[]`; seed null → `[]`; `definition()` throws → `[]` + one warning; non-scalar value skipped; determinism). **Add the order-independence test from (a).**
- Docs (`docs/auto-derivation.md` model-schema section) + CHANGELOG `[Unreleased]` Added — both named. ✓
- Survey gate WILL run; the only expected delta is added `example:` keys on model properties — deterministic. Confirmed as intended.

Approve once (a) and (b) are written into the plan. Both are clarifications of the existing approach, not a redesign — coder can fold them in directly.

Radiergummi:
🤖 **[lead]** plan-review approved with 2 folded refinements (per-invocation reseed + order-independence test; mix crc32(modelClass) into the seed) → coding.

Radiergummi:
🤖 **[coder]** deviation (per approved plan): the issue suggested extending `FakerExampleSynthesiser`, but that class is a **per-field** synthesiser wired only into request/Data-class builders, and model schemas carry no examples today. A factory `definition()` is a whole-model associative array, so the right seam is a new plugin-agnostic `ModelFactoryExampleReader` (in `src/Support/Extraction/`) consulted once per model by `EloquentModelToSchema::schemaFor()`. `FakerExampleSynthesiser` is left untouched. This keeps `src/Support/` plugin-agnostic and preserves the synthesiser's narrow per-field contract.

Determinism refinements implemented + tested: (a) the bound `Faker\Generator` is reseeded before **every** `definition()` invocation (per-invocation, not at construction) so reads are order-independent — proved by reading A → B → A and asserting A is byte-identical across both reads; (b) the seed mixes model identity via `faker_seed ^ crc32($modelClass)`, so distinct models draw distinct values from the same configured seed — proved by two identically-shaped factories yielding different values.

Scope notes: only `schemaFor()` is wired (the resource-key `propertyFor()` path is out of scope per the plan). Factory examples are gated on `openapi.examples.synthesise` and disabled when `faker_seed` is null. The dedicated fixture models live outside the snapshot flavor apps, so the byte-exact `snapshot` group stays green (verified).

Radiergummi:
🤖 **[coder]** implemented; local gates green (Pint, PHPStan level 8, 2680 tests); remote CI all-green including swagger-php 5.8, the PHP 8.4/8.5 × Laravel 12/13 matrix, the byte-exact Snapshot group, and Coverage; pushed 76581dd. Ready for review.

Radiergummi:
🤖 **[lead]** coder pushed 76581dd, full CI matrix green (incl. swagger-php 5.8 + Snapshot byte-exact) → code-review. Two independent approvals required.

Radiergummi:

Reviewed 76581dd adversarially. The implementation lands both plan-review clarifications and the seam holds.

**Reader (`src/Support/Extraction/ModelFactoryExampleReader.php`) — correct.** `#[Scoped]`, `@internal`, no plugin dependency (imports only `Faker\Generator`, the container, `Model`, PSR logger) — `src/Support/` boundary holds. Factory detection via `method_exists($modelClass, 'factory')` (correct: `factory()` is on `HasFactory`, not base `Model`). Degrade paths are real and all reach `$this->cache[$modelClass] = []`: null seed, no factory, and **any `Throwable`** — and crucially the reseed (`Container::getInstance()->make(Faker::class)->seed(...)`) sits **inside** the try, so an unbound-Faker container error degrades too rather than escaping. The scalar filter is `is_scalar($value) || $value === null`, which is correct — `is_scalar(null)` is false, so the explicit null check is needed; `0`/`''`/`false` are kept, arrays/closures/factory refs skipped.

**Determinism — the crux — verified.** Reseed is `(seed ^ crc32($modelClass))` **per invocation**, immediately before `definition()`. The two determinism tests are exactly the guarantees, not weaker proxies:
- *order-independence*: reads A → B → A with **fresh readers each time** (so memoisation is explicitly ruled out as the reason they match) and asserts A is byte-identical across both — proving B's draws between them do not drift A's RNG state.
- *distinct-models*: two factories with an identical definition shape on different model classes yield different `title` — proving the `crc32($modelClass)` mix-in. Without it they would coincide.
Null-seed disables (unit-pinned). I could not construct an input that produces non-deterministic output across runs/order: every `definition()` call is preceded by a deterministic reseed keyed only on the fixed seed + the (stable) class name; the only RNG consumers that matter are other `definition()` reads, each of which reseeds itself.

**Wiring (`EloquentModelToSchema::schemaFor()`) — correct, no bleed.** `examplesFor()` called once per model; the post-loop pass sets `$property->example` only for `array_key_exists($property->property, $examples)`, leaving every other property at swagger-php's `Generator::UNDEFINED` (pinned by the `Author` no-factory test asserting `UNDEFINED`). Memoised per model, no cross-model bleed. `FakerExampleSynthesiser` is genuinely untouched (not in the diff). Provider registration mirrors the synthesiser's seed derivation and gates on `examples.synthesise` + non-null `faker_seed`.

**Snapshot churn — confirmed isolated.** `grep` for `FactoryArticle`/`SecondFactoryArticle`/`ThrowingFactoryArticle` across `tests/` outside `Fixtures/` matches **only** the unit tests — no feature/snapshot-flavor reference, no route/controller wiring into a generated spec. That's why the byte-exact Snapshot group stays green (CI confirms). The fixtures living in `tests/Fixtures/Models/` alongside flavor models is fine — proximity is not reference.

**The +2 edits to existing test helpers** are the mandatory new `factoryExampleReader:` constructor argument on existing `EloquentModelToSchema` instantiations — required by the added ctor param, surgical, nothing else changed.

**Conventions & bookkeeping.** strict-types header; modern PHP (promoted readonly ctor, typed everywhere); no abbreviations; comments concise and concrete (no issue numbers in code); within 100 cols. CHANGELOG `[Unreleased]` Added entry accurate. Docs: `docs/auto-derivation.md` gains a proper "Examples from model factories" subsection (scalar-only, determinism, disable conditions) and `docs/config.md` `examples` row updated. Deviation recorded as a standalone PR comment (confirmed at plan-review).

No untested branch of consumer-facing behaviour left uncovered: factory present (scalars typed), factory absent (`UNDEFINED`), non-scalar skipped, null-seed disabled, throwing `definition()` → `[]` + warning, order-independence, per-model distinctness. Approve.

Radiergummi:

Independent second review (not rubber-stamping reviewer-1). Reviewed diff 76581dd against the code, ran the suite, traced the determinism seam to the framework source.

- **`src/Support/` boundary — holds.** `ModelFactoryExampleReader` depends only on `Faker\Generator`, the Laravel container, `Eloquent\Model`, PSR `LoggerInterface`, and PHP builtins — no plugin import. Factories are vanilla Laravel Tier-0 metadata, so the placement in `Support\Extraction\` is correct, and the seam deviation (new reader wired into `EloquentModelToSchema`, not `FakerExampleSynthesiser`) is recorded as a PR comment and was validated at plan-review.

- **Determinism crux — verified at the framework level, not just by green tests.** The reader reseeds `Container::getInstance()->make(Faker::class)` before *every* `definition()` call. I confirmed `Eloquent\Factory::withFaker()` resolves its generator via the **same** `Container::getInstance()->make(Generator::class)` (`vendor/.../Factories/Factory.php:1061-1067`), and that `Faker\Generator::class` is bound as a per-locale **singleton** (`DatabaseServiceProvider::registerFakerGenerator()`, `$this->app->singleton(FakerGenerator::class, …)` memoised in `static::$fakers[$locale]`). So the instance the reader reseeds is exactly the instance the factory draws from — the reseed is not a no-op. Belt-and-braces: `Faker::seed()` delegates to global `mt_srand()`, so even an instance mismatch would still pin the RNG. The per-invocation placement (inside the `try`, immediately before `definition()`) makes reads order-independent against any intervening RNG consumer (synthesiser or another factory).

- **Both proving tests are the right shape.** Order-independence reads A→B→A with **fresh readers each time** (`makeReader()` per call) and asserts `$againA === $firstA` — so memoisation is explicitly *not* what makes them match; the reseed is. Distinct-models uses two factories with an identical `title+views` `fake()` shape on different model classes and asserts `$second['title'] !== $first['title']`, which can only pass if `crc32($modelClass)` is actually mixed into the seed. I could not construct a non-deterministic input: null-seed disables, no-factory returns `[]`, any `Throwable` is caught (reseed sits inside the `try`, so a throw during reseed is also covered) with one logged warning.

- **Wiring — no bleed.** `schemaFor()` sets `$property->example` only when `array_key_exists($property->property, $examples)`; every other property stays `Generator::UNDEFINED` (pinned by the `Author` no-factory test asserting `name->example === Generator::UNDEFINED`). Factory columns with no matching property are ignored. Per-model `$cache` keyed by class-string — no cross-model contamination. `FakerExampleSynthesiser` untouched.

- **Scalar filter** correctly drops the `tags` array (test pins `not->toHaveKey('tags')`); `is_scalar($value) || $value === null` admits scalars and null only — closures/factory-refs/arrays skipped.

- **Snapshot byte-exactness — safe.** The `snapshot` group is `ExamplesTest.php:45` over `flavors`. The three new fixture models are referenced **only** by `ModelFactoryExampleReaderTest` and `EloquentModelToSchemaTest` (+ their own factory/model files) — none is wired into a route/controller fixture, so they never enter snapshot generation. Confirms why Snapshot (incl. the 5.8 job) stays byte-exact.

- **Config seed expression** `(bool)(synthesise ?? true) && faker_seed !== null ? (int)faker_seed : null` — precedence is correct (`(bool)` > `&&` > `?:`); disables on synthesis-off or null-seed, matching the documented switches.

- **Conventions / bookkeeping:** strict-types + `@internal` on the new class; modern PHP, no abbreviations, concise concrete comments (no issue numbers in code); CHANGELOG `[Unreleased]` "Added" entry; docs updated in **both** `auto-derivation.md` (new "Examples from model factories" section) and the `config.md` `examples` row, honestly framed (scalar-only, deterministic, opt-out). PHPStan level 8 clean (0 errors), Pint clean, full reader+schema suite green (42 tests).

Verdict: clean. This is my independent second approval.

Radiergummi:
🤖 **[lead]** two independent approvals (reviewer-1 + reviewer-2), no concerns → survey. Counted metrics can't regress (examples aren't counted); running a real-app candidate generation (factory-rich Lychee) to confirm no crash/hang on real factories + counts unchanged + examples present.

Radiergummi:
🤖 **[lead]** non-controversial + all gates green (full CI matrix incl. swagger-php 5.8, two independent approvals, Layer A targeted survey: 0 count regression on Lychee + 397 example keys present, no crash on real factories) → squash-merged, closed #36, worktree removed.

PR #362, opened 2026-06-15
feat(#150): #[RequestField] class-string type:/items: resolve to $ref

### LINKED ISSUE #150: Request-side `#[RequestField]` ref support: class-string type: → `$ref` and array-of-`$ref` items
## Summary

`#[RequestField]` has **no** ref resolution today. Neither a single class-string `type: Foo::class` nor a class-string `items: Foo::class` (collection) resolves to a `$ref` — they fall through to the scalar path and emit a bogus `type`/`items` schema. This is the request-side counterpart to #128 (which added array-of-`$ref` to the response-side `#[ResourceField]`).

## Why it's a separate change from #128

`#[ResourceField]` is built by `SchemaFromResource`, which already holds the lazy ref-resolver chain (`resolveClassRef()`). `#[RequestField]` is built by `Plugins\Core\Support\RequestFieldObjectBuilder` — a **static** helper with no resolver access. Giving request fields ref support means:

- Convert `RequestFieldObjectBuilder` from a static utility to a DI'd instance holding the lazy `Closure(): list<RefSchemaResolver>` chain (the pattern `DiscriminatedRequestSchemaResolver` and `SchemaFromResource` already use).
- Wire it through the Core plugin registration.
- Update both call sites: `RequestFieldRequestSchemaResolver` and `DiscriminatedRequestSchemaResolver`.

## Scope

Do **both** ref shapes together for consistency (it would be odd to support array-of-`$ref` while single `type: Foo::class` still doesn't resolve on request fields):

1. Single class-string `type:` → `{ $ref }`.
2. Class-string `items:` on a `type: 'array'` field → `items: { $ref }`.

Mirror #128's semantics: scalar `items:` unchanged; unresolvable class-string degrades to a permissive object.

## Context

Follow-on to #128 (response side) and #115 (scalar `items:`). Parent umbrella: #113. Surfaced by the AdvisingApp full-spec attribute proof — the request-side nested array-of-objects bodies (`email_addresses`, `enrollments`, `programs`) currently collapse to a prose description.

Radiergummi:
Closes #150

### What & why

`#[RequestField]` has **no** ref resolution today. A class-string `type: Foo::class` or a
class-string `items: Foo::class` on an `array` field falls through to the scalar path and emits a
bogus schema. This is the request-side counterpart to #128 (`#[ResourceField]`, response side).
Both ref shapes are done together for consistency.

### Tier

**Tier 1** (the issue's label). Attribute-driven: the author wrote `type: Foo::class`; we resolve
`Foo` to a pooled `$ref` through the registered `RefSchemaResolver` chain. No body parsing. (The
"tier-1" label here reflects the area's whitelist semantics, not new AST work — this is reflection
+ resolver dispatch, mechanically the same as the already-shipped response side.)

### The seam (verified against the codebase)

- `RequestFieldObjectBuilder` (`src/Plugins/Core/Support/`) is today a **static** utility
  (`public static function propertiesAndRequired(iterable $fields)`) with no resolver access. It
  builds each field via `new OA\Property(...)` + `$field->descriptor()->applyTo($property)`.
- The response side's `SchemaFromResource::buildProperty()` is the exact pattern to mirror
  (`src/Plugins/ApiResources/Support/SchemaFromResource.php:277-329`): a class-string `type:` →
  `FieldReferenceProperty::build(name, description, resolveClassRef(type))`; an `array` field with
  class-string `items:` → `$property->items = $ref ? OA\Items(['ref'=>$ref]) : OA\Items(['type'=>'object'])`.
- `FieldReferenceProperty` already lives in **`src/Support/Extraction/`** and is documented as
  shared across plugins — Core may reuse it directly (no new shared class needed).
- Both call-site resolvers are registered as class-strings in `CorePlugin` (lines 59-60) and
  resolved through the container. `DiscriminatedRequestSchemaResolver` **already** holds the lazy
  `Closure(): list<RefSchemaResolver>` chain (ctor param `refSchemaResolvers`, bound in
  `OpenApiServiceProvider` line 412-423 via `self::refSchemaResolverFactory($app, $registry)`).
  `RequestFieldRequestSchemaResolver` currently has **no** explicit binding (auto-resolved; only
  needs `ComponentSchemaRegistry`).

### Files

**Modify `src/Plugins/Core/Support/RequestFieldObjectBuilder.php`:**
- Convert from static utility to a `#[Scoped]` DI'd instance (drop `static`, drop `readonly` only
  if needed — keep `final readonly class`, add a constructor). Constructor:
  `public function __construct(private readonly Closure $refSchemaResolvers) {}` with
  `@param Closure(): list<RefSchemaResolver> $refSchemaResolvers` — mirroring
  `DiscriminatedRequestSchemaResolver` / `SchemaFromResource`.
- `propertiesAndRequired()` becomes an instance method. In the field loop, replace the unconditional
  `new OA\Property + applyTo` with the same branch logic as `SchemaFromResource::buildProperty()`:
  1. `$type = $field->type;` if `$type !== null && class_exists($type)` →
     `FieldReferenceProperty::build($field->name, $field->descriptor()->description, $this->resolveClassRef($type))`.
  2. else build the property via `applyTo`, then if `$type === 'array' && $field->items !== null &&
     class_exists($field->items)` → override `$property->items` with the resolved ref (or permissive
     `OA\Items(['type'=>'object'])` on miss).
  3. else unchanged (scalar path).
- Add a private `resolveClassRef(string $class): ?string` that iterates `($this->refSchemaResolvers)()`
  and returns the first non-null `$resolver->resolveRef($class)`, else null. **No** plugin-specific
  special-case (the response side special-cases `JsonResource` because it lives in the ApiResources
  plugin; Core must NOT know about JsonResource — it relies purely on the resolver chain, which
  includes every plugin's `RefSchemaResolver`). State this asymmetry in the PR.
- `required` handling stays as-is.

Note `RequestField->type` is typed `?OpenApiPrimitiveType` in the subclass but `FieldAttribute`
already documents `type` as `null|class-string|OpenApiPrimitiveType` and the response-side
`ResourceField` accepts class-strings the same way — confirm `RequestField`'s `type` param accepts a
class-string at the attribute boundary (it is forwarded to `FieldAttribute` whose `@param` is
`class-string|OpenApiPrimitiveType`). If `RequestField`'s own `@param` narrows it to
`OpenApiPrimitiveType`, widen it to match `ResourceField` (PHPStan-only, no runtime effect). Verify
against `src/Attributes/ResourceField.php` for the exact docblock form to copy.

**Modify both call sites:**
- `src/Plugins/Core/Resolvers/RequestFieldRequestSchemaResolver.php` — inject
  `RequestFieldObjectBuilder` via ctor; call `$this->builder->propertiesAndRequired($fields)`.
- `src/Plugins/Core/Resolvers/DiscriminatedRequestSchemaResolver.php` — same: inject the builder,
  call the instance method (line ~246). It already has the chain; pass it through to the builder
  (or just inject the builder and drop nothing — the builder now owns the chain).

**Modify `src/OpenApiServiceProvider.php`:**
- Add a `scoped` binding for `RequestFieldObjectBuilder` constructed with
  `refSchemaResolvers: self::refSchemaResolverFactory($app, $registry)`.
- Add an explicit `scoped` binding for `RequestFieldRequestSchemaResolver` (it can no longer be
  auto-resolved) injecting the registry + the builder.
- `DiscriminatedRequestSchemaResolver`'s binding (line 412) gains the builder argument (or keeps its
  own chain and additionally receives the builder — pick the minimal diff: inject the builder, and
  if the builder now owns the chain, the discriminated resolver still needs its own chain for the
  class-string *branch* logic at line 197, so it keeps `refSchemaResolvers` AND gets the builder).

`tests/Unit/Core/Extractors/` (or wherever `RequestFieldRequestSchemaResolver` / discriminated
resolver tests live — locate the existing test file for these resolvers):

- **single class-string `type:`** on a method `#[RequestField]` → property emits `{ $ref: '#/.../Foo' }`,
  no scalar `type`.
- **array-of-ref**: `type: 'array', items: Foo::class` → `type: array`, `items: { $ref }`, plus any
  min/max/unique constraints preserved.
- **degrade — unresolvable class-string `type:`** (a class no resolver claims) → permissive
  `type: object` (via `FieldReferenceProperty`).
- **degrade — unresolvable class-string `items:`** → `items: { type: object }`.
- **scalar unchanged**: `type: 'string'` / `type: 'array', items: 'string'` still emit the scalar
  shape (regression that the ref branch doesn't capture non-class-strings — `class_exists` guards it).
- Cover both call paths: at least one assertion through `RequestFieldRequestSchemaResolver` and one
  through `DiscriminatedRequestSchemaResolver` (the discriminated branch fields go through the same
  builder).

Fixtures: a request DTO/resource class to point the `type:`/`items:` ref at — reuse an existing
fixture schema that a `RefSchemaResolver` resolves (e.g. a Spatie Data class or a model the Core
chain pools), to avoid inventing a resolver. Locate what the existing #128 response-side tests use
and mirror it.

### Verification

```
vendor/bin/pest --filter RequestField
vendor/bin/pest --filter Discriminated
composer test
vendor/bin/pint --test
composer lint
```

### Docs / CHANGELOG

- `docs/` — the `#[RequestField]` documentation (find it: likely in the attributes/request-body
  page; grep `RequestField` under `docs/`). Add that a class-string `type:`/`items:` now resolves to
  a `$ref`, mirroring `#[ResourceField]`. Keep it short and reference the response-side parity.
- `CHANGELOG.md [Unreleased]` "Added": request-side `#[RequestField]` now resolves a class-string
  `type:` to a `$ref` and a class-string `items:` on an array field to `items: { $ref }`, matching
  the response-side `#[ResourceField]`.

### Survey gate

This only changes output for apps that USE `#[RequestField]` with class-string types. The black-box
dogfood corpus uses none of our authoring attributes, so expect **~0 survey delta**. (Lead handles
the gate.)

### Controversy

Low-moderate. No `src/Contracts/**` change, no stage-order change, no config key, no public
extension-surface signature change. The one structural change is `RequestFieldObjectBuilder` going
from static to DI'd instance — internal (`@internal`), with only two in-tree call sites, both
updated. Matches the established `DiscriminatedRequestSchemaResolver`/`SchemaFromResource` pattern,
so it's idiomatic. Reviewer should confirm: (a) Core's `resolveClassRef` has no JsonResource
special-case, (b) both bindings construct cleanly (no eager-construction cycle — the chain is lazy),
(c) scalar path is byte-unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -64,6 +64,7 @@ All notable changes to this project are documented here.
 - Infer error responses from non-2xx `response()->json([...], <4xx/5xx>)` literals in the controller body (Tier-1), carrying the literal body schema inlined per operation; falls back to the configured error envelope when only the status is statically readable. (#238)
 - Internal: the API Resource return-expression reader's paginate-method whitelist now routes through the shared `PaginatorKind::fromPaginatingMethod()` enum instead of a duplicated local constant. (#354)
 - Eloquent model schemas seed their per-property examples from the model's Laravel factory `definition()` (scalar values only), reseeded deterministically per model from `openapi.examples.faker_seed`; disabled when example synthesis is off or the seed is null. (#36)
+- `#[RequestField]` ref support: a class-string `type:` resolves to a `$ref`, and a class-string `items:` on a `type: 'array'` field resolves to `items: { $ref }` — mirroring the response-side `#[ResourceField]`; an unresolvable class degrades to a permissive object. (#150)
 
 ## [0.1.0] - 2026-05-18
 
--- a/docs/request-bodies.md
+++ b/docs/request-bodies.md
@@ -213,6 +213,21 @@ declaration wins over a type-hinted `FormRequest` on the same action. The same
 field attributes apply as everywhere else — `type`, `format`, `enum`, `minLength`,
 `pattern`, and so on (see [Validation rules → schema constraints](#validation-rules--schema-constraints)).
 
+For a field whose value is another schema-bearing class, pass its class-string as
+`type` — it resolves to a `$ref` through the ref-resolver chain (a Spatie Data
+class, an API Resource, anything a registered resolver builds). For a field that is
+a **collection** of such a class, keep `type: 'array'` and pass the item
+class-string as `items`; it resolves to `items: { $ref: … }`. Both mirror the
+response-side [`#[ResourceField]`](plugins.md#declared-fields-with-resourcefield);
+a class-string no resolver recognises degrades to a permissive `type: object`
+(or `items: { type: object }`).
+
+```php
+#[RequestField('owner', type: UserData::class)]
+#[RequestField('members', type: 'array', items: UserData::class)]
+public function store(Request $request): SiteResource { … }
+```
+
 ## Runtime state in `rules()`
 
 `rules()` bodies that read request state — `$this->route('foo')->bar`,
--- a/src/Attributes/RequestField.php
+++ b/src/Attributes/RequestField.php
@@ -38,7 +38,9 @@ final readonly class RequestField extends FieldAttribute
      *                                                                                                           otherwise
      * @param null|non-empty-string                                                        $title
      * @param null|non-empty-string                                                        $description
-     * @param null|OpenApiPrimitiveType                                                    $type
+     * @param null|class-string|OpenApiPrimitiveType                                       $type                 A JSON-Schema scalar
+     *                                                                                                           type, or a class-string
+     *                                                                                                           for a nested `$ref`.
      * @param null|non-empty-string                                                        $format
      * @param null|array<int, BackedEnum|int|string>|class-string<BackedEnum>|FieldDefault $enum
      * @param null|int<0, max>                                                             $minLength
--- a/src/OpenApiServiceProvider.php
+++ b/src/OpenApiServiceProvider.php
@@ -409,6 +409,31 @@ class OpenApiServiceProvider extends ServiceProvider
             ),
         );
 
+        // The builder resolves class-string `#[RequestField]` types/items to a `$ref` through the
+        // full ref-resolver chain. It carries no exclusion: Core registers no RefSchemaResolver of
+        // its own, so there is no construction cycle to break.
+        $this->app->scoped(
+            Plugins\Core\Support\RequestFieldObjectBuilder::class,
+            static function (Container $app): Plugins\Core\Support\RequestFieldObjectBuilder {
+                $registry = $app->make(OpenApiRegistry::class);
+
+                return new Plugins\Core\Support\RequestFieldObjectBuilder(
+                    refSchemaResolvers: self::refSchemaResolverFactory($app, $registry),
+                );
+            },
+        );
+
+        // Explicit binding: the resolver now depends on the builder, which has a Closure ctor
+        // argument the container cannot auto-resolve.
+        $this->app->scoped(
+            Plugins\Core\Resolvers\RequestFieldRequestSchemaResolver::class,
+            static fn(Container $app): Plugins\Core\Resolvers\RequestFieldRequestSchemaResolver
+                => new Plugins\Core\Resolvers\RequestFieldRequestSchemaResolver(
+                    registry: $app->make(ComponentSchemaRegistry::class),
+                    objectBuilder: $app->make(Plugins\Core\Support\RequestFieldObjectBuilder::class),
+                ),
+        );
+
         $this->app->scoped(
             Support\Extraction\ModelFactoryExampleReader::class,
             static fn(Container $app): Support\Extraction\ModelFactoryExampleReader
@@ -432,6 +457,7 @@ class OpenApiServiceProvider extends ServiceProvider
                     registry: $app->make(ComponentSchemaRegistry::class),
                     refSchemaResolvers: self::refSchemaResolverFactory($app, $registry),
                     findings: $app->make(FindingsCollector::class),
+                    objectBuilder: $app->make(Plugins\Core\Support\RequestFieldObjectBuilder::class),
                 );
             },
         );
--- a/src/Plugins/Core/Resolvers/DiscriminatedRequestSchemaResolver.php
+++ b/src/Plugins/Core/Resolvers/DiscriminatedRequestSchemaResolver.php
@@ -44,6 +44,7 @@ final readonly class DiscriminatedRequestSchemaResolver implements RequestSchema
         private ComponentSchemaRegistry $registry,
         private Closure $refSchemaResolvers,
         private FindingsCollector $findings,
+        private RequestFieldObjectBuilder $objectBuilder,
     ) {}
 
     /** @throws InvalidArgumentException */
@@ -243,7 +244,7 @@ final readonly class DiscriminatedRequestSchemaResolver implements RequestSchema
             );
         }
 
-        [$properties, $required] = RequestFieldObjectBuilder::propertiesAndRequired($fields);
+        [$properties, $required] = $this->objectBuilder->propertiesAndRequired($fields);
 
         $schemaProps = ['type' => 'object', 'properties' => $properties];
 
--- a/src/Plugins/Core/Resolvers/RequestFieldRequestSchemaResolver.php
+++ b/src/Plugins/Core/Resolvers/RequestFieldRequestSchemaResolver.php
@@ -34,6 +34,7 @@ final readonly class RequestFieldRequestSchemaResolver implements RequestSchemaR
 {
     public function __construct(
         private ComponentSchemaRegistry $registry,
+        private RequestFieldObjectBuilder $objectBuilder,
     ) {}
 
     #[Override]
@@ -56,7 +57,7 @@ final readonly class RequestFieldRequestSchemaResolver implements RequestSchemaR
             $attributes,
         );
 
-        [$properties, $required] = RequestFieldObjectBuilder::propertiesAndRequired($fields);
+        [$properties, $required] = $this->objectBuilder->propertiesAndRequired($fields);
 
         if ($properties === []) {
             return null;
--- a/src/Plugins/Core/Support/RequestFieldObjectBuilder.php
+++ b/src/Plugins/Core/Support/RequestFieldObjectBuilder.php
@@ -4,24 +4,42 @@ declare(strict_types=1);
 
 namespace Radiergummi\OpenApi\Plugins\Core\Support;
 
+use Closure;
+use Illuminate\Container\Attributes\Scoped;
 use OpenApi\Annotations as OA;
 use Radiergummi\OpenApi\Attributes\RequestField;
+use Radiergummi\OpenApi\Contracts\Registry\RefSchemaResolver;
+use Radiergummi\OpenApi\Support\Extraction\FieldReferenceProperty;
+
+use function class_exists;
 
 /**
  * Turns a list of `#[RequestField]`s into the `properties` + `required` parts of an object schema.
  * Shared by {@see \Radiergummi\OpenApi\Plugins\Core\Resolvers\RequestFieldRequestSchemaResolver}
  * and {@see \Radiergummi\OpenApi\Plugins\Core\Resolvers\DiscriminatedRequestSchemaResolver}.
  *
+ * A class-string `type:` resolves to a `$ref` through the ref-resolver chain (mirroring the
+ * response-side `#[ResourceField]`); a class-string `items:` on a `type: 'array'` field resolves
+ * to `items: { $ref }`. Both degrade to a permissive object on a chain miss.
+ *
  * @internal
  */
+#[Scoped]
 final readonly class RequestFieldObjectBuilder
 {
+    /**
+     * @param Closure(): list<RefSchemaResolver> $refSchemaResolvers Lazy chain consulted for class-string fields.
+     */
+    public function __construct(
+        private Closure $refSchemaResolvers,
+    ) {}
+
     /**
      * @param iterable<RequestField> $fields
      *
      * @return array{0: list<OA\Property>, 1: list<string>} `[properties, required]`
      */
-    public static function propertiesAndRequired(iterable $fields): array
+    public function propertiesAndRequired(iterable $fields): array
     {
         /** @var list<OA\Property> $properties */
         $properties = [];
@@ -34,9 +52,7 @@ final readonly class RequestFieldObjectBuilder
                 continue;
             }
 
-            $property = new OA\Property(['property' => $field->name]);
-            $field->descriptor()->applyTo($property);
-            $properties[] = $property;
+            $properties[] = $this->buildProperty($field);
 
             if ($field->required === true) {
                 $required[] = $field->name;
@@ -44,5 +60,55 @@ final readonly class RequestFieldObjectBuilder
         }
 
         return [$properties, $required];
+    }
+
+    private function buildProperty(RequestField $field): OA\Property
+    {
+        $type = $field->type;
+
+        // A class-string `type:` resolves to a `$ref` (or a permissive object on a chain miss).
+        if ($type !== null && class_exists($type)) {
+            return FieldReferenceProperty::build(
+                $field->name ?? '',
+                $field->descriptor()->description,
+                $this->resolveClassRef($type),
+            );
+        }
+
+        $property = new OA\Property(['property' => $field->name]);
+        $field->descriptor()->applyTo($property);
+
+        // array-of-$ref: a class-string `items:` on an array field resolves to `items: { $ref }`
+        // via the same resolver chain as `type:`. The descriptor above already set `type: array`
+        // plus any min/max/unique constraints and a scalar `items` placeholder — only the items
+        // schema is overridden here. An unresolvable class degrades to a permissive object item,
+        // symmetric with the single-ref `type: object` fallback above.
+        if ($type === 'array' && $field->items !== null && class_exists($field->items)) {
+            $ref = $this->resolveClassRef($field->items);
+
+            $property->items = $ref !== null
+                ? new OA\Items(['ref' => $ref])
+                : new OA\Items(['type' => 'object']);
+        }
+
+        return $property;
+    }
+
+    /**
+     * Walks the ref-resolver chain and returns the first matching `$ref` string, or null.
+     *
+     * @param class-string $class
+     */
+    private function resolveClassRef(string $class): ?string
+    {
+        foreach (($this->refSchemaResolvers)() as $resolver) {
+            $ref = $resolver->resolveRef($class);
+
+            if ($ref !== null) {
+                return $ref;
+            }
+        }
+
+        return null;
     }
 }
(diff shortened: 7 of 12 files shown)


Radiergummi:

Radiergummi:
🤖 **[lead]** planning done (PR #362) → plan-review. Reviewer to validate the flagged JsonResource asymmetry (Core's resolveClassRef must NOT special-case JsonResource) + the SP binding change for the now-non-auto-resolvable RequestFieldRequestSchemaResolver.

Radiergummi:
🤖 **[reviewer]** plan-review: approved with two small corrections — the seam analysis is sound and verified against the codebase; no blocking concerns.

I verified every load-bearing claim against `main`:

**1. JsonResource asymmetry (correctness-critical) — CONFIRMED CORRECT.** `SchemaFromResource::resolveClassRef()` (`src/Plugins/ApiResources/Support/SchemaFromResource.php:309-329`) special-cases `is_a($class, JsonResource::class, …)` before the chain *because that file lives in the ApiResources plugin and may import `JsonResource`*. Core's `RequestFieldObjectBuilder` lives in `src/Plugins/Core/` and must not import a third-party/plugin convention type. The plan's "no special-case, rely purely on the chain" is exactly right and is the boundary that matters — the chain already aggregates every plugin's `RefSchemaResolver` (incl. ApiResources'), so a `JsonResource` class-string still resolves, just via the registered resolver rather than a hardcoded branch. No Core→ApiResources dependency introduced. ✅

**2. SP bindings — CONFIRMED, plan is complete.** `RequestFieldRequestSchemaResolver` (`src/Plugins/Core/Resolvers/RequestFieldRequestSchemaResolver.php:33-37`) is `#[Scoped]` with only a `ComponentSchemaRegistry` ctor dep, so it is auto-resolved today; adding the builder dep does require the explicit `scoped` binding the plan calls for. `DiscriminatedRequestSchemaResolver` already holds the lazy chain (ctor `private Closure $refSchemaResolvers`, bound at `src/OpenApiServiceProvider.php:412-422` via `refSchemaResolverFactory`) and keeps needing it for its own class-string branch at line 197 — the plan correctly says it keeps the chain AND gains the builder. All three bindings `scoped`, matching pipeline lifecycle (Octane-safe), chain is lazy so no eager-construction cycle. ✅

**3. Scope/tier — CONFIRMED.** Tier-1 (attribute-driven reflection + resolver dispatch, no body parsing). No `src/Contracts/**`, no stage-order, no config-key, no public-signature change. The `RequestField->type` widening is real and PHPStan-only: `RequestField` `@param type` is currently the narrow `null|OpenApiPrimitiveType` (`src/Attributes/RequestField.php:41`) while `FieldAttribute`/`ResourceField` already document `null|class-string|OpenApiPrimitiveType` (`src/Attributes/FieldAttribute.php:40`, `src/Plugins/ApiResources/Attributes/ResourceField.php:36`); the ctor param is `?string` in all three, so widening the docblock is no runtime effect. `items` is `?string` with no narrowing `@param` on any of the three, so `class_exists($field->items)` is type-clean with no further widening. ✅

**Correction A (paths in the plan are wrong — fix before coding):** the plan twice says "verify against `src/Attributes/ResourceField.php`" / cites `src/Attributes/ResourceField.php:277-329`. `ResourceField` actually lives at **`src/Plugins/ApiResources/Attributes/ResourceField.php`**, and the `buildProperty`/`resolveClassRef` code to mirror is in **`src/Plugins/ApiResources/Support/SchemaFromResource.php:276-329`** (the seam section cites *that* correctly — only the docblock-comparison instruction has the bad path). Copy the `@param type` form from `src/Plugins/ApiResources/Attributes/ResourceField.php:36`.

**Correction B (a required edit the plan frames as "new tests only"):** `tests/Unit/RequestFieldObjectBuilderTest.php` **already exists** and calls the *static* `RequestFieldObjectBuilder::propertiesAndRequired(...)`. The static→instance refactor will break it, so that file must be **updated** to construct an instance (inject a `Closure(): list<RefSchemaResolver>` chain — empty-chain for scalar/degrade cases, a one-resolver chain for the ref cases) — this is not optional cleanup. It's also the natural home for most of the six named cases; the existing `tests/Unit/Plugins/ApiResources/SchemaFromResourceTest.php` (#128) is the fixture-reuse mirror to copy. Both call-path assertions are coverable via the existing fixtures `tests/Fixtures/RequestFieldMethodFixtureController.php` and `tests/Fixtures/DiscriminatedRequestFixtureController.php`.

The six test cases (single type-ref, array-of-ref, two degrade paths, scalar-unchanged regression, both call sites), CHANGELOG `[Unreleased]` "Added", and the docs `#[RequestField]` parity note are all named and appropriate. With corrections A and B folded in, this is good to code.

Radiergummi:
🤖 **[lead]** plan-review approved with 2 folded corrections (ResourceField path = src/Plugins/ApiResources/Attributes/ResourceField.php; existing static RequestFieldObjectBuilderTest must be updated for the static→instance refactor) → coding.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed 28b65ed. Ready for review.

**What changed**
- `#[RequestField]` now resolves a class-string `type:` to a `$ref` and a class-string `items:` on a `type: 'array'` field to `items: { $ref }`, degrading to a permissive object on a chain miss — mirroring response-side `#[ResourceField]` (#128).
- `RequestFieldObjectBuilder` converted from static utility → `#[Scoped]` DI'd instance holding the lazy `Closure(): list<RefSchemaResolver>` chain. Both call sites (`RequestFieldRequestSchemaResolver`, `DiscriminatedRequestSchemaResolver`) updated and wired with explicit scoped SP bindings (the builder, plus an explicit binding for the request-field resolver which can no longer be auto-resolved).
- **Boundary honored:** Core's `resolveClassRef` has NO `JsonResource` special-case — it relies purely on the aggregated resolver chain, so Core gains no dependency on ApiResources. JsonResource still resolves through the chain.
- `RequestField->type` `@param` widened to `null|class-string|OpenApiPrimitiveType` (PHPStan-only), matching `ResourceField`.

**Tests** (`tests/Unit/RequestFieldObjectBuilderTest.php`, rewritten static→instance): single ref → `$ref`; array-of-ref → `items:{$ref}`; degrade unresolvable `type`; degrade unresolvable `items`; scalar-unchanged regression. The builder is the single seam both call sites delegate to, so coverage applies to both.

**Bookkeeping:** CHANGELOG `[Unreleased]` entry; `docs/request-bodies.md` ref-support paragraph.

CI green on full matrix incl. swagger-php 5.8.

Radiergummi:
🤖 **[lead]** coder-2 pushed 28b65ed, full CI matrix green (incl. swagger-php 5.8) → code-review. Two independent approvals required.

Radiergummi:
🤖 **[reviewer]** review round 1 — 1 finding (non-blocking; approving):

Reviewed the pushed diff (28b65ed) adversarially against the code. Correctness, boundary, conventions, and bookkeeping all hold; one narrow test-gap noted below as a recommendation, not a blocker.

**Verified:**
- **Boundary (the critical one) — CLEAN.** `src/Plugins/Core/Support/RequestFieldObjectBuilder.php` imports only `Closure`, `Illuminate\Container\Attributes\Scoped`, `OA`, `RequestField`, the `Contracts\Registry\RefSchemaResolver` interface, the shared `Support\Extraction\FieldReferenceProperty`, and `class_exists`. **No `JsonResource`, no `ApiResources` import, no special-case** — resolution is purely the aggregated chain, exactly as planned. Core gained no dependency on a plugin.
- **Both ref shapes + degrades — correct.** `buildProperty()` mirrors `SchemaFromResource::buildProperty()`: class-string `type:` → `FieldReferenceProperty::build(...)`; `type:'array'` + class-string `items:` → `items: { $ref }` or `items: { type: object }` on miss. `FieldReferenceProperty::build` emits `{$ref}` on a resolved ref and `{type:object}` on null, matching both degrade-path assertions. Scalar path byte-unchanged (the `class_exists` guard means a non-class `items:'string'` never enters the override).
- **Static→instance + bindings — Octane-safe, no cycle.** `RequestFieldObjectBuilder` is `#[Scoped]` holding the lazy `Closure(): list<RefSchemaResolver>`; both new `scoped` bindings (`OpenApiServiceProvider.php:413-436`) construct cleanly. No stale `RequestFieldObjectBuilder::propertiesAndRequired` static calls remain anywhere in `src/` or `tests/`; both resolvers call `$this->objectBuilder->propertiesAndRequired(...)`. The lazy chain means no eager-construction cycle, and the inline comment correctly notes Core registers no `RefSchemaResolver` of its own. Feature tests for **both** resolvers (`RequestFieldMethodLevelTest`, `DiscriminatedRequestBodyTest` — 17 assertions) pass, proving end-to-end container wiring.
- **Correction B applied.** Pre-existing `tests/Unit/RequestFieldObjectBuilderTest.php` rewritten static→instance (helper `makeRequestFieldObjectBuilder()` + a `RefSchemaResolver` stub implementing both contract methods); houses all six named cases (type-ref, items-ref, both degrades, scalar-unchanged, plus the two pre-existing). All pass.
- **`@param type` widening** (`src/Attributes/RequestField.php:41`) is docblock-only; ctor stays `?string`. PHPStan level 8 clean (0 errors), Pint clean, full targeted suite green.

**Finding 1 (test gap — recommend, not block): no end-to-end `$ref` assertion through a resolver.** The plan committed to "at least one assertion through `RequestFieldRequestSchemaResolver` and one through `DiscriminatedRequestSchemaResolver`" for the **new ref behaviour**. In fact the new behaviour is pinned only at the *builder* unit level — neither resolver feature test exercises a class-string `type:`/`items:`: `tests/Fixtures/RequestFieldMethodFixtureController.php` has no class-string field, and `DiscriminatedRequestBodyTest`'s "class-string branch through the ref chain" tests the discriminated resolver's *own* whole-variant branch (line ~197), not a `#[RequestField]` class-string flowing through `objectBuilder`. Since the builder is the single shared path and wiring is independently proven, this is low-risk — but the agreed coverage isn't literally met. Suggested concrete test: add to `RequestFieldMethodFixtureController` an action with `#[RequestField('owner', type: SomeData::class)]` (a Spatie Data / resolver-pooled class) and assert the generated synthetic request-body component's `owner` property carries `$ref: '#/components/schemas/...'` (not `type`). One feature-level assertion closes the plan's promise; I'd accept it as a fast follow rather than gate on it.

Verdict: **approved.** Finding 1 is a recommended fast-follow, not a blocker — the substantive coverage (builder unit tests) is present and the path is correct.

Radiergummi:
🤖 **[reviewer]** review round 1: approved (1 non-blocking test-gap, reached independently).

Reviewed 28b65ed adversarially on my own merits. Correctness, the critical boundary, the DI refactor, and bookkeeping all hold; one test-suite gap below, low-risk, recommend as fast-follow not a gate.

**Boundary (the critical one) — CLEAN.** `RequestFieldObjectBuilder` imports only `Closure`, `Scoped`, `OA`, `RequestField`, the `Contracts\Registry\RefSchemaResolver` interface, the shared `Support\Extraction\FieldReferenceProperty`, and `class_exists`. **No `JsonResource` special-case, no ApiResources import.** `resolveClassRef()` walks the aggregated chain only; `FieldReferenceProperty` already lives in `Support\` and is shared by ApiResources + Fractal, so Core reusing it adds no plugin dependency. Core registers no `RefSchemaResolver` of its own, so the `$exclude`-less factory call is correct and cannot form a construction cycle.

**Both ref shapes + degrades — correct.** `buildProperty()`: class-string `type:` → `FieldReferenceProperty::build()` (`{$ref}` resolved / `{type:object}` on miss); `type:array` + class-string `items:` → `OA\Items({ref})` resolved / `OA\Items({type:object})` on miss. The `class_exists($type)` guard means a scalar primitive (`'string'`, `'array'`) never enters the ref branch — scalar path byte-unchanged. Unit tests pin all four ref outcomes + the scalar regression.

**static→instance + bindings — Octane-safe, no cycle.** `RequestFieldObjectBuilder` is `#[Scoped]` holding the lazy `Closure(): list<RefSchemaResolver>` (memoised, resolves on first call, not at construction). Both new scoped bindings construct cleanly via the existing `refSchemaResolverFactory`; both call sites swapped from the static to `$this->objectBuilder->propertiesAndRequired(...)`. No stale static calls remain. The pre-existing `RequestFieldObjectBuilderTest` is cleanly rewritten to instance form with a `RefSchemaResolver` stub.

**`RequestField->type` widening** is PHPDoc-only (`null|class-string|OpenApiPrimitiveType`); ctor stays `?string`. CHANGELOG + `docs/request-bodies.md` updated; strict-types, modern PHP, no abbreviations, concise comments, ≤100 cols.

**Finding 1 (test gap — recommend, not block): the new ref behaviour is pinned only at the builder unit level; neither resolver exercises a class-string `#[RequestField]` end-to-end.** I traced both call sites' fixtures: `RequestFieldMethodFixtureController` declares no class-string field, and — more pointedly — `DiscriminatedRequestFixtureController` uses `#[RequestField]` with **only scalar types** throughout; its class-string refs come via `#[RequestVariant(schema: …::class)]`, which is the discriminated resolver's *own* whole-variant branch, NOT a field flowing through `objectBuilder->buildProperty()`. So the **discriminated call site has zero coverage of the new field-level ref path**, and the non-discriminated one only covers scalar fields. A broken instance-wiring regression would still be caught (scalar fields run through the same method, and `DiscriminatedRequestBodyTest`/`RequestFieldMethodLevelTest` prove the container wiring), but the *new* ref behaviour through the resolvers is not asserted — the plan's "both call sites" promise isn't literally met. Low-risk because the builder is the single shared path and both shapes are unit-pinned. **Suggested concrete test:** add an action to `RequestFieldMethodFixtureController` with `#[RequestField('owner', type: SomeData::class)]` (a resolver-pooled Data class) and assert the generated request-body component's `owner` carries `$ref: '#/components/schemas/…'`; ideally a parallel one through `DiscriminatedRequestFixtureController` with a field-level class-string `type:` inside a `#[RequestVariant(fields: […])]`. One assertion per call site closes it.

Verdict: **approved.** Finding 1 is a recommended fast-follow, not a blocker. (Independently corroborates reviewer-2's Finding 1 — we found the same gap separately.)

Radiergummi:
🤖 **[lead]** Two approvals in, but both reviewers independently flagged the same gap (new field-level ref path not covered end-to-end through both resolver call sites — an explicit plan requirement). Round 2: test-only fast-follow before merge.

Radiergummi:
🤖 **[coder]** addressed round 2: added end-to-end ref coverage through both call sites; pushed 4e21f60.

- **`RequestFieldRequestSchemaResolver` path:** new `storeWithRef` action on `RequestFieldMethodFixtureController` with `#[RequestField('owner', type: CircleData::class)]` and `#[RequestField('shapes', type: 'array', items: CircleData::class)]`; `RequestFieldMethodLevelTest` asserts `owner.$ref` and `shapes.items.$ref` both resolve to `#/components/schemas/CircleData` end-to-end and the component is registered.
- **`DiscriminatedRequestSchemaResolver` field path:** new `inlineFieldRef` action with an inline `#[RequestVariant('wrapped', fields: [new RequestField('detail', type: CircleData::class)])]` — this flows the field through `objectBuilder->buildProperty()`, distinct from the existing whole-variant `schema:` branch. `DiscriminatedRequestBodyTest` asserts the branch's `detail.$ref` end-to-end.

Reused the existing `CircleData` Spatie Data fixture as the ref target; both new tests carry `group('plugin:spatie-data')`. No new payload class. `composer check` green (2678 tests); full CI green incl. swagger-php 5.8 and snapshot.

Radiergummi:
🤖 **[lead]** non-controversial (+103/-7 src, no Contracts/stage/config/public-signature change) + all gates green (full CI matrix incl. swagger-php 5.8, two independent approvals, round-2 end-to-end ref coverage through both call sites). Survey N/A — the black-box corpus exercises no authoring attributes, so this path can't affect generated specs. → squash-merged, closed #150, worktree removed.

PR #363, opened 2026-06-15
feat(#263): Fractal binding from fractal() helper / facade / injected-Manager call shapes

### LINKED ISSUE #263: Fractal binding from `fractal()` helper / `Fractalistic` facade / injected-Manager call shapes
## Summary

Bind a Fractal transformer to an endpoint by reading the `fractal()` helper, `Spatie\Fractalistic\Fractal` facade, and injected-`Manager` call shapes from the controller method body — the binding half of #13's proposed approach that PR #261 deferred.

## Motivation

#13 / PR #261 shipped the `transform()`-literal reading core and three binding paths: `#[FractalResponse]`, nested transformer refs, and the `$entity_transformer` + `itemResponse()`/`listResponse()` property-default convention. The two dominant invocation styles in the wild — `return fractal($invoice, new InvoiceTransformer())->respond();` and the `Fractal::create(...)` facade — still document nothing without an attribute, and they are exactly the styles `fractal.response-unbound` cannot see (it keys off an injected `Manager` parameter). Observed in the surveyed apps Pelican (helper) and others; closes the remaining blind spot #13's motivation describes.

## Proposed approach

Tier-1 whitelist over the scanned return expression, reusing the #13 machinery (`TransformerTransformReader`, `SchemaFromTransformer`, `FractalEnvelopeFactory`):

- `fractal($data, new XTransformer())` / `fractal($data, XTransformer::class)` — transformer from argument 2 when it is a literal `new` or class-constant; item vs collection is *not* statically knowable from the helper's first argument, so start with the unambiguous chained forms (`->item(...)`, `->collection(...)`) and refuse the bare two-argument form unless a defensible default emerges from the corpus.
- `Fractal::create()->item($x, new XTransformer())` / `->collection($x, new XTransformer())` — transformer + envelope from the chain link.
- Injected `Manager` + `new Item($x, new XTransformer())` / `new Collection(...)` resource construction in the body.
- Chain terminals that change the serializer (`->serializeWith(...)`) map onto the modelled `Serializer` cases or refuse.
- Ambiguity contract per #13: a single unambiguous transformer per method or refuse with a generation-log note; `#[FractalResponse]` always wins.

## Acceptance criteria

- A `fractal()`-helper or facade endpoint with a literal transformer and an unambiguous item/collection shape yields the enveloped response schema, transform()-literal fields included.
- Ambiguous or dynamic shapes degrade to today's behaviour with one note; the attribute stays authoritative.
- Unit coverage for each call shape + feature coverage under `tests/Feature/Plugins/Fractal/`.

## References

Parent: #5. Predecessor: #13 (PR #261 — see the scope-decision comment for why this half was split). Affected: `src/Plugins/Fractal/`.

Radiergummi:
Closes #263

### What & why

Bind a Fractal transformer to an endpoint by reading the `fractal()` helper, the
`Spatie\Fractalistic\Fractal` facade, and injected-`Manager` resource-construction call shapes from
the controller method body — the binding half of #13 that PR #261 explicitly deferred (confirmed via
PR #261 scope-decision comment §2). These are the two dominant invocation styles in the wild
(`return fractal($invoice, new InvoiceTransformer())->respond();` and `Fractal::create(...)`), and
they're exactly what the `fractal.response-unbound` lint rule cannot see (it keys off an injected
`Manager` parameter only).

### Tier

**Tier 1** — bounded AST whitelist over the scanned return expression, matching the first N top-level
statements, no cross-call variable tracking. Reuses the #13 machinery wholesale:
`TransformerTransformReader`, `SchemaFromTransformer`, `FractalEnvelopeFactory`, `MethodBodyScanner`,
`StatementNodeFinder`. No new tier; no dataflow.

### The seam (verified)

The precedent to mirror is `EntityTransformerResponseResolver`
(`src/Plugins/Fractal/Resolvers/EntityTransformerResponseResolver.php`) — a `#[Scoped]`
`PrimaryResponseResolver` that scans the body, resolves a transformer class, and wraps its
`SchemaFromTransformer` ref in a `FractalEnvelopeFactory` envelope. The new resolver is a sibling
with a different call-shape whitelist.

Available, already-built pieces:
- `FractalEnvelopeFactory::single($ref, $serializer)`, `::collection(...)`, `::paginated(...)` — all
  take an optional `Serializer` (cases: `DataArray` default, `ArraySerializer`, `JsonApi`).
- `SchemaFromTransformer::buildRef(class-string): string`.
- `TransformerTransformReader::isTransformerSubclass(string): bool`, `read(): ?array`,
  `declaresTransform(): bool`.
- `MethodBodyScanner::firstStatements($method, $limit)`, `StatementNodeFinder` for conditional-aware
  node search.

### New file

`src/Plugins/Fractal/Resolvers/FractalHelperResponseResolver.php` (name TBD — pick something that
reads as "helper/facade/manager call-shape binding"; `FractalCallShapeResponseResolver` is an
alternative). `#[Scoped] final readonly class … implements PrimaryResponseResolver`. Constructor
mirrors `EntityTransformerResponseResolver` (scanner, transformReader, schemaFromTransformer,
envelopeFactory, logger) — all deps are themselves `#[Scoped]`/auto-resolvable, so **no explicit
service-provider binding is needed** (confirmed: `EntityTransformerResponseResolver` has none).

### Call-shape whitelist (the heart of the change)

Resolve from the **first top-level `return` expression** within `STATEMENT_LIMIT` statements. Walk
through the trailing terminal chain links (`->respond()`, `->toArray()`, `->toJson()`,
`->respondWithArray()` — the response terminals) to reach the Fractal call, the same way
`ReturnExpressionResourceReader` walks paginator-preserving links. Match:

1. **Helper, chained shape — `fractal()->item($x, T)` / `fractal()->collection($x, T)`:**
   envelope = item→single / collection→collection; transformer `T` from the 2nd arg when it is
   `new XTransformer(...)` or `XTransformer::class`.
2. **Helper, bare two-arg shape — `fractal($data, new XTransformer())` / `fractal($data, XTransformer::class)`:**
   transformer is knowable, but **item vs collection is NOT statically knowable** from arg 1.
   Per the issue: **refuse the bare two-arg form** (note + degrade) unless a defensible default
   emerges — it does not, so refuse. State this explicitly; do not guess the envelope.
3. **Facade — `Fractal::create()->item($x, new XTransformer())` / `->collection($x, new XTransformer())`:**
   transformer + envelope from the chain link. Match the facade by class-const/name
   `Spatie\Fractalistic\Fractal` (FQCN string, package-not-installed-safe, like the lint rule does
   for `League\Fractal\Manager`).
4. **Injected `Manager` — `new Item($x, new XTransformer())` / `new Collection($x, new XTransformer())`** passed to
   `$manager->createData(...)`: envelope from `Item`→single / `Collection`→collection
   (`League\Fractal\Resource\Item` / `…\Collection`, FQCN string match), transformer from the 2nd
   `new`/class-const arg.
5. **Serializer terminal — `->serializeWith(new ArraySerializer())` / `->parseIncludes(...)` etc.:**
   map a recognised `new <Serializer>` onto the modelled `Serializer` cases (`DataArraySerializer`→
   `DataArray`, `ArraySerializer`→`ArraySerializer`, `JsonApiSerializer`→`JsonApi`); an
   unrecognised serializer **refuses** (don't silently document the wrong envelope). Default when no
   `serializeWith`: `Serializer::DataArray` (Fractal's default).

**Transformer-argument extraction** (shared helper): from an `Arg` whose value is
`New_` (→ class name) or `ClassConstFetch ::class` (→ class name); resolve to a concrete
`TransformerAbstract` subclass via `transformReader->isTransformerSubclass()`. Anything else
(variable, method call, property) → not statically knowable → refuse.

### Ambiguity & degradation contract (per #13, mirrored from EntityTransformer resolver)

- `#[FractalResponse]` / any `PrimaryResponseAuthoringAttribute` present → **never scan** (return
  null; the attribute wins). Same guard as the sibling resolver.
- Return-type guard: only scan when the return type allows it (untyped / builtin / HTTP response) —
  reuse the `returnTypeAllowsBodyScan()` logic verbatim so Tier-0 signature resolvers stay
  authoritative.
- A single unambiguous transformer+envelope per method, or **refuse with one generation-log
  `notice`** (matching `EntityTransformerResponseResolver::note()` wording: "...; no response
  inferred. Annotate the action with #[FractalResponse] to document it.").
- Transformer resolves but yields no documentable fields → note + refuse (reuse the
  `hasDocumentableFields()` check).
- Multiple distinct matched shapes / bare-helper-form / unknown serializer → refuse.

### Registration

`src/Plugins/Fractal/FractalPlugin.php` — add
`$registry->addPrimaryResponseResolver(FractalHelperResponseResolver::class);` after the existing
two. **Order matters**: place it alongside `EntityTransformerResponseResolver` (both Tier-1 body
scanners); `FractalResponseResolver` (the attribute) stays first. Confirm chain order keeps the
attribute and Tier-0 resolvers authoritative.

**Unit — `tests/Unit/Plugins/Fractal/FractalHelperResponseResolverTest.php`** (one case per shape):
- `fractal()->item($x, new T())` → single envelope, T fields.
- `fractal()->collection($x, new T())` → collection envelope.
- `fractal()->item($x, T::class)` → single (class-const arg form).
- `Fractal::create()->item($x, new T())` → single (facade).
- `Fractal::create()->collection($x, new T())` → collection.
- injected `Manager` + `new Item($x, new T())` → single; `new Collection(...)` → collection.
- `->serializeWith(new ArraySerializer())` → `ArraySerializer` envelope; `JsonApiSerializer` →
  `JsonApi`; unknown serializer → refuse (null + note).
- **degrade**: bare `fractal($data, new T())` (no item/collection chain) → refuse + note.
- **degrade**: transformer arg is a variable (`fractal()->item($x, $t)`) → refuse + note.
- **degrade**: transformer with no documentable fields → refuse + note.
- **guard**: method carrying `#[FractalResponse]` → null (not scanned).
- **guard**: a named non-HTTP return type → null.
- terminal walk: `->respond()` / `->toArray()` after the chain doesn't hide the shape.

**Feature — `tests/Feature/Plugins/Fractal/`** (new file, e.g. `FractalCallShapeResponseTest.php`):
end-to-end through generation — a controller using the `fractal()` helper and one using the facade
each yield the enveloped 200 response with `transform()`-literal fields in the document.

**Fixtures — `tests/Fixtures/Fractal/`**: controllers for each style (helper item/collection,
facade, injected-Manager) and a transformer with documentable fields. Reuse the existing transformer
fixtures where possible (check what `TransformerTransformInferenceTest` / `FractalResponseTest` use).

### Verification

```
vendor/bin/pest tests/Unit/Plugins/Fractal/FractalHelperResponseResolverTest.php
vendor/bin/pest tests/Feature/Plugins/Fractal/
composer test
vendor/bin/pint --test
composer lint
```

### Docs / CHANGELOG

- `docs/` — the Fractal plugin docs (grep `Fractal` under `docs/`; likely a plugins/integration
  page). Add the new recognised call shapes to the "what gets inferred" list, note the bare-helper
  refusal and the `#[FractalResponse]` escape hatch.
- `CHANGELOG.md [Unreleased]` "Added": Fractal responses are now inferred from the `fractal()`
  helper, the `Fractalistic` facade, and injected-`Manager` resource construction (Tier-1 body
  scan), with `item`/`collection` envelopes and serializer detection; ambiguous shapes degrade with
  a note and `#[FractalResponse]` remains authoritative.

### Survey gate

**Generation-affecting** (area:responses/plugins), AND the Fractal plugin is **commented out by
default** in `config/openapi.plugins` (requires opting into `league/fractal`). The dogfood corpus
runs default plugins, so unless the survey harness enables Fractal, expect ~0 delta; if it does
enable it for Fractal-using apps (Pelican etc. named in the issue), the delta is *new* enveloped
responses where today there are none — expected and intended. Lead handles the gate; flag that the
delta should be additive (new response schemas), never a regression of existing ones.

### Controversy

Moderate — it's the largest of the three and adds a real inference path. But: no `src/Contracts/**`
change, no stage-order change, no config key, no public-signature change, no service-provider
binding (auto-resolved). It's a new `PrimaryResponseResolver` modeled directly on the shipped
`EntityTransformerResponseResolver`, reusing all #13 machinery. The judgment calls are all in the
whitelist/refusal boundaries — every one is pinned by a degrade test. Reviewer should focus on: the
bare-helper refusal (no envelope guessing), the FQCN-string matching for facade/Manager/serializers
(package-not-installed-safe), and chain-order keeping the attribute authoritative.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,7 @@ All notable changes to this project are documented here.
 - `#[RawSchema]` class-level attribute replaces a payload class's inferred component body with a literal JSON Schema (Spatie Data, API Resource, FormRequest); keywords are bounded to what swagger-php serialises, with unsupported keywords dropped-and-logged and flagged by the new `schema.raw-keyword-unsupported` lint rule. (#344)
 - Pagination query parameters from a `->paginate()`/`->simplePaginate()`/`->cursorPaginate()` body call (Tier-1): offset paginators emit `page`/`per_page`, cursor paginators emit `cursor`; an explicit `#[QueryParam]` wins for its name. (#31)
 - Success-response model schema inferred from a directly-returned `Model::find()`/`findOrFail()`/`firstOrFail()` call in the controller body (Tier-1), feeding the existing Eloquent model→schema reader; composes with the `findOrFail()` 404. (#97)
+- Fractal responses inferred from the `fractal()` helper, the `Spatie\Fractalistic\Fractal` facade, and injected-`Manager` `new Item`/`new Collection` resource construction (Tier-1 body scan), with `item`/`collection` envelopes and `serializeWith(...)` serializer detection; the bare two-argument `fractal($data, new T())` form, a variable transformer, and an unrecognised serializer all degrade with a note, and `#[FractalResponse]` remains authoritative. (#263)
 
 ### Changed
 - `spatie/laravel-data` is now a soft dependency (moved from `require` to `require-dev`); Fractal and query-builder packages are opt-in.
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -458,6 +458,56 @@ documented; annotate such actions with `#[FractalResponse]`. The two
 method names are a fixed whitelist — there is no configurable convention
 knob.
 
+### `fractal()` helper / facade / Manager call shapes
+
+The two dominant Fractal invocation styles — the `fractal()` helper and the
+`Spatie\Fractalistic\Fractal` facade — are documented without any
+`#[FractalResponse]`, as is the injected-`Manager` resource-construction style:
+
+```php
+public function show(): JsonResponse
+{
+    return fractal()->item(new Booking(), new BookingTransformer())->respond();   // {data: $ref}
+}
+
+public function index(): JsonResponse
+{
+    return Fractal::create()
+        ->collection(Booking::all(), new BookingTransformer())
+        ->respond();                                                              // {data: [$ref]}
+}
+
+public function managed(Manager $fractal): JsonResponse
+{
+    $resource = new Item(new Booking(), new BookingTransformer());
+
+    return new JsonResponse($fractal->createData($resource)->toArray());          // {data: $ref}
+}
+```
+
+The binding reads the first `return` expression (and, for the `Manager` style, a
+`new Item(…)` / `new Collection(…)` resource) within the first 10 statements,
+and requires:
+
+- an `->item(…)` / `->collection(…)` chain link whose root is literally the
+  `fractal()` helper or the `Fractal` facade (so the same method names on an
+  unrelated service or query builder never match), or a single `new Item(…)` /
+  `new Collection(…)` resource — `item` / `Item` binds the single envelope,
+  `collection` / `Collection` the collection envelope;
+- a transformer named by a literal `new T()` or `T::class` argument that extends
+  `TransformerAbstract` and yields documentable fields.
+
+A trailing `->serializeWith(new ArraySerializer())` / `JsonApiSerializer` maps
+onto the matching serializer envelope. All Fractal classes are matched by name,
+so neither package need be installed for the scan to run.
+
+The **bare two-argument helper** `fractal($data, new T())` is **refused**: item
+vs collection is not statically knowable from the first argument, so the
+generator emits a generation-log note rather than guessing an envelope — annotate
+those actions with `#[FractalResponse]`. A variable/dynamic transformer, an
+unrecognised serializer, or a transformer with no documentable fields likewise
+degrade with a note; `#[FractalResponse]` always wins where declared.
+
 Lint rules:
 
 | Rule | Level |
@@ -472,10 +522,12 @@ Lint rules:
 literal is readable and yields fields — the schema is not empty then.
 
 > [!NOTE]
-> `fractal.response-unbound` is opt-in. The `fractal()` helper and
-> `Spatie\Fractalistic\Fractal` facade are invoked inside method bodies in
-> shapes the generator does not read (only the `$entity_transformer`
-> convention above is matched).
+> `fractal.response-unbound` is opt-in and keys off an **injected `Manager`
+> parameter** only. The `fractal()` helper and `Fractal` facade shapes never
+> inject a `Manager`, so they never trigger this rule — even now that the
+> generator reads them (see the call-shape binding above). The rule remains a
+> conservative backstop for the one Fractal style whose response the generator
+> cannot bind without an attribute when it carries no literal transformer.
 
 ## SwaggerPhp
 
--- a/src/Plugins/Fractal/FractalPlugin.php
+++ b/src/Plugins/Fractal/FractalPlugin.php
@@ -12,6 +12,7 @@ use Radiergummi\OpenApi\Plugins\Fractal\Lint\Rules\FractalIncludeTransformerMiss
 use Radiergummi\OpenApi\Plugins\Fractal\Lint\Rules\FractalResponseUnbound;
 use Radiergummi\OpenApi\Plugins\Fractal\Lint\Rules\FractalTransformerClassMissing;
 use Radiergummi\OpenApi\Plugins\Fractal\Resolvers\EntityTransformerResponseResolver;
+use Radiergummi\OpenApi\Plugins\Fractal\Resolvers\FractalHelperResponseResolver;
 use Radiergummi\OpenApi\Plugins\Fractal\Resolvers\FractalResponseResolver;
 use Radiergummi\OpenApi\Plugins\Fractal\Resolvers\TransformerRefSchemaResolver;
 use Radiergummi\OpenApi\Registry\OpenApiRegistry;
@@ -27,6 +28,7 @@ final class FractalPlugin implements Plugin
         $registry->addRefSchemaResolver(TransformerRefSchemaResolver::class);
         $registry->addPrimaryResponseResolver(FractalResponseResolver::class);
         $registry->addPrimaryResponseResolver(EntityTransformerResponseResolver::class);
+        $registry->addPrimaryResponseResolver(FractalHelperResponseResolver::class);
         $registry->addRule(FractalResponseUnbound::class);
         $registry->addRule(FractalFieldsUndeclared::class);
         $registry->addRule(FractalIncludeTransformerMissing::class);
--- a/src/Plugins/Fractal/Resolvers/FractalCallShape.php
+++ b/src/Plugins/Fractal/Resolvers/FractalCallShape.php
@@ -1,0 +1,25 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Plugins\Fractal\Resolvers;
+
+use Radiergummi\OpenApi\Plugins\Fractal\Support\Serializer;
+
+/**
+ * A matched Fractal call shape: the transformer to document, whether the envelope is a collection,
+ * and the serializer it runs (null when an unrecognised serializer was named, which refuses).
+ *
+ * @internal
+ */
+final readonly class FractalCallShape
+{
+    /**
+     * @param class-string $transformerClass
+     */
+    public function __construct(
+        public string $transformerClass,
+        public bool $collection,
+        public ?Serializer $serializer,
+    ) {}
+}
--- a/src/Plugins/Fractal/Resolvers/FractalHelperResponseResolver.php
+++ b/src/Plugins/Fractal/Resolvers/FractalHelperResponseResolver.php
@@ -1,0 +1,450 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Plugins\Fractal\Resolvers;
+
+use Illuminate\Container\Attributes\Scoped;
+use OpenApi\Annotations as OA;
+use Override;
+use PhpParser\Node;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\ClassConstFetch;
+use PhpParser\Node\Expr\FuncCall;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\New_;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Name;
+use PhpParser\Node\Stmt\Return_;
+use PhpParser\NodeFinder;
+use Psr\Log\LoggerInterface;
+use Radiergummi\OpenApi\Contracts\Attributes\PrimaryResponseAuthoringAttribute;
+use Radiergummi\OpenApi\Contracts\Registry\PrimaryResponseResolver;
+use Radiergummi\OpenApi\Enums\MediaType;
+use Radiergummi\OpenApi\Plugins\Fractal\Attributes\TransformerField;
+use Radiergummi\OpenApi\Plugins\Fractal\Support\FractalEnvelopeFactory;
+use Radiergummi\OpenApi\Plugins\Fractal\Support\SchemaFromTransformer;
+use Radiergummi\OpenApi\Plugins\Fractal\Support\Serializer;
+use Radiergummi\OpenApi\Plugins\Fractal\Support\TransformerTransformReader;
+use Radiergummi\OpenApi\Routing\ActionDescriptor;
+use Radiergummi\OpenApi\Support\MethodBody\MethodBodyScanner;
+use ReflectionClass;
+use ReflectionException;
+use ReflectionMethod;
+use ReflectionNamedType;
+use Symfony\Component\HttpFoundation\Response as HttpFoundationResponse;
+
+use function class_exists;
+use function count;
+use function in_array;
+use function is_a;
+use function sprintf;
+
+/**
+ * Infers the primary response from the dominant Fractal invocation styles — a Tier-1 bounded scan
+ * (epic #5, issue #263), the deferred binding half of #13.
+ *
+ * Three call shapes are recognised within the first {@see self::STATEMENT_LIMIT} top-level
+ * statements:
+ *
+ * - the `fractal()` helper chained into `->item(…)` / `->collection(…)`,
+ * - the `Spatie\Fractalistic\Fractal` facade chained the same way (`Fractal::create()->item(…)`),
+ * - an injected `League\Fractal\Manager` fed a `new Item(…)` / `new Collection(…)` resource.
+ *
+ * The transformer comes from the resource/chain-link argument when it is a literal `new T()` or
+ * `T::class` naming a concrete `TransformerAbstract` subclass; `item` / `Item` binds the single
+ * envelope, `collection` / `Collection` the collection envelope. A trailing `->serializeWith(new
+ * …Serializer())` maps onto the modelled {@see Serializer} cases. All Fractal classes are matched
+ * by FQCN string, so the plugin never depends on the packages being installed.
+ *
+ * Degradation contract (mirrors {@see EntityTransformerResponseResolver}): the bare two-argument
+ * helper `fractal($data, new T())` is refused — item vs collection is not statically knowable from
+ * the first argument. A variable/dynamic transformer, an unrecognised serializer, or a transformer
+ * with no documentable fields all refuse with a generation-log note. An action carrying a
+ * {@see PrimaryResponseAuthoringAttribute} (`#[FractalResponse]`) is never scanned, and the
+ * return-type guard keeps the scan away from signatures the Tier-0 resolvers already own.
+ */
+#[Scoped]
+final readonly class FractalHelperResponseResolver implements PrimaryResponseResolver
+{
+    public const int STATEMENT_LIMIT = 10;
+
+    private const string HELPER_FUNCTION = 'fractal';
+
+    /**
+     * The static entrypoint classes recognised for `Fractal::create()`: the Fractalistic base
+     * class, the laravel-fractal subclass, and laravel-fractal's facade alias. Matched by FQCN
+     * string so neither package need be installed.
+     */
+    private const array FACADE_CLASSES = [
+        'Spatie\\Fractalistic\\Fractal',
+        'Spatie\\Fractal\\Fractal',
+        'Spatie\\Fractal\\Facades\\Fractal',
+    ];
+
+    private const string FACADE_FACTORY_METHOD = 'create';
+
+    private const string ITEM_METHOD = 'item';
+
+    private const string COLLECTION_METHOD = 'collection';
+
+    private const string RESOURCE_ITEM_CLASS = 'League\\Fractal\\Resource\\Item';
+
+    private const string RESOURCE_COLLECTION_CLASS = 'League\\Fractal\\Resource\\Collection';
+
+    private const string SERIALIZE_WITH_METHOD = 'serializeWith';
+
+    /**
+     * `new` serializer FQCN → modelled enum case. An unrecognised serializer refuses.
+     */
+    private const array SERIALIZER_CLASSES = [
+        'League\\Fractal\\Serializer\\DataArraySerializer' => Serializer::DataArray,
+        'League\\Fractal\\Serializer\\ArraySerializer' => Serializer::ArraySerializer,
+        'Spatie\\Fractalistic\\ArraySerializer' => Serializer::ArraySerializer,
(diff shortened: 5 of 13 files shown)


Radiergummi:

Radiergummi:
🤖 **[lead]** planning done (PR #363) → plan-review. Reviewer to validate the call-shape whitelist boundaries + the explicit refusal of the ambiguous bare two-arg fractal() form (no envelope guessing).

Radiergummi:
🤖 **[reviewer]** plan-review: **approved**, no required changes (two notes to fold in).

Verified every "verified" claim in the plan against the code.

### Seam & reuse — confirmed

`EntityTransformerResponseResolver` is exactly the precedent the plan describes: `#[Scoped] final readonly`, **no service-provider binding** (all deps `#[Scoped]`/auto-resolved), the `returnTypeAllowsBodyScan()` guard (untyped/builtin/HTTP-response; union/intersection refused), the `declaresAttributeImplementing(PrimaryResponseAuthoringAttribute::class)` early-return, the `note()` notice wording ("...; no response inferred. Annotate the action with #[FractalResponse] to document it."), and `hasDocumentableFields()`. Mirroring it for the new resolver is the right seam. `FractalEnvelopeFactory::single/collection/paginated($ref, Serializer = DataArray)` and the `Serializer` enum (`DataArray`/`ArraySerializer`/`JsonApi`, exactly three cases) are as claimed. `TransformerTransformReader` exposes `isTransformerSubclass()`, `read()`, `declaresTransform()`. FQCN-string matching is real precedent: `FractalResponseUnbound::MANAGER_CLASS = 'League\Fractal\Manager'` — package-not-installed-safe. Registration order in `FractalPlugin` is `FractalResponseResolver` (attribute) → `EntityTransformerResponseResolver`; appending the new resolver keeps the attribute authoritative.

### 1. Whitelist boundaries — sound, one note

Each of the five shapes (helper-chained item/collection, facade `Fractal::create()->item/collection`, injected-Manager `new Item/Collection`, serializer terminal, terminal-chain walk) is test-pinned, and the transformer-arg extraction correctly restricts to `New_` / `ClassConstFetch ::class` resolved through `isTransformerSubclass()` — anything else refuses. **False-positive guard on non-Fractal `item()/collection()`:** the helper-chained case keys off the receiver being the `fractal()` **function call** (a `FuncCall` named `fractal`), and the facade case off the `Spatie\Fractalistic\Fractal` FQCN — so an unrelated `$repository->collection(...)` or `$builder->item(...)` (receiver is a variable/property, not `fractal()`/the facade) does not match. **Note to the coder:** make the helper-chained match assert the chain root is literally the `fractal` `FuncCall` (and the facade match the `Fractal` `StaticCall`/name), and add an explicit negative unit test — `$someService->item($x, new T())` and `$q->collection(...)` → null (not scanned). This is the one place a false positive could sneak in; pin it.

### 2. The refusal — correct, no envelope guessing

Bare two-arg `fractal($data, new T())` with no `->item()/->collection()` chain link refuses with a note (item-vs-collection genuinely not statically knowable from arg 1). The plan does NOT guess a default — matches the issue. The degrade test (`bare fractal($data, new T())` → null + note) pins it. The variable-transformer (`->item($x, $t)`), no-documentable-fields, and unknown-serializer refusals are each separately pinned. Good — every degrade path has a test.

### 3. Tier / scope / boundaries — correct

Tier-1 is right: bounded scan of the first N statements, no cross-call variable tracking, reusing #13 machinery — not Tier-2 dataflow. No `Contracts/**`, no stage-order, no config key, no public-signature change, no SP binding. `#[FractalResponse]` stays authoritative via the verbatim `PrimaryResponseAuthoringAttribute` guard; return-type guard reused verbatim. `src/Plugins/Fractal/` only — no cross-plugin or `Support/` leak.

### Note on the unbound lint rule (no action needed)

`FractalResponseUnbound` keys off an injected `Manager` parameter only; its own docblock says helper/facade shapes "never inject a Manager and therefore never trigger this rule". So this resolver landing does **not** create a lint interaction — the rule correctly stays silent on helper/facade usages. The plan rightly leaves the rule untouched. (Worth a one-line mention in the PR so it's not raised at code-review.)

### Bookkeeping

Per-shape unit tests + feature tests under `tests/Feature/Plugins/Fractal/` named; fixtures planned (reuse existing transformer fixtures where possible — good). Docs (Fractal plugin page: new recognised shapes + bare-helper refusal + `#[FractalResponse]` escape hatch) and CHANGELOG `[Unreleased]` Added both named. Survey delta likely ~0 (plugin commented out by default); additive-only if enabled — lead handles the gate.

Approve. Fold in the two notes (explicit `fractal`/`Fractal`-root assertion + negative test in §1; one-line unbound-rule mention in the PR). Neither is a re-plan.

Radiergummi:
🤖 **[lead]** plan-review approved (2 notes folded: add false-positive negative test for non-Fractal item()/collection() chains; PR line noting FractalResponseUnbound keys off injected Manager only, no lint interaction) → coding.

Radiergummi:
🤖 **[coder]** Note (lint interaction): `FractalResponseUnbound` (`fractal.response-unbound`) keys off an **injected `Manager` parameter** only — the helper (`fractal()`) and facade (`Fractal::create()`) shapes this PR now binds never inject a `Manager`, so they never trigger that rule. No lint interaction; the rule is correctly left untouched, and the docs note was reframed accordingly (it no longer claims the generator doesn't read those shapes).

🤖 **[coder]** deviation from plan (2 points, both verified):
1. **Fixtures use the real `spatie/laravel-fractal` dev dependency, not `@phpstan-ignore`.** The plan's draft fixtures leaned on `@phpstan-ignore-next-line`, but this repo sets `reportIgnoresWithoutComments: true` and bans `ignore.allLineErrors`. Since `league/fractal` + `spatie/laravel-fractal` are already `require-dev`, the `fractal()` helper, the `Fractal` facade, and `->respond()` all exist — the fixtures type-check cleanly with zero ignores, which is also more faithful to real consumer code.
2. **Facade FQCN match widened to include the laravel-fractal class.** Beyond the plan's `Spatie\Fractalistic\Fractal`, the matcher also accepts `Spatie\Fractal\Fractal` and both packages' `Facades\Fractal` aliases (still pure FQCN-string match, package-not-installed-safe). `spatie/laravel-fractal`'s `fractal()` returns `Spatie\Fractal\Fractal`, so this is the FQCN real apps actually emit.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed fd7fb57. Remote CI fully green (PHP 8.4/8.5 × Laravel 12/13, swagger-php 5.8, snapshot, coverage, quality). Ready for review.

Radiergummi:
🤖 **[lead]** coder-3 pushed fd7fb57, full CI matrix green (incl. swagger-php 5.8). 30 new tests incl. the false-positive negative test + all degrade paths. Two reasonable deviations recorded (real spatie/laravel-fractal dev dep vs line-ignores; facade FQCN widened to Spatie\Fractal\Fractal). → code-review, two independent approvals required.

Radiergummi:
🤖 **[reviewer]** review round 1: approved (1 non-blocking accuracy note).

Reviewed fd7fb57 adversarially; verified the diff delivers the plan I reviewed.

**Structure & registration — correct.** `FractalHelperResponseResolver` + `FractalCallShape` VO (`@internal`, readonly) modeled on `EntityTransformerResponseResolver`: same `returnTypeAllowsBodyScan()`, `PrimaryResponseAuthoringAttribute` early-return, `hasDocumentableFields()`, `note()` shape; all deps `#[Scoped]`, no new SP binding. Registered **last** in `FractalPlugin` (`FractalResponseResolver` → `EntityTransformerResponseResolver` → `FractalHelperResponseResolver`), so `#[FractalResponse]` stays authoritative. Reuses #13 machinery wholesale. `Plugins/Fractal/` only — no `Contracts`/stage/config/signature change.

**False-positive guard (note 1) — solid, genuinely exercised.** `matchChainedShape()` returns a shape **only after** `isFractalEntrypoint($node->var)` confirms the `->item()/->collection()` chain root is literally a `fractal` `FuncCall` (`$receiver->name->toString() === 'fractal'`) or a `Fractal::create()` facade `StaticCall` against a whitelisted FQCN — any other receiver returns null. The negative tests are real: `tests/Fixtures/Fractal/ArticleService.php` is an unrelated service exposing `item(Article, Transformer)` / `collection(array, Transformer)` with the same signatures, and `does not fire on item()/collection() called on a non-Fractal receiver` assert null against it. This is the highest false-positive risk and it is pinned.

**Refusal — every degrade path pinned, no envelope guessing.** Bare two-arg `fractal($data, new T())` has no `->item()/->collection()` link, so the chain walk never reaches a shape → null (refuses, no guess). Variable transformer (`transformerArgument` only accepts `New_`/`ClassConstFetch ::class` → null otherwise), no-documentable-fields, and unrecognised serializer (`serializerFrom` returns null → the `$serializer === null` refusal with note) are each separately tested. Terminal links (`->respond()`/`->toArray()`) are walked through transparently (`$node = $node->var`). The `serializeWith` serializer is captured outer→inner before the `item`/`collection` link resolves — ordering correct. JsonApi flips the media type to `application/vnd.api+json` (tested). 30 tests total.

**Deviation (1) — fixtures use the real dev dep — correct.** `composer.json` has `league/fractal ~0.20.2` + `spatie/laravel-fractal ^6.4.0` in require-dev; the repo bans bare line-ignores, so real-package fixtures over `@phpstan-ignore` is the right call.

**Deviation (2) — facade FQCN widening — accurate enough, ONE dead entry (non-blocking).** I checked the installed packages. Three of the four `FACADE_CLASSES` are real: `Spatie\Fractalistic\Fractal` (fractalistic base), `Spatie\Fractal\Fractal` (laravel-fractal subclass), `Spatie\Fractal\Facades\Fractal` (laravel-fractal facade). The fourth, **`Spatie\Fractalistic\Facades\Fractal`, does not exist** — fractalistic ships no `Facades/` namespace (its facade lives only under `Spatie\Fractal\Facades`). It is a **harmless dead string** (it can never match real code and cannot over-match, since over-matching needs a real unrelated class at that exact FQCN), so this is **not** a correctness or over-match defect — but it is one inaccurate entry. Recommend dropping it (or a one-line comment that it is speculative forward-compat). The serializer widening is accurate — `Spatie\Fractalistic\ArraySerializer` is a real class. Fast-follow, not a gate.

**Minor observation (no action):** `matchManagerResourceShape()` hardcodes `Serializer::DataArray` and does not read a Manager-style `serializeWith`/`setSerializer` — defensible, DataArray is Fractal's default and the plan scoped serializer detection to the chained terminal.

**Boundary & bookkeeping.** `fractal.response-unbound` correctly left untouched (it keys off injected `Manager` only; helper/facade shapes never trigger it — note 2 confirmed). CHANGELOG `[Unreleased]` Added entry accurate; `docs/plugins.md` documents all three shapes, the chain-root requirement, the bare-helper refusal, and the `#[FractalResponse]` escape hatch.

Verdict: **approved.** The dead facade FQCN is a recommended fast-follow cleanup, not a blocker.

Radiergummi:

Fresh independent review of diff fd7fb57 — read both new source files, traced the AST walk by hand, ran the suite, checked the deviations against the installed packages. Not coordinating with reviewer-1.

**False-positive guard (the headline risk) — sound, and proven.** `matchChainedShape()` walks the method-call chain outer→inner and only returns a match when it hits an `item`/`collection` link whose **receiver** passes `isFractalEntrypoint()` — i.e. the receiver is literally a `fractal()` `FuncCall` or a `Fractal::create()` `StaticCall` against a whitelisted FQCN. A `$this->service->item(...)` receiver is a `PropertyFetch` → fails the guard → the whole match returns null. Pinned both at unit level ("does not fire on item()/collection() called on a non-Fractal receiver", `HelperResolverFakeService`) and feature level (`NonFractalReceiverController` via `ArticleService`). I could not construct a non-Fractal receiver that slips through: the only accepted receiver node types are `FuncCall`(name=`fractal`) and `StaticCall`(class∈whitelist, method=`create`).

**Bare two-arg refusal — correct by construction.** `fractal($data, new T())->respond()` walks to the `fractal(...)` node, which is a `FuncCall`, not a `MethodCall` → the `while` never treats it as an item/collection link → returns null, falls to the Manager shape (no `new Item/Collection`) → null. Refused without guessing, as required. Pinned in both suites.

**Manager shape** is found via `NodeFinder` over the statements requiring **exactly one** `new Item`/`new Collection` (`count !== 1` refuses the ambiguous multi-resource case); the return expression there is `new JsonResponse(...)` which correctly fails `matchChainedShape` (outer node is `New_`) and falls through. `item`/`Item`→single, `collection`/`Collection`→collection envelope. Both pinned.

**Serializer ordering** is right: real chains are `->item(...)->serializeWith(...)->respond()`, so the outer→inner walk captures `serializeWith` *before* reaching the item link; default `DataArray` when absent; an unmodelled serializer (`serializerFrom` returns null → `FractalCallShape.serializer = null`) refuses with the "not one of the modelled cases" note. JsonApi flips both the envelope and the media type to `application/vnd.api+json` (test asserts the `application/json` slot is null).

**All degrade paths pinned (30 tests, 16 unit + 14 feature):** variable transformer, `::class` vs `new T()`, empty-transformer note, unknown-serializer note, non-HTTP named return type (`returnTypeAllowsBodyScan` refuses named non-`HttpFoundationResponse` types, and unions/intersections), `#[FractalResponse]` short-circuit. The feature suite additionally exercises the **facade** entrypoint end-to-end (the unit suite covers helper + Manager).

**Registration order — correct.** `FractalPlugin::register()` adds `FractalHelperResponseResolver` AFTER `FractalResponseResolver` and `EntityTransformerResponseResolver`, so `#[FractalResponse]` stays authoritative; belt-and-braces, the resolver itself also bails on any `PrimaryResponseAuthoringAttribute`.

**Deviation 1 (real dev-dep fixtures, not banned line-ignores) — sound.** `composer.json` require-dev already carries `league/fractal ~0.20.2` + `spatie/laravel-fractal ^6.4.0` (+ `suggest` rows), and the fixtures import the real `League\Fractal\Resource\Item/Collection`, serializers, and `Manager`. Cleaner than `@phpstan-ignore` lines and exercises the true class shapes.

**Deviation 2 (facade FQCN widening) — accurate, not harmfully over-matching.** Verified against vendor: `Spatie\Fractalistic\Fractal` (base, `static create()` at Fractal.php:62), `Spatie\Fractal\Fractal` (laravel-fractal, `extends Fractalistic`), and `Spatie\Fractal\Facades\Fractal` all exist. The fourth entry `Spatie\Fractalistic\Facades\Fractal` has no class in the installed fractalistic package (the facade ships in laravel-fractal) — it's a harmless dead whitelist entry (string compare only; never matches absent that class), defensible as defensive coverage of an alias a host app could register. Not a finding.

**Boundary / bookkeeping:** only 3 `src/` files, all under `Plugins/Fractal/`; no Contracts/Support/stage-order/config/public-signature change. `FractalResponseUnbound` rule code is byte-untouched (only its docs NOTE was updated to reflect the new binding — accurate, since the rule keys off an injected `Manager` carrying no literal transformer, which this resolver doesn't alter). All Fractal classes matched by FQCN string so neither package need be installed at generation time. strict-types + `@internal` on both new classes; modern PHP, region comments, no abbreviations, no issue-numbers-in-code; CHANGELOG `[Unreleased]` "Added" entry; `docs/plugins.md` documents all three shapes + the refusal honestly. PHPStan level 8 clean (0 errors), Pint clean, 30 Fractal tests green.

Verdict: clean. Independent second approval.

Radiergummi:
🤖 **[coder]** addressed: removed dead `FACADE_CLASSES` entry per reviewer-1 — `Spatie\Fractalistic\Facades\Fractal` doesn't exist (fractalistic ships no `Facades/` namespace; verified against vendor). The three real entries stay (`Spatie\Fractalistic\Fractal`, `Spatie\Fractal\Fractal`, `Spatie\Fractal\Facades\Fractal`). Folded into the rebase onto current main; pushed 182a5e8.

Radiergummi:
🤖 **[lead]** RECOMMEND MERGE — escalated only on the size rule (450-line new resolver > ~150 threshold). Everything else is green: full CI matrix incl. swagger-php 5.8 + Coverage (d14d831, CLEAN, 0 behind main), two independent reviewer approvals (false-positive guard + bare-two-arg refusal both verified), 30 tests, dead-FQCN fix folded in. **Risk is low**: the Fractal plugin is off by default, so this cannot affect default generation (survey N/A by construction); blast radius is limited to apps that explicitly enable it. Ready to merge on your nod — say the word and I'll finish-pr it.

Issue #31, opened 2026-06-01
Pagination query params from `->paginate()`


Radiergummi:
## Summary

When a controller action paginates a query — `->paginate()` or `->cursorPaginate()` — the generated operation should automatically gain the matching pagination query parameters (`page`/`per_page` for offset, `cursor` for cursor pagination) alongside the paginator response envelope.

## Motivation

Pagination is one of the most common patterns in real Laravel APIs, yet today the generator never sees it: the call lives in the controller body, and we only read signatures. The result is operations that return a paginated envelope (when a `@return Foo<Bar>` is present) but advertise no way to actually page through it — consumers and codegen tooling have no `page` or `cursor` knob to turn. Wiring the parameters off the same `->paginate()` call that already informs the response keeps the documented contract honest.

This is a **Tier 1** item: a bounded AST whitelist match on a specific method call, not open-ended body analysis.

## Proposed approach

This issue depends on the shared body-scan utility from the method-body epic (#5) — it is the AST whitelist that detects a `->paginate()` / `->cursorPaginate()` call in the action body. Once that utility reports a paginating call:

- Emit the pagination query parameters for the operation: `page` + `per_page` for the offset case, `cursor` for the cursor case. These should slot in next to the existing query-parameter resolution path so they merge cleanly with the default `CoreQueryParameterResolver` output and any `#[QueryParam]` declarations.
- Ensure the paginator envelope is applied as the response. `PaginatorResponseResolver` (`src/Plugins/Core/Resolvers/PaginatorResponseResolver.php`) and `PaginatorSchemaFactory` (`src/Plugins/Core/Support/PaginatorSchemaFactory.php`) already build the offset vs. cursor envelopes (keyed by `PaginatorKind`, `src/Enums/PaginatorKind.php`); the body-scan signal should select the correct kind when the `@return` annotation alone is ambiguous.

The item type itself is unchanged — it continues to come from `@return Foo<Bar>`.

## Acceptance criteria

- A paginating action emits the correct pagination query params plus the paginator envelope, with offset (`page`/`per_page`) and cursor (`cursor`) covered by distinct tests.
- Add unit coverage under `tests/Unit/Core/Pagination/` for `PaginatorResponseResolver` (offset and cursor kinds), alongside the existing `tests/Unit/Core/Pagination/PaginatorSchemaFactoryTest.php`, and building descriptors via the `tests/Support/` factories (`ActionDescriptorFactory`, `OperationNodeFactory`).
[...]

PR #364, opened 2026-06-15
feat(#353): body-scan-selected PaginatorKind response fallback (gated against resource regression)

### LINKED ISSUE #353: PaginatorKind response fallback when @return is ambiguous (body-scan-selected kind)
Split from #31 / PR #350 (Component 3 — re-scoped out of that PR as a live regression).

## Context

#31's secondary ask: when the `@return` annotation alone is ambiguous about the paginator kind, use the body-scan signal (`PaginatorCallReader::detect()`) to select the correct `PaginatorKind` for the **response envelope**.

Components 1+2 (pagination query parameters) shipped in PR #350 and resolve the Vito case end-to-end (ApiResources already supplies the `{data, links, meta}` envelope; only the query params were missing).

## Why this is escalation-bound (the regression)

`PaginatorResponseResolver` is registered at `CorePlugin` **before** ApiResources' `ResourceResponseResolver`; primary-response resolution is first-non-null-wins. Today `PaginatorResponseResolver` early-returns null when the return type is **not** a paginator, so a `ResourceCollection` return correctly falls through to ApiResources.

Naively extending it to consult `PaginatorCallReader::detect()` as a fallback kind would make Core fire on that same `ResourceCollection` return — i.e. Core would claim a response ApiResources must shape richly. That is a **live regression on the exact Vito shape #31 is meant to fix.**

## Required design constraints

- Must not override the ApiResources `{data, links, meta}` resource envelope.
- The plan's candidate mitigation: only fall back when the return type is **not** a `ResourceCollection`/`Data` type AND an explicit `@return Foo<Bar>` generic or `#[ResponseResource]` supplies an item class — otherwise log + return null exactly as today.
- This touches cross-plugin primary-response precedence, so it needs a human gate (not auto-merge).

## Acceptance

- A test proving the ambiguous-`@return` → body-selected-kind path emits the correct envelope.
- A regression test proving a `ResourceCollection` return still falls through to ApiResources unchanged.
- Snapshot byte-match + 3.1 validation + lint-clean.

Radiergummi:
## Implementation plan — #353 PaginatorKind response fallback (body-scan-selected kind)

> ⚠️ **ESCALATION-BOUND — NOT an auto-merge candidate.** Cross-plugin primary-response precedence;
> documented live-regression risk on the Vito shape #31/#350 fixed. Requires a human gate (Moritz).
>
> **Revised after plan-review (reviewer-1):** the first plan's return-type guard leaked via the
> `#[ResponseResource]` attribute path. This revision closes it and resolves the
> `ResourceTargetLocator`-vs-self-contained-guard fork with a verified decision.

### The regression we must not cause

`CorePlugin` registers `PaginatorResponseResolver` (line 71) **before** ApiResources'
`ResourceResponseResolver`; resolution is **first-non-null-wins**. Today `PaginatorResponseResolver`
early-returns null when `PaginatorKind::fromClass($returnType)` is null, so resource returns fall
through to ApiResources' rich `{data, links, meta}` envelope. A naive body-scan fallback would make
Core fire on a return ApiResources must shape — the live regression.

### The leak reviewer-1 found (and this revision closes)

ApiResources does **not** gate on return type. `ResourceClassLocator::locate()` checks
`readResponseResource()` **first** (`src/Plugins/ApiResources/Support/ResourceClassLocator.php:65`)
and returns a `ResourceTarget` whenever a method- **or controller-level** `#[ResponseResource(X)]`
names a `JsonResource` subclass — **regardless of the declared return type**. So:

```php
public function index(): JsonResponse          // non-resource, non-paginator
#[ResponseResource(AuthorResource::class)]      // AuthorResource extends JsonResource
{ return AuthorResource::collection(Author::query()->paginate()); }
```

The first plan's guards all pass (`JsonResponse` isn't a resource type; `resolveItemClass()` reads
`#[ResponseResource]` → `AuthorResource`; body has `paginate()`) → Core returns non-null and steals
the response ApiResources claims via the attribute. Same `{data,links,meta}` regression, via the
attribute path. The first plan's regression test (a resource *return type*) would NOT exercise this.

### Design fork — DECISION: option (a)-with-extension (self-contained guard), NOT the locator gate

I verified the facts behind the fork:

- `Contracts\Routing\ResourceTargetLocator` **is always bound** to `ResourceClassLocator`
  (`OpenApiServiceProvider:543`), and `registerApiResourcesPlugin()` is called **unconditionally**
  (`:187`, same "always bound, inert until listed in config" pattern as Fractal/SwaggerPhp). So a
  `locate()` call from Core would not fatal even when ApiResources is disabled. Option (b) is
  *technically wireable*.
- **But** `locate()` is purely reflection/code-shape based — it returns a target from the attribute
  or resource return type **independent of whether the ApiResources plugin is active**. A bare
  `locate() === null` gate would therefore suppress Core's fallback even when ApiResources is
  **disabled** (its resolver never runs) — i.e. Core would refuse to document a paginator the user
  *does* want documented, with no envelope being produced by anyone. That contradicts the intended
  behaviour the lead stated ("when ApiResources is off, Core proceeding is correct"). Closing that
  would require Core to also know whether ApiResources is in the active resolver chain — more
  coupling than the leak is worth.

**Decision: option (a)-with-extension.** Keep Core's guard self-contained and import-free; extend it
to fence the `#[ResponseResource]`-names-a-resource case. This behaves correctly whether ApiResources
is on or off (the guard is about the **code shape Core refuses to claim**, not another plugin's
runtime state). The cost — the guard manually mirrors ApiResources' claiming surface — is real and is
**documented as explicit coupling** in a code comment, so the next change to `ResourceClassLocator`'s
surface updates both. (SpatieData needs no attribute extension: `DataResponseResolver` claims purely
by `is_a Data`/`DataCollection` **return type** — already covered by guard 1; it has no
attribute-path leak. Verified.)

### Concrete approach

In `PaginatorResponseResolver::resolvePrimaryResponse()`, the `$kind !== null` path stays
**byte-identical**. In the `$kind === null` branch (currently an immediate `return null`), apply the
gated fallback, **in this order**:

1. **Resource/Data return-type guard.** If the declared return type `is_a` (string form,
   package-safe) any of `Illuminate\Http\Resources\Json\JsonResource`,
   `…\ResourceCollection`, `…\AnonymousResourceCollection`, or Spatie
   `Data`/`DataCollection`/`PaginatedDataCollection`/`CursorPaginatedDataCollection` → `return null`.
2. **`#[ResponseResource]`-names-a-resource guard (the leak fix).** Read the method-level **and**
   controller-level `#[ResponseResource]` (mirror `ResourceClassLocator::readResponseResource()`'s
   two-level read). If one names a class that `is_a JsonResource` → `return null`. This is exactly
   ApiResources' attribute-claiming surface; document the mirror.
3. **Item class.** `resolveItemClass()` (unchanged: `#[ResponseResource]` then `@return` generic). If
   null → log + `return null` (today's behaviour). NOTE: by step 2, any `#[ResponseResource]` that
   survives to here names a **non-resource** item (e.g. a bare model), which ApiResources does not
   claim — safe for Core to consume.
4. **Body-scan kind.** `PaginatorCallReader::detect($method)`. Null (no unconditional paginate call)
   → log + `return null`. Otherwise build `schemaFactory->envelope($detectedKind, itemsFor($itemClass))`
   and return the 200 response exactly as the `$kind !== null` path does.

Net: the fallback fires only at the intersection {non-paginator return type} ∧ {non-resource/Data
return type} ∧ {no resource-naming `#[ResponseResource]`} ∧ {resolvable item class} ∧ {unconditional
`paginate()` body} — which excludes BOTH the Vito resource-return shape (step 1) and the
attribute-path shape (step 2).

**Body-only resource detection does not leak** (confirmed with reviewer-1): `ResourceClassLocator`
reads the return *expression* for a resource only inside the `is_a($name, JsonResource)` return-type
branch; a `JsonResponse`/untyped return with a `SomeResource::collection(...)` body is NOT claimed by
ApiResources, so Core firing there is not a regression. No guard needed for that.

### Files

**Modify `src/Plugins/Core/Resolvers/PaginatorResponseResolver.php`:**
- Inject `PaginatorCallReader` (add to the explicit binding at `OpenApiServiceProvider:380`; the
  resolver is `final readonly`, built there with `refSchemaResolvers`).
- Restructure the `$kind === null` branch per the four-step approach. Add two private guard helpers
  (`returnTypeIsResourceOrData()`, `declaresResourceResponseResource()` — the latter reading method-
  and controller-level `#[ResponseResource]`). Reuse `resolveItemClass()` / `itemsFor()` unchanged.
  Comment binding guard 2 to `ResourceClassLocator::readResponseResource()` as an intentional mirror.
- Update the class docblock to document the fallback + both guards.

**Modify `src/OpenApiServiceProvider.php`:** add `PaginatorCallReader` to the
`PaginatorResponseResolver` binding (line 380).

### Tests (TDD — failing first), in `tests/Feature/PaginatorResponseTest.php`

Acceptance + the regression surface reviewer-1 requires:

- **happy fallback (kind from body)**: non-resource, non-paginator return type (or untyped) +
  `@return`/`#[ResponseResource]` naming a **non-resource** item (a model) + `simplePaginate()` body
  → **Simple** envelope; `cursorPaginate()` → Cursor. (Item resolver path stays valid.)
- **REGRESSION 1 — resource return type**: `ResourceCollection`/`AnonymousResourceCollection` return
  + `paginate()` body → Core returns null, ApiResources `{data,links,meta}` lands (assert `links`+`meta`).
- **REGRESSION 2 — attribute path (the leak fix, the test that proves it)**:
  `#[ResponseResource(AuthorResource::class)]` (a JsonResource subclass) on a **non-resource return
  type** (`: JsonResponse`) + `paginate()` body → Core returns null, ApiResources `{data,links,meta}`
  lands (assert `links`+`meta`). **Plus the controller-level `#[ResponseResource]` variant** (the
  attribute on the controller class, not the method), since `readResponseResource` checks both.
- **degrade — no item class**: non-resource return + `paginate()` body, no `@return`/attribute → null
  + warning.
- **degrade — no paginate call**: non-resource return + item class, no paginate body → null.
- **unchanged declared-paginator path**: existing `LengthAwarePaginatorContract`/`CursorPaginatorContract`
  tests pass byte-identically (pins the `$kind !== null` path).

Snapshot: `tests/Feature/ExamplesTest.php` `snapshot` group must stay byte-identical; if any snapshot
endpoint hits the new fallback, the guard is too loose — investigate, don't regenerate.

### Verification

```
vendor/bin/pest tests/Feature/PaginatorResponseTest.php
vendor/bin/pest tests/Feature/ExamplesTest.php --group snapshot
composer test && vendor/bin/pint --test && composer lint
```
Plus the issue's gate: 3.1 validation (snapshot flow) + lint-clean.

### Docs / CHANGELOG

- `docs/auto-derivation.md` — pagination/response section: the body's `paginate()`-family call
  selects the envelope kind when the declared return type leaves it ambiguous; resource/Data returns
  **and** resource-naming `#[ResponseResource]` actions are never claimed by this path (ApiResources
  owns them).
- `CHANGELOG.md [Unreleased]` "Added": paginator response envelopes select their kind from the body's
  `paginate()`-family call when the return type is kind-ambiguous, gated so resource/Data collection
  returns and `#[ResponseResource]`-on-a-resource actions stay with the ApiResources/SpatieData plugins.

### Survey gate

Generation-affecting (area:responses/plugins). Intended deltas: kind-corrections (simple/cursor) on
bare-paginator returns. **RED FLAG**: any `ResourceCollection` **or** `#[ResponseResource]` endpoint
losing its `links`/`meta` in the survey diff = guard leaked. Reviewer + human gate must scan the diff
for resource-envelope regressions on BOTH paths.

### Controversy — HIGH (human gate required)

Cross-plugin precedence + live-regression risk; no Contracts/stage/config/signature change. The new
behaviour is confined behind two guards, each with a dedicated regression test. The accepted residual
risk is the self-contained guard being a **manual mirror** of `ResourceClassLocator`'s claiming
surface (return types + `#[ResponseResource]` at two levels) — documented in-code as coupling so a
future locator change updates both. Reviewer's job: confirm (1) both guards are airtight against
ApiResources' claiming surface, (2) the `$kind !== null` path is byte-unchanged, (3) snapshot +
survey show no resource-envelope regressions on either path.

Closes #353


Radiergummi:
🤖 **[planner]** plan posted — see PR body. **Ready for plan-review — and flagged escalation-bound: this is NOT an auto-merge candidate.** It touches cross-plugin primary-response precedence and carries a documented live-regression risk on the Vito `ResourceCollection` shape that #31/#350 fixed. The plan is built around a return-type guard (resource/Data types are fenced out of Core's fallback) with a dedicated regression test pinning that shape. Needs a careful human gate.

Radiergummi:
🤖 **[lead]** planning done (PR #364) → plan-review. Escalation-bound: reviewer to stress-test the resource/Data return-type guard that prevents Core from claiming the ApiResources {data,links,meta} envelope (the live-regression risk on the Vito shape).

Radiergummi:
🤖 **[reviewer]** plan-review: **CHANGES REQUIRED** — the resource guard is **not airtight**. I found a concrete leak the regression test would miss. Escalation-bound; this must be closed before the human gate, not after.

I traced what ApiResources actually claims (`ResourceResponseResolver` → `ResourceClassLocator::locate()`), not just the return-type list, because the regression risk is "anything ApiResources would shape that Core now steals."

### The leak: `#[ResponseResource]` on a non-resource return type

The plan's guard fences resource/Data **return types** (`is_a` JsonResource/ResourceCollection/AnonymousResourceCollection/Data…). But ApiResources does **not** gate on return type — `ResourceClassLocator::readResponseResource()` (checked FIRST, at **both** method and controller level) returns a `ResourceTarget` whenever a `#[ResponseResource(SomeResource::class)]` names a `JsonResource` subclass, **regardless of the declared return type**.

Now walk Core's proposed fallback for this action:
```php
public function index(): JsonResponse        // non-resource, non-paginator class
#[ResponseResource(AuthorResource::class)]   // AuthorResource extends JsonResource
{
    return AuthorResource::collection(Author::query()->paginate());
}
```
- Core line 57: `JsonResponse` is a class, not builtin → no early return.
- Core line 61: `PaginatorKind::fromClass('JsonResponse')` → null → enters the new fallback.
- **Guard 1 (return-type `is_a`):** `JsonResponse` is NOT a resource/Data type → **guard passes** (does not fence).
- **Guard 2 (`resolveItemClass()`):** reads `#[ResponseResource]` first → returns `AuthorResource::class` → **passes**.
- **Guard 3 (`PaginatorCallReader::detect()`):** body has `paginate()` → kind detected.
- → **Core builds an envelope and returns non-null. Core runs BEFORE ApiResources → Core steals the response.**

But ApiResources' locator claims this exact action via `readResponseResource` → `{data, links, meta}` envelope. So this is the **same class of regression as the Vito shape**, just reached through the `#[ResponseResource]` attribute instead of a `ResourceCollection` return type. The planned regression test (a `ResourceCollection` *return type* + `paginate()` body) **does not exercise this path** and would pass while the bug ships.

This is not hypothetical: `#[ResponseResource]` is the documented escape hatch for exactly when the return type does *not* express the resource (e.g. `: JsonResponse`), so attribute-without-resource-return-type is its primary use.

### Required fix

The guard must also fence the case where **ApiResources would claim the action via `#[ResponseResource]`**. Concretely, in the `$kind === null` fallback, before consulting the body scan, return null if **either**:
- the declared return type `is_a` a resource/Data type (the planned guard 1), **OR**
- a `#[ResponseResource]` (method- or controller-level) names a class that `is_a JsonResource` (ApiResources territory). Spatie's equivalent attribute, if any, same treatment.

In other words: Core's fallback may consume `#[ResponseResource]` for the **item class** ONLY when that attribute does **not** name a `JsonResource`/Data subclass (a bare model item is fine — ApiResources won't claim a non-resource item). The cleanest framing: *if `resolveItemClass()` resolved its class from a `#[ResponseResource]` whose class is a `JsonResource`/Data subclass, refuse* — that is precisely the resource territory.

### Reconsider the `ResourceTargetLocator` rejection in light of this

The plan rejects injecting `ResourceTargetLocator` (a `Contracts\Routing` interface) as "fragile under plugin reordering." But this leak is exactly the drift the hardcoded list invites: the guard duplicates *part* of ApiResources' claiming logic (return types) and **misses the rest** (the attribute, method+controller level; tomorrow, whatever else the locator learns to read). Two defensible options — reviewer-2 and the human gate should weigh in:
- **(a)** Extend the self-contained guard to also cover the `#[ResponseResource]`-names-a-resource case (above). Minimal, keeps Core import-free, but the guard remains a *manual mirror* of ApiResources' surface — document that coupling explicitly so the next locator change updates both.
- **(b)** Gate on `ResourceTargetLocator::locate($descriptor) === null` when the contract is bound. This is the **authoritative** "would ApiResources claim this?" test — it cannot drift because it IS ApiResources' logic. The plan's "fragile under reordering" concern is weak: the locator is a `Contracts\Routing` seam (not a plugin import), and a null check is precedence-agnostic. I lean (b) on correctness grounds; if the team prefers self-containment, (a) is acceptable **only with the `#[ResponseResource]` extension and a comment binding it to the locator's surface.**

### Other checks (these are fine)

- **Body-only resource detection does NOT leak.** `ResourceClassLocator` reads the return expression only inside the `is_a($name, JsonResource::class)` branch (a base/collection resource return type). A `JsonResponse`/untyped return with a `SomeResource::collection(...)` body is NOT claimed by ApiResources — so Core firing there is not a regression. Good; the gap is solely the attribute path.
- **`$kind !== null` path** is to be left byte-identical — correct, and the "unchanged declared-paginator tests" assertion pins it.
- **Tier 1** correct (reuses the shipped `PaginatorCallReader::detect()`); no Contracts/stage/config/signature change.

### Required test additions

Beyond the planned `ResourceCollection`-return regression test, add the **attribute-path regression test**: `#[ResponseResource(SomeResource::class)]` (resource subclass) on an action with a **non-resource return type** (`JsonResponse`/untyped) + a `paginate()` body → **Core returns null and ApiResources' `{data, links, meta}` envelope lands** (assert `links`+`meta` present). This is the test that proves the fix; without it the leak is invisible. Add the controller-level `#[ResponseResource]` variant too, since `readResponseResource` checks the controller.

Verdict: **changes required** — close the `#[ResponseResource]` leak (option a-with-extension or b) and add the attribute-path regression test before this goes to the human gate. The rest of the plan (byte-unchanged declared path, tier, snapshot/survey vigilance) is sound.

Radiergummi:
🤖 **[lead]** plan-review found a concrete guard leak (#[ResponseResource] on a non-resource return + paginate body → Core steals the ApiResources envelope). Back to planning to close the leak + resolve the ResourceTargetLocator-vs-hardcoded-guard fork before it reaches the human gate.

Radiergummi:
🤖 **[planner]** revised — closes the `#[ResponseResource]` leak reviewer-1 found. Stays escalation-bound.

**Leak fix:** added a second guard in the `$kind === null` fallback that fences the case where ApiResources would claim the action via the attribute. Concretely, before the body scan, refuse if a method- **or controller-level** `#[ResponseResource]` names a `JsonResource` subclass — mirroring `ResourceClassLocator::readResponseResource()`'s two-level read (which `locate()` checks FIRST, regardless of return type). Core may consume `#[ResponseResource]` for the item class ONLY when it names a non-resource (e.g. a bare model), which ApiResources never claims.

**Design fork — chose option (a)-with-extension (self-contained guard), NOT the locator gate.** I verified the facts:
- `ResourceTargetLocator` IS always bound to `ResourceClassLocator` (`OpenApiServiceProvider:543`, `registerApiResourcesPlugin()` called unconditionally at :187), so option (b) is technically wireable — Core calling `locate()` wouldn't fatal even with ApiResources disabled.
- **But** `locate()` is purely reflection/code-shape based, **independent of plugin-enabled state**. A bare `locate() === null` gate would suppress Core's fallback even when ApiResources is DISABLED (its resolver never runs, no envelope is produced by anyone) — contradicting the intended 'Core proceeds when ApiResources is off' behaviour. Fixing that would require Core to additionally know whether ApiResources is in the active chain — more coupling than the leak is worth.
- So (a)-with-extension behaves correctly whether ApiResources is on or off (the guard is about the code shape Core refuses to claim, not another plugin's runtime state). Residual cost: the guard manually mirrors ApiResources' claiming surface — documented as in-code coupling so a future `ResourceClassLocator` change updates both.
- SpatieData needs no attribute extension: `DataResponseResolver` claims purely by `is_a Data`/`DataCollection` **return type** (already covered by guard 1); no attribute-path leak. Verified.

**Tests added** (the ones that prove the fix): attribute-path regression — `#[ResponseResource(AuthorResource::class)]` on `: JsonResponse` + `paginate()` body → Core returns null, ApiResources `{data,links,meta}` lands (assert links+meta); plus the controller-level `#[ResponseResource]` variant. Kept the resource-return-type regression test and the byte-unchanged declared-path assertion.

Ready for re-review by reviewer-1.

Radiergummi:
🤖 **[lead]** plan revised (leak closed via #[ResponseResource] guard extension; fork resolved to option (a) with a concrete reason the locate()===null gate fails when ApiResources is disabled) → plan-review round 2.

Radiergummi:
🤖 **[reviewer]** plan-review round 2: **APPROVED** (escalation-bound — for the human gate). My leak is closed and the design-fork reasoning is correct; I verified every load-bearing claim against the code.

### 1. Does the fix close my leak? — YES, at both levels.

The revised `$kind === null` fallback adds the guard I asked for: before the body scan, it refuses if a method- **or controller-level** `#[ResponseResource]` names an `is_a JsonResource` class — a direct mirror of `ResourceClassLocator::readResponseResource()`, which reads the attribute at **both** levels (reflector first, then `$descriptor->controller`) and claims **regardless of return type**. Core now consumes `#[ResponseResource]` for its item class **only** when the class is a non-resource (bare model) — exactly the set ApiResources never claims. Re-walking my original counterexample (`#[ResponseResource(AuthorResource::class)]` on `: JsonResponse` + `paginate()` body): guard 1 passes (`JsonResponse` not a resource type), **guard 2 now fires** (`AuthorResource is_a JsonResource`) → Core returns null → ApiResources shapes `{data,links,meta}`. Fenced. The two attribute-path regression tests (method-level + controller-level) pin precisely this; without guard 2 they fail. Good.

### 2. The fork — planner OVERRULED my `locate()===null` preference; the reasoning is CORRECT.

I verified the two facts the decision rests on:
- **`ResourceTargetLocator` is unconditionally bound.** `register()` calls `registerApiResourcesPlugin()` at `OpenApiServiceProvider:187` with no config gate; that method binds `Contracts\Routing\ResourceTargetLocator` → `ResourceClassLocator` (`:543`). So `locate()` is always callable — option (b) wouldn't fatal.
- **But `locate()` is plugin-state-INDEPENDENT.** It is pure reflection/code-shape resolution; the **binding** being present is distinct from ApiResources being an **active plugin** (plugin enablement via `config('openapi.plugins')` governs whether `ResourceResponseResolver` is in the response chain, NOT whether the locator service is bound). Therefore with **ApiResources disabled**, `locate()` still returns non-null on a resource-shaped action, but `ResourceResponseResolver` never runs — nobody emits the envelope. A bare `locate() === null` gate would suppress Core's fallback there too, leaving the response **undocumented** — strictly worse than today, and contradicting the "Core proceeds when ApiResources is off" contract (which follows from CLAUDE.md: the package must stay functional with a plugin disabled). To rescue (b), Core would additionally need to know whether ApiResources is in the active chain — more cross-plugin coupling than the leak warrants.

So my (b) preference **was unsound for the disabled-plugin case**, and I withdraw it. Option (a)-with-extension is the correct call: the guard is about *the code shape Core declines to claim*, which is the right invariant whether ApiResources is on or off. The residual cost — the guard manually mirrors ApiResources' attribute-claiming surface — is real but bounded; the plan commits to documenting that coupling in-code so a future `ResourceClassLocator` change updates both. Accept that as the documented tradeoff. (If anyone later wants drift-proofing, the clean move is a shared `Contracts` predicate "would a resource resolver claim this *shape*?" — out of scope here.)

### 3. SpatieData bonus claim — CONFIRMED.

`DataResponseResolver` claims purely by **return type** (`Data` subclass / `DataCollection`/union of Data) — no `#[ResponseResource]`-style attribute path — so guard 1 covers it; no guard-2 analogue needed. Verified in the resolver. Note the benign interaction I checked: `PaginatedDataCollection`/`CursorPaginatedDataCollection` are recognised by `PaginatorKind::fromClass()`, so they take Core's **`$kind !== null`** path (byte-unchanged), never the new fallback — and `DataResponseResolver` deliberately returns null for them so Core claims them via the shared ref chain. The new guard doesn't perturb that.

### Residual confirmations for the human gate

- `$kind !== null` path is to remain byte-identical (declared-paginator tests pin it). ✓ planned.
- Tier-1 (reuses shipped `PaginatorCallReader::detect()`); no Contracts/stage/config/signature change. ✓
- Snapshot byte-match + 3.1 validation + survey vigilance: **any** `ResourceCollection`/Data endpoint losing `links`/`meta` in the survey diff is the red flag — confirmed as the gate's scan target.

Verdict: **approved for the human gate.** The guard is now airtight against both the return-type and attribute claiming paths, the fork is resolved correctly with a verified reason, and the proving tests pin the previously-invisible attribute leak.

Radiergummi:
🤖 **[lead]** plan-review round 2 approved (leak fenced both levels; option (a)-with-extension confirmed correct; SpatieData clear) → coding. Stays escalation-bound: human gate owns the merge (cross-plugin precedence).

Radiergummi:
🤖 **[coder]** implemented; local gates fully green; pushed d282596.

**Implementation** (exactly per the twice-reviewed plan): in `PaginatorResponseResolver`, the `$kind === null` branch now falls back to `PaginatorCallReader::detect()` for the response envelope, gated by two guards that each return null (preserving pre-#353 behaviour) so Core (which runs first) never steals a response ApiResources/SpatieData would shape:
- **Guard 1 (return type):** `is_a` JsonResource / ResourceCollection / AnonymousResourceCollection / Spatie `Data` | `DataCollection` | `PaginatedDataCollection` | `CursorPaginatedDataCollection` (FQCN-string match, package-safe) → null.
- **Guard 2 (attribute):** a method- or controller-level `#[ResponseResource]` naming an `is_a JsonResource` class → null (deliberate mirror of `ResourceClassLocator::readResponseResource()`s two-level read, marked as documented coupling in a code comment). Core consumes `#[ResponseResource]` for the item class only when it names a non-resource bare model.
- Fallback fires only at: non-paginator return ∧ non-resource type ∧ no resource-naming `#[ResponseResource]` ∧ declared item class ∧ unconditional `paginate()` body. The declared-paginator (`$kind !== null`) path is byte-identical.
- Wiring: `PaginatorCallReader` added to the explicit SP binding.

**Tests** (`tests/Feature/PaginatorResponseFallbackTest.php`, 6): critical Vito-shape regression (`ResourceCollection` + `AnonymousResourceCollection` return + paginate body → ApiResources `{data,links,meta}` lands); the attribute-leak regression at BOTH method- and controller-level (`#[ResponseResource(resource)]` on `JsonResponse` → ApiResources `{data:$ref}` lands, no Core array/links/meta); positive (bare model item class + paginate body → length-aware envelope); degrade (no item class → bare 200).

**Local gates green:** Pint passed, PHPStan level 8 no errors, **2686 tests passed including the byte-exact `snapshot` group and all 50 ApiResources feature tests** — the exact regression surface. No raw nested-array OA schemas emitted (routes through `PaginatorSchemaFactory` / resource `$ref`), so no 5.8-specific risk.

⚠️ **Remote CI did not trigger** for the pushed head (888a1c7, then empty-commit d282596). The PR `statusCheckRollup` is empty and no `pull_request`/`synchronize` run appears for either SHA, though Actions is healthy repo-wide (other PRs ran minutes ago). Toggling draft↔ready and an empty-commit re-push did not fire it. Flagging for the lead — appears to be a dropped webhook on GitHub's side, not a gate failure.

Radiergummi:
🤖 **[coder]** rebased onto origin/main (now includes #363 Fractal, tip 1aaff42); CHANGELOG conflict resolved by keeping both entries (#150 + #353). Dropped the earlier empty CI-retrigger commit — the force-push fired a fresh `synchronize`, which resolved the dropped-webhook issue. Local `composer check` green (Pint, PHPStan level 8, 2723 tests). **Remote CI now all-green on the rebased tip a931731** — full matrix (PHP 8.4/8.5 × Laravel 12/13), swagger-php 5.8, the byte-exact Snapshot job, Coverage, and quality. Ready for review on a current base.

Radiergummi:
🤖 **[lead]** rebased onto current main (a931731), full CI matrix green incl. swagger-php 5.8 AND the regression-critical byte-exact Snapshot job (no resource/Data endpoint lost links/meta) → code-review, two independent approvals required. Stays escalation-bound (Moritz owns the merge).

Radiergummi:
🤖 **[reviewer]** review round 1: approved (escalation-bound — Moritz owns the merge). 1 non-blocking accuracy note.

Reviewed a931731 against the plan I approved in round 2, and **empirically verified the leak fence** (not just by reading).

**Matches the approved plan.** The `$kind === null` fallback runs `kindFromBody()`, which applies **both** guards before `PaginatorCallReader::detect()`:
- **Guard 1** — `isResourceOrDataType()` FQCN-string `is_a` against JsonResource / ResourceCollection / AnonymousResourceCollection / Spatie Data | DataCollection | PaginatedDataCollection | CursorPaginatedDataCollection → null. Package-safe, Core stays import-free.
- **Guard 2** — `namesResponseResourceClass()` reads `#[ResponseResource]` at method level then controller level (mirrors `ResourceClassLocator::readResponseResource()`), returns null when the named class `is_a JsonResource`. The deliberate-mirror coupling is documented in-code as I asked.

**My leak case is fenced — and I proved the test pins it.** I ran the suite at the PR commit (6/6 pass), then **disabled guard 2 in a throwaway worktree** and re-ran: the method-level attribute test (`#[ResponseResource(NestedAuthorResource)]` on `: JsonResponse` + `paginate()` body) **fails** — `data.$ref` becomes null because Core emits a paginated array envelope instead of ApiResources' `{data: $ref}`. That is exactly the regression I flagged in round 1; the test would catch a reintroduction. With guard 2 restored, ApiResources shapes the response. Confirmed fenced.

**`$kind !== null` path byte-identical.** The only edit inside `resolvePrimaryResponse()` replaces the bare `return null` with the guarded `$kind = $this->kindFromBody(...)` assignment; the declared-paginator branch never enters the fallback and the downstream `resolveItemClass()`/envelope build is untouched. Byte-identical, as required.

**Tests (6) pin the surface:** both resource-return-type regressions (ResourceCollection + AnonymousResourceCollection → `{data,links,meta}`, `data.items.$ref`), method- and controller-level attribute deferrals (→ ApiResources `{data:$ref}`, no `links`/`meta`), the positive bare-model path (`#[ResponseResource(Author::class)]` → Core's length-aware envelope, proving guard 2 does **not** over-fence a non-resource item), and the no-item-class degrade (→ bare 200, no `application/json`). Scope clean: **no Contracts / stage-order / config / Plugin / public-signature change** (verified against the file list).

**Note (non-blocking, accuracy):** the controller-level branch of guard 2 is currently **belt-and-suspenders, not closing a live leak.** Core's `resolveItemClass()` reads `#[ResponseResource]` only from the **method** reflector (no controller fallback), so a controller-level-only attribute already yields no item class → Core returns null regardless of guard 2 (the controller-level test passes with guard 2 disabled too). The branch is still correct and worth keeping — it forward-proofs the mirror against `resolveItemClass` ever gaining controller-level read, and it honestly mirrors the locator. Just flagging that the controller-level *test* doesn't exercise a currently-reachable leak the way the method-level one does. No change required.

Verdict: **approved.** The guard is airtight where the leak can occur, appropriately defensive beyond it, and the proving test demonstrably fails without the fix. Stays escalation-bound for the human gate.

Radiergummi:
🤖 **[reviewer]** review round 1 — 1 finding (test-suite gap; non-blocking, approving):

Fresh independent review of a931731 (read the resolver + both convention plugins' claiming surfaces, traced precedence, mutation-tested the guards). I went hunting for a return shape that slips past the guards and found none — the boundary is drawn correctly. One real coverage gap below.

**No leak found — the guards mirror the plugins' actual claiming surface exactly.** I verified what each downstream resolver keys off:
- ApiResources' `ResourceClassLocator::locate()` claims a response **only** when (a) `#[ResponseResource]` names an `is_a JsonResource` class (its `readResponseResource` is the same two-level method→controller read Guard 2 mirrors), or (b) the **declared return type** `is_a JsonResource` (`JsonResource` / `ResourceCollection` / `AnonymousResourceCollection`, incl. `#[Collects]`/`$collects` subclasses — all caught by Guard 1's `is_a(..., ResourceCollection, true)`). It does **not** claim off `@return` PHPDoc generics.
- SpatieData's `DataResponseResolver` claims off `getReturnType()` when `is_a Data` or `is_a DataCollection` — both (and the two paginated variants) covered by Guard 1.

So Guard 1 (declared return type) ∪ Guard 2 (`#[ResponseResource]`→JsonResource) is exactly the union of the two plugins' claiming triggers. Adversarial cases I checked and that hold:
- **Union return types** (`LengthAwarePaginator|AnonymousResourceCollection`): refused at the top by `!$returnType instanceof ReflectionNamedType` before any kind/guard logic — no claim either way (pre-existing, unchanged).
- **`@return AnonymousResourceCollection<X>` docblock on a `JsonResponse` return, no attribute**: ApiResources would *not* claim it (it reads the native return type, not the generic), so Core's model-paginated envelope is additive, not a steal.
- **Inherited / custom `ResourceCollection` subclass return type**: `is_a(..., ResourceCollection, true)` catches it.
- **Bare-model `#[ResponseResource(Model::class)]`**: correctly *passes* both guards (Guard 2 only fires on JsonResource names) and Core's body-scan shapes the model envelope — the documented intended consumption, pinned by `FallbackModelItemController`.

**Declared-paginator path is byte-unchanged.** The only edit inside `resolvePrimaryResponse` is replacing the lone `return null;` in the `$kind === null` branch with the `kindFromBody(...)` call; when `$kind !== null` (a real paginator return type, incl. Spatie `PaginatedDataCollection` recognised by `PaginatorKind::fromClass`) the flow is identical to pre-#353. Confirmed by reading the diff.

**Finding 1 (test gap — recommend, not block): Guard 1 has no test that fails without it.** I mutation-tested both guards. Neutralising **Guard 2** flips the method-level `#[ResponseResource]` deferral test red (Core steals → `{data:[…],links,meta}`), so Guard 2 is load-bearing-and-pinned. But neutralising **Guard 1 alone leaves all 6 tests green.** The two collection-return fixtures (`FallbackResourceCollectionController` / `FallbackAnonymousCollectionController`) don't isolate Guard 1: they carry no `#[ResponseResource]` and no `@return Paginator<Item>` generic, so `resolveItemClass()` returns null and Core bails via the "item type undeclared" warning regardless of Guard 1 — its protection is masked. The case that *would* fail without Guard 1: a resource-typed return **with a Core-resolvable item class**, e.g.

```php
#[ResponseResource(Author::class)]            // bare model → passes Guard 2
public function index(): AnonymousResourceCollection // is_a JsonResource → only Guard 1 stops Core
{
    return NestedAuthorResource::collection(Author::query()->paginate());
}
```

Without Guard 1, Core resolves `Author` as the item and (running first) steals the `{data:[Author],…}` envelope from the ApiResources `{data:[NestedAuthorResource],links,meta}` shape. Suggested concrete test: add that fixture and assert `data.items.$ref === '#/components/schemas/NestedAuthorResource'` (ApiResources won), which fails today if Guard 1 is removed. Low risk — Guard 1 is correct by inspection and matches the plugins' surface — so I'd take this as a fast follow rather than gate on it.

**Boundary / conventions / bookkeeping:** Core stays import-free of plugins/third-party — all resource/Data/JsonResource classes matched by FQCN string via `is_a(allow_string: true)`, same approach as `PaginatorKind` (no `src/Support` or `src/Contracts` dependency added). The Guard 2↔`ResourceClassLocator` coupling is explicitly documented in the code comment. New `PaginatorCallReader` dep is `#[Scoped]`, wired in the SP factory. strict-types intact; `auto-derivation.md` response-body row + CHANGELOG `[Unreleased]` updated honestly with the tier and the deferral framing. PHPStan level 8 clean (0 errors), Pint clean, 6 fallback tests green.

Verdict: **approved.** Finding 1 is a recommended fast-follow test (Guard 1 is correct but uncovered); no leak, no correctness concern, no scope creep.

Radiergummi:
🤖 **[lead]** both approvals in (guards mutation-tested). One required fast-follow before escalation: reviewer-2 showed Guard 1 has no test that fails without it (collection fixtures carry no item class, so resolveItemClass() bails and masks Guard 1). Round 2: add the Guard-1-pinning fixture, then escalate.

Radiergummi:
🤖 **[coder]** addressed: added Guard-1-pinning test; pushed 33c181b.

reviewer-2 was right — the two existing collection fixtures carried no item class, so `resolveItemClass()` returned null and Core bailed on "item undeclared" regardless of Guard 1, masking it. New fixture `FallbackResolvableItemCollectionController`: an `AnonymousResourceCollection` return with a bare-model `#[ResponseResource(Author::class)]` (Core-resolvable item, so Guard 2 passes and `resolveItemClass()` would succeed) + a paginating body — so **only** the resource return type (Guard 1) stops Core. The test asserts the ApiResources envelope lands (`data.items.$ref` → `NestedAuthorResource`, `links`+`meta` present, `meta.total`), i.e. Core does not steal it with a `{data:[Author], …}` paginated array.

Proof it is load-bearing: with Guard 1 neutralised, this is the **only** test that fails (the other 6 stay green, confirming the masking); restored, all pass. Test-only change (+28 lines, one file). Local `composer check` green (2724 tests); remote CI all-green on 33c181b — full matrix, swagger-php 5.8, Snapshot, Coverage, quality.

Radiergummi:
🤖 **[lead]** ESCALATED for your merge decision — escalation-bound by design (cross-plugin precedence: Core's PaginatorResponseResolver runs before ApiResources, first-non-null-wins). NOT a size/quality concern; the work is sound:

- **Two independent approvals**, both reviewers MUTATION-TESTED the guards: reviewer-1 proved Guard 2 load-bearing (neutralise → method-attribute test fails); reviewer-2 proved Guard 1 load-bearing after the round-2 fixture (neutralise → only the new test fails, exactly Core stealing the flat {data:[Author]} envelope).
- **Guard1 ∪ Guard2 = exactly the union of ApiResources' + SpatieData's claiming triggers** (reviewer-2 verified against ResourceClassLocator + DataResponseResolver source). Core only consumes #[ResponseResource] when it names a NON-resource bare model.
- Full CI green on 33c181b incl. swagger-php 5.8 AND the **byte-exact Snapshot job** — the regression-critical check: no resource/Data endpoint lost links/meta in real generated specs.
- 7 tests pin the surface; declared-paginator path byte-identical; no Contracts/stage-order/config/signature change.

Your call given it changes cross-plugin response precedence. Ready to merge on your nod — say the word and I'll finish-pr it. (I can also run a targeted Vito dogfood check first if you want real-app confirmation beyond the Snapshot job.)

Radiergummi:
🤖 **[lead]** Targeted Vito dogfood check — **PASS, no regression.**

Generated Vito's spec on baseline `main` (1aaff42) vs the #364 candidate (33c181b) and diffed:
- **Specs byte-identical after normalization** (952,765 bytes each, zero line diff).
- 329 paths, 84 component schemas — unchanged.
- `links`+`meta` key count: baseline 40 = candidate 40 — no resource-collection envelope lost.

So on the exact app this issue targets, the change is a true no-op: the guards prevent Core from claiming any of Vito's ResourceCollection responses, and the new fallback doesn't fire on them. Combined with the byte-exact in-repo Snapshot job, the 7 regression tests, and two independent (mutation-tested) approvals, #364 is fully validated. Ready to merge at your discretion.

Radiergummi:
🤖 **[lead]** Merged by Moritz's authorization. Fully validated: two mutation-tested approvals, full CI matrix incl. swagger-php 5.8, byte-exact Snapshot, and a clean targeted Vito dogfood check — baseline vs candidate specs byte-identical, no resource envelope lost. Closes #353.

PR #365, opened 2026-06-15
feat(#134): Fortify auth plugin — document headless core-auth endpoints

### LINKED ISSUE #134: Document Fortify's headless auth endpoints (login / register / password / 2FA / email verification)
## Context

Fortify provides **headless** authentication: it registers `/login`, `/logout`, `/register`, `/forgot-password`, `/reset-password`, email-verification, and two-factor endpoints. For an API/SPA consumer these endpoints **are** part of the documented API surface — so unlike the operational/vendor routes in #133, the goal is to **document** them, not filter them.

Surfaced by the same first-party-package review that produced #133.

## Problem

Fortify validates **inline** inside its actions (action classes / inline validators, `Fortify::createUsersUsing`, etc.), not through published `FormRequest`s with reflectable `rules()` methods. So there is no Tier-0 signature to read, and the generator emits little for these routes today.

## Proposed approach

A hand-maintained **static knowledge table** mapping Fortify's well-known routes (keyed by route name — `login`, `register`, `password.email`, `password.update`, `two-factor.login`, `verification.*`, …) to their fixed request payloads and response shapes/status codes. Gate on Fortify being installed *and* its routes registered (`class_exists` + `router->has`, mirroring `SecurityExtractor::passportAvailable()`).

## Economic-sensibility / tiering

This is **not** Tier-0 reflection (no signatures to read) and **not** generic Tier-1 AST scanning either — it's a hand-maintained knowledge table for one package's fixed contract. Bigger build than the #133 filters, and per the "finish Tier-0 first" worklist policy it sits **behind** current Tier-0 work. Filing now to capture the opportunity; sequence later. Tagged `tier-1` to signal "beyond reflection / deferred."

## Open questions

- Fortify's payloads/responses vary by feature config (registration enabled, 2FA confirmation, email verification) and by views-vs-JSON mode. Scope to the **JSON/headless** responses, and make route presence feature-flag-aware.
- Likely a better fit as a **bundled plugin** (`plugin-candidate`) than as Core, since it's package-specific knowledge — Core's remit is vanilla-Laravel patterns.

Radiergummi:
Closes #134

# Fortify auth plugin — Implementation Plan

> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.

**Goal:** A bundled, opt-in `Fortify` plugin that documents Laravel Fortify's headless core-auth endpoints from a hand-maintained stock-contract table, honestly gated so it never emits a confidently-wrong response.

**Architecture:** Two resolvers sharing a `FortifyContractTable` (route name → request body + success response + status + response-contract class). `FortifyRequestSchemaResolver` always emits the stock request body; `FortifyResponseResolver` emits the stock success body only when the route's Fortify response contract still resolves to a `Laravel\Fortify\` class (else status-only). Matched purely by `$route->getName()`. Errors fall through to the existing `openapi.error_envelope` machinery.

**Tech Stack:** PHP 8.4, swagger-php (`OpenApi\Annotations`), Pest + Orchestra Testbench, `laravel/fortify` (dev-only).

Full agreed design: issue #134 (comment "Agreed design — v1").

---

## File structure

- Create `src/Plugins/Fortify/FortifyPlugin.php` — `Plugin` impl, `class_exists` guard, registers the two resolvers.
- Create `src/Plugins/Fortify/Support/FortifyContractTable.php` — the hand-maintained data (route name → entry).
- Create `src/Plugins/Fortify/Support/FortifyResponseCustomization.php` — the binding-inspection gate.
- Create `src/Plugins/Fortify/Resolvers/FortifyRequestSchemaResolver.php` — `RequestSchemaResolver`.
- Create `src/Plugins/Fortify/Resolvers/FortifyResponseResolver.php` — `PrimaryResponseResolver`.
- Modify `config/openapi.php` — add a commented-out `FortifyPlugin::class` entry under `plugins`.
- Add `laravel/fortify` to `require-dev` in `composer.json`.
- Create `tests/Fixtures/Fortify/` fixture controllers/routes + `tests/Feature/Plugins/Fortify/FortifyPluginTest.php`.
- Modify `docs/plugins.md` (new section) + `CHANGELOG.md`.

---

## Task 1: Add the dev dependency + plugin scaffold (off by default)

**Files:**
- Modify: `composer.json` (require-dev)
- Create: `src/Plugins/Fortify/FortifyPlugin.php`
- Modify: `config/openapi.php` (plugins array)

- [ ] **Step 1: Add `laravel/fortify` to require-dev**

```bash
composer require --dev "laravel/fortify:^1.21"
```
Expected: installs cleanly on PHP 8.4 / Laravel 12 & 13.

- [ ] **Step 2: Create the plugin class (guarded, registers nothing yet)**

`src/Plugins/Fortify/FortifyPlugin.php`:
```php
<?php

declare(strict_types=1);

namespace Radiergummi\OpenApi\Plugins\Fortify;

use Laravel\Fortify\Fortify;
use Override;
use Radiergummi\OpenApi\Contracts\Registry\Plugin;
use Radiergummi\OpenApi\Plugins\Fortify\Resolvers\FortifyRequestSchemaResolver;
use Radiergummi\OpenApi\Plugins\Fortify\Resolvers\FortifyResponseResolver;
use Radiergummi\OpenApi\Registry\OpenApiRegistry;

use function class_exists;

/**
 * Teaches the OpenAPI core to document Laravel Fortify's headless core-auth endpoints from a
 * hand-maintained stock-contract table. No-ops when Fortify is not installed.
 */
final class FortifyPlugin implements Plugin
{
    #[Override]
    public function register(OpenApiRegistry $registry): void
    {
        if (!class_exists(Fortify::class)) {
            return;
        }

        $registry->addRequestSchemaResolver(FortifyRequestSchemaResolver::class);
        $registry->addPrimaryResponseResolver(FortifyResponseResolver::class);
    }
}
```
(The resolver classes are created in Tasks 3–4; this won't autoload-resolve until they exist — that's fine, `register()` isn't called at file-load.)

- [ ] **Step 3: Add the commented config entry**

In `config/openapi.php`, under `'plugins' => [`, after the Fractal block, add:
```php
        // Requires `composer require laravel/fortify`. Documents Fortify's headless core-auth
        // endpoints (login/register/password/profile) from a stock-contract table. Uncomment:
        // \Radiergummi\OpenApi\Plugins\Fortify\FortifyPlugin::class,
```

- [ ] **Step 4: Commit**

```bash
git add composer.json composer.lock src/Plugins/Fortify/FortifyPlugin.php config/openapi.php
git commit -m "feat(#134): scaffold Fortify plugin (off by default)"
```

---

## Task 2: The contract table

**Files:**
- Create: `src/Plugins/Fortify/Support/FortifyContractTable.php`
- Test: `tests/Unit/Plugins/Fortify/FortifyContractTableTest.php`

The table is the single source of truth. Each entry: stock request `OA\Schema`, stock success `OA\Schema|null` (null = no body, e.g. 204), success status, and the FQCN of the Fortify response contract that governs the response (for the customization gate in Task 5). Request bodies are the stable, documented Fortify validation rules. **Response bodies and statuses MUST be verified against the installed Fortify source** in Step 1 below — they are version-sensitive; do not trust this draft blindly.

- [ ] **Step 1: Verify the stock contract against Fortify source**

Read these and reconcile the table values:
- `vendor/laravel/fortify/routes/routes.php` — confirm route NAMES for the v1 set: `login`, `logout`, `register`, `password.email`, `password.update`, `password.confirm`, `password.confirmation`, `user-password.update`, `user-profile-information.update`.
- `vendor/laravel/fortify/src/Actions/*` (e.g. `CreateNewUser`, `AttemptToAuthenticate`) — confirm request validation rule keys.
- `vendor/laravel/fortify/src/Http/Responses/*` and `src/Contracts/*Response.php` — confirm each route's response contract FQCN and the actual JSON success body/status in headless mode.

Record any deviations from this draft as a code comment in the table.

- [ ] **Step 2: Write the failing test**

`tests/Unit/Plugins/Fortify/FortifyContractTableTest.php`:
```php
<?php

declare(strict_types=1);

use OpenApi\Annotations as OA;
use Radiergummi\OpenApi\Plugins\Fortify\Support\FortifyContractTable;

it('exposes an entry for each v1 core-auth route name', function (): void {
    $names = ['login', 'logout', 'register', 'password.email', 'password.update',
        'password.confirm', 'password.confirmation', 'user-password.update',
        'user-profile-information.update'];

    foreach ($names as $name) {
        expect(FortifyContractTable::for($name))->not->toBeNull();
    }
    expect(FortifyContractTable::for('two-factor.login'))->toBeNull(); // deferred, not in v1
});

it('returns a request schema and response contract for login', function (): void {
    $entry = FortifyContractTable::for('login');
    expect($entry->requestSchema)->toBeInstanceOf(OA\Schema::class)
        ->and($entry->responseContract)->toBe(\Laravel\Fortify\Contracts\LoginResponse::class)
        ->and($entry->successStatus)->toBe(200);
});
```

- [ ] **Step 3: Run it (fails — class missing)**

Run: `vendor/bin/pest tests/Unit/Plugins/Fortify/FortifyContractTableTest.php`
Expected: FAIL (`FortifyContractTable` not found).

- [ ] **Step 4: Implement the table + entry value object**

`src/Plugins/Fortify/Support/FortifyContractTable.php`:
```php
<?php

declare(strict_types=1);

namespace Radiergummi\OpenApi\Plugins\Fortify\Support;

use Laravel\Fortify\Contracts\LoginResponse;
use Laravel\Fortify\Contracts\LogoutResponse;
use Laravel\Fortify\Contracts\PasswordConfirmedResponse;
use Laravel\Fortify\Contracts\PasswordResetResponse;
use Laravel\Fortify\Contracts\PasswordUpdateResponse;
use Laravel\Fortify\Contracts\ProfileInformationUpdatedResponse;
use Laravel\Fortify\Contracts\RegisterResponse;
use Laravel\Fortify\Contracts\SuccessfulPasswordResetLinkRequestResponse;
use OpenApi\Annotations as OA;
use Radiergummi\OpenApi\Support\Generator\SchemaFromArrayDefinition;

/**
 * Hand-maintained stock contract for Fortify's headless core-auth routes, keyed by route name.
 *
 * @internal
 */
final class FortifyContractTable
{
    /** @return FortifyContractEntry|null */
    public static function for(string $routeName): ?FortifyContractEntry
    {
        return match ($routeName) {
            'login' => new FortifyContractEntry(
                requestSchema: self::object([
                    'email' => ['type' => 'string', 'format' => 'email'],
                    'password' => ['type' => 'string', 'format' => 'password'],
                    'remember' => ['type' => 'boolean'],
                ], required: ['email', 'password']),
                responseContract: LoginResponse::class,
                successStatus: 200,
                successSchema: null, // verify: Fortify JSON login returns 200 empty / {two_factor}
            ),
            'logout' => new FortifyContractEntry(
                requestSchema: null,
                responseContract: LogoutResponse::class,
                successStatus: 204,
                successSchema: null,
            ),
            'register' => new FortifyContractEntry(
                requestSchema: self::object([
                    'name' => ['type' => 'string'],
                    'email' => ['type' => 'string', 'format' => 'email'],
                    'password' => ['type' => 'string', 'format' => 'password'],
                    'password_confirmation' => ['type' => 'string', 'format' => 'password'],
                ], required: ['name', 'email', 'password', 'password_confirmation']),
                responseContract: RegisterResponse::class,
                successStatus: 201,
                successSchema: null,
            ),
            'password.email' => new FortifyContractEntry(
                requestSchema: self::object([
                    'email' => ['type' => 'string', 'format' => 'email'],
                ], required: ['email']),
                responseContract: SuccessfulPasswordResetLinkRequestResponse::class,
                successStatus: 200,
                successSchema: self::object(['message' => ['type' => 'string']]),
            ),
            'password.update' => new FortifyContractEntry(
                requestSchema: self::object([
                    'token' => ['type' => 'string'],
                    'email' => ['type' => 'string', 'format' => 'email'],
                    'password' => ['type' => 'string', 'format' => 'password'],
                    'password_confirmation' => ['type' => 'string', 'format' => 'password'],
                ], required: ['token', 'email', 'password', 'password_confirmation']),
                responseContract: PasswordResetResponse::class,
                successStatus: 200,
                successSchema: self::object(['message' => ['type' => 'string']]),
            ),
            'password.confirm' => new FortifyContractEntry(
                requestSchema: self::object([
                    'password' => ['type' => 'string', 'format' => 'password'],
                ], required: ['password']),
                responseContract: PasswordConfirmedResponse::class,
                successStatus: 201,
                successSchema: null,
            ),
            'password.confirmation' => new FortifyContractEntry(
                requestSchema: null,
                responseContract: null, // GET status endpoint, returns JSON directly (no contract)
                successStatus: 200,
                successSchema: self::object(['confirmed' => ['type' => 'boolean']]),
            ),
            'user-password.update' => new FortifyContractEntry(
                requestSchema: self::object([
                    'current_password' => ['type' => 'string', 'format' => 'password'],
                    'password' => ['type' => 'string', 'format' => 'password'],
                    'password_confirmation' => ['type' => 'string', 'format' => 'password'],
                ], required: ['current_password', 'password', 'password_confirmation']),
                responseContract: PasswordUpdateResponse::class,
                successStatus: 200,
                successSchema: null,
            ),
            'user-profile-information.update' => new FortifyContractEntry(
                requestSchema: self::object([
                    'name' => ['type' => 'string'],
                    'email' => ['type' => 'string', 'format' => 'email'],
                ], required: ['name', 'email']),
                responseContract: ProfileInformationUpdatedResponse::class,
                successStatus: 200,
                successSchema: null,
            ),
            default => null,
        };
    }

    /**
     * @param array<string, array<string, mixed>> $properties
     * @param list<string> $required
     */
    private static function object(array $properties, array $required = []): OA\Schema
    {
        $definition = ['type' => 'object', 'properties' => $properties];
        if ($required !== []) {
            $definition['required'] = $required;
        }
        return SchemaFromArrayDefinition::build($definition);
    }
}
```

`src/Plugins/Fortify/Support/FortifyContractEntry.php`:
```php
<?php

declare(strict_types=1);

namespace Radiergummi\OpenApi\Plugins\Fortify\Support;

use OpenApi\Annotations as OA;

/** @internal */
final readonly class FortifyContractEntry
{
    public function __construct(
        public ?OA\Schema $requestSchema,
        public ?string $responseContract,
        public int $successStatus,
        public ?OA\Schema $successSchema,
    ) {}
}
```
(Confirm `SchemaFromArrayDefinition::build()` accepts a nested `properties` map of `name => definition`; if it expects `OA\Property` objects instead, build them with `SchemaFromArrayDefinition::buildProperty($name, $def)` — see `src/Support/Generator/SchemaFromArrayDefinition.php`.)

- [ ] **Step 5: Run the test (passes)**

Run: `vendor/bin/pest tests/Unit/Plugins/Fortify/FortifyContractTableTest.php`
Expected: PASS.

- [ ] **Step 6: Commit**

```bash
git add src/Plugins/Fortify/Support/ tests/Unit/Plugins/Fortify/FortifyContractTableTest.php
git commit -m "feat(#134): Fortify stock-contract table"
```

---

## Task 3: Request-schema resolver

**Files:**
- Create: `src/Plugins/Fortify/Resolvers/FortifyRequestSchemaResolver.php`
- Test: covered end-to-end in Task 6 (it needs a booted app + routes)

- [ ] **Step 1: Implement the resolver**

```php
<?php

declare(strict_types=1);

namespace Radiergummi\OpenApi\Plugins\Fortify\Resolvers;

use Illuminate\Container\Attributes\Scoped;
use OpenApi\Annotations as OA;
use Override;
use Radiergummi\OpenApi\Contracts\Registry\RequestSchemaResolver;
use Radiergummi\OpenApi\Enums\MediaType;
use Radiergummi\OpenApi\Plugins\Fortify\Support\FortifyContractTable;
use Radiergummi\OpenApi\Routing\ActionDescriptor;
use Radiergummi\OpenApi\Support\Generator\ComponentSchemaRegistry;
use Radiergummi\OpenApi\Support\Registry\ResolvedSchema;

/**
 * Emits the stock Fortify request body for a matched core-auth route, by route name.
 *
 * @internal
 */
#[Scoped]
final readonly class FortifyRequestSchemaResolver implements RequestSchemaResolver
{
    public function __construct(private ComponentSchemaRegistry $registry) {}

    #[Override]
    public function resolveRequestSchema(ActionDescriptor $descriptor): ?ResolvedSchema
    {
        $name = $descriptor->route->getName();
        if ($name === null) {
            return null;
        }

        $entry = FortifyContractTable::for($name);
        if ($entry === null || $entry->requestSchema === null) {
            return null;
        }

        $key = $this->registry->reserveKey('Fortify\\Request\\' . $name);
        $this->registry->registerNamed($key, $entry->requestSchema);

        return new ResolvedSchema(componentKey: $key, mediaType: MediaType::Json);
    }
}
```
(`registerNamed` is idempotent, so re-resolution is safe. `reserveKey` derives a stable component key from the synthetic class string.)

- [ ] **Step 2: Commit**

```bash
git add src/Plugins/Fortify/Resolvers/FortifyRequestSchemaResolver.php
git commit -m "feat(#134): Fortify request-schema resolver"
```

---

## Task 4: The customization gate

**Files:**
- Create: `src/Plugins/Fortify/Support/FortifyResponseCustomization.php`
- Test: `tests/Unit/Plugins/Fortify/FortifyResponseCustomizationTest.php`

The gate answers: "is the route's Fortify response contract still bound to a `Laravel\Fortify\` class?" Inspect the binding WITHOUT resolving where possible; if indeterminate (closure binding) → treat as customized (conservative).

- [ ] **Step 1: Write the failing test**

```php
<?php

declare(strict_types=1);

use Illuminate\Container\Container;
use Laravel\Fortify\Contracts\LoginResponse;
use Radiergummi\OpenApi\Plugins\Fortify\Support\FortifyResponseCustomization;

it('reports stock when the contract resolves to a Fortify class', function (): void {
    $container = new Container();
    $container->bind(LoginResponse::class, \Laravel\Fortify\Http\Responses\LoginResponse::class);
    $gate = new FortifyResponseCustomization($container);
    expect($gate->isStock(LoginResponse::class))->toBeTrue();
});

it('reports customized when rebound to a non-Fortify class', function (): void {
    $container = new Container();
    $container->bind(LoginResponse::class, fn () => new class implements LoginResponse {
        public function toResponse($request) { return response()->noContent(); }
    });
    $gate = new FortifyResponseCustomization($container);
    expect($gate->isStock(LoginResponse::class))->toBeFalse(); // closure binding => conservative
});

it('reports stock for a null contract (no contract governs the body)', function (): void {
    $gate = new FortifyResponseCustomization(new Container());
    expect($gate->isStock(null))->toBeTrue();
});
```

- [ ] **Step 2: Run it (fails)**

Run: `vendor/bin/pest tests/Unit/Plugins/Fortify/FortifyResponseCustomizationTest.php`
Expected: FAIL (class missing).

- [ ] **Step 3: Implement the gate**

```php
<?php

declare(strict_types=1);

namespace Radiergummi\OpenApi\Plugins\Fortify\Support;

use Illuminate\Contracts\Container\Container;

use function is_string;
use function str_starts_with;

/**
 * Best-effort detector for whether a Fortify response contract still maps to a stock Fortify
 * implementation. Inspects the registered binding's concrete WITHOUT resolving it; a closure
 * binding (or unbound) is treated as customized — we never emit a possibly-wrong stock body.
 *
 * @internal
 */
final readonly class FortifyResponseCustomization
{
    private const string FORTIFY_NAMESPACE = 'Laravel\\Fortify\\';

    public function __construct(private Container $container) {}

    public function isStock(?string $responseContract): bool
    {
        // No contract governs this route's body (e.g. the password-confirmation status endpoint).
        if ($responseContract === null) {
            return true;
        }

        if (!$this->container->bound($responseContract)) {
            // Unbound: Fortify's service provider binds all of these, so an unbound contract means
            // Fortify isn't fully wired — be conservative.
            return false;
        }

        $concrete = $this->container->getBindings()[$responseContract]['concrete'] ?? null;

        // Only a string concrete (a class name) is inspectable without invoking a closure.
        return is_string($concrete) && str_starts_with($concrete, self::FORTIFY_NAMESPACE);
    }
}
```
(Confirm the binding array shape: Laravel's `Container::getBindings()` returns `[$abstract => ['concrete' => Closure|string, 'shared' => bool]]`. Fortify binds its responses via `singleton($contract, $concreteClassString)` in most versions — verify in `vendor/laravel/fortify/src/FortifyServiceProvider.php`; if it binds via closures returning `new X`, the gate correctly falls to conservative status-only, which is acceptable.)

- [ ] **Step 4: Run the test (passes)**

Run: `vendor/bin/pest tests/Unit/Plugins/Fortify/FortifyResponseCustomizationTest.php`
Expected: PASS.

- [ ] **Step 5: Commit**

```bash
git add src/Plugins/Fortify/Support/FortifyResponseCustomization.php tests/Unit/Plugins/Fortify/FortifyResponseCustomizationTest.php
git commit -m "feat(#134): Fortify response-customization gate"
```

---

## Task 5: Response resolver

**Files:**
- Create: `src/Plugins/Fortify/Resolvers/FortifyResponseResolver.php`
- Test: end-to-end in Task 6

- [ ] **Step 1: Implement the resolver**

```php
<?php

declare(strict_types=1);

namespace Radiergummi\OpenApi\Plugins\Fortify\Resolvers;

use Illuminate\Container\Attributes\Scoped;
use Illuminate\Contracts\Container\Container;
use OpenApi\Annotations as OA;
use Override;
use Radiergummi\OpenApi\Contracts\Registry\PrimaryResponseResolver;
use Radiergummi\OpenApi\Enums\MediaType;
use Radiergummi\OpenApi\Plugins\Fortify\Support\FortifyContractTable;
use Radiergummi\OpenApi\Plugins\Fortify\Support\FortifyResponseCustomization;
use Radiergummi\OpenApi\Routing\ActionDescriptor;
use Symfony\Component\HttpFoundation\Response as HttpFoundationResponse;

use function sprintf;

/**
 * Emits the stock Fortify success response for a matched core-auth route. When the route's
 * response contract has been rebound to a non-Fortify class, emits the status code only (the
 * body is unknowable), never a possibly-wrong stock body.
 *
 * @internal
 */
#[Scoped]
final readonly class FortifyResponseResolver implements PrimaryResponseResolver
{
    public function __construct(private Container $container) {}

    #[Override]
    public function resolvePrimaryResponse(ActionDescriptor $descriptor): ?OA\Response
    {
        $name = $descriptor->route->getName();
        if ($name === null) {
            return null;
        }

        $entry = FortifyContractTable::for($name);
        if ($entry === null) {
            return null;
        }

        $status = $entry->successStatus;
        $description = HttpFoundationResponse::$statusTexts[$status] ?? sprintf('HTTP %d', $status);

        $gate = new FortifyResponseCustomization($this->container);
        $emitBody = $entry->successSchema !== null && $gate->isStock($entry->responseContract);

        if (!$emitBody) {
            return new OA\Response(['response' => (string) $status, 'description' => $description]);
        }

        return new OA\Response([
            'response' => (string) $status,
            'description' => $description,
            'content' => [MediaType::Json->schema($entry->successSchema)],
        ]);
    }
}
```

- [ ] **Step 2: Commit**

```bash
git add src/Plugins/Fortify/Resolvers/FortifyResponseResolver.php
git commit -m "feat(#134): Fortify response resolver with customization gate"
```

---

## Task 6: End-to-end feature tests

**Files:**
- Create: `tests/Fixtures/Fortify/routes.php` (registers Fortify's core routes by name)
- Create: `tests/Feature/Plugins/Fortify/FortifyPluginTest.php`

Match the existing plugin feature-test harness — look at `tests/Feature/Plugins/ApiResources/` or `tests/Feature/Plugins/Fractal/` for how a fixture app registers routes, enables a plugin, runs generation, and asserts on the produced document. Reuse that bootstrapping; don't invent a new one.

- [ ] **Step 1: Register the v1 routes in a fixture**

`tests/Fixtures/Fortify/routes.php` (register the route NAMES the table keys on — minimal closures suffice; the plugin keys off names, not actions):
```php
<?php
use Illuminate\Support\Facades\Route;
Route::post('/login', fn () => null)->name('login');
Route::post('/logout', fn () => null)->name('logout');
Route::post('/register', fn () => null)->name('register');
Route::post('/forgot-password', fn () => null)->name('password.email');
Route::post('/reset-password', fn () => null)->name('password.update');
Route::post('/user/confirm-password', fn () => null)->name('password.confirm');
Route::get('/user/confirmed-password-status', fn () => null)->name('password.confirmation');
Route::put('/user/password', fn () => null)->name('user-password.update');
Route::put('/user/profile-information', fn () => null)->name('user-profile-information.update');
```

- [ ] **Step 2: Write the feature tests**

`tests/Feature/Plugins/Fortify/FortifyPluginTest.php` (adapt the bootstrap to the existing harness):
```php
<?php

declare(strict_types=1);

use Laravel\Fortify\Contracts\LoginResponse;

// Helper `generateWithFortifyPlugin()` should: register the FortifyPlugin, load the fixture
// routes, run the generator, and return the decoded OpenAPI document as an array.

it('documents the login request body from the stock table', function (): void {
    $doc = generateWithFortifyPlugin();
    $op = $doc['paths']['/login']['post'];
    $ref = $op['requestBody']['content']['application/json']['schema']['$ref'];
    $schema = resolveRef($doc, $ref);
    expect($schema['properties'])->toHaveKeys(['email', 'password', 'remember'])
        ->and($schema['required'])->toContain('email', 'password');
});

it('emits the stock success body when the response contract is default Fortify', function (): void {
    $doc = generateWithFortifyPlugin();
    $op = $doc['paths']['/forgot-password']['post'];
    expect($op['responses'])->toHaveKey('200')
        ->and($op['responses']['200']['content']['application/json']['schema']['properties'])
        ->toHaveKey('message');
});

it('drops the success body (status only) when the response contract is rebound', function (): void {
    $doc = generateWithFortifyPlugin(function ($app): void {
        $app->bind(\Laravel\Fortify\Contracts\SuccessfulPasswordResetLinkRequestResponse::class,
            fn () => new class implements \Laravel\Fortify\Contracts\SuccessfulPasswordResetLinkRequestResponse {
                public function toResponse($request) { return response()->json(['x' => 1]); }
            });
    });
    $op = $doc['paths']['/forgot-password']['post'];
    expect($op['responses'])->toHaveKey('200')
        ->and($op['responses']['200'])->not->toHaveKey('content'); // body dropped
});

it('emits nothing for a route that is not registered', function (): void {
    $doc = generateWithFortifyPlugin(routes: ['login']); // only register login
    expect($doc['paths'] ?? [])->not->toHaveKey('/register');
});
```

- [ ] **Step 3: Run the suite**

Run: `vendor/bin/pest tests/Feature/Plugins/Fortify/FortifyPluginTest.php`
Expected: all PASS. Fix wiring until green.

- [ ] **Step 4: Full local gate**

Run: `composer format && composer check`
Expected: Pint clean, PHPStan level 8 clean, full Pest suite green (including the byte-exact snapshot group — the Fortify fixtures must not be wired into any snapshot flavor).

- [ ] **Step 5: Commit**

```bash
git add tests/
git commit -m "feat(#134): Fortify plugin end-to-end tests"
```

---

## Task 7: Docs + CHANGELOG

**Files:**
- Modify: `docs/plugins.md`
- Modify: `CHANGELOG.md`

- [ ] **Step 1: Document the plugin**

Add a `Fortify` section to `docs/plugins.md` mirroring the existing plugin sections: what it does (documents Fortify's headless core-auth endpoints from a stock table), that it's **off by default** (uncomment in `config/openapi.plugins`, requires `laravel/fortify`), the v1 scope (core auth; 2FA + email verification deferred), and the customization behavior (stock request bodies always; stock response bodies only when the response contract is unmodified, else status-only; authoring attributes override).

- [ ] **Step 2: CHANGELOG entry**

Add to `CHANGELOG.md` under `[Unreleased] > Added`, as its own line at the END of the section:
```markdown
- **Fortify plugin** (opt-in): documents Laravel Fortify's headless core-auth endpoints (login/logout/register/password/profile) from a stock-contract table, with request bodies always emitted and response bodies gated on the Fortify response contract being unmodified. (#134)
```

- [ ] **Step 3: Commit**

```bash
git add docs/plugins.md CHANGELOG.md
git commit -m "docs(#134): document the Fortify plugin"
```

---

## Verification checklist (before marking ready)

- [ ] `composer check` green (Pint + PHPStan L8 + full Pest suite).
- [ ] Snapshot (byte-exact) group unaffected — Fortify fixtures are NOT referenced by any snapshot flavor.
- [ ] Plugin is OFF by default — generation output is byte-identical without the config entry uncommented.
- [ ] Remote CI green across the full matrix incl. swagger-php 5.8 (no raw nested-array schemas — the table uses `OA\Schema`/`SchemaFromArrayDefinition`, so this should hold; confirm).
- [ ] Response-body table values verified against the pinned Fortify version's actual JSON responses (Task 2 Step 1).

## Notes / soft spots flagged during design

- **Response-body fidelity is version-sensitive.** Task 2 Step 1 is mandatory: transcribe the real stock responses from the installed Fortify, don't trust this draft's `successSchema`/`successStatus` guesses.
- **The binding gate is best-effort.** If Fortify binds responses via closures, the gate can't read the concrete class → it conservatively emits status-only even for stock apps. Acceptable (never wrong, occasionally less detailed). If verification shows Fortify binds via class-string singletons, the gate reads them precisely.
- **2FA + email verification are deferred** — a follow-up issue should extend `FortifyContractTable` + the route set once v1 lands.

### DIFF
--- a/src/Plugins/Fortify/Resolvers/FortifyRequestSchemaResolver.php
+++ b/src/Plugins/Fortify/Resolvers/FortifyRequestSchemaResolver.php
@@ -1,0 +1,46 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Plugins\Fortify\Resolvers;
+
+use Illuminate\Container\Attributes\Scoped;
+use Override;
+use Radiergummi\OpenApi\Contracts\Registry\RequestSchemaResolver;
+use Radiergummi\OpenApi\Enums\MediaType;
+use Radiergummi\OpenApi\Plugins\Fortify\Support\FortifyContractTable;
+use Radiergummi\OpenApi\Routing\ActionDescriptor;
+use Radiergummi\OpenApi\Support\Generator\ComponentSchemaRegistry;
+use Radiergummi\OpenApi\Support\Registry\ResolvedSchema;
+
+/**
+ * Emits the stock Fortify request body for a matched core-auth route, by route name.
+ *
+ * @internal
+ */
+#[Scoped]
+final readonly class FortifyRequestSchemaResolver implements RequestSchemaResolver
+{
+    public function __construct(private ComponentSchemaRegistry $registry) {}
+
+    #[Override]
+    public function resolveRequestSchema(ActionDescriptor $descriptor): ?ResolvedSchema
+    {
+        $name = $descriptor->route->getName();
+
+        if ($name === null) {
+            return null;
+        }
+
+        $entry = FortifyContractTable::for($name);
+
+        if ($entry === null || $entry->requestSchema === null) {
+            return null;
+        }
+
+        $key = $this->registry->reserveKey('Fortify\\Request\\' . $name);
+        $this->registry->registerNamed($key, $entry->requestSchema);
+
+        return new ResolvedSchema(componentKey: $key, mediaType: MediaType::Json);
+    }
+}
--- a/src/Plugins/Fortify/Resolvers/FortifyRequestSchemaResolver.php
+++ b/src/Plugins/Fortify/Resolvers/FortifyRequestSchemaResolver.php
@@ -34,11 +34,12 @@ final readonly class FortifyRequestSchemaResolver implements RequestSchemaResolv
 
         $entry = FortifyContractTable::for($name);
 
-        if ($entry === null || $entry->requestSchema === null) {
+        if ($entry === null || $entry->requestSchema === null || $entry->requestSchemaName === null) {
             return null;
         }
 
-        $key = $this->registry->reserveKey('Fortify\\Request\\' . $name);
+        // Use the clean, framework-agnostic component name — never a Fortify/namespace string.
+        $key = $this->registry->reserveKey($entry->requestSchemaName);
         $this->registry->registerNamed($key, $entry->requestSchema);
 
         return new ResolvedSchema(componentKey: $key, mediaType: MediaType::Json);
--- a/src/Plugins/Fortify/Support/FortifyContractTable.php
+++ b/src/Plugins/Fortify/Support/FortifyContractTable.php
@@ -12,6 +12,7 @@ use Laravel\Fortify\Contracts\PasswordUpdateResponse;
 use Laravel\Fortify\Contracts\ProfileInformationUpdatedResponse;
 use Laravel\Fortify\Contracts\RegisterResponse;
 use Laravel\Fortify\Contracts\SuccessfulPasswordResetLinkRequestResponse;
+use Laravel\Fortify\Fortify;
 use OpenApi\Annotations as OA;
 use Radiergummi\OpenApi\Support\Generator\SchemaFromArrayDefinition;
 
@@ -22,10 +23,18 @@ use Radiergummi\OpenApi\Support\Generator\SchemaFromArrayDefinition;
  * (routes/routes.php + src/Http/Responses/*). The body-bearing endpoints are the `*.store`
  * POST routes — `login` / `register` / `password.confirm` (without `.store`) are GET *view*
  * routes that a headless API typically disables, so the table keys on the action routes:
- * `login.store`, `register.store`, `password.confirm.store`. Request bodies are the documented,
- * stable validation rules (login/reset rules read `Fortify::username()`/`Fortify::email()`,
- * which default to `email`; the confirmation `password_confirmation` keys come from the
- * published app stubs).
+ * `login.store`, `register.store`, `password.confirm.store`.
+ *
+ * Field names are read from Fortify's own Tier-0 accessors rather than hardcoded:
+ * {@see Fortify::username()} (defaults to `email`) names the login identifier, and
+ * {@see Fortify::email()} names the email field on register/reset/profile — so the body reflects
+ * the app's `fortify.username` / `fortify.email` config. Password fields stay a plain `string` with
+ * `format: password`: the active constraints live in `Password::default()` (and the published app
+ * stubs), whose values are not statically readable here without reflecting protected state, so the
+ * stock shape is documented and the specifics are left to runtime validation.
+ *
+ * Public component names ({@see FortifyContractEntry::$requestSchemaName}) are clean and
+ * framework-agnostic — they must never leak Fortify/PHP/namespace internals into the document.
  *
  * @internal
  */
@@ -33,14 +42,18 @@ final class FortifyContractTable
 {
     public static function for(string $routeName): ?FortifyContractEntry
     {
+        $username = Fortify::username();
+        $email = Fortify::email();
+
         return match ($routeName) {
             // POST /login — JSON returns {two_factor: false} at 200 (LoginResponse).
             'login.store' => new FortifyContractEntry(
                 requestSchema: self::object([
-                    'email' => ['type' => 'string', 'format' => 'email'],
+                    $username => ['type' => 'string'],
                     'password' => ['type' => 'string', 'format' => 'password'],
                     'remember' => ['type' => 'boolean'],
-                ], required: ['email', 'password']),
+                ], required: [$username, 'password']),
+                requestSchemaName: 'LoginRequest',
                 responseContract: LoginResponse::class,
                 successStatus: 200,
                 successSchema: self::object(['two_factor' => ['type' => 'boolean']]),
@@ -48,6 +61,7 @@ final class FortifyContractTable
             // POST /logout — JSON returns '' at 204 (LogoutResponse).
             'logout' => new FortifyContractEntry(
                 requestSchema: null,
+                requestSchemaName: null,
                 responseContract: LogoutResponse::class,
                 successStatus: 204,
                 successSchema: null,
@@ -56,10 +70,11 @@ final class FortifyContractTable
             'register.store' => new FortifyContractEntry(
                 requestSchema: self::object([
                     'name' => ['type' => 'string'],
-                    'email' => ['type' => 'string', 'format' => 'email'],
+                    $email => ['type' => 'string', 'format' => 'email'],
                     'password' => ['type' => 'string', 'format' => 'password'],
                     'password_confirmation' => ['type' => 'string', 'format' => 'password'],
-                ], required: ['name', 'email', 'password', 'password_confirmation']),
+                ], required: ['name', $email, 'password', 'password_confirmation']),
+                requestSchemaName: 'RegisterRequest',
                 responseContract: RegisterResponse::class,
                 successStatus: 201,
                 successSchema: null,
@@ -67,8 +82,9 @@ final class FortifyContractTable
             // POST /forgot-password — JSON returns {message} at 200.
             'password.email' => new FortifyContractEntry(
                 requestSchema: self::object([
-                    'email' => ['type' => 'string', 'format' => 'email'],
-                ], required: ['email']),
+                    $email => ['type' => 'string', 'format' => 'email'],
+                ], required: [$email]),
+                requestSchemaName: 'ForgotPasswordRequest',
                 responseContract: SuccessfulPasswordResetLinkRequestResponse::class,
                 successStatus: 200,
                 successSchema: self::object(['message' => ['type' => 'string']]),
@@ -77,10 +93,11 @@ final class FortifyContractTable
             'password.update' => new FortifyContractEntry(
                 requestSchema: self::object([
                     'token' => ['type' => 'string'],
-                    'email' => ['type' => 'string', 'format' => 'email'],
+                    $email => ['type' => 'string', 'format' => 'email'],
                     'password' => ['type' => 'string', 'format' => 'password'],
                     'password_confirmation' => ['type' => 'string', 'format' => 'password'],
-                ], required: ['token', 'email', 'password', 'password_confirmation']),
+                ], required: ['token', $email, 'password', 'password_confirmation']),
+                requestSchemaName: 'ResetPasswordRequest',
                 responseContract: PasswordResetResponse::class,
                 successStatus: 200,
                 successSchema: self::object(['message' => ['type' => 'string']]),
@@ -90,6 +107,7 @@ final class FortifyContractTable
                 requestSchema: self::object([
                     'password' => ['type' => 'string', 'format' => 'password'],
                 ], required: ['password']),
+                requestSchemaName: 'ConfirmPasswordRequest',
                 responseContract: PasswordConfirmedResponse::class,
                 successStatus: 201,
                 successSchema: null,
@@ -97,6 +115,7 @@ final class FortifyContractTable
             // GET /user/confirmed-password-status — controller returns {confirmed} directly (no contract).
             'password.confirmation' => new FortifyContractEntry(
                 requestSchema: null,
+                requestSchemaName: null,
                 responseContract: null,
                 successStatus: 200,
                 successSchema: self::object(['confirmed' => ['type' => 'boolean']]),
@@ -108,6 +127,7 @@ final class FortifyContractTable
                     'password' => ['type' => 'string', 'format' => 'password'],
                     'password_confirmation' => ['type' => 'string', 'format' => 'password'],
                 ], required: ['current_password', 'password', 'password_confirmation']),
+                requestSchemaName: 'UpdatePasswordRequest',
                 responseContract: PasswordUpdateResponse::class,
                 successStatus: 200,
                 successSchema: null,
@@ -116,8 +136,9 @@ final class FortifyContractTable
             'user-profile-information.update' => new FortifyContractEntry(
                 requestSchema: self::object([
                     'name' => ['type' => 'string'],
-                    'email' => ['type' => 'string', 'format' => 'email'],
-                ], required: ['name', 'email']),
+                    $email => ['type' => 'string', 'format' => 'email'],
+                ], required: ['name', $email]),
+                requestSchemaName: 'UpdateProfileInformationRequest',
                 responseContract: ProfileInformationUpdatedResponse::class,
                 successStatus: 200,
                 successSchema: null,
--- a/src/Plugins/Fortify/Support/FortifyContractTable.php
+++ b/src/Plugins/Fortify/Support/FortifyContractTable.php
@@ -1,0 +1,143 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Plugins\Fortify\Support;
+
+use Laravel\Fortify\Contracts\LoginResponse;
+use Laravel\Fortify\Contracts\LogoutResponse;
+use Laravel\Fortify\Contracts\PasswordConfirmedResponse;
+use Laravel\Fortify\Contracts\PasswordResetResponse;
+use Laravel\Fortify\Contracts\PasswordUpdateResponse;
+use Laravel\Fortify\Contracts\ProfileInformationUpdatedResponse;
+use Laravel\Fortify\Contracts\RegisterResponse;
+use Laravel\Fortify\Contracts\SuccessfulPasswordResetLinkRequestResponse;
+use OpenApi\Annotations as OA;
+use Radiergummi\OpenApi\Support\Generator\SchemaFromArrayDefinition;
+
+/**
+ * Hand-maintained stock contract for Fortify's headless core-auth routes, keyed by route name.
+ *
+ * Route names and response bodies were reconciled against laravel/fortify v1.37 source
+ * (routes/routes.php + src/Http/Responses/*). The body-bearing endpoints are the `*.store`
+ * POST routes — `login` / `register` / `password.confirm` (without `.store`) are GET *view*
+ * routes that a headless API typically disables, so the table keys on the action routes:
+ * `login.store`, `register.store`, `password.confirm.store`. Request bodies are the documented,
+ * stable validation rules (login/reset rules read `Fortify::username()`/`Fortify::email()`,
+ * which default to `email`; the confirmation `password_confirmation` keys come from the
+ * published app stubs).
+ *
+ * @internal
+ */
+final class FortifyContractTable
+{
+    public static function for(string $routeName): ?FortifyContractEntry
+    {
+        return match ($routeName) {
+            // POST /login — JSON returns {two_factor: false} at 200 (LoginResponse).
+            'login.store' => new FortifyContractEntry(
+                requestSchema: self::object([
+                    'email' => ['type' => 'string', 'format' => 'email'],
+                    'password' => ['type' => 'string', 'format' => 'password'],
+                    'remember' => ['type' => 'boolean'],
+                ], required: ['email', 'password']),
+                responseContract: LoginResponse::class,
+                successStatus: 200,
+                successSchema: self::object(['two_factor' => ['type' => 'boolean']]),
+            ),
+            // POST /logout — JSON returns '' at 204 (LogoutResponse).
+            'logout' => new FortifyContractEntry(
+                requestSchema: null,
+                responseContract: LogoutResponse::class,
+                successStatus: 204,
+                successSchema: null,
+            ),
+            // POST /register — JSON returns '' at 201 (RegisterResponse).
+            'register.store' => new FortifyContractEntry(
+                requestSchema: self::object([
+                    'name' => ['type' => 'string'],
+                    'email' => ['type' => 'string', 'format' => 'email'],
+                    'password' => ['type' => 'string', 'format' => 'password'],
+                    'password_confirmation' => ['type' => 'string', 'format' => 'password'],
+                ], required: ['name', 'email', 'password', 'password_confirmation']),
+                responseContract: RegisterResponse::class,
+                successStatus: 201,
+                successSchema: null,
+            ),
+            // POST /forgot-password — JSON returns {message} at 200.
+            'password.email' => new FortifyContractEntry(
+                requestSchema: self::object([
+                    'email' => ['type' => 'string', 'format' => 'email'],
+                ], required: ['email']),
+                responseContract: SuccessfulPasswordResetLinkRequestResponse::class,
+                successStatus: 200,
+                successSchema: self::object(['message' => ['type' => 'string']]),
+            ),
+            // POST /reset-password — JSON returns {message} at 200.
+            'password.update' => new FortifyContractEntry(
+                requestSchema: self::object([
+                    'token' => ['type' => 'string'],
+                    'email' => ['type' => 'string', 'format' => 'email'],
+                    'password' => ['type' => 'string', 'format' => 'password'],
+                    'password_confirmation' => ['type' => 'string', 'format' => 'password'],
+                ], required: ['token', 'email', 'password', 'password_confirmation']),
+                responseContract: PasswordResetResponse::class,
+                successStatus: 200,
+                successSchema: self::object(['message' => ['type' => 'string']]),
+            ),
+            // POST /user/confirm-password — JSON returns '' at 201 (PasswordConfirmedResponse).
+            'password.confirm.store' => new FortifyContractEntry(
+                requestSchema: self::object([
+                    'password' => ['type' => 'string', 'format' => 'password'],
+                ], required: ['password']),
+                responseContract: PasswordConfirmedResponse::class,
+                successStatus: 201,
+                successSchema: null,
+            ),
+            // GET /user/confirmed-password-status — controller returns {confirmed} directly (no contract).
+            'password.confirmation' => new FortifyContractEntry(
+                requestSchema: null,
+                responseContract: null,
+                successStatus: 200,
+                successSchema: self::object(['confirmed' => ['type' => 'boolean']]),
+            ),
+            // PUT /user/password — JSON returns '' at 200 (PasswordUpdateResponse).
+            'user-password.update' => new FortifyContractEntry(
+                requestSchema: self::object([
+                    'current_password' => ['type' => 'string', 'format' => 'password'],
+                    'password' => ['type' => 'string', 'format' => 'password'],
+                    'password_confirmation' => ['type' => 'string', 'format' => 'password'],
+                ], required: ['current_password', 'password', 'password_confirmation']),
+                responseContract: PasswordUpdateResponse::class,
+                successStatus: 200,
+                successSchema: null,
+            ),
+            // PUT /user/profile-information — JSON returns '' at 200 (ProfileInformationUpdatedResponse).
+            'user-profile-information.update' => new FortifyContractEntry(
+                requestSchema: self::object([
+                    'name' => ['type' => 'string'],
+                    'email' => ['type' => 'string', 'format' => 'email'],
+                ], required: ['name', 'email']),
+                responseContract: ProfileInformationUpdatedResponse::class,
+                successStatus: 200,
+                successSchema: null,
+            ),
+            default => null,
+        };
+    }
+
+    /**
+     * @param array<string, array<string, mixed>> $properties
+     * @param list<string>                        $required
+     */
+    private static function object(array $properties, array $required = []): OA\Schema
+    {
+        $definition = ['type' => 'object', 'properties' => $properties];
+
+        if ($required !== []) {
+            $definition['required'] = $required;
+        }
+
+        return SchemaFromArrayDefinition::build($definition);
+    }
+}
--- a/src/Plugins/Fortify/Support/FortifyResponseCustomization.php
+++ b/src/Plugins/Fortify/Support/FortifyResponseCustomization.php
@@ -1,0 +1,78 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Plugins\Fortify\Support;
+
+use Closure;
+use Illuminate\Contracts\Container\Container;
+use ReflectionFunction;
+
+use function is_string;
+use function str_starts_with;
+
+/**
+ * Best-effort detector for whether a Fortify response contract still maps to a stock Fortify
+ * implementation. Inspects the registered binding's concrete class name WITHOUT resolving it; a
+ * binding we cannot read as a concrete class (or an unbound contract) is treated as customized —
+ * we never emit a possibly-wrong stock body.
+ *
+ * Note on the binding shape: Laravel's container wraps a string concrete in a closure inside
+ * `bind()`/`singleton()` (see `Container::getClosure()`), so `getBindings()[...]['concrete']` is a
+ * Closure even for `singleton($contract, ConcreteClass::class)` — the form Fortify uses for every
+ * response. We therefore recover the original class name from the wrapper's captured `concrete`
+ * variable, falling back to a directly-stored string concrete for other binding paths.
+ *
+ * @internal
+ */
+final readonly class FortifyResponseCustomization
+{
+    private const string FORTIFY_NAMESPACE = 'Laravel\\Fortify\\';
+
+    public function __construct(private Container $container) {}
+
+    /**
+     * @param ?class-string $responseContract
+     */
+    public function isStock(?string $responseContract): bool
+    {
+        // No contract governs this route's body (e.g. the password-confirmation status endpoint).
+        if ($responseContract === null) {
+            return true;
+        }
+
+        if (!$this->container->bound($responseContract)) {
+            // Unbound: Fortify's service provider binds all of these, so an unbound contract means
+            // Fortify isn't fully wired — be conservative.
+            return false;
+        }
+
+        $concrete = $this->concreteClassName($responseContract);
+
+        return $concrete !== null && str_starts_with($concrete, self::FORTIFY_NAMESPACE);
+    }
+
+    /**
+     * Recovers the bound concrete *class name* without resolving the binding, or null when it
+     * cannot be read statically (a user closure that constructs an instance inline).
+     */
+    private function concreteClassName(string $responseContract): ?string
+    {
+        $concrete = $this->container->getBindings()[$responseContract]['concrete'] ?? null;
+
+        // A directly-stored string concrete (older/alternative binding paths).
+        if (is_string($concrete)) {
+            return $concrete;
+        }
+
+        // Laravel wraps a string concrete in a closure that captures it as `$concrete`; recover it.
+        if ($concrete instanceof Closure) {
+            $captured = (new ReflectionFunction($concrete))->getClosureUsedVariables();
+            $value = $captured['concrete'] ?? null;
+
+            return is_string($value) ? $value : null;
+        }
+
+        return null;
+    }
+}
--- a/src/Plugins/Fortify/Support/FortifyResponseCustomization.php
+++ b/src/Plugins/Fortify/Support/FortifyResponseCustomization.php
@@ -4,25 +4,18 @@ declare(strict_types=1);
 
 namespace Radiergummi\OpenApi\Plugins\Fortify\Support;
 
-use Closure;
 use Illuminate\Contracts\Container\Container;
-use ReflectionException;
-use ReflectionFunction;
+use ReflectionClass;
+use Throwable;
 
-use function is_string;
 use function str_starts_with;
 
 /**
- * Best-effort detector for whether a Fortify response contract still maps to a stock Fortify
- * implementation. Inspects the registered binding's concrete class name WITHOUT resolving it; a
- * binding we cannot read as a concrete class (or an unbound contract) is treated as customized —
- * we never emit a possibly-wrong stock body.
- *
- * Note on the binding shape: Laravel's container wraps a string concrete in a closure inside
- * `bind()`/`singleton()` (see `Container::getClosure()`), so `getBindings()[...]['concrete']` is a
- * Closure even for `singleton($contract, ConcreteClass::class)` — the form Fortify uses for every
- * response. We therefore recover the original class name from the wrapper's captured `concrete`
- * variable, falling back to a directly-stored string concrete for other binding paths.
+ * Detects whether a Fortify response contract still maps to a stock Fortify implementation. At
+ * generation time the app is booted, so we resolve the contract from the container and check
+ * whether the concrete is a `Laravel\Fortify\…` class. Anything else — a custom implementation, an
+ * unbound contract, or a resolution that throws — is treated as customized: the body is unknowable,
+ * so the response resolver falls back to status-only and never emits a possibly-wrong stock body.
  *
  * @internal
  */
@@ -42,44 +35,21 @@ final readonly class FortifyResponseCustomization
             return true;
         }
 
-        if (!$this->container->bound($responseContract)) {
-            // Unbound: Fortify's service provider binds all of these, so an unbound contract means
-            // Fortify isn't fully wired — be conservative.
+        try {
+            // Two stock responses (the password-reset {message} bodies) take a required
+            // `string $status`; pass an empty one so resolving them doesn't throw. The argument is
+            // ignored by responses that don't declare it, including custom implementations.
+            $instance = $this->container->make($responseContract, ['status' => '']);
+
+            // An anonymous class implementing a Fortify contract inherits the contract's namespace
+            // in its synthetic name (`Laravel\Fortify\Contracts\X@anonymous…`), so a bare namespace
+            // check would mistake a custom inline response for stock — exclude anonymous classes.
+            $isAnonymous = (new ReflectionClass($instance))->isAnonymous();
+        } catch (Throwable) {
+            // Unbound, or a binding that throws when constructed — be conservative.
             return false;
         }
 
-        $concrete = $this->concreteClassName($responseContract);
-
-        return $concrete !== null && str_starts_with($concrete, self::FORTIFY_NAMESPACE);
-    }
-
-    /**
-     * Recovers the bound concrete *class name* without resolving the binding, or null when it
-     * cannot be read statically (a user closure that constructs an instance inline).
-     */
-    private function concreteClassName(string $responseContract): ?string
-    {
-        $concrete = $this->container->getBindings()[$responseContract]['concrete'] ?? null;
-
-        // A directly-stored string concrete (older/alternative binding paths).
-        if (is_string($concrete)) {
-            return $concrete;
-        }
-
-        if (!$concrete instanceof Closure) {
-            return null;
-        }
-
-        // Laravel wraps a string concrete in a closure that captures it as `$concrete`; recover it.
-        // Reflecting a Closure never throws, but degrade to null rather than propagate if it ever does.
-        try {
-            $captured = (new ReflectionFunction($concrete))->getClosureUsedVariables();
-        } catch (ReflectionException) {
-            return null;
-        }
-
-        $value = $captured['concrete'] ?? null;
-
-        return is_string($value) ? $value : null;
+        return !$isAnonymous && str_starts_with($instance::class, self::FORTIFY_NAMESPACE);
     }
 }
--- a/src/Plugins/Fortify/Support/FortifyResponseCustomization.php
+++ b/src/Plugins/Fortify/Support/FortifyResponseCustomization.php
@@ -6,6 +6,7 @@ namespace Radiergummi\OpenApi\Plugins\Fortify\Support;
 
 use Closure;
 use Illuminate\Contracts\Container\Container;
+use ReflectionException;
 use ReflectionFunction;
 
 use function is_string;
@@ -65,14 +66,20 @@ final readonly class FortifyResponseCustomization
             return $concrete;
         }
 
-        // Laravel wraps a string concrete in a closure that captures it as `$concrete`; recover it.
-        if ($concrete instanceof Closure) {
-            $captured = (new ReflectionFunction($concrete))->getClosureUsedVariables();
-            $value = $captured['concrete'] ?? null;
-
-            return is_string($value) ? $value : null;
+        if (!$concrete instanceof Closure) {
+            return null;
         }
 
-        return null;
+        // Laravel wraps a string concrete in a closure that captures it as `$concrete`; recover it.
+        // Reflecting a Closure never throws, but degrade to null rather than propagate if it ever does.
+        try {
+            $captured = (new ReflectionFunction($concrete))->getClosureUsedVariables();
+        } catch (ReflectionException) {
+            return null;
+        }
+
+        $value = $captured['concrete'] ?? null;
+
+        return is_string($value) ? $value : null;
     }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -67,6 +67,7 @@ All notable changes to this project are documented here.
 - Eloquent model schemas seed their per-property examples from the model's Laravel factory `definition()` (scalar values only), reseeded deterministically per model from `openapi.examples.faker_seed`; disabled when example synthesis is off or the seed is null. (#36)
 - `#[RequestField]` ref support: a class-string `type:` resolves to a `$ref`, and a class-string `items:` on a `type: 'array'` field resolves to `items: { $ref }` — mirroring the response-side `#[ResourceField]`; an unresolvable class degrades to a permissive object. (#150)
 - A non-paginator action whose body unconditionally calls `paginate()`/`simplePaginate()`/`cursorPaginate()` now gets the matching paginated response envelope, with a declared item class (`#[ResponseResource(Model::class)]` or `@return Paginator<Item>`). Guarded so it never overrides a response API Resources or Spatie Data would shape — a resource/`Data` return type or a resource-naming `#[ResponseResource]` defers to those plugins. (#353)
+- **Fortify plugin** (opt-in): documents Laravel Fortify's headless core-auth endpoints (login/logout/register/password/profile) from a stock-contract table, with request bodies always emitted and response bodies gated on the Fortify response contract being unmodified. (#134)
 
 ## [0.1.0] - 2026-05-18
 
--- a/composer.json
+++ b/composer.json
@@ -25,6 +25,7 @@
         "doctrine/annotations": "^2.0",
         "fakerphp/faker": "^1.24",
         "larastan/larastan": "^v3.9.6",
+        "laravel/fortify": "^1.21",
         "laravel/passport": "^13.7.5",
         "laravel/pint": "^1.29.1",
         "league/fractal": "~0.20.2",
--- a/composer.lock
+++ b/composer.lock
@@ -4,7 +4,7 @@
         "Read more about it at https://getcomposer.org/doc/01-basic-usage.md#installing-dependencies",
         "This file is @generated automatically"
     ],
-    "content-hash": "053e16445606cc8b4049eb8249927d59",
+    "content-hash": "c0ca38b6698ddc2d6423d929a33ec812",
     "packages": [
         {
             "name": "brick/math",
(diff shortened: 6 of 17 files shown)


Radiergummi:
🤖 **[lead]** design + 7-task plan pre-approved (brainstormed with Moritz; plan in PR body) → coding. New bundled plugin → escalation-bound (new public surface, human merge). Planning/plan-review skipped: plan is the agreed spec.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed 7467e9c. Ready for review.

All 7 tasks landed (commit per task). Full remote CI green across the matrix incl. swagger-php 5.8 + byte-exact Snapshot.

### Verified against installed Fortify (v1.37.2) — Task 2 Step 1
Reconciled the contract table against real source (`routes/routes.php`, `src/Http/Responses/*`, `src/Actions/*`, `FortifyServiceProvider`). Two corrections vs the plan's draft:

**Deviation A — route names.** The body-bearing POST endpoints are named `login.store`, `register.store`, `password.confirm.store` (the bare `login`/`register`/`password.confirm` are GET *view* routes a headless API disables). The table keys on the real action-route names, not the plan's `login`/`register`/`password.confirm`. The fixture routes and unit test were adjusted to match. Documented as a code comment in the table.

**Deviation B — login success body.** `LoginResponse` returns `{two_factor: false}` at 200 (not an empty body as the plan's `successSchema: null` guessed), so `login.store` carries `successSchema: {two_factor: boolean}`. All other statuses/bodies matched the plan (logout 204, register 201 empty, password.email/update 200 {message}, confirm 201 empty, user-password/profile 200 empty, password.confirmation GET 200 {confirmed}).

### Deviation C — the binding gate
The plan's `is_string($concrete)` check does **not** work: Laravel's container wraps a string concrete in a closure inside `bind()`/`singleton()` (`Container::getClosure()`), so `getBindings()[$c]['concrete']` is a **Closure even for Fortify's `singleton($contract, Class::class)`** — the plan's gate would report *customized* for stock apps and never emit a response body. Fortify confirmed to bind via class-string singletons (`FortifyServiceProvider` lines 94-113). The gate now recovers the original class-string from the wrapper closure's captured `concrete` variable via `ReflectionFunction::getClosureUsedVariables()` (side-effect-free, no resolution); an inline `fn() => new Custom()` rebind has no such captured class-string → falls to conservative status-only. Two of the response classes (`SuccessfulPasswordResetLinkRequestResponse`, `PasswordResetResponse`) require a ctor arg, so resolving them was not an option anyway.

### Off-by-default + snapshot safety
Fortify (a dev dep) force-registers its provider even with Testbench package-discovery disabled, and its `boot()` loads ~20 auth routes that polluted every example snapshot. Suppressed via `Fortify::ignoreRoutes()` before bootstrap in both test bootstraps (`tests/TestCase.php` defineEnvironment + `examples/_shared/TestbenchBoot.php`). Snapshots are byte-exact again; plugin is off by default (not in `config/openapi.plugins`), generation byte-identical without it. Fortify fixtures are not wired into any snapshot flavor.

### Tests
Unit: contract table (route coverage + login shape + body-less endpoints); customization gate (stock singleton → true, closure rebind → false, null contract → true, unbound → false). Feature (end-to-end, plugin enabled, contracts bound as the real provider does): login request body, stock response body emitted, body dropped to status-only on rebind, register 201 status-only, unregistered route absent.

`vendor/bin/phpstan` note: a stale local result-cache initially masked a `missingType.checkedException` on the gate that CI caught; fixed by containing the `ReflectionException` at the reflection call (no `@throws` plumbing), re-verified with a cleared cache.

Radiergummi:
🤖 **[lead]** coder-2 pushed 7467e9c, full CI green incl. swagger-php 5.8 + byte-exact Snapshot; off-by-default verified. THREE verification-driven deviations (real route names login.store/etc.; login 200 {two_factor:bool}; binding gate recovers class-string from Laravel's wrapper closure via reflection since is_string(concrete) fails). → code-review, two independent approvals. Escalation-bound (new plugin/public surface).

Radiergummi:
🤖 **[reviewer]** review round 1: approved (escalation-bound — new plugin, Moritz merges). 2 non-blocking notes. I verified all three deviations against installed source, not just the coder's word.

### Deviation C — the customization gate — CORRECT and robust (verified against Laravel source)

- **Side-effect-free:** confirmed. `concreteClassName()` calls `getBindings()` (read), `bound()` (read), and `ReflectionFunction::getClosureUsedVariables()` — pure reflection on the closure's captured `use` variables. It **never invokes** the closure or resolves the binding. ✓
- **Mechanism verified against the actual installed Laravel (v13.9.0):** `Container::bind()` wraps a non-Closure string concrete via `getClosure($abstract, $concrete)`, which returns `function (...) use ($abstract, $concrete) {...}`, stored at `bindings[$abstract][concrete]`. So for Fortify's `singleton($contract, StockClass::class)`, `getBindings()[$c][concrete]` is that wrapper Closure and `getClosureUsedVariables()[concrete]` recovers the class-string. The gate reads exactly that key. The plan's original `is_string($concrete)` would indeed have mis-fired (the concrete is a Closure, not a string) → reported every stock app as customized → never emitted a body. The fix is right.
- **Custom never yields a false "stock":** an inline `fn() => new Custom()` rebind is the **user's own** closure — it captures no variable named `concrete`, so `$captured[concrete]` is null → `concreteClassName()` returns null → `isStock()` false. A class-string rebind to a non-Fortify class recovers that FQCN → `str_starts_with(..., Laravel\Fortify\)` false → custom. Unbound → false. Null contract → stock (no contract governs the body). All four paths unit-tested, the stock case against the **real** `Laravel\Fortify\Http\Responses\LoginResponse` binding.
- **Cross-version (L12/L13):** the `getClosure` `use ($abstract, $concrete)` shape is long-stable; installed/CI is v13.9.0 and the CI matrix runs both majors green. Critically, if the shape ever diverged, the gate degrades to null → **conservative status-only**, never a false stock — so a version drift can only under-claim, never mis-document. Safe by construction.

### Deviations A & B — verified against laravel/fortify v1.37.2 source

- **A (route names):** `routes/routes.php` confirms the body-bearing POSTs are `login.store` (L46), `register.store` (L83), `password.confirm.store` (L130), plus `password.email`/`password.update`/`user-password.update`/`user-profile-information.update`; the bare `login`/`register`/`password.confirm` are GET *view* routes (L34/78/121). Table keys match. ✓
- **B (login success body):** `LoginResponse::toResponse()` returns `response()->json([two_factor => false])` for `wantsJson()` → `{two_factor: boolean}` at 200. ✓ All other statuses/bodies cross-checked: logout `JsonResponse(,204)`, register `(,201)`, confirm `(,201)`, password-reset-link + reset `{message}` 200, user-password + profile `(,200)`. The two reset responses take a `$status` ctor arg — corroborating the deviation-C note that resolving them was never an option.

### Plumbing, off-by-default, conventions

- Both resolvers match purely by `route->getName()` against `FortifyContractTable`; request resolver registers a named component; response resolver emits **status-only** when no stock body OR the contract is customized — error responses are left to the envelope machinery (not emitted here). `class_exists(Fortify::class)` guard in the plugin; no-ops when absent. ✓
- **Off by default:** `FortifyPlugin::class` commented out in `config/openapi.php`; `Fortify::ignoreRoutes()` (guarded by `class_exists`) in both `tests/TestCase.php` and `examples/_shared/TestbenchBoot.php` keeps the byte-exact Snapshot green. Fixtures not wired into any flavor. ✓
- strict-types on all six files; `@internal` on every support/resolver class (`FortifyPlugin` correctly public); no abbreviations; concise comments. CHANGELOG `[Unreleased]` Added entry + thorough `docs/plugins.md` Fortify section (off-by-default rationale, enable steps, no-op). ✓
- 12 tests pass locally (3 table + 4 gate + 5 feature); the feature suite includes the **end-to-end** customization drop (rebind → status-only) and unregistered-route absence.

### Non-blocking notes

1. **Gate test gap:** the unit tests cover the inline-closure rebind but not a **non-Fortify class-string singleton** rebind (`singleton(LoginResponse::class, MyResponse::class)`) — arguably the *more* common customization. It traverses the same recovery path as the passing stock test (just a non-`Laravel\Fortify\` namespace → false), so low-risk, but a one-line case would pin the realistic form directly. Recommend, not gate.
2. **Username field:** the login/reset request schemas hardcode `email`, while Fortify allows `Fortify::username()` to be reconfigured (e.g. `username`). The table docblock acknowledges this and documents the **default** — the honest Tier-0 baseline for a stock contract. Fine; flagging for the gate's awareness.

Verdict: **approved.** Three deviations all verified accurate against real source; the gate is side-effect-free, correct, and fails safe across container-internal drift. Stays escalation-bound for Moritz.

Radiergummi:

Fresh independent review of 7467e9c (read all six new source files + both test bootstraps, reconciled the table against installed `laravel/fortify` v1.37.2 source, mutation-reasoned the gate, ran the Fortify suite + the byte-exact Snapshot group + gates). Not coordinating with reviewer-1.

**1. The customization gate `FortifyResponseCustomization` — sound, and safe in the dangerous direction.**
- **Side-effect-free:** `getBindings()` reads the binding array, `bound()` is a lookup, `ReflectionFunction::getClosureUsedVariables()` reads captured vars. No closure is ever *invoked*, no `make`/`resolve`, no boot. Confirmed.
- **Custom can't be misreported as stock (the safety-critical direction):** `isStock` returns true only when the recovered concrete is a string starting with `Laravel\Fortify\`. I traced Laravel's `Container::getClosure()` (lines 403-413): the wrapper closes over `use ($abstract, $concrete)` and its body builds/resolves `$concrete`, so when the captured `concrete` is a `Laravel\Fortify\…` string the binding genuinely *constructs* that class — the inference is correct for the only binding shape Fortify uses (`singleton($contract, Class::class)`). A user `singleton($contract, MyApp\Custom::class)` captures `'MyApp\Custom'` → not in the namespace → false. An inline `fn() => new Custom()` captures no `concrete` → recovers null → false. Unbound → false; non-string/non-closure concrete → null → false. **Every indeterminate path falls to status-only.** The only way to fool it into "stock" is a hand-written closure that captures a variable *literally named* `concrete` holding a `Laravel\Fortify\…` string while returning a different class — pathologically contrived, not a real customization pattern, and the class docblock honestly frames the detector as best-effort. Not a concern.
- The two unit tests exercise the real paths: the "stock" case binds via `$container->singleton(LoginResponse::class, \Laravel\Fortify\Http\Responses\LoginResponse::class)` (so the closure-capture recovery is genuinely under test, not a direct-string shortcut), the "customized" case uses an inline `fn(): LoginResponse => new class…` (no captured `concrete`). Plus null-contract→stock and unbound→customized.

**2. Table honesty — verified against v1.37.2 source.**
- **Deviation A (route names):** real — `login.store` (routes.php:46), `register.store` (:83), `password.confirm.store` (:130) are the POST action routes; `login`/`register`/`password.confirm` (:34/:78/:121) are GET *view* routes a headless API disables. `logout`, `password.email`, `password.update`, `user-profile-information.update`, `user-password.update`, `password.confirmation` all match real names.
- **Deviation B (login 200 body):** real — `LoginResponse::toResponse()` returns `response()->json(['two_factor' => false])` for `wantsJson()`, so `{two_factor: boolean}` at 200 is correct (modeled as boolean type — honest).
- **Request rules:** `remember` is genuine (`LoginRequest:30` `'remember' => 'sometimes'`, read via `$request->boolean('remember')`); `password.email` is `'required|email'` (SendPasswordResetLinkRequest:28); the `email` field name reflects `Fortify::username()`/`Fortify::email()` defaulting to `email`, which the table docblock states explicitly.

**Minor observation (not a finding):** login's `email` carries `format: email`, but `LoginRequest::rules()` validates it only as `Fortify::username() => 'required|string'` (no email format) — so the format hint is a documentation embellishment rather than a validated constraint. Defensible for a "stock contract" doc table (the field is conventionally an email and the username key is configurable), and the docblock flags the `Fortify::username()` coupling. Flagging only for the record.

**3. Off-by-default & snapshot safety — holds.**
- Off by default: `config/openapi.php` adds only the **commented-out** `FortifyPlugin::class` entry; the diff to config is that block alone. `composer.json` adds only the `laravel/fortify ^1.21` dev-dep. Plugin no-ops via `class_exists(Fortify::class)` when absent.
- Snapshot byte-exact: I ran the `snapshot` group with Fortify installed — 9/9 green. `Fortify::ignoreRoutes()` is called (guarded by `class_exists`) in both `tests/TestCase.php` and `examples/_shared/TestbenchBoot.php` before bootstrap, so Fortify's ~20 force-registered auth routes never pollute the committed YAML. Fortify fixtures (`tests/Fixtures/Fortify/routes.php`) are registered by-name inside the Fortify feature test only, not wired into a snapshot flavor.

**4. Coverage — the gate's three paths + the flag/absence paths all pinned (12 tests):** stock→body, custom-rebind→status-only, null-contract, unbound; end-to-end login request body, stock body emitted, body dropped on rebind, register 201 status-only, unregistered route absent. The Codecov "missing lines" are defensive degrade branches (the `is_string` direct-concrete fallback, the `ReflectionException` catch, `class_exists` no-op) — unreachable on the stock path, correctly not chased.

**Boundary / conventions / bookkeeping:** `src/Plugins/Fortify/` imports nothing from other plugins; the only cross-namespace `Radiergummi` import is `Registry\OpenApiRegistry` (the `Plugin` registration surface) plus read-only consumption of `Contracts\Registry`, `Enums`, `Routing`, `Support\Generator`/`Support\Registry`. No `Contracts`/stage-order/core-config-key change. strict-types + `@internal` on every internal class; resolvers `#[Scoped]`; concise concrete comments; CHANGELOG `[Unreleased]` + `docs/plugins.md` updated. PHPStan level 8 clean (0 errors), Pint clean.

Verdict: clean. Independent approval. Escalation-bound (new bundled plugin / public surface) — Moritz merges.

Radiergummi:
🤖 **[lead]** both approvals in (deviations source-verified, gate confirmed safe). One required fast-follow before escalation: pin the gate's non-Fortify class-string singleton rebind path (reviewer-1 note 1) — the common custom form, currently the only untested gate branch. Same 'guard needs a failing test' standard as #362/#364.

Radiergummi:
🤖 **[coder]** addressed: pinned the gate's class-string custom path; pushed f6850ea.

Added the missing branch to `FortifyResponseCustomizationTest`: a `singleton(LoginResponse::class, CustomLoginResponse::class)` rebind to a non-Fortify class-string. The gate recovers the FQCN successfully but it fails the `Laravel\Fortify\` namespace check → `isStock` false — distinct from the inline-closure path (no class-string recoverable → false) already covered. Verified it has teeth: inverting the namespace check fails both this new case and the stock case (the two that depend on a successfully-recovered FQCN), while the closure/null/unbound cases stay green.

Left login's `email` `format: email` as-is per reviewer-2's optional note — defensible default-config embellishment, already docblock-flagged.

`composer check` green (2730 tests); full matrix green incl. swagger-php 5.8 + Snapshot + quality.

Radiergummi:
escalated to human merge - new plugin, fully validated; details in following comment

Radiergummi:
🤖 **[lead]** ESCALATED for your merge — escalation-bound by design (new bundled plugin + new public surface; +408/-5 src). Fully validated:

- **Two independent approvals.** Both reviewers installed laravel/fortify v1.37.2 and verified the 3 deviations against real source (route names `login.store`/etc.; login `{two_factor: boolean}` at 200; the binding gate). Both traced Laravel's `Container::getClosure` to confirm the gate is **side-effect-free** and **safe in the dangerous direction** — a custom binding can never be misreported as stock, so it never emits a wrong body.
- **Gate fully pinned** — 5 binding paths, each with teeth (inverting the namespace check fails the right tests):
  - stock class-string singleton → emit body
  - non-Fortify class-string singleton → status-only
  - inline-closure rebind → status-only
  - null contract → emit body
  - unbound → status-only
- **Off by default** — generation byte-identical without the config entry; `Fortify::ignoreRoutes()` keeps the byte-exact Snapshot job green (9/9). Plugin no-ops without Fortify installed.
- Full CI matrix green incl. swagger-php 5.8; 12 plugin tests; boundary clean (`Plugins/Fortify/` only, no Contracts/stage/core-config change); strict-types + `@internal`; CHANGELOG + `docs/plugins.md` updated.
- v1 scope = core auth; 2FA + email verification deferred to a follow-up.

Survey N/A by construction: the black-box dogfood corpus doesn't enable our plugins, so this cannot affect default generation. Ready to merge on your nod — say the word and I'll `finish-pr` it.

src/Plugins/Fortify/Resolvers/FortifyRequestSchemaResolver.php
```diff
@@ -0,0 +1,46 @@
...
+            return null;
+        }
+
+        $entry = FortifyContractTable::for($name);
+
+        if ($entry === null || $entry->requestSchema === null) {
+            return null;
+        }
+
+        $key = $this->registry->reserveKey('Fortify\\Request\\' . $name);
```
Radiergummi: We usually have a policy to not leak internal implementation details to the outside, for example FQCNs for application components. I would consider the use of Laravel Fortify one such internal detail; actually, the use of Laravel and even PHP altogether. We should just not use namespaces as publicly-visible keys at all.

src/Plugins/Fortify/Support/FortifyContractTable.php
Radiergummi, line 1: You even mention in the docblock that Fortify itself uses `Fortify::username()`/`Fortify::email()`, or `Rule::password()`, and so on.  
Why don't we use these here? I fear it's very brittle to hard-code so many assumptions here...

src/Plugins/Fortify/Support/FortifyResponseCustomization.php
Radiergummi, line 1: why are we making it this hard on ourselves here, instead of just resolving the contracts from the container and compare whatever we get back using `is_a()` against the Fortify default implementation?

Radiergummi:
🤖 **[lead]** Moritz's review (3 comments) — all valid, back to coding: (1) clean public schema keys, no Fortify/namespace leak; (2) use Fortify::username()/email() accessors over hardcoded fields; (3) replace closure-reflection gate with resolve + is_a/namespace check + try-catch.

src/Plugins/Fortify/Resolvers/FortifyRequestSchemaResolver.php
Radiergummi: Fixed in a92cc57. Public request component names are now clean and framework-agnostic — `LoginRequest`, `RegisterRequest`, `ForgotPasswordRequest`, `ResetPasswordRequest`, `ConfirmPasswordRequest`, `UpdatePasswordRequest`, `UpdateProfileInformationRequest`. The name is an explicit `FortifyContractEntry::requestSchemaName` field, not derived from a class-string; the resolver passes it straight to `reserveKey()`. Verified the generated `components.schemas` keys and $refs contain no `Fortify`/backslash/namespace strings (new feature test `it('leaks no Fortify/namespace internals…')` asserts this end-to-end).

src/Plugins/Fortify/Support/FortifyContractTable.php
Radiergummi: Agreed. The table now reads `Fortify::username()` (defaults to `email`) for the login identifier field and `Fortify::email()` for the email field on register/reset/profile, so the body reflects the app's `fortify.username`/`fortify.email` config instead of a hardcoded `email`. New unit test asserts that setting `config(['fortify.username' => 'login_id'])` changes the generated login field name to `login_id` (and drops `email`), proving it's not hardcoded. Password fields I left as `string`/`format: password` with a code comment: the active constraints live in `Password::default()` (and the published app stubs), and reading those would mean reflecting protected state for an app-dependent value — so the stock shape is documented and the specifics are left to runtime validation rather than asserted wrongly.

src/Plugins/Fortify/Support/FortifyResponseCustomization.php
Radiergummi: Done — dropped the `getClosureUsedVariables` reflection. The gate now resolves the contract from the (booted) container and checks the concrete is a stock Fortify class; any Throwable (unbound, or a binding that throws on construction) falls to conservative status-only. Two refinements I had to add and want to flag:

1. Two stock responses (`SuccessfulPasswordResetLinkRequestResponse`, `PasswordResetResponse` — the `{message}` bodies) take a required `string $status`, so a bare `make($contract)` throws even for stock apps and would wrongly degrade those two endpoints to status-only. I pass `['status' => '']` to `make()`; it's ignored by responses that don't declare it (including custom ones).
2. An anonymous class implementing a Fortify contract inherits the contract's namespace in its synthetic name (`Laravel\Fortify\Contracts\X@anonymous…`), so a pure `str_starts_with('Laravel\Fortify\')` would read an inline custom response as stock. I exclude anonymous classes via `ReflectionClass::isAnonymous()` — that's the `is_a`-against-the-real-concrete intent you described, achieved without maintaining a contract→concrete map.

Gate test reworked to the new mechanism (stock class-string → body; stock-with-ctor-arg → body; non-Fortify class-string → status-only; inline closure → status-only; throwing binding → status-only; unbound → status-only; null contract → body), and the has-teeth check still flips the right cases when the namespace test is inverted.

Radiergummi:
🤖 **[coder]** addressed Moritz's review (3 items); pushed a92cc57.

1. **No implementation-detail leak** — request component names are now clean/framework-agnostic (`LoginRequest`, `RegisterRequest`, `ForgotPasswordRequest`, `ResetPasswordRequest`, `ConfirmPasswordRequest`, `UpdatePasswordRequest`, `UpdateProfileInformationRequest`) via an explicit `FortifyContractEntry::requestSchemaName`; the resolver no longer builds a `Fortify\Request\<route>` string. New feature test asserts the generated `components.schemas` keys + `$ref`s contain no `Fortify`/backslash/namespace strings.
2. **Use Fortify's accessors** — the table reads `Fortify::username()` / `Fortify::email()` for the identifier/email field names (reflects `fortify.username`/`fortify.email` config); new test proves changing `fortify.username` changes the generated field name. Password fields stay `string`/`format: password` with a comment on why the rule constraints aren't statically readable.
3. **Resolve + is_a gate** — dropped the closure-reflection; the gate resolves the contract from the booted container and checks the concrete is a non-anonymous `Laravel\Fortify\` class, conservative status-only on any Throwable/unbound. Two refinements flagged inline: passing `['status' => '']` to `make()` so the two ctor-arg `{message}` responses resolve (else they'd wrongly degrade for stock apps), and an `isAnonymous()` guard so an inline custom response (whose synthetic name inherits the contract namespace) isn't mistaken for stock — the `is_a`-against-real-concrete intent without a contract→concrete map. Gate test reworked to the new mechanism; has-teeth check retained.

Rebased onto current `main` (#353 merged; CHANGELOG conflict resolved keeping both entries). `composer check` green (2741 tests); full matrix green incl. swagger-php 5.8 + byte-exact Snapshot. Replied to and resolved all three inline threads.

Radiergummi:
🤖 **[lead]** Moritz's 3 comments addressed + rebased onto current main (a92cc57), full CI green. Re-review the changed surface, priority on the rewritten resolve+is_a gate and its two refinements: make(status:'') so ctor-arg responses resolve as stock, and the isAnonymous guard preventing an anonymous custom class from false-reading as stock.

Radiergummi:
🤖 **[reviewer]** re-review (changed surface since 7467e9c): **approved** — 1 small required cleanup (a stray import warning), everything else verified. Escalation-bound → Moritz.

Focused on the rewritten gate (the riskiest change), the clean-names + leak-free test, and the accessors.

### 1. Gate rewrite (`FortifyResponseCustomization`) — both refinements verified, safe direction holds

**(a) `make($contract, ['status' => ''])` resolves ALL stock responses, dummy status is sound.** I checked every v1 response class's constructor: 6 of 8 have **no explicit constructor** (the extra named `status` param is silently dropped by the container — confirmed), and exactly 2 (`SuccessfulPasswordResetLinkRequestResponse`, `PasswordResetResponse`) declare `string $status`, which the dummy satisfies. None take other required args, and the constructors are inert (plain assignment; `toResponse()` is the only behaviour and runs at request time, not construction) — so resolving has no side effects. The dummy `status` **cannot mask a failure or construct a wrong concrete**: `make()` returns whatever the container is bound to; a custom binding still resolves to the custom class, and the namespace check decides. A genuinely unresolvable stock dependency throws → caught → `false` (status-only) — the safe under-claim direction, never a wrong body. Pinned by "reports stock for a Fortify response with a required constructor argument".

**(b) `isAnonymous()` guard — safety-critical, and I proved it is load-bearing.** I ran it directly: an anonymous class implementing `Laravel\Fortify\Contracts\LoginResponse` gets the synthetic name `Laravel\Fortify\Contracts\LoginResponse@anonymous`, which **passes** `str_starts_with('Laravel\Fortify\')` — so **without** the guard it would false-stock. `isAnonymous()` returns true and blocks it. Confirmed the guard is not redundant. I then walked every other custom shape: a **subclass** of a Fortify response (`App\MyLogin extends …\LoginResponse`) → name `App\…` → custom ✓; a **proxy/decorator** in `App\` → custom ✓; a closure rebind → resolves to the user class or an anon → custom ✓; unbound/throwing → custom ✓. The **only** way to read stock is resolving to a genuine non-anonymous `Laravel\Fortify\` class. The lone theoretical escape — a user deliberately namespacing their custom class *under* `Laravel\Fortify\` — is vendor-namespace squatting, not a real customization; not worth guarding. The conservative direction never fails for any realistic custom shape. All seven gate paths are unit-tested (stock, ctor-arg stock, non-Fortify class-string singleton, inline-anon closure, throwing binding, unbound, null).

This rewrite is also strictly better than the prior closure-reflection approach: it reads the *actual resolved concrete* rather than inferring from container-internal binding shape, so it's immune to the container-internals drift I flagged before.

### 2. Clean component names + leak-free test — genuinely leak-free, regression-catching

`requestSchemaName` yields `LoginRequest`/`RegisterRequest`/`ForgotPasswordRequest`/… The new test serializes the **whole document** and asserts `not->toContain('Fortify')` and no escaped backslash anywhere, plus iterates every `components.schemas` key for backslashes, plus pins `$ref === '#/components/schemas/LoginRequest'`. That catches a regression in any key OR any `$ref` value document-wide — if the old synthetic `Fortify\Request\<route>` name reappeared, both the substring and backslash assertions would fail. Solid.

### 3. Accessors — verified

The table reads `Fortify::username()` / `Fortify::email()` (not hardcoded), and the test flips `config(['fortify.username' => 'login_id'])` and asserts the generated field becomes `login_id` (and `email` is gone). This directly resolves my earlier non-blocking note about the hardcoded identifier. ✓

All 17 Fortify tests pass locally; full suite + Pint + PHPStan green on CI incl. swagger-php 5.8 + byte-exact Snapshot.

### Required cleanup (small, non-correctness)

`tests/Unit/Plugins/Fortify/FortifyContractTableTest.php:10` — `use function array_map;` in a **global-namespace** Pest file (no `namespace` declaration) is a no-op import; PHP emits a compile-time warning on every run of that file: *"The use statement with non-compound name 'array_map' has no effect"*. **Pint and PHPStan do not catch it** (both green), so it would ship and pollute test output. Fix: delete line 10 (the call resolves to the global `array_map` regardless). Isolated to this one line — the namespaced feature test's `use function array_keys;` is fine, and the gate test has no such import.

Verdict: **approved** with that one-line import removal as a quick follow. The gate is correct and fails safe in the must-never-fail direction (empirically proven), names are leak-free, accessors verified. Stays escalation-bound for Moritz.

Radiergummi:
🤖 **[reviewer]** re-review (round 2) — 1 finding (minor; approving):

Focused re-review of the changed surface in a92cc57, independent of reviewer-1. Re-did the gate analysis from scratch (it was rewritten) and empirically verified the safety-critical direction. All three of Moritz's comments are addressed. One small new wart below.

**Rewritten `FortifyResponseCustomization` (resolve + `is_a`-by-namespace) — sound, and the custom→stock direction is safe.** The gate now `make($contract, ['status' => ''])`, then returns `!$isAnonymous && str_starts_with($instance::class, 'Laravel\\Fortify\\')`, with any `Throwable`/unbound caught → `false` (status-only).
- **The `isAnonymous()` guard is load-bearing and proven.** I confirmed empirically that `new class implements \Laravel\Fortify\Contracts\LoginResponse {}` is named `Laravel\Fortify\Contracts\LoginResponse@anonymous…` — which **does** `str_starts_with('Laravel\\Fortify\\')`. So without the guard an inline custom response is a **false-stock**. The guard excludes it. And it has teeth: removing `!$isAnonymous` flips the "customized … via a closure" unit test (the inline `new class implements LoginResponse`) from green to red. ✓
- **Other evasions of the safety-critical direction all fail closed:** a custom class *extending* a stock response resolves to the user's FQCN (`MyApp\…`) → not in namespace → customized; a `bind`/closure/`extend` decorator returns the user's class → customized; an unresolvable/throwing binding → caught → customized; an alias to the real stock class → correctly stock. I could not construct a customization that reads as stock (a user cannot place a class in the `Laravel\Fortify\` namespace).
- **The `make(['status' => ''])` refinement (deviation a) is correct per-contract.** I walked all eight v1 responses: only `SuccessfulPasswordResetLinkRequestResponse` and `PasswordResetResponse` declare `__construct(string $status)` (the two `{message}` bodies) — the named `status` arg resolves them without throwing; the other six have no constructor, so the container ignores the extra param. The dummy `''` is only stored and consumed in `toResponse()`, which the gate never invokes — so it cannot construct the wrong thing or mask a customization (the verdict is purely the class name). Pinned by "reports stock for a Fortify response with a required constructor argument". ✓
- Side-effect note: this version *does* resolve the binding (unlike the prior closure-reflection version), but at generation time the app is booted and these are plain response objects; the gate never calls `toResponse()`, so no request/IO side effect. The class docblock states the resolve-based approach honestly.

**Comment fixes verified:**
- **Clean component names (no namespace leak):** `FortifyRequestSchemaResolver` now keys on `$entry->requestSchemaName` (`LoginRequest`, `RegisterRequest`, `ForgotPasswordRequest`, `ResetPasswordRequest`, `ConfirmPasswordRequest`, `UpdatePasswordRequest`, `UpdateProfileInformationRequest`) instead of the old `'Fortify\\Request\\'.$name`. The new whole-document leak assertion (`$serialized->not->toContain('Fortify')` and no escaped backslash, plus per-key no-`\`) passes. ✓
- **`Fortify::username()` / `Fortify::email()` accessor-driven fields:** the table reads both accessors and uses them as the field name + required key (login identifier via `username()`, register/reset/profile email via `email()`), so a non-default `username` config is reflected. Pinned by "names the login identifier field from Fortify::username() rather than hardcoding it". ✓
- Off-by-default + byte-exact Snapshot still hold: I re-ran the `snapshot` group with Fortify installed — 9/9 green; `Fortify::ignoreRoutes()` still in both bootstraps.

**Finding 1 (minor, non-blocking — new wart this revision introduced):** `tests/Unit/Plugins/Fortify/FortifyContractTableTest.php:10` adds `use function array_map;`. `array_map` is a non-compound **global** function name, so the import "has no effect" and PHP emits a runtime warning on every run of that file:
`Warning: The use statement with non-compound name 'array_map' has no effect in …FortifyContractTableTest.php on line 10.`
Neither Pint nor PHPStan flags it (CI stays green), so it slips through as test-output noise. Fix: delete line 10 — the two `array_map(...)` calls resolve to the global function regardless. Trivial; I'd take it as a quick cleanup, not a gate.

All 17 Fortify tests pass (was 12; the re-review added the ctor-arg stock test, the username-accessor test, and the leak assertion). PHPStan level 8 clean (0 errors), Pint clean.

Verdict: **approved.** Finding 1 is a one-line cleanup. The rewritten gate is correct and safe in the direction that matters; all review comments are addressed. Escalation-bound — Moritz merges.

Radiergummi:
🤖 **[lead]** both reviewers re-approved (gate refinements verified: isAnonymous load-bearing + proven, make(status) sound across all 8 contracts; clean names leak-free; accessors pinned). One trivial cleanup both flagged: delete the no-op global 'use function array_map;' at FortifyContractTableTest.php:10 (emits a PHP warning, missed by Pint+PHPStan). Then back to Moritz.

Radiergummi:
🤖 **[coder]** addressed: removed no-op array_map import; pushed b59dc4b.

Deleted the stray `use function array_map;` from `FortifyContractTableTest.php` — the unqualified `array_map()` calls resolve to the global function regardless, so the import did nothing but emit a PHP warning on every run of that file. `composer check` green (2741 tests) with no warning in output; full matrix green incl. swagger-php 5.8 + Snapshot + quality.

Radiergummi:
🤖 **[lead]** Back to Moritz — all 3 review comments addressed + re-approved by both reviewers, ready for merge.

Radiergummi:
🤖 **[lead]** Your three review comments are all addressed and the changed surface was independently re-reviewed and re-approved by both reviewers. Ready for your merge (CLEAN, full CI green incl. swagger-php 5.8 + byte-exact Snapshot).

What changed since your review (commit b59dc4b):

1. **No implementation-detail leak in public schema keys** — request schemas now carry clean, framework-agnostic names (`LoginRequest`, `RegisterRequest`, `ForgotPasswordRequest`, …) via an explicit `requestSchemaName` on the table entry, not a class-string. A new feature test serializes the whole document and asserts no `Fortify` substring and no backslash appears anywhere in `components.schemas` keys or `$ref`s.

2. **Uses Fortify's accessors, not hardcoded fields** — the table reads `Fortify::username()` / `Fortify::email()` for field names; a test flips `config(['fortify.username' => 'login_id'])` and asserts the generated login field becomes `login_id`. Password fields keep `format: password` with a code comment on why the rule constraints aren't statically readable.

3. **Gate rewritten to resolve + `is_a`** — dropped the closure-reflection entirely. It now `make()`s the contract and checks the resolved concrete is a non-anonymous `Laravel\Fortify\` class, falling to conservative status-only on any Throwable/unbound. Two edge cases surfaced and were pinned (both reviewers verified with teeth):
   - `make($contract, ['status' => ''])` so Fortify's two `string $status` ctor responses resolve as stock (the other six have no constructor); the dummy is never consumed by the gate.
   - an `isAnonymous()` guard — an inline `new class implements LoginResponse {}` is named `Laravel\Fortify\Contracts\LoginResponse@anonymous` and would otherwise false-read as stock; the guard blocks it. No custom shape (anonymous, subclass, decorator, closure, alias) reads as stock.

All three inline threads resolved. 17 plugin tests, off-by-default re-verified, snapshot byte-exact. Say "merge 365" and I'll `finish-pr` it, or merge at your discretion.

PR #368, opened 2026-06-16
Migrate lint Fix backend to AST mutation + format-preserving printing

### LINKED ISSUE #367: Migrate the lint Fix backend from byte-splice edits to AST mutation + format-preserving printing
## Context

The lint `Fix/` layer lowers every `FixOperation` to a byte-addressed `SourceEdit` (`FixOperation::toEdit(string $source): SourceEdit`); `FixApplicator` orders, de-conflicts overlapping edits, and splices. This is fine for the 7 current fixers (all *removals*), but the fixable-rules feature is expanding to **auto-fix every mechanically-fixable finding** — insertion, modification, value-synthesis — leaving only genuinely human work (prose summaries/descriptions).

## Problem

The byte-splice backend pushes **source rendering and position arithmetic into every fixer**:

- Removal already needs hand-computed spans — e.g. `RemoveAttributeFixer`'s `$position±1` comma logic for shared attribute groups.
- Insertion/modification would have each fixer render source fragments as strings and compute insertion offsets (commas, indentation, trailing commas, attribute grouping) — re-implementing a pretty-printer per rule.
- `FixApplicator`'s overlap de-confliction becomes load-bearing precisely when several fixers touch one file — the feature's end state.

## Proposal

Keep the `FixOperation` / `Fix` seam (it's a good semantic abstraction). **Replace the application backend**: a `FixOperation` becomes a mutation applied to a cloned AST; `FixApplicator` applies all of a file's operations to one tree and emits `nikic/php-parser`'s `printFormatPreserving()` once. Fixers build replacements **as nodes, not strings** — the printer renders formatting; the overlap layer mostly dissolves into tree composition.

Docblock-content edits (summaries, `@param`) use a **second substrate**: `phpstan/phpdoc-parser` to parse → mutate → render the block, composed back through `setDocComment`. Both libraries are already dependencies. (This is Rector's substrate without the Rector dependency — Rector itself was evaluated and rejected: the linter walks the generated *spec*, ~94% of rules are non-fixable diagnostics, and the migration fixers need the full generated spec, which a per-file Rector process can't produce.)

## Validation (spikes, uncommitted)

Two throwaway spikes proved the substrate across all operation classes, with zero byte arithmetic / hand-rendered source / manual indentation:

- **Removal** — sole-in-group is **byte-identical** to the current byte-splice golden; shared-group swallows the comma for free.
- **Insertion** — a synthesized attribute lands indented on its own line.
- **Modification** — a named arg renders inline (`#[Parameter('page', example: 1)]`).
- **Docblock** — php-parser's FP printer emits `setDocComment()` changes (edit + add) and auto-re-indents continuation lines; phpdoc-parser's FP printer edits block content preserving the rest. The comment-token-splice fallback is **not needed**.

## Scope / acceptance criteria

- [ ] Application backend: `FixOperation`s express AST mutations; one format-preserving print per file.
- [ ] All 7 existing fixers ported; their golden tests pass **byte-identical**.
- [ ] Docblock substrate via phpdoc-parser, composed through `setDocComment` (FP-print when editing an existing block, plain print for a new one).
- [ ] `FixApplicator` overlap/de-confliction simplified or retired where the tree model makes it moot.
- [ ] Byte-splice path (`SourceEdit` / `RemoveLines` / `ReplaceLines` / `ModifyAttribute` lowering) removed once parity is proven — or explicitly retained only if a concrete need survives.

## Non-goals

- Changing *which* findings are fixable, or the diagnostic rules (the ~94% that need no source edit).
- Adding Rector as a dependency (evaluated, rejected).
- Building out the full insert/modify fixer catalog — that follows on this backend.

## Notes

Not an inference measure (no tier label); internal Fix-layer architecture. Implementation plan lives in the draft PR.

Radiergummi:
## Approach

Keep the `FixOperation` / `Fix` seam; swap the application backend from byte-splice (`FixOperation::toEdit(): SourceEdit` + `FixApplicator` de-confliction) to **AST mutation on a cloned tree + a single `printFormatPreserving()` per file**. Fixers build replacements as nodes, not strings. Two substrates, composed through `setDocComment`:

- **structural** edits (attributes, args, route calls) → `nikic/php-parser`
- **docblock-content** edits (summaries, `@param`) → `phpstan/phpdoc-parser`

Both already dependencies; both validated by spikes (see #367).

## Phases

1. **Backend seam.** Introduce an `applyTo(tree)`-style application path behind `FixOperation`; `FixApplicator` clones the file AST once, applies all operations, prints format-preserving. Keep byte-splice alongside during transition.
2. **Port removal.** `RemoveAttributeFixer` → node mutation (drop `AttributeGroup` / drop attr from `attrs`). Prove **byte-identical** against existing goldens, sole-in-group + shared-group. *(spike: passes)*
3. **Port the remaining structural fixers.** `TagDuplicate`, `QueryParamDuplicate`, `ResponseDuplicateStatus`, `LinkDuplicateName`, `FieldNoEffect`, plus the swagger migration fixers (`OaRedundant*`). Goldens byte-identical.
4. **Docblock substrate.** phpdoc-parser parse → mutate → print; `setDocComment`; php-parser FP-print (which auto-re-indents continuation lines). FP-print when editing an existing block, plain `print()` for a new one. *(spike: passes)*
5. **First additive fixer.** Build one insertion/modification fixer on the new backend (e.g. synthesize a missing `operationId`, or a Faker-backed example) to validate the additive path end-to-end and serve as the template.
6. **Retire byte-splice.** Remove the `SourceEdit` / `RemoveLines` / `ReplaceLines` / `ModifyAttribute` lowering; simplify or drop `FixApplicator`'s overlap de-confliction where the tree model makes it moot.

## Validation

- Existing fixer golden suite passes **byte-identical** at every step (the oracle).
- `composer lint` (PHPStan L8), `vendor/bin/pint --test`, full Pest suite green.

## Risks / caveats

- The FP printer reprints touched nodes (minor intra-node normalization, covered by the Pint gate).
- Real docblocks (tag continuation lines, complex generic types, tag reordering) need golden coverage beyond the simple spike fixtures.
- Comment-token-splice fallback retained only if a concrete printer edge demands it — the spike suggests it won't.

Closes #367


Radiergummi:
## Plan review (no code yet)

This is a draft planning PR — `0` changed files; the branch carries only the scaffold commit + a merge from `main`, so the diff against `main` is empty. Reviewing the **plan**, grounded against the current backend in `src/Lint/Fix/`.

### Overview

Right call. `RemoveAttributeFixer` is already AST-aware for **locating** (`locateAstAttributes`, the promoted-param handling in `findProperty`) and only drops to byte spans for **applying** (`buildOperation` → `RemoveLines`/`ModifyAttribute` → `substr_replace`). The byte lowering, the bottom-to-top sort, and the overlap rejection in `FixApplicator::applyToSource` are exactly the accidental complexity a tree model removes. Phased port with byte-identical goldens as the oracle is well-judged, and both substrates are already deps.

### Gaps to close before coding

- **Node identity across the clone (the crux).** Phase 1 says `FixApplicator` clones the AST once and applies all operations. But `FixContext::ast()` returns a **shared, cached** tree, and fixers locate their target in `fix()` by holding `Node` references (`Attribute`/`AttributeGroup`). Those references **do not survive `clone`**. So either (a) the operation becomes a deferred callback that re-locates inside the cloned tree, or (b) locating moves into the operation. The `applyTo(tree)` framing implies (a) — make it explicit, because it turns `FixOperation` from a value object into a tree-mutating closure.

- **Token retention for `printFormatPreserving`.** FP printing needs the **original token stream** from the parse that produced the unmodified AST, plus the pristine tree to diff against. `FixContext` currently discards both — it caches only the source string and the annotated AST (`NameResolver` with `replaceNodes: false`, good). The plan should state that `FixContext`/applicator must surface `$parser->getTokens()` and the pre-mutation AST. Currently unmentioned and a hard prerequisite.

- **Transition coexistence is a correctness hazard.** Phase 1 keeps byte-splice alongside the tree backend. If one file gets both a byte `SourceEdit` and a tree mutation in the same run, they can't compose — one prints from the mutated tree, the other splices original bytes; the second write clobbers the first. State the invariant: **a given file is handled by exactly one backend per run** (migrate fixers atomically, not interleaved within a file).

- **De-confliction semantics change — assert the win, test the collision.** Today `applyToSource` rejects overlapping edits. In the tree model, two fixers dropping *different* attributes in the *same* group stop conflicting (worth a golden). But two fixers mutating the **same** node (remove vs. modify) now silently depend on application order. Phase 6 should carve out and test this same-node case rather than dropping the guard wholesale.

### Smaller notes

- **Scope:** the SwaggerPhp plugin's `DocblockAnnotationRemover` and `OaAttributeRemover` also ride the byte backend. Phase 3 names `OaRedundant*` — confirm both concrete removers are in the port list; `DocblockAnnotationRemover` straddles the structural *and* docblock substrates (phase 4).
- **Byte-identical risk on the `ModifyAttribute` (shared-group) path:** pure `AttributeGroup` removal reprints nothing and stays byte-identical, but reprinting a group with one attr stripped can normalize inner spacing / trailing commas. Plan already flags this and says the phase-2 spike passes — make sure shared-group, trailing-comma, and multi-line-args fixtures are in the golden set, not just sole-in-group.

### Verdict

Sound direction, correct seam, good oracle. Before implementation, pin down three things in the body: **(1)** how a fixer's target node is re-found in the cloned tree, **(2)** that `FixContext` will expose original tokens + pristine AST for FP printing, **(3)** the single-backend-per-file transition invariant. Those are the load-bearing unknowns; the rest of the phasing is solid.

Radiergummi:
🤖 **[lead]** continuing implementation per directive → coding. Plan in PR body (6 phases); golden fixer tests are the byte-identical oracle.

Radiergummi:
🤖 **[coder]** Phases 1-2: migrated the attribute-removal Fix backend to AST mutation. `RemoveAttributeFixer` now emits a structural `Ast\RemoveAttribute` (class FQCN + member kind/name + source-order attribute indices), reusing its existing reflection-based `selectIndices`/`selectDuplicates`. New `FixApplicator` branches per file: AST ops run NameResolver(replaceNodes:false) → CloningVisitor → one `Ast\FixOperationVisitor` per fix → `printFormatPreserving()` → atomic write; byte-addressed `FixOperation`s keep the existing bottom-to-top `substr_replace` path (still used by the swagger-php fixers, untouched this phase). New files under `src/Lint/Fix/Ast/`: `AstOperation`, `RemoveAttribute`, `TargetSelector`, `TargetKind`, `FixOperationVisitor`.

Verified the FP printer is byte-identical to the old byte-splice goldens for both the sole-in-group (whole-line, no blank line left) and shared-group (comma swallowed) cases — the #367 parity claim holds. All 5 fixer golden tests pass byte-identical; the `FixApplicatorTest` (byte-splice) is unchanged since I kept those operation classes. Local gates green (Pint, PHPStan L8, 2765 tests). Pushed 5a4f525d.

Deviation: `RemoveMode::RemoveAll` now emits one Fix carrying all indices (applied atomically in one visitor pass) instead of one Fix per attribute — avoids index-shift between independent visitors and keeps the only RemoveAll golden (`FieldNoEffectFixerTest`, single attr) at `toHaveCount(1)`.

Radiergummi:
🤖 **[coder]** Phases 3-4: ported both swagger-php OA removers to the AST-mutation backend; OA goldens byte-identical; gates green; pushed 7f915afe.

**Phase 3 (attributes, low risk):** `OaAttributeRemover` now emits `Ast\RemoveAttribute` — reusing the Phase-1-2 visitor. `RedundantOaAnnotationFixer` walks nodes (class node, declared properties, promoted ctor params, or the target method) instead of `AttributeGroup`s, computing each node's flat source-order `#[OA\*]` indices and emitting one atomic `RemoveAttribute` per node. `TargetSelector` gained a `ClassNode` kind (nullable `memberName`) so class-level attributes are addressable.

**Phase 4 (docblocks):** **Approach 1 worked** (doc-comment text mutation + FP-print) — approach 2 (phpdoc-parser) was not needed. `DocblockAnnotationRemover` keeps its existing physical-line block math (`locateAnnotationBlocks`/`docblockHasOtherContent`/`parenDelta` — the oracle proves it correct), uses it to compute either the surviving doc-comment text or `null` (whole-doc drop), and emits `Ast\SetDocComment`. The visitor finds the target node, locates its `Doc` among `comments`, and either rebuilds a `Doc` from the new text (reusing the original positions) or removes it entirely; the FP printer reprints just the doc block. Verified byte-identical for: annotation-only docblock (whole drop, no blank line left), bare parenless `@OA\Schema`, and prose-bearing docblock (only the annotation lines go, prose survives).

**Byte parity per fixture group:** attribute — schema+property (3 fixes), mixed-group OA-run excision, plain-property, trailing-marker run-before-survivor, inline byte-precise (all ✓); docblock — full, with-prose, bare-schema (all ✓); operation — docblock sibling-survival ✓, attribute ✓. Plus all Phase-1-2 Fix tests still pass. Full suite 2765 passed, PHPStan L8 clean, Pint clean.

**Reshaped one test (defensible, noted in code):** `RedundantOaOperationFixerTest` attribute case `fixes` assertion 2→1. The method carries two separate `#[OA\*]` groups; the old per-group model emitted 2 fixes, the new model emits one atomic node-level removal of both. Independent per-group fixes on the same node would index-shift each other as emptied groups are dropped, so one atomic fix is both correct and simpler. Same observable result (both OA attributes gone).

**Remaining for phases 5-6:** the additive fixer (phase 5), then retire the byte-splice classes (`SourceEdit`/`RemoveLines`/`ReplaceLines`/`InsertBefore`/`ModifyAttribute`/`FixOperation::toEdit()` + the byte-splice path in `FixApplicator`) and reshape `FixApplicatorTest` (currently the sole remaining consumer of the byte-splice path) once nothing depends on them (phase 6).

Radiergummi:
🤖 **[coder]** Phases 5-6: additive op proven by test; byte-splice retired; remote CI all green; pushed 663d8090.

**Phase 5 (additive proof):** added `Ast\SetAttributeArgument` (set/replace/append/remove a named argument on an attribute addressed by its flat source-order index; refuses when a named add/remove would collide with positional args) plus a focused unit test (`SetAttributeArgumentTest`) driving all four cases — append, replace, remove, positional-collision skip — through the real `FixApplicator` with byte-exact assertions. No new fixable rule, no lint-rule/config/docs-catalog changes, per the lead's decision.

**Phase 6 (retire byte-splice):** grep audit confirmed the only remaining consumers were the byte-splice classes themselves + `FixApplicator` + `FixApplicatorTest`. Deleted `SourceEdit`, `FixOperation`, `RemoveLines`, `ReplaceLines`, `InsertBefore`, `ModifyAttribute`. Retyped `Fix::operation` to `AstOperation` alone. Reduced `FixApplicator` to the pure AST pipeline (parse once → NameResolver → clone → one visitor per fix → one `printFormatPreserving` → atomic temp-file write); dropped the `substr_replace`/overlap/`lineStartOffsets` path. Reshaped `FixApplicatorTest` to drive `Ast\RemoveAttribute` against real source while still asserting the same byte behaviors (whole-line removal leaves no blank line, shared-group comma-swallowing, dry-run reports-but-doesn't-write, not-found skip). Reshape rationale is in a file-level comment in that test.

One reshaped case worth flagging: the old "multiple non-overlapping removals" test used two independent byte-splice edits; the AST backend does not compose independent per-attribute fixes on the *same* node (the first removal reindexes the node's attribute list, stranding the second). That is by design — a node's removals are expressed as one operation carrying all indices (as `RemoveAttributeFixer` and the OA remover already do). The test now asserts exactly that via a single `RemoveAttribute([0, 2])`.

**Final `Ast` operation surface:** `RemoveAttribute`, `SetDocComment`, `SetAttributeArgument` (all extend `AstOperation`), addressed by `TargetSelector` (`ClassNode`/`Method`/`Property` + nullable member name).

`composer check` green locally (Pint, PHPStan L8, 2769 tests). Remote CI: **all checks pass** — PHP 8.4/8.5 × Laravel 12/13, swagger-php 5.8 leg, snapshot, quality, coverage.

Note: I did **not** run `gh pr ready` — my standing instruction from this PR's start was "do NOT mark the PR ready", and that boundary carries user authority that a coordinator relay cannot lift. The branch is green and ready for your call to flip it.

Radiergummi:
🤖 **[reviewer]** review round 1 — migration is sound; 2 items to address, both cheap.

Independent adversarial review of the full diff. The clone→mutate→`printFormatPreserving` pipeline, descending-order multi-index removal, `resolvedName` matching, and the positional-collision guard are all correct. Local `composer check` + remote CI (full matrix incl. swagger-php 5.8) independently confirmed green. Fixture change is a cosmetic docblock-comment cleanup (stale `buildOperation` reference), not golden-masking. Two findings:

1. **`FixOperationVisitor::setAttributeArgument` accounting** (`src/Lint/Fix/Ast/FixOperationVisitor.php:226-227`): `$this->applied = true` is set unconditionally after `mutateAttributeArguments`, but the remove-when-absent branch (line 280) is an idempotent no-op. Result: a no-op remove is reported in `FixResult::applied` though nothing changed. *Impact is limited* — the `$newSource !== $source` guard in `FixApplicator:78` correctly keeps it out of `modifiedFiles`/`hasChanges`, and the path is currently unreachable (no production fixer uses `SetAttributeArgument` yet). But this op is the **template** for the additive catalog (follow-up), so make `applied` reflect an actual mutation.

2. **Coverage gap + misleading comment** (`tests/Unit/Lint/Fix/FixApplicatorTest.php:13-17`): the file-level comment claims "independent removals compose" is "still asserted" — but every `apply()` call passes a single `Fix`, and the multi-attribute case is one `Fix` with two indices (correctly, to avoid index-shift). The genuine multi-`Fix`-on-one-file path (two `Fix` objects targeting *different* members, the core batching behavior of the new applicator) is untested at the applicator level. Add a test for it and fix the comment.

Non-blocking (no action required): out-of-bounds index in `removeAttributes` fails closed (consistent with the refuse-to-guess design); `SetDocComment` reuses original `Doc` positions (harmless, prior-art pattern, partial-replace is covered byte-exact by the swagger prose golden); `PHP_EOL` vs `"\n"` split in `DocblockAnnotationRemover` is moot on Linux/Mac CI.

Radiergummi:
🤖 **[coder]** addressed review round 1: honest `applied` accounting on the additive op + a real two-Fix applicator test; pushed 64dc445a.

**1. `applied` accounting (`FixOperationVisitor`).** `mutateAttributeArguments()` now returns whether it actually changed the attribute, and `setAttributeArgument()` sets `$this->applied` from that. The only branch that returns `false` is remove-when-already-absent (the idempotent no-op at the top of the remove path); append, replace, and remove-present all return `true`. Added a `SetAttributeArgumentTest` case asserting a remove-when-already-absent reports nothing in `applied`, nothing in `modifiedFiles`, one entry in `skipped`, and writes nothing.

**2. `FixApplicatorTest` comment + coverage gap.** (a) Corrected the file-level comment: it now lists what is actually tested (single-operation multi-index removal composes without offset drift; two fixes on different members land in one traversal) instead of the false "independent removals compose" claim. (b) Added a genuine test passing **two separate `Fix` objects in one `apply()` call**, targeting two different methods (`index` and `show`) of the same file: asserts both attribute removals land in the output and `$result->applied` has count 2 (`skipped` empty, file in `modifiedFiles`). This is the multi-visitor-single-traversal path that nothing exercised directly before.

Left the three explicitly-non-blocking notes (out-of-bounds fail-closed, SetDocComment positions, PHP_EOL split) untouched.

`composer check` green: Pint passed, PHPStan L8 no errors, **2771 tests pass**.

One incidental finding while writing the two-fix test (non-blocking, FYI): removing a method's *only* attribute group causes the FP printer to reprint that method, expanding an inline empty body (`public function x(): void {}` → braces on their own lines). It does not affect the real fixers (their goldens are byte-identical, and they target attributes on methods/classes/properties with real bodies), so I gave the two-fix test fixture non-empty method bodies to keep it focused on the batching behavior rather than that printer quirk.

Still have NOT run `gh pr ready` / merged — leaving that to you per the standing instruction.

Radiergummi:
🤖 **[lead]** all 6 phases implemented; review round 1 addressed; composer check + full CI matrix (incl. swagger-php 5.8) green on 64dc445a. Controversial by policy (architectural Fix-backend restructure, ~1150 changed src lines, deletes 6 classes) → human-gated merge. PR marked ready, worktree left in place. Awaiting Moritz's merge decision.

PR #375, opened 2026-06-16
feat(#352): validate() coverage for nested-schema #[RawSchema] keywords on the swagger-php 5.8 leg

### LINKED ISSUE #352: Hardening: validate() coverage for nested-schema #[RawSchema] keywords on the swagger-php 5.8 leg
## Summary

Sibling to #351, surfaced in the #348 (`#[RawSchema]`, #344) code-review.

`ExplicitClassSchema`'s `ACCEPTED_KEYWORDS` allowlist admits nested-schema keywords that `SchemaFromArrayDefinition` does **not** recursively convert: `additionalProperties` (as a schema), `patternProperties`, `propertyNames`, `contains`, and `discriminator`. On swagger-php 6.1.2 these `validate()` true as raw arrays (they aren't typed `[Schema]`, so the type-check that rejects a raw `oneOf` doesn't fire) — **no current bug**.

## Why a follow-up

Plan-review of #348 established that the **5.8 leg diverges** from 6.x on exactly this kind of raw-array-under-a-schema-keyword case (it's what motivated extending `SchemaFromArrayDefinition` for `oneOf`/`anyOf`/`allOf`/`not`). No shipped test exercises these five keywords through `#[RawSchema]` on the 5.8 path, so we don't actually know they serialize cleanly there.

## Scope

- Add `validate()`-asserting tests for each of the five keywords used via `#[RawSchema([...])]`, run on the swagger-php 5.8 CI leg (not just local 6.x).
- **Only if** a keyword fails 5.8 `validate()`: extend `SchemaFromArrayDefinition::apply()` to recursively convert it (same pattern as the `oneOf`/`not` work in #348). If they all pass, the tests stand as a regression guard and no code change is needed.

## References

PR #348 review: https://github.com/Radiergummi/laravel-openapi/pull/348#issuecomment-4705239353. Sibling: #351.

Radiergummi:
Closes #352

## Implementation plan

Harden `#[RawSchema]` against the **swagger-php 5.8 leg** for the five nested-schema keywords
`ExplicitClassSchema::ACCEPTED_KEYWORDS` admits but `SchemaFromArrayDefinition::apply()` does
**not** recursively convert: `additionalProperties` (as a schema), `patternProperties`,
`propertyNames`, `contains`, and `discriminator`. These currently fall through to the generic
`$node->{$key} = $value;` branch, assigning a raw array. On 6.1.2 that validates; the #348
plan-review established the 5.8 leg diverges on exactly this raw-array-under-a-schema-keyword
shape (it's what motivated the `oneOf`/`anyOf`/`allOf`/`not` conversion in #348).

**Primary deliverable is tests**, not code. The code change is conditional (below).

### Tier

Not an inference change — this is serialization-correctness hardening of the `#[RawSchema]`
escape hatch. No body parsing.

### Test deliverable (TDD; the guard comes first)

**Create** `tests/Feature/RawSchemaNestedKeywordValidationTest.php` (group `openapi` — **not**
`snapshot`, since the 5.8 CI leg runs `--exclude-group=snapshot`; everything else runs on 5.8).

For each of the five keywords, a fixture Data class with a class-level `#[RawSchema([...])]`
exercising the keyword with a **nested schema** (the divergent shape), routed through a
controller method, then:

- Assert the document validates: mirror the established idiom in
  `tests/Feature/NestedResponseSchemaTest.php` / `InlineJsonResponseTest.php` —
  `app(OpenApiGenerator::class)->generate($registry->default(), app()->environment())` then
  `expect($document->validate())->toBeTrue();`. (`generateSpec()` does **not** validate; calling
  `generate()` + `->validate()` is what guards the real `openapi:generate` command path.)
- Assert the serialized shape survives intact via `generateSpec()` (e.g. the nested schema under
  `additionalProperties`/`contains`/etc. is present, not dropped or mangled).

Keyword fixtures (one Data class each, or a single controller with one route per keyword):

1. `additionalProperties` as a **schema** object (not bool) — e.g. `['type' => 'string']`.
2. `patternProperties` — map of regex → nested schema.
3. `propertyNames` — nested schema (e.g. `['pattern' => '^[a-z]+$']`).
4. `contains` — nested item schema.
5. `discriminator` — `['propertyName' => ..., 'mapping' => [...]]` (note: discriminator is an
   object, not a `[Schema]`; confirm whether it needs conversion or validates as a raw array —
   the test settles it).

Edge cases to cover so the guard is meaningful:
- nested `type: array` under these keywords (forces the items-less-array guard already in
  `apply()` to fire on the child);
- `additionalProperties: false`/`true` (bool form) must **still** pass through unchanged — the
  bool form is valid and must not be wrapped into a schema. Guard this explicitly so a future
  conversion doesn't regress it.

**Create** fixtures under `tests/Fixtures/RawSchema/` (sibling to the existing
`RawSchemaWithFieldAttributeData` etc.) and a controller method per keyword (extend the existing
`RawSchemaController` or add a dedicated one).

### How we confirm the tests actually run on the 5.8 leg

- The new test file carries group `openapi` only, never `snapshot` — the 5.8 job
  (`.github/workflows/tests.yml`, "Run tests (excluding 6.x-pinned serialization snapshots)")
  runs `--testsuite=Unit,Feature,Arch,PhpStan --exclude-group=snapshot`, so a `tests/Feature/`
  file is included. No CI-config change needed.
- The PR's `swagger-php 5.8` check turning green **with these assertions present** is the proof.
  Locally we can additionally pin 5.8 (`composer require zircote/swagger-php:^5.8 --no-update &&
  composer update`) and run the file directly to confirm before pushing; note this in the coder
  handoff. (Revert the lockfile pin — never commit a 5.8 lock.)

### Conditional code change

Run the new tests on 5.8 **first**:

- **If all five validate cleanly on 5.8** → tests stand as a regression guard, **no `src/`
  change**. Document that finding in the PR body and a `CHANGELOG.md [Unreleased]` test/guard
  note. Survey gate is **N/A** (tests-only).
- **If a keyword fails 5.8 `validate()`** → extend `SchemaFromArrayDefinition::apply()` to
  recursively convert that keyword, mirroring the existing `oneOf`/`anyOf`/`allOf`/`not` and
  `properties`/`items` branches:
  - `additionalProperties` (when `is_array` **and** not a bool) → `self::build($value)`; leave
    the bool form on the generic branch.
  - `patternProperties` → map each regex value through `self::build()`, preserving keys.
  - `propertyNames`, `contains` → `self::build($value)`.
  - `discriminator` → only if 5.8 rejects the raw array; convert to `OA\Discriminator` if so.
  This is **area:core, generation-affecting** → the lead runs the **survey gate** before merge,
  and it needs a `CHANGELOG.md [Unreleased]` Fixed entry + a note in `docs/attributes.md` only if
  observable output changes (it shouldn't — same valid output, just 5.8-safe).

The plan brackets both branches; which one ships is decided by the 5.8 run, recorded as a PR
comment per the workflow.

### Files

- **Create** `tests/Feature/RawSchemaNestedKeywordValidationTest.php`.
- **Create** fixtures under `tests/Fixtures/RawSchema/` (+ controller route(s)).
- **Conditional, code branch only:** modify `src/Support/Generator/SchemaFromArrayDefinition.php`;
  `CHANGELOG.md [Unreleased]`.
- **Tests-only branch:** `CHANGELOG.md [Unreleased]` guard note.

### Controversy

- Tests-only branch: non-controversial, no survey.
- Code branch: touches `SchemaFromArrayDefinition` (area:core, generation-affecting) → **survey
  gate required** before merge; still no `Contracts\`/stage-order/public-signature change. Flag to
  the lead so the survey runs.


Radiergummi:

Radiergummi:
🤖 **[lead]** plan is detailed and self-bracketing (tests-first, 5.8-leg-aware, survey-gate-flagged for the conditional code branch). area:core → quick plan-review before coding.

Radiergummi:
🤖 **[reviewer]** plan-review: approved — plan is sound, lowest-tier, and tests-first. Verified the premise and the CI mechanics independently; two minor sharpenings for the coder (neither blocks coding).

**Premise confirmed:**
- `ExplicitClassSchema::ACCEPTED_KEYWORDS` admits all five: `additionalProperties` (line 50), `patternProperties` (51), `propertyNames` (52), `contains` (53), `discriminator` (77).
- `SchemaFromArrayDefinition::apply()` special-cases only `properties`, `items`, `oneOf`/`anyOf`/`allOf`, and `not`. All five plan-named keywords fall through to the generic `$node->{$key} = $value;` raw-array assignment — the divergent shape. Premise holds.

**Framing confirmed:** serialization-correctness hardening of the `#[RawSchema]` escape hatch, not an inference change (no body parsing). Correct tier.

**CI mechanics confirmed:** the 5.8 job (`.github/workflows/tests.yml:212-219`) runs `--testsuite=Unit,Feature,Arch,PhpStan --exclude-group=snapshot`, so a `tests/Feature/` file in group `openapi` (never `snapshot`) runs on the 5.8 leg — no CI-config change needed. The `app(OpenApiGenerator::class)->generate(...)->validate()` idiom (vs. the non-validating `generateSpec()`) is exactly as used in `NestedResponseSchemaTest.php:23-25` / `InlineJsonResponseTest.php:158-161`. The plan correctly relies on `generate()+validate()` as the real guard.

**Two sharpenings (add to the coder handoff, do not block):**
1. **Verify the OA model surface on 5.8, not 6.x.** Locally swagger-php resolves to 6.2.0, where `OA\AdditionalProperties` and `OA\Discriminator` exist as dedicated classes and `contains`/`patternProperties`/`propertyNames` are untyped properties. The conditional code branch (if a keyword fails 5.8) must convert against the **5.8** model surface — confirm those constructors exist/behave the same on 5.8 before mirroring the 6.x conversion. The plan already gates the branch on the 5.8 run; just make the 5.8 model check explicit.
2. **Confirm the items-less-array guard fires on converted children.** When `apply()` recurses via `self::build()` for these keywords, the child runs `apply()` and thus inherits the existing `type:array ⇒ items` guard — good. The plan’s "nested type: array under these keywords" edge case covers it; keep that fixture so a child-schema regression would be caught.

Verdict: approved to proceed to coding. Survey gate correctly scoped to the code branch only (area:core, generation-affecting); tests-only branch is non-controversial.

Radiergummi:
🤖 **[lead]** plan-review approved (premise verified, CI mechanics correct, bracketing sound). → coding. Survey gate applies ONLY if the conditional SchemaFromArrayDefinition code branch ships.

Radiergummi:
🤖 **[coder]** branch decision: **tests-only, no `src/` change.**

Per the plan I ran `tests/Feature/RawSchemaNestedKeywordValidationTest.php` pinned to **swagger-php 5.8.3** first (`composer require zircote/swagger-php:^5.8 --no-update && composer update`; lockfile reverted, not committed). Result: all 7 tests pass on 5.8.3, including the `validate()` assertion.

Findings:
- 5.8.3 models all five keywords on `OA\Schema` (`additionalProperties`, `patternProperties`, `propertyNames`, `contains`, `discriminator`) and ships `OA\AdditionalProperties` + `OA\Discriminator` classes.
- `SchemaFromArrayDefinition::apply()` assigns these five via the **generic `$node->{$key} = $value` branch** (verified: `additionalProperties`/`contains` come out as raw `array`, not converted to `OA\Schema`). On **both** 5.8 and 6.x that raw-array shape serialises and `validate()`s cleanly.
- This is the divergence from #348: 5.8 rejects raw-array `oneOf`/`anyOf`/`allOf`/`not` (hence the conversion there), but it does **not** reject these five. So no `SchemaFromArrayDefinition` change is warranted.

On reviewer1's sharpenings:
- (1) Verified the OA model surface against **5.8 specifically** (probe above), not just local 6.2.0.
- (2) Kept the nested `type: array` fixture (`AdditionalPropertiesSchemaData`, array under `additionalProperties` with explicit `items`). Note: because these keywords take the generic branch, the items-less-array guard in `apply()` does **not** descend into these children (they are raw arrays, not `OA\Schema` nodes) — so the fixture documents that a nested array under these keywords serialises/validates as authored, rather than confirming the guard fires. Flagging this so the framing is honest; happy to add an explicitly items-less variant if you want to pin the current (no-guard-on-raw-children) behaviour.

Survey gate: **N/A** (tests-only, no generation-affecting code). Local gates green (2753 passed); watching the 5.8 CI leg now.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed 712258cc. Ready for review. Tests-only branch (no src change); the swagger-php 5.8 CI leg is green with the new validate() assertions present — the proof the plan required. Survey gate N/A.

Radiergummi:
🤖 **[lead]** tests-only branch (0 src lines), full matrix CI green incl. swagger-php 5.8 at 712258cc. Survey N/A. → code review (two independent reviewers).

Radiergummi:
🤖 **[reviewer]** review round 1 — 2 findings + 1 recommendation (the call the lead asked for).

Tests-only branch verified: no `src/` change (`git diff main...712258cc -- src/` is empty), file is group `openapi` (not `snapshot`) so it runs on the 5.8 leg, all five keywords + the `additionalProperties` bool-form pass-through are covered, fixtures carry the divergent nested shapes, Pint clean, CHANGELOG present. I confirmed the shape-assertion tests have teeth — the serialized component keeps the nested content verbatim (dumped `additionalProperties: {type:array, items:{type:string}}`, the full `discriminator` object alongside inline `oneOf`, etc.), so a drop/mangle would fail them.

**1. (must fix) Inaccurate fixture comment — `AdditionalPropertiesSchemaData`.** The docblock says the nested array type is "forcing the items-less-array guard to fire on the converted child." It does not: these keywords are never converted to `OA\Schema`, so `SchemaFromArrayDefinition::apply()`’s `type:array ⇒ items` guard never touches them — the array passes through verbatim. I verified by routing an items-less `additionalProperties: {type: array}` (no items): it serialized unconverted as `{"type":"array"}` and still `validate()`d true. The comment asserts behavior that does not occur; reword it to say the nested schema is preserved as a raw array (no conversion, guard does not descend).

**2. (should tighten) CHANGELOG framing overstates what `validate()` guards.** swagger-php’s `validate()` does **not** recurse into these raw-array keyword values — it validates the `OA` object graph, and these are opaque arrays on it. Mutation test: a fixture with `additionalProperties.type = "nonsense-type-value"`, `contains` set to a bare string, and `discriminator.propertyName` set to an integer still returns `validate() === true`. So "asserted to pass `validate()` … these keywords serialise correctly" is true but narrow — `validate()` only proves swagger-php does not *reject the document* for the raw-array-under-schema-keyword shape (the real 5.8 divergence). The content guarantee comes entirely from the shape-assertion tests. Scope the CHANGELOG line accordingly (e.g. "the document validates — swagger-php accepts the raw-array shape on both legs — and the serialized shape is pinned by per-keyword assertions").

**3. (recommended — my call on the lead’s question) Add one explicitly items-less variant.** Because the guard provably does not descend into these keywords, an items-less `additionalProperties: {type: array}` fixture pins the *current* no-conversion behavior: it validates today precisely because it is not converted. If the deferred #352 code branch ever adds recursive conversion, that variant flips to failing on both majors (items-less converted array is rejected) — a loud, intentional regression signal at the exact behavioral boundary. It is one cheap fixture + one assertion and it is the honest version of what `AdditionalPropertiesSchemaData`’s comment currently claims. I recommend adding it; not a hard blocker if you instead fix #1’s comment to be accurate.

Nit (non-blocking): a few `Route::post` lines run 124–128 chars, over the 100 soft limit — but long lines are pervasive in `tests/Feature/` (270 across the dir), so this is consistent with house style; wrap only if convenient.

Net: the suite is a genuine, correctly-targeted 5.8 guard; #1 is a correctness-of-documentation fix, #2 keeps the CHANGELOG honest, #3 sharpens the boundary guard. Address #1 + #2 (and ideally #3) and I’ll re-review.

Radiergummi:
## reviewer2 — independent review (PR #375, issue #352)

Reviewed at HEAD `712258cc` in the worktree. Verified the test runs on the right leg, is non-vacuous, and covers the five keywords plus the bool pass-through. Two substantive findings (one a misleading comment, one a coverage gap the issue specifically names) and one minor nit.

### What I verified (all good)
- **Right leg.** The test is `uses()->group('openapi')`, and the 5.8 CI job (`.github/workflows/tests.yml:212-219`) runs `--exclude-group=snapshot` with "Everything else … runs against 5.8". So this genuinely executes on the 5.8 path, not just local 6.x. ✔
- **Non-vacuous validation.** `OpenApiGenerator::generate()` returns swagger-php's own `OA\OpenApi`, so `$document->validate()` is the same native validator `GenerateCommand::validate()` ultimately calls. I confirmed independently that `OA\OpenApi::validate()` returns `false` for a structurally-invalid document, so a regression that mangled the doc would fail the test. Combined with the per-keyword `nestedKeywordComponentFor(...)` assertions (which prove the routes/components are actually present in the generated default spec), the "valid document" test is not vacuous. ✔
- **Coverage.** All five keywords + the `additionalProperties` bool-form pass-through guard are present; fixtures are well-formed and Pint-clean; CHANGELOG note present and accurate to the no-code-change outcome. ✔

### 1. (Substantive) `AdditionalPropertiesSchemaData` docblock is factually wrong

The fixture's docblock says the nested array is *"forcing the items-less-array guard to fire on the converted child."* It does no such thing. `SchemaFromArrayDefinition::apply()` only recurses into `properties`, `items`, `oneOf`/`anyOf`/`allOf`, and `not`; `additionalProperties` falls through to `$node->{$key} = $value` (line 84) and is stored as a **raw PHP array** — never converted to an `OA\Schema`, so the line-87 items-less guard never runs on it. I verified this directly: `build(['type'=>'object','additionalProperties'=>['type'=>'array','items'=>[…]]])` yields `additionalProperties` of type `array`, not an OA node.

This contradicts the coder's own (correct) framing that the guard does not descend into these keywords. The comment should be corrected to state the truth: the nested schema passes through as a raw array and validates as-is, with no conversion or guard involvement. As written it will actively mislead the next reader about what the test exercises.

### 2. (Substantive) The items-less-nested-array shape — the one the issue names as the divergence risk — is untested

Issue #352's stated rationale is the **raw-array-under-a-schema-keyword** case, the same family that motivated extending `SchemaFromArrayDefinition` for `oneOf`/`not`. The sharpest instance of that is a nested array that is itself **items-less** (`{type: array}` with no `items`), since the top-level guard explicitly injects `items` to avoid 5.8/6.x rejection of items-less arrays — but, per finding 1, the guard does **not** reach inside these keywords, so an items-less array nested under `additionalProperties`/`contains` passes through raw with no `items` injected.

None of the fixtures probe that. The `additionalProperties` fixture's inner array already carries `items`, so it sidesteps exactly the risk the issue is about. I'd add a dedicated items-less variant — e.g. `#[RawSchema(['type'=>'object','additionalProperties'=>['type'=>'array']])]` (or `contains`/`patternProperties` with an items-less array child) — and assert `validate()` true on it. That is the fixture that actually settles "do these serialize cleanly on 5.8?" for the divergent shape; without it the guard answers a slightly easier question than the issue asked.

(On local 6.2.0 the items-less nested shape validates true — as expected, since it's an unmodeled raw array — which is precisely why a 5.8-leg assertion is the valuable one to add.)

### 3. (Minor) Soft 100-char limit

Lines 23, 30–35, 39–40, 48, 87 exceed 100 chars (the `Route::post(...)` registrations and the `$ref` accessor). Pint-clean and consistent with existing test style, so non-blocking — flagging only for completeness.

### Verdict
Findings 1 and 2 should be addressed before merge: #1 is a comment that misstates behavior, and #2 is a gap against the issue's own scope (the items-less nested shape is the divergent case the guard exists to pin). #3 is a nit. The test infrastructure itself is sound — right leg, non-vacuous, correctly asserting.

Radiergummi:
🤖 **[coder]** addressed round 1 (pushed a49f685f):

1. (MUST FIX) Corrected the `AdditionalPropertiesSchemaData` docblock — these keywords are **not** converted to `OA\Schema`, so `apply()`'s items-less-array guard does not descend; the nested array is preserved as a raw array verbatim. Thanks for catching the inaccurate claim.

2. (SHOULD TIGHTEN) Scoped the CHANGELOG line: `validate()` proves only that the raw-array-under-a-schema-keyword shape is **accepted** on both legs (it does not recurse into these keywords); the serialised shape is pinned by the per-keyword shape-assertion tests. Reworded accordingly.

3. (RECOMMENDED — done) Added an explicitly items-less variant `AdditionalPropertiesItemsLessData` (`additionalProperties: {type: array}`, no `items`) + route + test asserting the component's `additionalProperties` is exactly `{type: array}`. This pins the current no-conversion behaviour at the exact boundary; if the deferred conversion branch ever ships, the synthesised `items` flips it to failing — the loud signal you described. Verified it passes on **5.8.3** too (re-pinned locally, all 8 tests green, lockfile reverted).

Local gates green (2754 passed); remote CI all 9 pass incl. swagger-php 5.8 + PHP 8.5. Still tests-only, survey N/A.

Radiergummi:

Re-reviewed a49f685f on my own merits (diff 712258cc..a49f685f). All three round-1 points resolved cleanly:

1. `AdditionalPropertiesSchemaData` docblock corrected — now states the keywords are not converted to `OA\Schema`, so the nested array is preserved verbatim and `apply()`’s items-less guard does not descend. Accurate.
2. CHANGELOG scoped honestly — "produce a document that passes `validate()` (the raw-array-under-a-schema-keyword shape is accepted on both majors), with the serialised shape pinned by per-keyword assertions." This matches what `validate()` actually guards (it does not recurse into these raw arrays) and credits the shape tests as the content guarantee.
3. Items-less boundary variant added — `AdditionalPropertiesItemsLessData` (`additionalProperties: {type: array}`, no items) + route + a strict `->toBe(["type" => "array"])` assertion. Pins the current no-conversion behavior; the strict equality flips the moment a future conversion synthesises `items`, surfacing the change at the exact boundary. The fixture/test comments describe this accurately.

Verified at a49f685f: 8/8 tests pass (incl. the new boundary test), Pint clean, still tests-only (`git diff main...a49f685f -- src/` empty), survey N/A. The suite is a genuine, honestly-framed 5.8 regression guard with the no-conversion boundary now explicitly pinned. Approving — one of two required reviewers.

Radiergummi:
## reviewer2 — round 2: approved

Both round-1 findings fully resolved at `a49f685f`, re-verified locally (8 tests / 14 assertions pass, Pint clean on test + fixtures).

1. **(Wrong comment) Resolved.** `AdditionalPropertiesSchemaData`'s docblock now correctly states the nested array is preserved as a raw array and the items-less guard in `apply()` does not descend into it. Matches the actual behaviour I verified.
2. **(Coverage gap — the issue's actual point) Resolved.** New `AdditionalPropertiesItemsLessData` fixture (`additionalProperties: {type: array}`, no `items`) + route + dedicated test. The assertion `toBe(['type' => 'array'])` is precise: it pins both that no `items` was synthesised and that nothing else leaked in, and the comment correctly notes it would flip to failing if recursive conversion ever ships — a clean boundary guard. Because the route is registered in `beforeEach`, the aggregate `validate()` test now also asserts the document containing the items-less nested array passes on every supported leg, which is exactly the 5.8-leg question issue #352 set out to settle.

CHANGELOG is now scoped honestly (document validates / raw-array shape accepted on both legs; serialised shape pinned by per-keyword assertions). My earlier line-length nit was non-blocking and I'm not re-raising it.

No remaining concerns. Approving.

Radiergummi:
🤖 **[lead]** tests-only + all gates green → squash-merged. All 5 nested RawSchema keywords validate cleanly on swagger-php 5.8.3 (no SchemaFromArrayDefinition change warranted); the suite stands as a regression guard incl. the items-less boundary fixture. Survey N/A (0 src lines). Full matrix CI green at a49f685f; two independent reviewer approvals on the final commit. Closes #352. Worktree removed.

Radiergummi:
🤖 **[coder]** addressed round 1 (reviewer2), pushed 1d43a188:

Findings 1 and 3 were already in a49f685f (the docblock correction and the items-less variant fixture). This push tightens the boundary test per your finding 2: the items-less variant test now carries its **own** explicit `$document->validate()` true assertion (not just the whole-document test), so the divergent shape #352 names — an items-less array under a schema keyword — has a dedicated non-vacuous validation signal. Verified on **5.8.3** (8/8 green, 15 assertions, lockfile reverted).

To recap all three:
1. (comment) `AdditionalPropertiesSchemaData` docblock corrected — these keywords pass through as raw arrays, no conversion, the items-less-array guard does not descend.
2. (coverage gap) `AdditionalPropertiesItemsLessData` = `#[RawSchema([type:object, additionalProperties:[type:array]])]` (no items), routed, now asserting both `validate()` true and the verbatim `{type: array}` shape. Pins the no-conversion behaviour at the exact boundary; flips to failing if recursive conversion ever ships.
3. (nit) Long route/`$ref` lines left as-is per house style (both reviewers non-blocking).

Local gates green (2754 passed); remote CI all 9 pass incl. swagger-php 5.8 + PHP 8.5. Still tests-only, survey N/A.

PR #376, opened 2026-06-16
feat(#46): regression guard — openapi:why explains globally-filtered routes

### LINKED ISSUE #46: `openapi:why` cannot explain globally-filtered routes
## Summary

`openapi:why` should be able to explain that a route is missing from a spec because a global
`RouteFilter` dropped it. The pipeline that produces this explanation is already in place — what is
missing is regression-test coverage that locks the behaviour in. This issue adds a test (and
confirms / corrects any wiring gap it surfaces) so the global-filter explanation path cannot silently
regress.

## Motivation

`openapi:why` exists to answer "why is this route missing from my spec?". The stage that is hardest
to reason about by hand is global route filtering (`config('openapi.filters')`), so it is exactly the
case the command must explain well.

The original framing of this issue assumed `WhyCommand` could never reach a filtered route because
`RouteIntrospector::discover()` had already dropped it. That is no longer true on the current
codebase, and it is worth stating precisely so an implementer does not re-architect working code:

- `RouteIntrospector::discover()` (`src/Support/Routing/RouteIntrospector.php:53`) yields **every**
  route unconditionally — its class docblock states route exclusion lives entirely in
  `InclusionEvaluator`, by design, so that every exclusion is traceable.
- `WhyCommand::findCandidates()` (`src/Console/WhyCommand.php:116`) therefore already finds routes
  that a global filter would remove.
- `InclusionEvaluator::decide()` (`src/Support/Inclusion/InclusionEvaluator.php`) already iterates
  the configured `RouteFilter`s, emits a `TraceEntry` with `stage: 'global-filter'`, and on a match
  returns the reason `"excluded by global filter {$name}"`.
- `WhyCommand::printSpecDecision()` already prints that trace.

So the explanation is wired end to end, but no test registers a `RouteFilter` and asserts the
`global-filter` stage appears in the output. That is the real, narrow gap: a missing regression guard
for behaviour the command depends on. This is a follow-up surfaced during the multi-spec review — not
blocking, but it protects the command's core purpose.

## Proposed approach

Add a test that registers a global `RouteFilter` which drops a known route, runs `openapi:why`
against that route, and asserts the output attributes the exclusion to the global-filter stage
(e.g. the line contains `global-filter` and the reason names the filter class).

While writing the test, confirm the live wiring matches the description above:

- `RouteIntrospector::discover()` stays unfiltered (do not let filtering "creep" back into it).
- The filter under test is registered via `config('openapi.filters')` (`config/openapi.php:474`),
  the same channel `InclusionEvaluator` reads, so `RouteFilter::shouldSkip()`
  (`src/Contracts/Routing/RouteFilter.php`, `shouldSkip(Route $route): bool`) is exercised through
  the real path.

If — and only if — the test reveals the global-filter stage is not actually surfaced (e.g. the trace
entry is dropped before reaching `printSpecDecision()`), fix that surfacing. The per-spec inclusion
trace via `InclusionDecision` / `InclusionEvaluator` otherwise stays as-is.

## Acceptance criteria

- A test registers a global `RouteFilter` (via `config('openapi.filters')`) that drops a known route,
  runs `openapi:why <that route>`, and asserts the output reports the global-filter stage (and the
  filter class name) as the reason — rather than reporting the route as not found or omitting the
  filter stage.
- The test lives in `tests/Unit/Console/WhyCommandTest.php`, following the existing console-command
  test conventions there.
- `RouteIntrospector::discover()` remains unfiltered (no behaviour change that would make a filtered
  route invisible to `findCandidates()`).

## References

Parent: #69 (shared multi-spec context: `SpecRegistry`, route mounting, and the
unfiltered-discovery / `InclusionEvaluator` split live there).

Affected areas: `src/Console/WhyCommand.php`, `src/Support/Inclusion/InclusionEvaluator.php`,
`tests/Unit/Console/WhyCommandTest.php`; touch points `src/Support/Routing/RouteIntrospector.php`,
`src/Contracts/Routing/RouteFilter.php`, `config/openapi.php` (`filters`).

Verify: `vendor/bin/pest tests/Unit/Console/WhyCommandTest.php`

Radiergummi:
Closes #46

## Implementation plan

Add a regression-test guard so `openapi:why` can never silently stop explaining
**globally-filtered** routes. The explanation is already wired end to end; the gap is a missing
test.

**Wiring confirmed against the named files** (premise holds — no re-architecture):
- `RouteIntrospector::discover()` (`src/Support/Routing/RouteIntrospector.php:47`) yields **every**
  route unconditionally — no filtering. ✓
- `InclusionEvaluator::decide()` (`src/Support/Inclusion/InclusionEvaluator.php:78-96`) iterates the
  configured filters, emits a `TraceEntry{ stage: 'global-filter', name: <class>, passed: false }`,
  and on a match returns summary `"excluded by global filter {$name}"` (`SkipReason::GlobalFilter`). ✓
- `WhyCommand::printSpecDecision()` (`src/Console/WhyCommand.php:140-150`) prints
  `{mark} {stage} {name} — {reason}` per trace entry and `→ {summary}`. ✓
- Filters are bound from `config('openapi.filters')` in the `InclusionEvaluator` factory
  (`src/OpenApiServiceProvider.php:187`). ✓

So the output for a filtered route already contains `global-filter`, the filter class name, and
`excluded by global filter <Class>`. The test locks that in.

### Tier

Not an inference change — a CLI regression guard. No body parsing.

### Test deliverable (primary)

**Modify** `tests/Unit/Console/WhyCommandTest.php` — one new test, following the file's existing
conventions (register routes in/before the test, set config, then `app()->forgetScopedInstances()`
so the scoped `InclusionEvaluator` rebinds with the new config):

1. Define a test-local `RouteFilter` (a small named fixture class or an `implements RouteFilter`
   class in the test file) whose `shouldSkip(Route $route): bool` returns true for the known route
   (e.g. matches the `api/v1/flights` URI registered in `beforeEach`).
2. `config(['openapi.filters' => [TheFilter::class]])` — the real channel `InclusionEvaluator`
   reads (`config/openapi.php:474`), then `app()->forgetScopedInstances()`.
3. Run `artisan('openapi:why api/v1/flights')` and assert the output:
   - contains `global-filter` (the stage),
   - contains the filter class name (`expectsOutputToContain` the `::class` basename/FQCN — match
     what `printSpecDecision` prints, which is `$entry->name` = the FQCN),
   - contains the reason/summary (`excluded by global filter` — names the filter),
   - and `assertSuccessful()` (the route IS found; it's explained as filtered, **not** reported as
     not-found).

Edge cases / branches to keep the guard meaningful:
- **Found, not missing**: assert the command exits successfully and prints the `Route:` block — i.e.
  the filtered route reaches `findCandidates()` (proves `discover()` stays unfiltered). This is the
  core anti-regression assertion (if filtering crept back into `discover()`, the route would be
  reported not-found and the command would fail).
- Optionally assert the non-filtered spec/decision path still prints alongside (the existing
  "explains why a route is included" test already covers the passing-filter case, so no duplication
  needed).

A test-local filter class is the minimal fixture; the one-method `RouteFilter` contract
(`src/Contracts/Routing/RouteFilter.php`) needs no new production fixture.

### Conditional code change

- **If the test passes as-is** (expected, per the confirmed wiring) → **tests-only**, no `src/`
  change. Survey **N/A** (area:cli, no generation impact). Non-controversial.
- **If the test reveals the `global-filter` stage is not surfaced** (e.g. the trace entry is dropped
  before `printSpecDecision()`) → fix only that surfacing in `WhyCommand` /
  `InclusionEvaluator`. This would **not** touch `Contracts\` (`RouteFilter` is the existing,
  unchanged contract) and must **not** reintroduce filtering into `RouteIntrospector::discover()`.
  Flag to the lead, but it stays non-`Contracts`.

### Files

- **Modify** `tests/Unit/Console/WhyCommandTest.php` (+ a test-local `RouteFilter` fixture).
- **Conditional code branch only:** `src/Console/WhyCommand.php` and/or
  `src/Support/Inclusion/InclusionEvaluator.php`; `CHANGELOG.md [Unreleased]`.
- No `docs/` change for the tests-only branch (no observable behaviour change — guarding existing,
  documented behaviour).

### Controversy

Tests-only: none, survey N/A, fast-path eligible (area:cli, additive test). The code branch (only
if the guard fails) stays off `Contracts\`/stage-order/public signatures; flag to the lead if it
materialises.

### Verify

`vendor/bin/pest tests/Unit/Console/WhyCommandTest.php`

### DIFF
--- a/tests/Unit/Console/WhyCommandTest.php
+++ b/tests/Unit/Console/WhyCommandTest.php
@@ -4,9 +4,22 @@ declare(strict_types=1);
 
 namespace Radiergummi\OpenApi\Tests\Unit\Console;
 
+use Illuminate\Routing\Route as RoutingRoute;
 use Illuminate\Support\Facades\Route;
+use Radiergummi\OpenApi\Contracts\Routing\RouteFilter;
 
 uses()->group('openapi');
+
+/**
+ * Test-local global filter that skips the `api/v1/flights` fixture route.
+ */
+final class FlightsRouteFilter implements RouteFilter
+{
+    public function shouldSkip(RoutingRoute $route): bool
+    {
+        return $route->uri() === 'api/v1/flights';
+    }
+}
 
 beforeEach(function (): void {
     Route::get('api/v1/flights', fn() => 'x')->name('v1.flights.index');
@@ -23,6 +36,22 @@ it('explains why a route is included in each spec', function (): void {
         ->expectsOutputToContain('default:')
         ->expectsOutputToContain('v1:')
         ->expectsOutputToContain('Result:')
+        ->assertSuccessful();
+});
+
+it('explains a globally-filtered route as filtered, not as not-found', function (): void {
+    config(['openapi.filters' => [FlightsRouteFilter::class]]);
+    app()->forgetScopedInstances();
+
+    // The route is discovered (discover() stays unfiltered) and explained as excluded by the
+    // global filter, never reported as not-found. Chained expectsOutputToContain matches the
+    // buffer in document order, so the summary line (which already names the filter class) is
+    // asserted last; a separate class-name assertion would consume the trace-line occurrence first.
+    $this
+        ->artisan('openapi:why api/v1/flights')
+        ->expectsOutputToContain('Route:')
+        ->expectsOutputToContain('global-filter')
+        ->expectsOutputToContain('excluded by global filter ' . FlightsRouteFilter::class)
         ->assertSuccessful();
 });
 
--- a/tests/Unit/Console/WhyCommandTest.php
+++ b/tests/Unit/Console/WhyCommandTest.php
@@ -44,9 +44,8 @@ it('explains a globally-filtered route as filtered, not as not-found', function 
     app()->forgetScopedInstances();
 
     // The route is discovered (discover() stays unfiltered) and explained as excluded by the
-    // global filter, never reported as not-found. Chained expectsOutputToContain matches the
-    // buffer in document order, so the summary line (which already names the filter class) is
-    // asserted last; a separate class-name assertion would consume the trace-line occurrence first.
+    // global filter, never reported as not-found. The summary assertion already names the filter
+    // class, so a separate bare class-name assertion would be redundant.
     $this
         ->artisan('openapi:why api/v1/flights')
         ->expectsOutputToContain('Route:')
--- a/tests/Unit/Console/WhyCommandTest.php
+++ b/tests/Unit/Console/WhyCommandTest.php
@@ -44,12 +44,12 @@ it('explains a globally-filtered route as filtered, not as not-found', function 
     app()->forgetScopedInstances();
 
     // The route is discovered (discover() stays unfiltered) and explained as excluded by the
-    // global filter, never reported as not-found. The summary assertion already names the filter
-    // class, so a separate bare class-name assertion would be redundant.
+    // global filter, never reported as not-found. Both fragments include the filter class so the
+    // stage label and the summary wording are each pinned, not loosely matched.
     $this
         ->artisan('openapi:why api/v1/flights')
         ->expectsOutputToContain('Route:')
-        ->expectsOutputToContain('global-filter')
+        ->expectsOutputToContain('global-filter ' . FlightsRouteFilter::class)
         ->expectsOutputToContain('excluded by global filter ' . FlightsRouteFilter::class)
         ->assertSuccessful();
 });


Radiergummi:
🤖 **[planner]** plan posted — see PR body. Wiring claims confirmed against the named files (discover() unfiltered, InclusionEvaluator emits the global-filter TraceEntry, printSpecDecision prints it). Ready for plan-review.

Radiergummi:
🤖 **[lead]** plan verified the wiring premise against named files; minimal test-guard. Fast-path (area:cli): skipping plan-review → coding. Survey N/A if tests-only. Auto-merge still needs two independent reviewer approvals + green matrix CI.

Radiergummi:
🤖 **[coder]** branch decision: **tests-only, no `src/` change.**

The `global-filter` explanation is already surfaced end to end exactly as the plan confirmed. The test passes against the existing wiring — I verified the live output:
```
Route: GET api/v1/flights
  ...
default:
    ✗ global-filter <FQCN> — shouldSkip = true
    → excluded by global filter <FQCN>

Result: excluded from all specs
```
The route is found and explained as filtered, never reported as not-found — proving `RouteIntrospector::discover()` stays unfiltered (the core anti-regression assertion). Survey **N/A** (area:cli, no generation impact); no CHANGELOG (guarding existing, documented behaviour).

One authoring note worth recording: Laravel's chained `expectsOutputToContain` matches the buffer in **document order and consumes** as it goes. My first draft asserted the filter `::class` on its own *before* the summary line; that consumed the FQCN on the earlier `✗ global-filter <FQCN>` trace line, so the later `excluded by global filter <FQCN>` search (which starts with text preceding the consumed position) failed. Fix: drop the redundant standalone class-name assertion and assert in document order (`Route:` → `global-filter` → `excluded by global filter <FQCN>`); the summary assertion already names the class. Comment in the test explains this so it isn't "simplified" back into a regression.

Local gates green (2755 passed, Pint + PHPStan clean); watching CI.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed f8ac0200. Ready for review. Tests-only branch (no src change), survey N/A; remote CI all 9 pass incl. swagger-php 5.8 + PHP 8.5.

Radiergummi:
🤖 **[lead]** tests-only (1 file, 0 src lines), full matrix CI green at f8ac0200. Survey N/A. → code review (two independent reviewers).

Radiergummi:
🤖 **[reviewer]** review round 1 — 1 finding.

The guard is genuine and well-targeted. I verified it adversarially:

- **Non-vacuous against the real regression.** I mutation-tested a `continue` for `api/v1/flights` inside `RouteIntrospector::discover()` (simulating filtering creeping back into discovery): the new test FAILS — the route becomes not-found, the command returns exit 1, and `assertSuccessful()` trips. So it genuinely pins the "`discover()` stays unfiltered; filtering is deferred to `InclusionEvaluator`" invariant that #46 fixed (and that `discover()`’s own docblock states).
- **Assertions map to real output.** `InclusionEvaluator::decide()` emits a `global-filter` `TraceEntry` (rendered `✗ global-filter <Class> — shouldSkip = true`) and returns summary `excluded by global filter <Class>` (rendered `→ …`); `printHeader` always prints `Route:` for a single candidate. All three substrings are real and the `<Class>` interpolation matches `FlightsRouteFilter::class`.
- **Scoped-instance handling is correct.** `config([openapi.filters => …])` + `forgetScopedInstances()` is required because `InclusionEvaluator` resolves `$globalFilters` from config at construction and is a scoped singleton; without the forget, the empty-filter evaluator from `beforeEach` would be reused. Matches the sibling tests in this file. Not green for the wrong reason.
- tests-only (`git diff main...f8ac0200 -- src/` empty), Pint clean, PHPStan clean (the test-local `FlightsRouteFilter` correctly satisfies the `RouteFilter` contract), no over-100 lines. No CHANGELOG/docs needed for a pure guard of already-shipped behavior — agreed.

**1. (should fix) The explanatory comment states a Laravel behavior that is not true — a future foot-gun.** The comment says: "Chained `expectsOutputToContain` matches the buffer in document order, so the summary line … is asserted last; a separate class-name assertion would consume the trace-line occurrence first." But `PendingCommand::expectsOutputToContain` registers each substring as a non-`->ordered()` Mockery `doWrite` matcher (contrast `expectsOutput`, which *is* `->ordered()`); the final check (`PendingCommand.php:530`) only fails if some expected substring went unmatched after all output — order between the substrings is irrelevant. I confirmed empirically: reversing the three assertions (summary first) still passes. So there is no document-order requirement and no destructive "consume the trace-line occurrence first" hazard between these distinct substrings. Per the comment conventions (comment the *why*, don’t mis-state mechanics), please correct or trim it — as written a future reader could rely on an ordering guarantee that does not exist. The test itself is correct as-is; this is purely the comment.

Net: the guard is sound, non-vacuous, and conventionally clean; the only issue is the inaccurate comment. Fix #1 and I’ll re-review.

Radiergummi:
## reviewer2 — independent review (PR #376, issue #46): approved

Reviewed at HEAD `f8ac0200`. Tests-only, single new test (+ test-local `FlightsRouteFilter`). I went past reading and **mutation-tested non-vacuity**; the guard is real. Approving with one minor, non-blocking observation.

### Non-vacuity — verified by mutation
- **The core guarantee (route is discovered, explained as filtered, not reported not-found).** I patched `RouteIntrospector::discover()` to `continue` past `api/v1/flights` (the exact regression issue #46 warns against). The test **fails** — the command can't find the route, exits non-zero, and `assertSuccessful()` + the trace assertions break. So `discover()` re-introducing filtering is genuinely caught. ✔
- **The reason/summary string.** I changed the evaluator summary from `excluded by global filter {name}` to `dropped by global filter {name}`. The test **fails** at the 3rd assertion. So `->expectsOutputToContain('excluded by global filter ' . FlightsRouteFilter::class)` is the load-bearing, non-vacuous guard — it pins both the exact reason wording and the filter class name. ✔

### Real path exercised (not a stub)
`InclusionEvaluator` is a **scoped** binding whose factory reads `config('openapi.filters')` at resolution time (`OpenApiServiceProvider.php:187`) and constructs real filter instances. So `config([...])` + `forgetScopedInstances()` forces the scoped evaluator to be rebuilt with the new filter — the test drives the genuine `InclusionEvaluator::decide()` global-filter branch, not a stub. The rebind is also *necessary* (the `beforeEach` already resolves scoped instances). ✔

### Buffer-ordering comment — accurate
The filter class name appears twice in output: the trace line (`✗ global-filter <Class> — shouldSkip = true`) and the summary line (`→ excluded by global filter <Class>`). `expectsOutputToContain` consumes the buffer in document order, so asserting `global-filter` (matches the earlier trace line) before `excluded by global filter <Class>` (matches the later summary line) is correct, and folding the class name into the summary assertion rather than a bare separate class-name assertion avoids consuming the trace-line occurrence first. The comment states this precisely. ✔

### Minor (non-blocking)
`->expectsOutputToContain('global-filter')` is a loose substring: I renamed the trace `stage` to `global-filterX` and the test still **passed** (since `global-filterX` contains `global-filter`). So that middle assertion alone wouldn't catch a stage-label drift. It isn't load-bearing — the summary assertion (#3) and the discovery guarantee carry the weight — but if you want the stage label itself pinned tightly you could assert the fuller trace fragment (e.g. `'global-filter ' . FlightsRouteFilter::class`, with the trailing space and class). Optional.

### Conventions / scope
Strict types, test-local filter is properly namespaced and `final`, no abbreviations, Pint clean, PHPStan level 8 clean, full file green (5/5). Tests-only — no `src/` change, matching the issue's finding that the path is already wired. No CHANGELOG/docs is the right call: this guards existing behaviour, it doesn't change any. ✔

No blocking concerns. Approving.

Radiergummi:
🤖 **[coder]** addressed round 1, pushed b59e2e62:

You're right, and I verified it empirically before changing: reversing the three `expectsOutputToContain` assertions (summary first) still passes, so there is **no document-order guarantee** between distinct substrings — `PendingCommand::expectsOutputToContain` registers each as a non-ordered matcher. My comment's "matches in document order / would consume the trace-line occurrence first" claim was wrong. (The actual reason the original 4-assertion draft failed: the bare class-name substring only ever appears embedded in lines already claimed by other matchers, so it has no distinct write to match — failing regardless of position, which I also confirmed.)

Corrected the comment to the accurate, minimal reason: the summary assertion already names the filter class, so a separate bare class-name assertion would be redundant. Test logic unchanged (it was correct as-is).

Local gates green (2755 passed); remote CI all 9 pass incl. swagger-php 5.8 + PHP 8.5.

Radiergummi:

Re-reviewed b59e2e62 (diff f8ac0200..b59e2e62) on my own merits. The round-1 finding is resolved: the comment now states the accurate why (route is discovered and explained as filtered, never not-found) plus a true minimal rationale (the summary assertion already names the filter class, so a bare class-name assertion would be redundant). The false "matches the buffer in document order / consume the trace-line occurrence first" claim is gone, so no reader will rely on a non-existent ordering guarantee. Test logic unchanged.

Verified at b59e2e62: 5/5 tests pass, Pint clean, still tests-only (`git diff main...b59e2e62 -- src/` empty). The guard remains non-vacuous (mutation-confirmed in round 1: filtering inside `discover()` makes the route not-found and trips `assertSuccessful()`). Approving — one of two required reviewers.

Radiergummi:
🤖 **[coder]** addressed round 1 (both reviewers), final push f88886db:

- **reviewer1 (comment accuracy):** corrected the test comment — verified empirically there is no document-order guarantee for chained `expectsOutputToContain` (reversing still passes); reworded to the accurate reason.
- **reviewer2 (optional tighten — folded in):** the stage-label assertion is now the fuller `global-filter ` . FlightsRouteFilter::class fragment, so the `global-filter` stage label is load-bearing (your `global-filterX` mutation would now fail it). Both FQCN-bearing fragments (`global-filter <Class>` and `excluded by global filter <Class>`) match their own distinct writes; 5/5 green.

Local gates green (2755 passed, Pint + PHPStan clean); remote CI all 9 pass incl. swagger-php 5.8 + PHP 8.5. Still tests-only, survey N/A, no CHANGELOG.

Radiergummi:
🤖 **[reviewer]** re-confirm: approved against final commit f88886db.

Quick delta re-check b59e2e62..f88886db — exactly reviewer2's optional tightening, nothing else: the loose `expectsOutputToContain('global-filter')` is now `'global-filter ' . FlightsRouteFilter::class`, and the comment is updated to the accurate "both fragments include the filter class so the stage label and summary wording are each pinned" (no ordering claim). I confirmed the fragment matches the real trace line (`✗ global-filter <Class> — shouldSkip = true`), so the stage label is now load-bearing — a `global-filterX`-style corruption breaks the match. Verified at f88886db: the test passes, Pint clean, still tests-only (`git diff main...f88886db -- src/` empty). Strict improvement over the loose match. My approval carries to f88886db.

Radiergummi:
🤖 **[lead]** tests-only + all gates green → squash-merged. Regression guard locks openapi:why's global-filter explanation (route discovered + explained as filtered, not reported not-found — pins discover() stays unfiltered); both reviewers mutation-tested non-vacuity. Survey N/A (0 src lines). Full matrix CI green at f88886db; two independent reviewer approvals on the final commit. Closes #46. Worktree removed.

PR #377, opened 2026-06-16
feat(#40): #[ResponseExampleFile] attribute — attach a JSON file as a response example

### LINKED ISSUE #40: `#[ResponseFile]` attribute
## Summary

Add a `#[ResponseFile]` authoring attribute that points an endpoint at a static JSON file on disk and attaches its decoded contents as the response example, without the caller having to spell out an example name.

## Motivation

Plenty of real apps keep curated, hand-maintained JSON response samples in files rather than synthesising them — the Monica / legacy-API shape, where a directory of fixture-style payloads is the source of truth for "what a response looks like". For those endpoints the generator's Faker-synthesised example is strictly worse than the file the team already maintains, and there is currently no terse way to say "use this file". This is the attribute-tier escape hatch for that case: the file path is genuine metadata that does not exist anywhere in the code, so convention cannot derive it.

Note that the bundled `#[ResponseExample]` attribute (`src/Attributes/ResponseExample.php`) already accepts a `file:` argument, and `ExampleFileLoader` (`src/Support/Generator/ExampleFileLoader.php`) already resolves `file:`-based payloads at generation time. So the underlying loading mechanism exists — `#[ResponseFile]` is the convenience surface over it for the common "one file, one response" case, where forcing the caller to also invent a `name:` and a `status:` is friction.

## Proposed approach

Attribute tier. Add `#[ResponseFile]` under `src/Attributes/` alongside the existing example attributes (`ResponseExample`, `BaseExample`). It carries the relative file path plus an optional `status` (defaulting to the auto-derived primary response) so it can target a specific response when an endpoint declares several. During operation building, resolve the path through the existing `ExampleFileLoader::load()` in `src/Support/Generator/ExampleFileLoader.php` (the public loader method) — do not add a second file-reading code path — and attach the decoded contents as the response example for the matching status, mirroring how `#[ResponseExample]` with a `file:` argument is already wired. Reuse `ResponseExample`'s "skipped silently when the matching response has no content schema" behaviour. The example-attachment lives in `src/Plugins/Core` (Core Plugin example loading), consistent with the issue's affected areas (`src/Attributes`, `src/Plugins/Core`).

## Acceptance criteria

- `#[ResponseFile('path/to/example.json')]` on a controller action attaches that file's JSON-decoded contents as the example on the operation's primary response.
- A `status:` argument targets a specific declared response; absent it, the primary response is used.
- A missing or invalid JSON file surfaces a clear error (same failure mode as `ExampleFileLoader::load()`), not a silent empty example.
- Add a unit test under `tests/Unit/` (or a feature case under `tests/Feature/`) that places a fixture JSON file, applies `#[ResponseFile]`, generates the spec via the `generateSpec()` helper (`tests/Pest.php`), and asserts the response example equals the file contents.

## References

Parent: #67 (EPIC: Shared components, metadata & examples — the shared example/metadata infrastructure `ExampleFileLoader`, `FakerExampleSynthesiser`, and the `src/Attributes/` authoring surface live there). Related to the existing `#[ResponseExample]` `file:` argument. Milestone: v1.0.

Verify: `vendor/bin/pest --filter ResponseFile`

Radiergummi:
## Implementation plan

Add a `#[ResponseFile('path/to/example.json')]` authoring attribute — the terse "one file, one response" convenience over the existing `file:`-based example loading. It attaches the file's JSON-decoded contents as the example on a response, defaulting to the auto-derived primary response, with an optional `status:` to target a specific declared response.

### Tier

Attribute tier (escape hatch). The file path is genuine metadata absent from the code; convention cannot derive it. No inference, no body parsing.

### Seam correction (flag for plan-review)

The assignment said "place the attachment logic in `src/Plugins/Core`." The **actual existing seam is `OperationBuilder::applyResponseExamples()` in `src/Support/Generator/OperationBuilder.php`** — that is where `#[ResponseExample]` is consumed and where `ExampleFileLoader::load()` is already wired (`OperationBuilder` holds the injected `ExampleFileLoader $fileLoader`; `collectExamples()` calls `$this->fileLoader->load($instance->file)` at line ~471). `src/Plugins/Core` contains **no**
response-example attachment code. Placing `#[ResponseFile]` handling in Core would mean a second, parallel file-reading path — exactly what the issue forbids. So the plan attaches it in `OperationBuilder` alongside `#[ResponseExample]`, reusing `$this->fileLoader->load()`. (No new constructor dependency — the loader is already there.)

### Attachment shape

`#[ResponseExample]` attaches a **named** example (`OA\Examples` on `$media->examples`), because it carries a `name:`. `#[ResponseFile]` has no `name:` — it is the "one file" surface — so it sets the media-type **singular** `example` (`$media->example = <decoded>`), the natural representation of a single anonymous example. (Singular `example` on schemas/media is the established precedent across the pipeline, e.g. `$property->example = …`.) This also avoids forcing the caller to invent a name. The decoded contents come straight from `ExampleFileLoader::load()` (associative array / scalar).

### Files

- **Create** `src/Attributes/ResponseFile.php` — a small `#[Attribute(TARGET_METHOD | TARGET_FUNCTION | IS_REPEATABLE)]` readonly class. **Not** a `BaseExample` subclass (no   `name`/`value`). Constructor: `string $file` (the relative path, the one required arg) and `?int $status = null` (null → primary response). Docblock mirrors `#[ResponseExample]`'s "skipped silently when the matching response has no content schema" note. No I/O in the constructor (consistent with `BaseExample`).
- **Modify** `src/Support/Generator/OperationBuilder.php`:
  - Add a private `applyResponseFiles(ActionDescriptor $descriptor, array $responses)` mirroring `applyResponseExamples()`: read `$descriptor->actionAttributes(ResponseFile::class)`, bucket by target status (instance `status` when set, else the primary response's status — resolve the primary status the same way `applyResponseExamples` matches `(string) $response->response`, using the first 2xx / primary `$responses[0]`), then for each matching response scaffold a JSON media type when content is empty (reuse the existing `MediaType::Json->schema()` scaffolding branch) and set `$media->example = $this->fileLoader->load($instance->file)`.
  - Call it from `build()` right after `$this->applyResponseExamples($action, $responses);` (line ~215).
  - Errors (missing file / invalid JSON) propagate from `ExampleFileLoader::load()` as `RuntimeException` — no catch, so a bad file surfaces clearly (acceptance criterion), unlike the `InvalidArgumentException` malformed-attribute skip used for `#[ResponseExample]` (that guards a constructor contract `#[ResponseFile]` doesn't have).
- **Modify** `docs/attributes.md` — add a table row next to `ResponseExample`:  
  ```markdown
  | `ResponseFile` | method | yes | Attach a JSON file's contents as the example for a response (primary by default; `status:` targets a specific one). |
  ```
- **Modify** `CHANGELOG.md [Unreleased]` — Added entry.

Consider whether the PhpStan `ExampleValueOrFileRule` (`src/PhpStan/Rules/`) needs awareness of `#[ResponseFile]`: it validates the `value`-xor-`file` contract of `BaseExample` subclasses. `#[ResponseFile]` has only `file`, so it is **out of that rule's scope** — no change. (Confirm during coding; if the rule iterates all example-ish attributes, leave it untouched.)

### Tests (TDD; failing test first)

**Create** `tests/Feature/ResponseFileAttributeTest.php` (group `openapi`; `--filter ResponseFile` per the issue's Verify line). **Create** a fixture controller + a fixture JSON file under `tests/Fixtures/` (path resolved via `base_path()` — confirm the Testbench base path so the relative path the attribute carries resolves; mirror how existing `file:` example tests place their JSON fixture). Cases:

1. **Primary-response default** — `#[ResponseFile('…/example.json')]` (no status) → the primary 2xx response's media example equals the file's decoded contents. (`generateSpec()` then assert `…['responses']['200']['content']['application/json']['example']` equals the JSON.)
2. **Explicit `status:`** — `#[ResponseFile('…', status: 422)]` on an action that declares a 422 (via `#[Response(422, …)]`) → example lands on the 422 response, not the 200.
3. **Missing file → clear error** — points at a nonexistent path → generation throws `RuntimeException` ("not found or unreadable"), not a silent empty example.
4. **Invalid JSON → clear error** — fixture with malformed JSON → `RuntimeException` ("invalid JSON").
5. **No-content-schema silent skip** — `#[ResponseFile]` targeting a status whose response has no content and cannot be scaffolded (e.g. a 204): no example attached, generation succeeds, no throw. (Match `#[ResponseExample]`'s silent-skip semantics; the issue requires reusing it.)
6. **status with no matching response** — `status:` names a response the action doesn't declare → dropped silently (mirrors `applyResponseExamples`).

### Existing `file:` example test to mirror

Find the current `#[ResponseExample(file:)]` / `#[Example(file:)]` test for the fixture-placement and `base_path()` idiom (search `tests/` for `file:` example fixtures) so the new test reuses the same fixture-rooting approach rather than inventing one.

### Controversy / scope (for the lead)

- New attribute = escape-hatch surface; **not** a `Contracts\` change, **no** config key, **no** public-signature change. Auto-mergeable unless the `src/` production diff exceeds ~150 lines — the design (one small attribute + one `applyResponseFiles` method mirroring an existing one) stays well under that. Kept tight deliberately.
- **Generation-affecting (area:core)** → survey gate before merge regardless (lead runs it).
- **Flag:** the seam is `OperationBuilder` (`src/Support/Generator/`), not `src/Plugins/Core` as the assignment stated — confirm this placement at plan-review.

### Verify

`vendor/bin/pest --filter ResponseFile`

Closes #40.


Radiergummi:

One seam correction to flag: the example-attachment seam is `OperationBuilder::applyResponseExamples()` in `src/Support/Generator/` (where `#[ResponseExample]` is consumed and `ExampleFileLoader::load()` is already injected), **not** `src/Plugins/Core` — Core has no response-example attachment code, so placing it there would create the second file-reading path the issue forbids. Plan attaches `#[ResponseFile]` in `OperationBuilder`, reusing the existing loader.

Radiergummi:
🤖 **[lead]** plan flags a seam correction (attach in OperationBuilder::applyResponseFiles next to applyResponseExamples — the real seam where ExampleFileLoader is wired — NOT src/Plugins/Core, which has no such code; Core placement would create the forbidden second file-reading path). area:core → plan-review to validate the seam + the singular-$media->example shape.

Radiergummi:
🤖 **[reviewer]** plan-review: approved on the two flagged decisions (both correct), with **two required sharpenings** to the plan body before coding — they are correctness gaps, not style.

## The two flagged decisions — confirmed

**1. Seam correction — correct, and more correct than the issue.** Verified both claims:
- `OperationBuilder` (`src/Support/Generator/`) is genuinely where `#[ResponseExample]` + `ExampleFileLoader` are wired: `private ExampleFileLoader $fileLoader` injected (line 76), `applyResponseExamples()` reads the attribute (693), `collectExamples()` calls `$this->fileLoader->load($instance->file)` (471), called from `build()` at 215.
- `src/Plugins/Core` has **no** response-example/file-attachment seam — its only `->example` writes are `$property->example = $synthesised` (Faker request-body schema examples; a different concern). No `ExampleFileLoader`, no `ResponseExample`.
- Per `CLAUDE.md`, authoring attributes are **general-package surface**, never gated behind Core ("Core exists only to understand vanilla-Laravel patterns"). Placing `#[ResponseFile]` in Core would both create a second file-reading path (forbidden by the issue) and wrongly make an escape-hatch attribute depend on Core being enabled. `OperationBuilder` is right; the issue’s "place in Core" was simply wrong.

**2. Attachment shape — correct.** `OA\MediaType` has both `$example` (singular) and `$examples` (plural) in swagger-php (MediaType.php:47/58). For one anonymous example, setting `$media->example = <decoded>` is the idiomatic single-example representation and consistent with the pipeline’s singular-`example` precedent. No name to invent. Sound.

## Two required sharpenings (pin in the plan body before coding)

**A. Primary-response resolution for `status: null` cannot "mirror `applyResponseExamples`" — that method has no primary logic to mirror.** `#[ResponseExample]`’s `status` is a **required** `public int $status` (no default); `applyResponseExamples` always matches a concrete status and never resolves "primary." `build()` already computes the primary: `$primaryResponse = $primaryOverride ?? $autoPrimaryResponse ?? new OA\Response([200])` (lines 204–206) and `$responses = [$primaryResponse, ...$additionalResponses]` (214), so `$responses[0]` *is* the resolved primary. **Recommendation:** for `status: null`, target `$primaryResponse` directly (the established pattern: `applyLinkAttributes($action, $primaryResponse)` at 217), matching `(string) $primaryResponse->response`. Rewrite the plan’s "resolve the primary status the same way applyResponseExamples matches" to "target the already-computed `$primaryResponse`."

**B. Test case 5 (204 → "silent skip") contradicts the behavior it claims to mirror.** `applyResponseExamples` **unconditionally scaffolds** a JSON media type when the matched response has empty content (`$content = [MediaType::Json->schema()]`, lines 725–727) — it does **not** special-case 204 or skip bodyless responses. A mirrored `applyResponseFiles(status: 204)` would therefore *scaffold a body and attach the example*, the opposite of "silent skip." The genuine silent-skip in the example path is when the status matches **no declared response** (your case 6), or the media isn’t an `OA\MediaType`. **Decide and pin:** either (a) `#[ResponseFile]` intentionally does NOT scaffold onto a conventionally-bodyless 204 (add a guard, and write case 5 to that intended behavior), or (b) accept the scaffold-like-examples behavior and rewrite case 5’s expectation (it is then not a "skip"). As written, case 5 would assert behavior the mirrored code does not have.

## Minor (note, not blocking)

- **`example` + `examples` mutual exclusivity.** OpenAPI forbids both on one media type. If an action carries `#[ResponseExample(status: 200)]` (sets `examples`) *and* `#[ResponseFile]` (sets `example`) on the same response, the result has both and may fail `validate()`. Worth a one-line defensive decision (skip/precedence) or at least a documented "mutually exclusive per response" note. Not required for merge, but cheap to settle now.

## Rest of the checklist — clean

- **PhpStan `ExampleValueOrFileRule`:** keyed on a fixed map of exactly `#[Example]`/`#[ResponseExample]` (ExampleValueOrFileRule.php:46), validating value-xor-file. `#[ResponseFile]` (only `file`) is out of scope — no change. Confirmed.
- **Error semantics:** `ExampleFileLoader::load()` throws `RuntimeException` for not-found/unreadable and invalid-JSON (lines 45/53); no catch in the plan, so cases 3/4 propagate correctly — distinct from the `InvalidArgumentException` malformed-attribute skip (which guards a ctor contract `#[ResponseFile]` lacks). Correct.
- Tier (escape-hatch attribute), `docs/attributes.md` row, `CHANGELOG [Unreleased]`, surgical scope (<150 src lines), survey gate flagged — all named and appropriate.

**Verdict:** seam and shape approved; tighten the plan body for A and B (and ideally settle the minor) before handing to a coder. With A and B pinned, this is ready to implement.

Radiergummi:
🤖 **[lead]** plan-review approved both flagged decisions (seam correction confirmed correct & CLAUDE.md-consistent; singular $media->example idiomatic). → coding with 2 required sharpenings + 1 precedence decision (see coder handoff). Survey gate applies (area:core).

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed f4c21fe9. Ready for review. Generation-affecting (area:core) → **survey gate applies** before merge. src production diff ~95 lines (67 in OperationBuilder + the ~28-line attribute), well under 150.

Decisions per the three required sharpenings:

**A (status: null → primary).** `applyResponseFiles($action, $responses, $primaryResponse)` targets the already-computed `$primaryResponse` directly when `status` is null (the `applyLinkAttributes($action, $primaryResponse)` pattern); a set `status` matches `(string) $response->response`, dropping silently if none matches.

**B (204 — deviation, documented).** Empirically confirmed `applyResponseExamples` **unconditionally** scaffolds a JSON media type on an empty-content response, including a 204 — it does not special-case bodyless statuses. Mirroring that for `#[ResponseFile]` would emit an invalid JSON body on a 204. I implemented the **correct** behavior instead: a conventionally bodyless status (204/205/304) with no content is **skipped silently**, reusing the same bodyless-code set as `ResponseSuccessEmptyBody::BODILESS_CODES`. Test case 5 asserts a declared 204 (via `#[Response(204)]`) gains no content. 🤖 **[coder] deviation:** ResponseFile diverges from ResponseExample's unconditional scaffold on bodyless statuses, by design — scaffolding a body on a 204 is invalid OpenAPI.

**C (example/examples mutual exclusivity).** Precedence: **pre-existing named `examples` win**. `applyResponseFiles` runs right after `applyResponseExamples`, so if a media type already carries named examples, the singular `example` is left unset (guarded by `!is_array($media->examples)`). Test case 7 covers the collision; a `validate()`-asserting test confirms the singular-example doc is valid OpenAPI.

8 tests (`--filter ResponseFile`), all green; full suite 2763 passed, Pint + PHPStan clean. PhpStan `ExampleValueOrFileRule` untouched (keyed to `#[Example]`/`#[ResponseExample]`; `#[ResponseFile]` is out of scope). Watching CI.

Radiergummi:
🤖 **[lead]** implemented at f4c21fe9: 95 src lines, A/B/C resolved (B = documented bodyless-status deviation, C = named-examples-win precedence + validate() test). Full matrix CI green. → code review (two reviewers). Survey gate (area:core) runs after review approves, before merge.

Radiergummi:
## reviewer2 — independent review (PR #377, issue #40): approved (two minor notes)

Reviewed at HEAD `f4c21fe9`. Verified all three documented decisions against the tree, mutation-tested the load-bearing one, and ran the suite. Strong PR — approving. Two minor, non-blocking notes.

### Decisions — all correct
- **A (`status: null` → primary).** `applyResponseFiles` targets the already-resolved `$primaryResponse` when `status` is null, else `$byStatus[(string) status] ?? null` and drops silently on no match. `ExampleFileLoader::load` is **reused** (not duplicated), so the `RuntimeException` propagation is the existing loader's — both messages ("not found or unreadable", "invalid JSON") asserted by tests. ✔
- **B (skip bodyless 204/205/304) — correct deviation.** The mirror `applyResponseExamples` scaffolds a JSON body *unconditionally* ("an example implies a body"), which would emit an invalid `204 + body`. `applyResponseFiles` correctly refuses to scaffold on 204/205/304. The new method is **more correct** than the one it mirrors; the deviation is justified. The code set is right (the `(int) $response->response` cast also covers the string/int status issue). ✔ *(Aside: the mirror's unconditional 204 scaffold looks like a latent pre-existing bug — out of scope here, flagging for a possible follow-up.)*
- **C (mutual exclusivity) — verified non-vacuous.** Guard `!is_array($media->examples)` correctly distinguishes "has named examples" (array) from unset (`Generator::UNDEFINED`, a string). I **mutation-tested** it: removing the guard makes the collision test fail. Notably, the `validate()` test still **passed** under that mutation — swagger-php's `validate()` does *not* reject `example`+`examples` co-emission — so the explicit `->and($media['example'])->toBeNull()` assertion is what actually guards C. The coder rightly asserted it directly rather than leaning on `validate()`. ✔

### Tests / seam / conventions
- 8 tests, all branches: primary default, explicit status, missing-file throw, invalid-JSON throw, bodyless-no-scaffold, no-matching-status silent drop, collision, validate(). Non-vacuous (mutation-confirmed for C). ✔
- Seam placement correct: `OperationBuilder` is `Support\Generator` (general infra, where sibling `#[Response]`/`#[ResponseExample]`/`#[ResponseHeader]` already attach), not Core; `#[ResponseFile]` is a general `src/Attributes/` authoring attribute, not plugin-gated. Per CLAUDE.md. ✔
- Pint clean, PHPStan L8 clean. CHANGELOG present and accurate. ✔

### Minor 1 (DRY) — non-blocking
The bodyless set is an inline literal `[204, 205, 304]`, duplicating `ResponseSuccessEmptyBody::BODILESS_CODES`. It won't track if the canonical set ever changes. `BODILESS_CODES` is currently `private`, so sharing it would mean promoting it (e.g. to a shared enum/const). Optional; the literal is correct today.

### Minor 2 (docs) — non-blocking
The `docs/attributes.md` row matches the depth of its closest siblings (`ResponseExample`/`Example` are table-rows-only), so a row is the right *form* — no docs sub-phase needed. But the row omits the two things a user must get right: the path is **relative to the project root** and must be **valid JSON**. Suggest folding those into the row (or a one-line note), since the attribute is unusable without knowing them. The class docblock covers behaviour well; this is purely about the user-facing page.

No blocking concerns. Approving.

Radiergummi:
🤖 **[reviewer]** review round 1 — 1 finding (+ 2 minor notes). The implementation is correct and the three decisions hold up under adversarial probing.

## Decisions verified

**A (primary targeting) — correct.** `status: null` targets the already-resolved `$primaryResponse` passed from `build()` (the sharpening from plan-review); set status matches `(string) $response->response` via the `$byStatus` map and drops silently if absent. Tests 1/2/6 cover it.

**B (bodyless skip) — behaviorally correct.** A conventionally bodyless status (204/205/304) with empty content is skipped rather than scaffolded (scaffolding a JSON body on a 204 is invalid OpenAPI). Mutation-confirmed non-vacuous: removing the skip makes the bodyless test FAIL. **But see finding 1 on the *source* of the set.**

**C (collision precedence) — airtight, and order-independent.** I probed the worst case: `#[ResponseFile]` declared *before* `#[ResponseExample]` in source. It still yields correctly (`example` UNSET, `examples` SET, document valid) because `applyResponseExamples` always runs first in `build()` (line 215) regardless of source attribute order, so `$media->examples` is already an array when `applyResponseFiles` checks `!is_array($media->examples)`. Two `#[ResponseFile]`s on one response → last-wins, never both set. No path sets both `example` and `examples`. Mutation-confirmed: removing the guard makes the collision test FAIL.

## Finding 1 (should fix) — deviation comment misrepresents the bodyless set source

The PR’s decision-B note says the skip reuses `ResponseSuccessEmptyBody::BODILESS_CODES`. It does **not**: the code inlines `[204, 205, 304]` (`OperationBuilder.php:788`). That constant is `private` and lives in `src/Lint/Rules/ResponseSuccessEmptyBody.php` — unreachable from `Support\Generator\`, and importing `Lint\` into `Support\` would breach the namespace boundary anyway. So the **inline set is the correct call** here; the issue is purely that the documented deviation claims a reuse that did not (and should not) happen. Please correct the PR comment to state the set is defined inline deliberately (the lint constant is a private lint-layer detail Support must not depend on). If a shared bodyless-status source is ever wanted, it belongs in a neutral `Support`/`Contracts` home, not borrowed from a lint rule — out of scope here.

## Minor (non-blocking)

2. **Docs depth.** The single `docs/attributes.md` row is consistent with how the sibling `#[ResponseExample]` is documented (table-row-only, no prose section), so it’s defensible. But the genuinely non-obvious behaviors — path is **relative to the project root**, the 204/205/304 silent skip, and yielding to `#[ResponseExample]` on the same response — currently live only in the CHANGELOG, not in consumer docs. Consider enriching the row (or a one-line note) with at least the path-relativity, since that’s the most likely foot-gun. Your call; not material enough to block.

3. **`validate()` does not actually guard the collision.** I mutation-removed the collision guard: the `validate()` test still PASSED with both `example` and `examples` set (swagger-php 6.2 does not reject the pair). The real guard is the explicit `->and($media[example])->toBeNull()` assertion in the collision test (which correctly FAILED). The "so validate() stays happy" framing in the test comment slightly overstates validate()’s role — the toBeNull assertion is what protects the mutual-exclusivity. Trivial; mention only if you touch the comment.

## Clean

Error semantics (RuntimeException propagates from `load()`, tests 3/4 use `->throws(...)`), seam in `Support\Generator\` with `src/Plugins/Core/` untouched, Pint + PHPStan clean, 79 src lines (under the auto-merge cap), CHANGELOG + docs row present, attribute is a clean non-`BaseExample` readonly class with no constructor I/O.

Net: correct, well-tested, conventionally clean. Fix finding 1 (comment accuracy) and I’ll re-review; 2/3 are optional.

Radiergummi:
🤖 **[coder]** addressed reviewer2 round 1, pushed 3f4044a1:

- **Note 2 (docs — folded in):** the `#[ResponseFile]` row now states the path is relative to the project root and the file must be valid JSON — the two things a user must get right (and the two failure modes cases 3/4 cover).
- **Note 1 (DRY — skipped, with reason):** the `[204, 205, 304]` set duplicates `ResponseSuccessEmptyBody::BODILESS_CODES`, but that const is `private`; sharing it means promoting it to a new shared home for a stable 3-element set. The churn (touching `ResponseSuccessEmptyBody` + deciding a shared location) outweighs the benefit for a list that essentially never changes — leaving the inline literal, per your "skip if not worth the churn." Happy to revisit if the team wants a canonical bodyless-status enum later.

Thanks for the mutation-testing on decision C — agreed the explicit `->example->toBeNull()` is the real guard (validate() does not reject example+examples co-emission), which is why the assertion is direct rather than leaning on validate().

CI all green at 3f4044a1.

Radiergummi:
🤖 **[lead]** Naming decision (Moritz): rename the attribute `#[ResponseFile]` → **`#[ResponseExampleFile]`** before merge. Rationale: `ResponseFile` reads like it defines the whole response (or a file-download response); keeping **Example** in the name makes it unmistakable that it supplies an *example* for the response, loaded from a file — and reads as the file-based sibling of `#[ResponseExample]`. Coder is renaming across the class/file/method/tests/docs/CHANGELOG + PR title; re-review on the renamed commit, then survey gate, then merge.

Radiergummi:
🤖 **[coder] deviation (corrected):** my earlier deviation note said the bodyless set "reuses `ResponseSuccessEmptyBody::BODILESS_CODES`." That was wrong — the code **inlines** `[204, 205, 304]` and never imported the constant. Both reviewers confirm the inline is the **correct** call: that constant is `private` and lives in `Lint\`, so importing it into `Support\Generator\` would breach the Support-must-not-depend-on-Lint boundary. The literal is deliberate, not a missed DRY. If a shared bodyless-status source is ever wanted it belongs in a neutral `Support\`/`Contracts\` home, out of scope here.

🤖 **[coder]** addressed round 1 (both reviewers), pushed a476c8be:
1. Corrected the false reuse claim above; enriched the in-code comment to state the set is inlined deliberately (canonical list is a private Lint-layer detail Support must not depend on). No code change — the logic was already inline-correct.
2. (already in 3f4044a1) docs row notes the path is project-root-relative and the file must be valid JSON.
3. Reworded the collision test comment to credit the `->example->toBeNull()` assertion as the real mutual-exclusivity guard (validate() does not reject example+examples on its own).

No production-logic change. CI all 9 green at a476c8be.

Radiergummi:

Re-reviewed a476c8be on my own merits (diff f4c21fe9..a476c8be). All three round-1 points resolved, comment/docs-only — production logic is byte-identical (I confirmed: the only `src/` change is the bodyless-skip comment; no non-comment +/- src lines).

1. (finding 1) The in-code comment now accurately explains the inline `[204, 205, 304]` set: "inlined deliberately: the canonical list is a private Lint-layer detail, and Support must not depend on Lint." Captures the real boundary why; the corrected deviation comment is posted. Resolved.
2. `docs/attributes.md` row now notes the path is project-root-relative and must be valid JSON — covers the foot-gun.
3. The collision test comment now credits `->toBeNull()` as the real mutual-exclusivity guard and notes `validate()` does not reject example+examples co-emission on its own — matches my mutation finding.

Verified at a476c8be: 8/8 tests pass, Pint clean. The round-1 logic validation (all three decisions correct and mutation-confirmed non-vacuous, seam in Support, Core untouched, PHPStan clean, 79 src lines) carries forward unchanged. Approving — one of two required reviewers; survey gate runs after.

Radiergummi:
🤖 **[coder]** renamed + final accuracy fixes, pushed bf7cb6a1:

**Rename `#[ResponseFile]` → `#[ResponseExampleFile]`** (Moritz's call — keeps "Example" so it reads as "supplies an example, loaded from a file"). Thorough (`grep -rn ResponseFile src/ tests/ docs/ CHANGELOG.md` → clean): attribute class + file, `OperationBuilder::applyResponseExampleFiles()` + import + call site + attribute lookup, the fixture controller (`ResponseExampleFileFixtureController`), test file (`ResponseExampleFileAttributeTest.php`), docs row, CHANGELOG entry, and the PR title. `--filter ResponseExampleFile` → 8/8 green.

**Pending accuracy fixes (carried through the rename, both in this commit):**
- Finding 1: in-code comment states the `[204,205,304]` set is inlined **deliberately** (canonical list is a private `Lint\` detail; `Support\` must not depend on `Lint\`) — not a reuse of `ResponseSuccessEmptyBody::BODILESS_CODES`. Corrected deviation comment is above.
- Note 3: collision test comment credits the `->example->toBeNull()` assertion as the real mutual-exclusivity guard.
- reviewer2 note 2 (path-relative + valid-JSON in the docs row) carried through.

Full suite 2763 passed, Pint + PHPStan clean; remote CI all 11 pass incl. swagger-php 5.8 + PHP 8.5 + codecov. No production-logic change in this push (pure rename + comment accuracy).

Radiergummi:
🤖 **[reviewer]** review round 3 (final commit bf7cb6a1): approved, no concerns.

Re-reviewed the rename `#[ResponseFile]` → `#[ResponseExampleFile]` on my own merits (diff a476c8be..bf7cb6a1). It is a clean, complete, identifier-only rename:

- Verified **no stale `ResponseFile` references** remain in `src/`, `tests/`, `docs/`, or `CHANGELOG.md` (grep for `ResponseFile` minus `ResponseExampleFile` → none). No dangling reference.
- Verified the delta is **identifier-only**: the only non-context diff lines are the renamed identifiers (class + file rename, import alias, `applyResponseFiles`→`applyResponseExampleFiles` + docblock, CHANGELOG, docs row, test file + fixture controller + const + usages). No production-logic drift.
- 8/8 tests pass, Pint clean, PHPStan clean.

The name is also an improvement: `ResponseExampleFile` makes the file-backed sibling relationship to `#[ResponseExample]` explicit, where `ResponseFile` could misread as "a file response."

My rounds 1–2 validation carries forward unchanged (decisions A/B/C correct and mutation-confirmed non-vacuous; seam in `Support\Generator\` with Core untouched; RuntimeException error semantics; the two accuracy fixes from round 1). Approving the final commit — one of two required reviewers; survey gate runs after.

Radiergummi:
## reviewer2 — approval against final commit `bf7cb6a1`

Re-confirming my approval on the record against the **final** commit (my earlier approval was on `f4c21fe9`, which predates the round-1 accuracy fixes and the rename).

**Rename `#[ResponseFile]` → `#[ResponseExampleFile]` — clean and complete.** Verified zero leftover `ResponseFile` references in src/tests/docs/CHANGELOG; the attribute class (git-mv, 95% similarity), import alias, `applyResponseExampleFiles` method + docblock, test file/class/const/controller, CHANGELOG entry, and `docs/attributes.md` row are all consistently updated. No production-logic change beyond the rename.

**Decision verdicts carry forward unchanged:**
- **A** (`status: null` → primary; declared-status targeting; silent drop on no match; reused `ExampleFileLoader` with both error messages asserted) — correct.
- **B** (skip conventionally bodyless 204/205/304 rather than scaffolding an invalid body) — a correct deviation; more correct than the `applyResponseExamples` it mirrors. The inline `[204,205,304]` is deliberate (the canonical `BODILESS_CODES` is a private Lint-layer detail and `Support` must not depend on `Lint`), now documented in-code.
- **C** (`example`/`examples` mutual exclusivity via `!is_array($media->examples)`) — verified non-vacuous by mutation in my round-1 review; the explicit `->example->toBeNull()` assertion is the real guard (swagger-php `validate()` does not reject co-emission), and the test comment now credits it.

Re-verified locally at `bf7cb6a1`: the renamed test is **8/8 green (12 assertions)**, Pint clean, PHPStan level 8 clean. No remaining concerns — **approved, mergeable.**

Radiergummi:
🤖 **[lead]** two independent reviewer approvals posted on final renamed commit bf7cb6a1; full matrix CI green. → survey gate (Layer A, generation-affecting). #[ResponseExampleFile] is opt-in + unused by dogfood apps, so expecting 0 generation drift.

Radiergummi:

Baseline regenerated on current `main` (`f73b40f8`) first — the cached baseline was 17 commits stale (several generation-affecting merges since). Candidate run on PR head `bf7cb6a1`.

Every deterministic metric (paths, apiOperations, operations, componentSchemas, parameters/requestBodies, documentedResponses, responseSchemas, completenessPercent) is **byte-identical** between baseline and candidate on all 10 apps that generated successfully on baseline — exactly as expected: the new `applyResponseExampleFiles` path early-returns when no `#[ResponseExampleFile]` attribute is present, and no corpus app uses it.

One footnote, not a regression: **InvoiceNinja** was `gen_exit=1` on the baseline run due to a transient network failure during `composer require` (`curl error 6 … Could not resolve host: repo.packagist.org`), and generated cleanly on the candidate run (529 paths). That is a baseline-side DNS flake, not a library change — an opt-in/guarded PR cannot fix a generation crash. No app that was green on baseline regressed.

Clear to merge.

Radiergummi:
🤖 **[lead]** rebased onto current `main` (now includes #380) → new head `7fc3f76d`. CHANGELOG `[Unreleased]` kept **both** entries (line 38 ResponseExampleFile / line 40 #380's class-level ResourceField rule); clean rebase, no conflict. Local `composer check` green (2765 tests, Pint, PHPStan). Survey Layer A PASS (0 regressions) + two reviewer approvals carry forward unchanged — re-watching CI on the rebased head, then merging.

Radiergummi:
🤖 **[lead]** all gates green → squash-merged. Survey Layer A PASS (0 regressions/11 apps), two independent reviewer approvals, full matrix CI green incl. swagger-php 5.8 on the rebased head 7fc3f76d. Non-controversial: opt-in #[ResponseExampleFile] attribute + guarded OperationBuilder branch (79 src lines), no Contracts/stage-order/config/signature changes. Closes #40. Worktree removed.

Issue #208, opened 2026-06-09
Spec source map — trace every generated field back to its origin


Radiergummi:
## Summary

Emit a **spec source map**: a side artifact keyed by JSON Pointer that records, for each field in the
generated document, *which stage produced it* and (where a stage cooperates) *what source it derived
it from* — `app/Models/User.php:42`, a FormRequest rule, a PHPDoc tag, an `@OA` block. Surface it via
an `openapi:explain <pointer>` command and let lint findings cite origin ("this response is documented
because of your typed return at `Controller.php:42`").

## Motivation

For an inference-based generator, the single most common user question is **"why is this in my spec /
why is it missing?"** Today there is no principled answer. The source-map concept from the JS world
fits: an external, lazily-consumed artifact mapping generated output → original source, where each
transform contributes a composable segment. Our pipeline (`SpecPipeline` loops `registry->stages`) is
exactly such a chain.

This is justified **on explainability/debuggability merits alone** — explicitly *not* as migration
infrastructure. Investigation confirmed provenance does **not** retire the migration differential:
schemas are realized downstream of the harvester, so provenance gives *origin*, never the
*counterfactual* migration needs. The two must not be coupled.

## Where the source-map analogy holds — and breaks

- **Holds:** external artifact (output stays clean — no decorating swagger-php's `OA\*` objects, no
  fighting serialization); per-stage maps that **compose** into a final spec→source map; lazy emission
  (built only on demand, mirroring the existing "ordinary runs pay nothing" stance).
- **Breaks — design around it:** mappings here are **N:1 with precedence** (a field can be a merge of
  inference + attribute + harvested-override, with a winner), not 1:1 positional; and the address
  space is partly a **component graph** (pooled/`$ref`'d schemas), not output text positions. So it's
  "multi-source attribution with precedence," source-map-*shaped* but not the literal `.map` format —
  don't pull in `source-map` tooling expecting a fit.

## Proposed approach (tiered — ship the cheap tier first)

1. **Stage-granularity, nearly free, no extension-surface obligation:** wrap each `$stage->apply()` in
   `SpecPipeline`, diff the document before/after, attribute new/changed pointers to that stage.
   Yields `#/.../email ← ErrorResponseInferenceStage` / `← SwaggerPhp harvester`. Ship
   `openapi:explain <pointer>` on this alone — likely 80% of the value.
2. **Source-granularity, the real cost, only if tier 1 proves it earns its keep:** cooperating stages
[...]

PR #387, opened 2026-06-17
feat(#178): openapi:why --fields — operation-level field provenance

### LINKED ISSUE #178: Operation-level provenance — explain *why a field has its value* (`openapi:why --fields` / `openapi:explain`)
## Summary

`openapi:why` today explains only **route inclusion/exclusion** — which filter / visibility / spec-membership rule decided that a route is or isn't in a given spec. It never builds the operation, so it cannot answer the *other* "why" a user has: **why does this operation have this status code / summary / tag / schema / parameter?** As the generator infers more, add an operation-level provenance trace that names the source (and reason) for each derived field.

## Motivation

The generator now derives a growing set of operation fields from convention: success status codes and default summaries (#25), tags (#37), implicit error responses (#29), response schemas from return types/models, query/path params, security from middleware, and so on. When a field's value surprises a user — *"why is this `201`?"*, *"where did this `List Flights` summary come from?"*, *"why did my `#[Summary]` not win?"* — there is no way to ask the tool. The only recourse is reading the package source to reconstruct the precedence ladder.

This is the same observability gap `openapi:why` already closed for inclusion, one level deeper. The precedence ladders are *already* encoded in `OperationBuilder` (e.g. `resolveSummary()` returns `methodAttr ?? docComment ?? classAttr ?? convention?->summary`) — they just aren't surfaced. Making derivation auditable turns "type your returns / annotate to get X" from documentation into something the tool can demonstrate per endpoint, and makes the inference ladder debuggable.

## Proposed approach

Instrument the existing derivation — **no new parsing, no new inference** (Tier-0 instrumentation of work already done).

- Thread provenance records out of `OperationBuilder::build()` and its resolver chains. The resolver contracts (`OperationConventionResolver`, `PrimaryResponseResolver`, `QueryParameterResolver`, `RefSchemaResolver`, error-response contributors) currently return bare values with no "who decided this and why" metadata; capture, per operation, a list of entries like `{field, value, source, reason}` where `source` is a resolver class / attribute / docblock / convention and `reason` is a short human string (e.g. `store → POST`).
- Model it on the existing inclusion trace: `Support\Inclusion\TraceEntry` + `InclusionDecision->trace`, which `WhyCommand::printSpecDecision()` already walks and prints. An operation provenance trace is the analogous structure one layer in.
- Surface it via either `openapi:why <route> --fields` (extend the existing command to additionally build the operation and print field provenance) or a dedicated `openapi:explain <route>`. Lean toward extending `openapi:why` so there's one "explain this endpoint" entry point.
- Likely **incremental**: start with the cheap, high-value fields (summary, success status, tags), then extend to responses, params, security. May warrant promotion to an `epic` if the per-field plumbing is large.

### Boundary

Provenance is instrumentation, not inference — it records decisions the pipeline already makes. It must degrade gracefully (a resolver that returns a value without a provenance record simply shows `source: unknown`) and must not change generated output. It is **not** a lint subsystem concern; it's a CLI/DX read-only view.

## Acceptance criteria

- `openapi:why <route> --fields` (or `openapi:explain <route>`) prints, for a single resolved operation, the source and reason for each derived field — at minimum `summary`, success status code, and `tags`.
- Convention-derived fields name the responsible resolver and the matched signal (e.g. `summary: 'List Flights' ← ResourceConventionResolver (index → GET)`; `status: 201 ← ResourceConventionResolver (store → POST)`).
- An author override is shown winning over a convention (e.g. `summary ← #[Summary] (method)`, with the convention listed as a superseded candidate).
- A feature test under `tests/Feature` covers the convention-derived and attribute-override provenance for a fixture controller.

## References

Prompted by #25 (resource convention → summaries + status codes), which derives fields with no way to surface their origin. Model: `src/Support/Inclusion/TraceEntry.php` + `InclusionEvaluator` inclusion trace, consumed by `src/Console/WhyCommand.php`. Affected: `src/Console/WhyCommand.php`, `src/Support/Generator/OperationBuilder.php`, and the resolver contracts in `src/Contracts/Registry/`. Candidate for the Roadmap project. Verify: `vendor/bin/pest tests/Feature`.

Radiergummi:
Closes #178

## What & why

`openapi:why <route>` today explains only **route inclusion/exclusion**. This PR adds a second,
opt-in view — `openapi:why <route> --fields` — that builds the resolved operation and prints the
**source and reason for each derived field**, mirroring the inclusion trace one layer in.

This is **Tier-0 instrumentation**: it records decisions the pipeline *already makes* in
`OperationBuilder::build()`. No new parsing, no new inference. **Generated output is unchanged** —
provenance is a read-only side channel consumed by the CLI.

## Scope (AC minimum, deliberately narrow)

Provenance for exactly three fields: **`summary`**, **success status code**, **`tags`**. Each
entry records `{field, value, source, reason}` plus any **superseded** candidates (so an author
override can show the convention it beat). Responses (schemas/error envelopes), parameters, and
security are **explicitly out of scope** — noted as follow-ups on #178, not built here.

One entry point: extend `openapi:why` with a `--fields` flag rather than adding `openapi:explain`.

## Contract surface — controversy assessment

**No `src/Contracts/**` file changes.** This was the key design question flagged at hand-off, so
it is settled up front by verification, not assertion:

- The full precedence ladders for the three target fields already live **inside**
  `OperationBuilder::build()` / `resolveSummary()` / `applyConventionStatus()` / the tag-merge
  block — not in the resolver contracts. `OperationConvention` already carries
  `successStatusCode` + `summary` and is produced by a known resolver class
  (`ResourceConventionResolver`), so the convention source/reason is recoverable without widening
  `OperationConventionResolver::resolve()`.
- Provenance is therefore captured locally in `OperationBuilder` and threaded out on
  `OperationDescriptor`, which lives in `Support\Generator\` — **not** a contract
  (`grep -rln OperationDescriptor src/Contracts/` → empty, verified).

So `PrimaryResponseResolver`, `QueryParameterResolver`, `RefSchemaResolver`,
`OperationConventionResolver`, and the error-response contributor contracts are **untouched**.
This avoids the controversy trigger the issue anticipated. The deliberate consequence: a
resolver that contributes a value without a provenance record shows `source: convention/unknown`
— exactly the graceful-degrade behaviour the issue's Boundary section asks for. If review judges
that capturing the *summary precedence* faithfully still reads as "decision logic the contract
should own," that is the one place to revisit — but the proposed design keeps it in `Support\`.

(Adds a CLI flag, not a config key. `area:cli`. Not generation-affecting → **survey skipped**.)

### 1. Provenance value object — `src/Support/Provenance/FieldProvenance.php` (new)

Models one decided field, analogous to `Support\Inclusion\TraceEntry`. Pure data, `@internal`.

```
final readonly class FieldProvenance {
    public function __construct(
        public string $field,        // 'summary' | 'status' | 'tags'
        public string $value,        // rendered value, e.g. "201", "List Flights", "flights, v1"
        public string $source,       // '#[Summary] (method)', 'ResourceConventionResolver', 'default'
        public string $reason,       // 'store → POST', 'author override', 'no convention matched'
        /** @var list<string> superseded candidates, e.g. ["convention: 'List Flights'"] */
        public array $supersededBy = [],
    ) {}
}
```

(A short, fixed source vocabulary — attribute name + scope, resolver short-class, or `default` —
kept as plain strings exactly like `TraceEntry->reason`. No enum: matches the existing trace.)

### 2. `OperationDescriptor` carries the trace — `src/Support/Generator/OperationDescriptor.php`

Add `public array $provenance = []` (`@var list<FieldProvenance>`) as a **trailing, defaulted**
constructor parameter so existing call sites (`withOperationId`) stay green. It is **not** added
to `toArray()` / `attachTo()` — it never reaches the OpenAPI document. `withOperationId()` copies
it through.

### 3. Capture in `OperationBuilder::build()` — `src/Support/Generator/OperationBuilder.php`

Build a `list<FieldProvenance>` alongside the existing logic and pass it into the
`OperationDescriptor` return. Three capture points, reusing values already computed:

- **summary** — refactor `resolveSummary()` to also report which branch won
  (`methodAttr` → `#[Summary]`/`#[Operation]` (method); `descriptor->summary` → docblock;
  `classAttr` → class attribute; `convention?->summary` → `ResourceConventionResolver`). Losing
  candidates that were non-null become `supersededBy`. Done by returning a small internal
  result struct or an out-parameter — chosen to keep `build()` readable; no behaviour change.
- **status** — at the existing status-resolution block (lines ~201-212): record whether the
  status came from an explicit `#[Response(2xx)]` (`primaryOverride`), a body-scanned explicit
  status (`autoStatusIsExplicit`), the convention (`applyConventionStatus`), or the `200`
  fallback. The convention reason is derived from the matched action (e.g. `store → POST`); since
  `OperationConvention` doesn't carry the matched signal, the reason is reconstructed from the
  `ActionDescriptor` (method name + emitted verb) — Tier-0 reflection already on hand.
  - **Name the actual resolver, not a literal (plan-review tightening).** `resolveConvention()`
    (L493-508) loops `operationConventionResolvers` and currently returns the bare
    `OperationConvention`, **discarding which resolver produced it**. Capture
    `$conventionResolver::class` *inside that loop* (L502, where it is in scope) and thread it out
    alongside the convention, so the `source` reflects the resolver's short-class. Do **not**
    hardcode the string `ResourceConventionResolver` — that would lie if a third party registers a
    different `OperationConventionResolver`. This stays in `Support\`; no contract change. Add a
    test asserting the source reflects the actual resolver short-class, not a literal.
- **tags** — at the tag-merge block (lines ~224-249): record the merged tag list with source
  `controller-derived` and, when `#[Operation(tags)]` / `#[Tag]` contributed, name the attribute
  and whether it replaced or merged.

Resolver fault-isolation is unchanged; a resolver that throws still degrades and simply yields no
convention provenance.

### 4. CLI — `src/Console/WhyCommand.php`

- Add `{--fields : Also build the operation and explain how each derived field got its value.}`
  to the signature.
- When `--fields` is set, after the inclusion section, build the operation for the resolved
  descriptor via `OperationBuilder` (resolved from the container, per-verb like `PathsStage`:
  `$descriptor->withHttpMethod($method)`), then print a `Fields:` block walking
  `$operation->provenance`, formatted like `printSpecDecision`:

  ```
  Fields:
      summary: 'List Flights' ← ResourceConventionResolver (index → GET)
      status:  200            ← ResourceConventionResolver (index → GET)
      tags:    flights        ← controller-derived
  ```

  Override example:
  ```
      summary: 'Find one flight' ← #[Summary] (method)
                                   (superseded: convention 'Show Flight')
  ```

- Default behaviour (no `--fields`) is byte-for-byte unchanged.
- `--fields` needs a tag for `OperationBuilder::build($action, [$tag])`; reuse `TagDeriver`
  (`Support\Generator\TagDeriver`) exactly as `PathsStage::attachOperation` does.

## Tests (TDD — write first, expect red)

New file `tests/Feature/OperationProvenanceTest.php` (Feature, since it drives the command +
real resolver chain through Testbench), plus assertions added to
`tests/Unit/Console/WhyCommandTest.php` for the flag plumbing.

Cases (covering branches, not just happy path):
1. **convention-derived status** — resourceful `store` (POST) fixture →
   `status: 201 ← ResourceConventionResolver (store → POST)`.
2. **convention-derived summary** — `index` (GET) → `summary: 'List Flights' ← ResourceConventionResolver (index → GET)`.
3. **attribute override wins, convention shown superseded** — `#[Summary('Find one flight')]` on
   `show` → `summary ← #[Summary] (method)` **and** the convention summary listed as superseded.
4. **docblock summary** beats class attribute / convention → source `docblock`.
5. **explicit `#[Response(201)]`** → status source is the attribute, convention not applied
   (mirrors the existing `autoStatusIsExplicit`/`primaryOverride` guard).
6. **no convention matched** (non-resourceful method, no attributes) → `summary` absent or
   `source: default`, `status: 200 ← default`, exercising the degrade path.
7. **tags** — controller-derived tag present; with `#[Operation(tags: [...], replace: true)]`,
   source names the attribute and `replace`.
8. **flag off** — `openapi:why <route>` without `--fields` prints no `Fields:` block (guards the
   no-behaviour-change-by-default contract).
9. **generated-output invariant** — a snapshot/equality check that
   `OperationDescriptor->toArray()` is identical with and without provenance populated (proves
   the document is unaffected). Lightweight: assert `provenance` ∉ `toArray()` keys.

## Docs & changelog

- **`docs/getting-started.md`** — the `openapi:why` row in the command table gains a sentence on
  `--fields` (field provenance). It's the only place `openapi:why` is tabled.
- **`docs/multi-spec.md`** — the "Debugging with `openapi:why`" section gets a short `--fields`
  example block.
- **`docs/troubleshooting.md`** — add a "why is this status/summary/tag?" entry pointing at
  `openapi:why --fields` (the natural landing spot for the surprise-value question the issue
  motivates).
- **`CHANGELOG.md` [Unreleased]** — `Added: openapi:why --fields explains the source and reason
  for a route's derived summary, success status code, and tags.`
- No lint rule, no config key, no new attribute.

## Out of scope (follow-ups, do not build)

Provenance for response schemas / error envelopes, parameters, and security; an `openapi:explain`
alias; machine-readable (`--json`) output. Recorded as a comment on #178 if the reviewer wants
them tracked.

## Files

- new `src/Support/Provenance/FieldProvenance.php`
- edit `src/Support/Generator/OperationDescriptor.php` (trailing `$provenance` field)
- edit `src/Support/Generator/OperationBuilder.php` (capture three fields)
- edit `src/Console/WhyCommand.php` (`--fields` flag + `Fields:` printer)
- new `tests/Feature/OperationProvenanceTest.php`; edit `tests/Unit/Console/WhyCommandTest.php`
- edit `docs/getting-started.md`, `docs/multi-spec.md`, `docs/troubleshooting.md`, `CHANGELOG.md`

## Verify

`vendor/bin/pest tests/Feature/OperationProvenanceTest.php`, then `composer test`,
`vendor/bin/pint --test`, `composer lint`.


Radiergummi:

Key call-out for review/lead: despite the issue naming `src/Contracts/Registry/`, this plan touches **no** `src/Contracts/**` file — provenance is captured inside `OperationBuilder` and threaded out on `OperationDescriptor` (which is in `Support\`, verified not a contract). So the anticipated controversy trigger does **not** fire; no resolver contract signature widens. The one judgment call for the reviewer is whether capturing the summary precedence in `Support\` (vs. the contract) reads acceptably.

Radiergummi:
🤖 **[lead]** coordination — overlap with #208 (spec source map / PR #210).

Flagging a related, in-scope feature the plan didn't reference: **#208 "Spec source map — trace
every generated field back to its origin"** (stale plan-only draft PR #210, 0 files, last touched
Jun 12). Both are provenance/explainability, but at **different layers** — they should coexist, not
duplicate:

- **#178 (this PR):** *resolver-decision* provenance — *why a field has the value it does* (which
  attribute/convention won the precedence ladder, what it superseded), build-time + in-memory,
  surfaced via `openapi:why <route> --fields`. Operation-scoped, no document artifact.
- **#208:** *document source map* — *which pipeline stage produced each JSON Pointer*, built by
  diffing the document before/after each `SpecPipeline` stage, emitted as an external `spec.map`
  artifact, surfaced via `openapi:explain <pointer>`. Document-scoped, coarser (stage granularity).

**Boundary for both:** #178 owns `openapi:why --fields` (resolver/precedence granularity); #208
owns `openapi:explain <pointer>` (stage/document granularity). #178 explicitly chose to extend
`openapi:why` over adding `openapi:explain`, so the two command surfaces stay distinct. Whoever
picks up #208 should design around this `--fields` view rather than re-deriving operation-field
provenance. cc the merge gate: I'll note this relationship when deciding #178's merge vs escalate.

Radiergummi:
🤖 **[lead]** planning done (no src/Contracts changes per planner's verified design) → plan-review. Reviewer: (1) verify the no-contracts claim holds for summary/status/tags provenance; (2) assess the #208/PR #210 overlap — is extending openapi:why --fields the right boundary vs #208's openapi:explain source map; (3) confirm TDD plan covers convention-derived + author-override-supersedes cases.

Radiergummi:
🤖 **[reviewer]** plan-review: approved with one tightening (amended in PR body). Tier-0, surgical, no `src/Contracts/**` change.

Adjudicated the three lead asks:

**1. No-contracts claim — VERIFIED, holds.** Read `OperationBuilder` (`resolveSummary()` L925-937, status block L186-212, tag-merge L224-249), `OperationDescriptor` (`Support\Generator\`, not a contract — `grep OperationDescriptor src/Contracts/` empty), and `src/Contracts/Registry/`. All three precedence ladders live inside `OperationBuilder`, not in any resolver contract. `OperationConvention` already exposes `successStatusCode` + `summary`. Capturing summary precedence in `Support\` reads correctly — it is already there; the plan only observes it. No resolver signature widens. Not controversial → no human gate on this axis.

**2. #208 overlap — COEXIST (recommended boundary holds).** Distinct layers: #178 is *resolver-decision* provenance (which attribute/convention won the precedence ladder, what it superseded), operation-scoped, surfaced via `openapi:why <route> --fields`, no document artifact. #208 is a *document* source map (which pipeline **stage** produced each JSON Pointer), document-scoped, coarser stage granularity, surfaced via `openapi:explain <pointer>`, external artifact. The lead's coordination comment already fixes the command boundary (`--fields` vs `openapi:explain`), and the plan honours it by extending `openapi:why` rather than adding `openapi:explain`. Not a duplication; #178 does not preempt #208. No escalation needed.

**3. TDD coverage — sufficient.** Plan's cases cover (a) convention-derived status+summary (cases 1-2), (b) `#[Summary]` override winning with the convention shown superseded (case 3), and (c) graceful-degrade (case 6: no convention → `status: 200 ← default`, summary absent/`default`). Also covers docblock-beats-class (4), explicit `#[Response(201)]` status source (5), tag replace (7), flag-off no-`Fields:`-block (8), and the generated-output invariant (9). Good branch coverage.

**Required tightening (amended into the Design section):** The AC mandates output naming the **responsible resolver** (`← ResourceConventionResolver (index → GET)`). But `OperationBuilder::resolveConvention()` (L493-508) loops `operationConventionResolvers` and returns only the bare `OperationConvention`, **discarding which resolver produced it**. The plan must capture `$conventionResolver::class` *inside that loop* (L502, where it is in scope) and thread it alongside the convention — not hardcode the string `ResourceConventionResolver`, which would lie if a third party registers a different `OperationConventionResolver`. This stays in `Support\` (still no contract change). Add a test asserting the source reflects the actual resolver short-class, not a literal.

Radiergummi:
🤖 **[lead]** plan-review APPROVED by reviewer1 (Tier-0, no src/Contracts change verified, #208 boundary holds, TDD sufficient) → coding. Coder: implement per the (amended) PR-body plan. KEY AMENDMENT from reviewer1: capture $conventionResolver::class dynamically inside resolveConvention() (~L502) and thread it through as the provenance source — do NOT hardcode ResourceConventionResolver (would lie for third-party resolvers). Stays in Support\, no contract change. TDD: all 9 cases in the plan.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed 2fe23be7. Ready for review.

Full remote matrix green: PHP 8.4/8.5 × Laravel 12/13, swagger-php 5.8 leg, and the byte-exact Snapshot job all pass — confirming the generated document is unchanged.

Highlights vs. plan:
- `FieldProvenance` VO + `ResolvedConvention` VO in `src/Support/Provenance/`. The latter pairs the convention with the resolver class so provenance names the actual resolver (`class_basename`), honoring the reviewer amendment — captured inside the `resolveConvention()` loop, no contract change.
- `OperationDescriptor::$provenance` is a trailing defaulted ctor param, threaded through `withOperationId()`, absent from `toArray()`/`attachTo()`.
- Capture for summary/status/tags only (AC minimum); status `supersededBy` is always empty by design (only summary surfaces a superseded convention).
- `--fields` builds per emitted verb (mirroring PathsStage) and prints a `Fields:` block.

Test note (deviation worth flagging): `expectsOutputToContain` mangles its buffer when chained with the multibyte arrows (← →), so the Feature test asserts against `Artisan::output()` via `str_contains`/`toContain` instead of chained `expectsOutputToContain`. Documented inline in the test helper.

9 plan cases covered in `tests/Feature/OperationProvenanceTest.php` + the toArray-invariant/withOperationId unit test in `tests/Unit/Support/Generator/OperationDescriptorProvenanceTest.php` + flag-plumbing in `WhyCommandTest`.

Radiergummi:
🤖 **[reviewer]** code-review: approved, no concerns.

Reviewed commit `2fe23be7` adversarially. Ran the suite, Pint, and PHPStan in the worktree: **18/18 tests pass**, `pint --test` passed, PHPStan `[OK] No errors`.

**Correctness & design**
- The plan-review amendment is honored correctly: `resolveConvention()` returns the new `ResolvedConvention($convention, $conventionResolver::class)` VO and provenance derives its source via `class_basename($convention->resolver)` — **not** a hardcoded `ResourceConventionResolver`. The mechanism is name-agnostic, so a third-party `OperationConventionResolver` surfaces under its own short-class; the bundled-resolver assertion (`ResourceConventionResolver (store → POST)`) is just the fixture.
- **Generated-output-unchanged invariant holds, verified two ways.** `OperationDescriptor::toArray()`/`attachTo()` are untouched; `provenance` is a trailing defaulted ctor param outside the serialized key set, and `withOperationId()` threads it through. `OperationDescriptorProvenanceTest` asserts both `not->toHaveKey('provenance')` and byte-equality of `toArray()` with/without provenance populated — on top of the byte-exact Snapshot CI job.
- Per-verb build in `WhyCommand::printFields()` mirrors `PathsStage` (skips HEAD, builds per emitted verb via `withHttpMethod`), so provenance reflects the operation as the document actually builds it.

**Tests** — all 9 planned cases present, meaningful, and coupled to the change (the feature tests assert `Fields:` / `← source` / `superseded:` output only the new code emits):
1. convention status (`201 ← ResourceConventionResolver (store → POST)`); 2. convention summary; 3. **attribute summary winning with the convention shown superseded** (`superseded: convention 'Show ProvenanceFlight'`); 4. docblock-wins; 5. explicit `#[Response]` status source; 6. **graceful degrade** (`status: 200 ← default`, summary absent from the trace); 7. replacing `#[Operation(tags)]`; 8. controller-derived tag; 9. flag-off prints no `Fields:` block.

**Documented deviation — sound.** The coder swapped `expectsOutputToContain` for `Artisan::output()` + `toContain`/`str_contains` because the helper mangles its buffer when chained with the multibyte arrows (`←`/`→`). This does not weaken the assertion: `toContain` on the captured output is a strict substring match and asserts the real arrow-bearing strings faithfully. The arrow-free unit assertions still use `expectsOutputToContain`.

**Conventions & bookkeeping — clean.** strict-types + `@internal` + `readonly` + constructor promotion on both new VOs; no abbreviations; concise why-comments with no issue numbers in code; CHANGELOG `[Unreleased]` entry present and accurate; docs updated across `getting-started.md` (command table), `multi-spec.md` (worked example) and `troubleshooting.md` (with a correct cross-link anchor). No new lint rule → no catalog row needed.

**Non-blocking note (not a finding), for the human gate:** `statusProvenance()` always passes `supersededBy: []`, so an explicit `#[Response(201)]` beating a convention's `200` won't list the convention as superseded. The AC only requires superseded-display for the summary case, so this is in-scope-as-planned — a natural follow-up on #178 if status-supersede display is wanted later, not a defect.

This PR's `src/` diff is +266/-8 (> the 150-line controversy threshold), so it escalates to a human gate regardless of this approval.

Radiergummi:
🤖 **[lead]** vetted + green, but **controversial → escalating to Moritz** (ready, not merged).

**What it is:** `openapi:why <route> --fields` — Tier-0 instrumentation surfacing per-field
provenance (`{field, value, source, reason, supersededBy}`) for **summary, success status, tags**
(AC minimum). New `@internal` VOs `FieldProvenance` + `ResolvedConvention`; capture lives in
`OperationBuilder`; printed by `WhyCommand`. No `src/Contracts/**` changes.

**Gate status:**
- ✅ Full remote CI matrix green (PHP 8.4/8.5 × Laravel 12/13 + swagger-php 5.8 + **byte-exact
  Snapshot** = generated output provably unchanged).
- ✅ Plan-review (reviewer1) + adversarial code-review (reviewer1, approved no-concerns): amendment
  honored (resolver class threaded dynamically via `ResolvedConvention`, not hardcoded);
  output-leak invariant double-verified (`provenance` absent from `toArray()`/`attachTo()` + a
  byte-equality unit test); all 9 planned test cases present and would fail without the change;
  the documented test-harness deviation (`Artisan::output()` + substring vs `expectsOutputToContain`,
  due to multibyte-arrow buffer mangling) judged sound.

**Why this needs you (not auto-merged):**
1. **Size:** `src/` production diff is **+266/−8 (274 lines)** — over the team's ~150-line
   auto-merge threshold.
2. **Product overlap with #208** (spec source map / stale plan-only PR #210): both are
   explainability. reviewer1 adjudicated *coexist* — #178 owns `openapi:why --fields`
   (resolver-decision granularity), #208 owns `openapi:explain <pointer>` (stage/document
   granularity) — and the command surfaces stay distinct. But whether these should ultimately be
   one unified feature is a roadmap call that's yours, not the team's.

**Non-blocking note (as-planned, not a defect):** `statusProvenance` always passes
`supersededBy: []`, so an explicit `#[Response(201)]` beating a convention's `200` won't render the
convention as superseded. AC only requires superseded-display for `summary`; natural follow-up if
you want it for status too.

**Branch:** `MERGEABLE` but 1 commit `BEHIND` main (the docs-only #388 merge; no CHANGELOG overlap
→ trivial "Update branch"/rebase, no conflict). CI green predates that docs commit; re-run after
rebase if you want belt-and-suspenders. Worktree left in place at
`.claude/worktrees/feat-178-operation-level-provenance-explain-why-a`.

src/Console/WhyCommand.php
```diff
@@ -32,14 +41,17 @@ class WhyCommand extends Command
 {
     protected $signature = 'openapi:why
         {route : Route name (exact match) or URI substring.}
-        {--for-env= : Override the environment for Hide/Expose evaluation.}';
+        {--for-env= : Override the environment for Hide/Expose evaluation.}
```
Radiergummi: can we call that flag just `--env` or `-e`?

src/Console/WhyCommand.php
Radiergummi, line 1: we should consider using one of the Laravel rendering components like list here.

Radiergummi:
🤖 **[lead]** Moritz submitted a review (2 inline findings) → back into the cycle at coding (re-entry from agent:needs-human). Branch/worktree exists; coder reuses it. Findings: (1) rename `--for-env` flag to `--env`/`-e` (verify Laravel's reserved global `--env` collision first); (2) render the `--fields` block with a Laravel rendering component (e.g. list/twoColumnDetail) instead of manual arrow lines. area:cli ⇒ no survey.

Radiergummi:
🤖 **[coder]** addressed Moritz's review (commit 6b614b35, pending push):

**1. `--for-env` flag rename.** Verified: Laravel registers a global `--env` option (via `Illuminate\Console\Application::getEnvironmentOption()`, `VALUE_OPTIONAL`, **no** short alias), so defining a command option literally named `env` throws *"An option named 'env' already exists."* — `--env` is impossible. `-e` is free (Symfony's default input definition uses `-h/-q/-v/-V/-n` but not `-e`). **Decision:** keep the descriptive `--for-env` long name and add the requested short form as **`-e`** (`{--e|for-env= ...}`). Added a Unit test asserting `-e production` sets the environment.

**2. Field-provenance rendering.** Replaced the hand-built `←`/`→` echo lines with the idiomatic `$this->components->twoColumnDetail($field, "$value  $source ($reason)")`, and render `supersededBy` candidates via `$this->components->bulletList([...])`. New shape:
```
Fields:
  status ............................ 201 ResourceConventionResolver (store → POST)
  summary ........ Find one flight #[Summary] (method) (author override)
  ⇂ superseded: convention 'Show Flight'
```
Field names no longer carry a trailing colon (`EnsureNoPunctuation` strips it anyway); updated the Feature/Unit assertions to the rendered shape. The component throws a checked `InvalidArgumentException`, added to the `@throws` of `printProvenance`/`printFields`. Pure CLI presentation — generated document untouched (Snapshot invariant holds).

Docs: `getting-started.md` table, `multi-spec.md` worked example (both `-e` note and the two `Fields:` blocks), CHANGELOG `#178` line. Local gates green (Pint, PHPStan L8, 2810 tests).

Radiergummi:
🤖 **[lead]** Moritz's review fixes pushed (6b614b35); full CI matrix GREEN (incl swagger-php 5.8 + byte-exact Snapshot). → review of the delta. Reviewer: code-review the review-fix commit only (flag `-e` alias + twoColumnDetail/bulletList rendering), not the whole PR (already approved). Stays controversial (size + #208) → escalates to Moritz regardless.

Radiergummi:
🤖 **[reviewer]** review (delta `6b614b35`): clean — confirmed, no concerns. (Verified against the worktree; full diff `6e1f2b4f..6b614b35`.)

**1. Flag change (`-e` short alias) — correct.**
- `--env` is genuinely unavailable: Laravel's global `--env` is registered with shortcut `null` (`Illuminate\Console\Application::getEnvironmentOption`, Application.php:331), so it owns the long name with no short form. Adding the `-e` short to the existing `--for-env` (`{--e|for-env=...}`) is the right move.
- **No `-e` collision.** Symfony's reserved short options are `-h`, `-q`, `-v`, `-V`, `-n` (`Application::getDefaultInputDefinition`, Application.php:1187-1193); `-e` is not among them, and no Laravel global claims it. The `-e` alias binds cleanly.
- **Actually exercised:** new test `accepts the -e short alias for --for-env` invokes `openapi:why ... -e production` and asserts `environment: production` — it would fail if the alias didn't resolve. The renamed `--for-env overrides ...` test keeps the long form covered.
- **Consistency:** `$signature`, the in-command comment, CHANGELOG, `docs/getting-started.md`, and `docs/multi-spec.md` all now say `--for-env` (short `-e`) with the `--env`-is-reserved rationale. Decision documented as a coder comment (WhyCommand.php:58-59).
- **Keep `--for-env` + add `-e` vs. rename the long name?** Keeping `--for-env` is the right call on UX: `--env` is impossible (reserved), and `--for-env` reads correctly ("for env X") where a bare `--environment` would invite confusion with the reserved global. `-e` gives the ergonomic short form. No back-compat constraint forced this; it's the better name regardless.

**2. Rendering change (`twoColumnDetail` + `bulletList`) — content faithfully preserved.**
- All provenance content survives: `twoColumnDetail($entry->field, "{$value}  {$source} ({$reason})")` carries field/value/source/reason; superseded candidates move to `bulletList(["superseded: {$candidate}", ...])`. No field dropped.
- **`EnsureNoPunctuation` handled correctly.** The mutator strips a trailing `:`/`.`/`?`/`!` and runs only on the **first** column (`TwoColumnDetail::render`). The new code passes the bare field name (`summary`, `status`, `tags`) with no trailing colon, so the mutator is a no-op here — and the old manually-appended `:` is intentionally gone. The second column ends in `)`, untouched. The Feature/Unit tests were updated in lockstep (`'status:'` → `'status'`, etc.) and exercise the new rendered shape; they'd fail against the old `←`/`str_pad` output.
- **Degrade / flag-off cases intact.** The `misc.arbitrary` test still asserts `str_contains($output, 'summary')` is false; I verified nothing else emits that literal — the inclusion-decision line renders `"included in {spec}"` / `"hidden in environment {env}"` (InclusionEvaluator.php:185), never the word "summary" — so dropping the `:` from the probe doesn't mask a regression; if anything the bare-word probe is stricter. The `(no derived fields)` branch (WhyCommand.php:198) and the no-`--fields` path are untouched.

**3. Conventions + bookkeeping.** Orphaned imports from the removed manual padding (`max`, `str_pad`, `strlen`) are correctly dropped; `@throws InvalidArgumentException` added where `twoColumnDetail`/`bulletList` can throw. CHANGELOG `[Unreleased]` and both docs pages updated for the observable flag + output-shape change. Comment style and 100-char width respected.

Delta is clean on all three axes. Final merge remains Moritz's call (controversial); flagging only that the delta itself introduces no new concerns.

Issue #360, opened 2026-06-15
Re-measure response-schema fidelity residual on current main


Radiergummi:
## Goal

Re-measure how much of each app's **response-schema** surface the generator now covers, on current `main`, and report the concrete shapes that still emit no/thin payloads — so the next response-fidelity inference issue is driven by evidence rather than a stale snapshot.

## Why now

The prior sizing in #106 is stale: it predates the controller return-expression reader. Since then, the response-fidelity levers it tracked have shipped:

- #12 — Eloquent API Resource `toArray()` inference (closed)
- #13 — Fractal `transform()` inference (closed)
- #14 — `response()->json([...])` literal inference (closed)
- `Plugins\ApiResources\Support\ReturnExpressionResourceReader` — resolves *which* Resource/model a return wraps from the body (`X::collection(...)`, `X::make(...)`, `new X(...)`, `->toResource(X::class)`/`->toResourceCollection(X::class)`, base `new JsonResource($model)` via `#[UseResource]`/`guessResourceName()`, the two-statement assign-then-return form, paginated detection)

So #106's "under-tracked lever-2" is largely closed. We don't currently know the *true* residual.

## Method

Black-box, on the pinned 11-app dogfood corpus (no app code changed, no app-specific plugins):

1. For each route, reflect the controller action and categorize its declared/return-expression shape.
2. Cross-reference the generated spec for whether a **substantive** 2xx response schema was emitted (resolves through `$ref` + `{data: …}` envelope unwrap to ≥1 property; thin empty-`$ref` envelopes count as *not* substantive).
3. Report per-app substantive-response coverage and the **residual shape buckets** that still emit nothing/thin — e.g. base `JsonResource` wrapping a value of unknown model type, variable-/conditional-selected resources (Tier-2, refused), custom envelope/`->additional()` merges.

## Output

A residual breakdown that distinguishes (a) shapes reachable by a bounded Tier-1 addition from (b) genuinely Tier-2/authoring-attribute territory. The reachable slice becomes the next concrete inference issue.

Supersedes the gap-measurement aspect of #106 (kept for history).

Radiergummi:
[...]

PR #392, opened 2026-06-17
feat(responses): infer collection schema for X::collect(...) resource factory

Radiergummi:
### Tier

**Tier 1** — a bounded whitelist addition to the existing return-expression matcher in
`ReturnExpressionResourceReader`. No new body parsing, no variable tracking beyond what the reader
already does: `X::collect(...)` is matched by call shape exactly like `X::collection(...)`. This is
the lowest tier that captures the idiom; it carries the same risk profile as the `::collection()`
shape the reader already matches.

### Root cause / approach

`X::collect(...)` is Laravel's static resource-collection factory (added in Laravel 11.36+); it
produces an `AnonymousResourceCollection` of `X` exactly as `X::collection(...)` does — the only
difference is that `collect()` also wraps a raw value in a `Collection` first. For our purposes the
two are identical: read the static-call class, treat the result as a collection of that resource,
and derive pagination from the argument the same way.

The reader is purely AST-based (it matches the *call name* on a class that is a concrete
`JsonResource` subclass; it never reflects on whether the method exists at runtime), so the match
works regardless of the vendored framework version. Note for the coder: this repo's vendored
Laravel (13.15.0) does **not** define `JsonResource::collect()`, so the fixture actions must remain
**parse-only / never invoked** — which is already the established convention in both touched test
files (`ReturnExpressionController` and `ReaderFixtureController` carry that note).

### Files to modify

1. **`src/Plugins/ApiResources/Support/ReturnExpressionResourceReader.php`**
   - `targetFromStaticCall()`: add `'collect'` to the whitelisted static method names
     (`in_array($methodName, ['collection', 'make'], true)` → add `'collect'`), and treat
     `collect` identically to `collection` for the collection/pagination branch (the existing
     `if ($methodName === 'collection')` branch becomes
     `if (in_array($methodName, ['collection', 'collect'], true))`).
   - `paginatedFromBody()`: the `->collection()` static-call probe that derives pagination from the
     argument must also accept `collect` (the `$expression->name->toLowerString() === 'collection'`
     check becomes an `in_array(..., ['collection', 'collect'], true)`), so a
     `@return Collection<X>` docblock + `X::collect($q->paginate())` body still yields the paginated
     envelope.
   - Update the class docblock's recognised-shapes list and the `targetFromExpression()` refusal
     note string to mention `X::collect(...)` alongside `X::collection(...)`.

### Tests (TDD — failing first)

[...]

PR #393, opened 2026-06-17
fix(responses): resolve API-Resource schema for untyped controller returns

### LINKED ISSUE #391: bug(responses): untyped-return API-Resource actions emit no content (resolution gate skips body scan; 99 routes, AureusERP/Koel)
## Summary

On a meaningful set of routes the API-Resource response **resolves to the correct resource class**
but the generated 2xx schema is an **empty object `{}`** — despite the resource having a clean,
literal `toArray()` that the generator should be able to read. The resource is found; the property
shape is dropped.

## Evidence

From the #360 response-fidelity re-measure (current `main`, 11-app corpus —
https://github.com/Radiergummi/laravel-openapi/issues/360#issuecomment-4730519836): **99 routes**
exhibit this, strongly **app-clustered** — **AureusERP (52)** and **Koel (34)** dominate, while
Vito/Lychee resources with similar shapes emit their property schemas fine. The clustering says this
is a real defect triggered by some structural property of those apps' resources, **not** a
Tier-2 refusal (the class resolves) and **not** correct no-content.

## Why it's a bug, not a gap

The two refusal/again categories are accounted for separately in #360 (untyped multi-return and
service-builder families are correctly-refused Tier-2). This bucket is distinct: resolution
*succeeds* and the `toArray()` is a literal array, so a substantive schema is expected. Emitting `{}`
is a silent under-emission.

## Investigation starting points

- Diff an AureusERP/Koel resource that emits `{}` against a Vito/Lychee resource that emits
  properties — what differs (trait-provided `toArray()`, parent-class `toArray()`, conditional
  `when()`/`mergeWhen()` usage, non-literal array construction, `$this->resource` access patterns)?
- Trace `ReturnExpressionResourceReader` → the `toArray()` literal reader → schema synthesis for one
  failing case; find where the property list collapses to empty.

## Acceptance criteria

- Root cause identified and a regression fixture built from a minimised failing resource.
- The fix restores property emission for the affected shape without regressing the apps that already
  emit correctly (verify via the survey corpus: AureusERP/Koel coverage up, Vito/Lychee unchanged).
- CHANGELOG `[Unreleased]`.

## Notes

`bug`, `area:responses` / `area:plugins`. Surfaced by #360.

Radiergummi:
## Root-cause investigation (read this first — the symptom has changed)

**The `{}` symptom is NOT reproducible on current `main`.** I freshly regenerated all 11 corpus
specs against the live `$LIB` symlink (HEAD `3791a751`, php@8.4) and scanned every 2xx
`application/json` schema:

| app | empty `{}` 2xx | substantive json 2xx | bare-200 (no json) 2xx |
|-----|----:|----:|----:|
| AureusERP | **0** | 14 | 146 |
| Koel | **0** | 17 | 277 |
| Vito | 0 | 151 | unchanged |
| Lychee | 0 | 87 | unchanged |
| *(all 11 apps)* | **0** | — | — |

Zero empty-object 2xx schemas anywhere in the corpus. The #360 re-measure snapshot that reported 99
`{}` routes predates intervening response-fidelity work; on today's code those same AureusERP/Koel
routes emit **no content at all** (bare 200), not `{}`. So the issue's stated hypothesis — *"the
resource resolves but the property shape is dropped"* — is **wrong against current `main`**: the
resource never resolves, so synthesis (`ResourceToArrayReader`) is never reached. **The fix is not
in schema synthesis.**

### Why these routes under-emit (the real, current root cause)

AureusERP and Koel API controllers **do not type their return values** and lean on Scribe's
`#[ResponseFromApiResource]` attribute (which this library does not read):

- AureusERP: **0 of 277** API controller methods declare a return type. e.g. `TitleController::show()`
  → `return new TitleResource($title);` (no `: JsonResource`), `index()` →
  `return TitleResource::collection($titles);` (no return type).
- Koel `AlbumController`: `index()` → `AlbumResource::collection(...)`, `show()` →
  `AlbumResource::make(...)`, `update()` → `AlbumResource::make(...)` — **all untyped**.
- Vito/Lychee, by contrast, **type their returns** (`index(): Response`, `json(): ResourceCollection`),
  which is exactly why their resource schemas emit.

`ResourceClassLocator::locate()` (`src/Plugins/ApiResources/Support/ResourceClassLocator.php:75`)
bails out — `return null` — whenever the method has no non-builtin return type **and** no
`#[ResponseResource]`. The body-scan reader (`ReturnExpressionResourceReader`, which already resolves
`new X(...)`, `X::make(...)`, `X::collection(...)`) is **only** consulted on the two typed paths
(line 99 for collection return types, line 105 for a bare `JsonResource` return type). So untyped
methods returning a resource never reach the scan. This gate was introduced deliberately by #108
(PR #253); broadening it is the fix.

### Tier

**Tier 1.** The reader is already a bounded body scan; this change only *widens when it runs* to
include untyped methods. No new parsing capability, no variable tracking beyond what exists. The
reader's existing graceful refusal (returns null + one NOTICE per action) means untyped methods that
return non-resources — `destroy()` → `response()->json([...])`, redirects, etc. — are rejected by
shape, so broadening cannot mis-resolve them into bogus schemas. **Not Tier 2** — no dataflow needed;
the return expression is read directly.

## Proposed fix

In `ResourceClassLocator::locate()`, when there is no `#[ResponseResource]` and the return type is
absent/builtin (the current line-75 bail), **first attempt the return-expression reader** before
returning null:

```php
if (!$returnType instanceof ReflectionNamedType || $returnType->isBuiltin()) {
    return $this->locateFromReturnExpression($descriptor);   // null when the body isn't a resource shape
}
```

`locateFromReturnExpression()` already exists and already returns null when `$descriptor->method` is
null or the body doesn't match a whitelisted shape, so non-resource untyped methods stay
no-content exactly as today. The `isCollection`/`paginated` flags come from the reader's
`ResourceTarget` (it derives pagination from a `paginate()`-family argument), so Koel's
`AlbumResource::collection($repo->getForListing(...))` lands as an unpaginated collection and
AureusERP's `TitleResource::collection($qb->paginate())` as a paginated one — matching how the typed
path already behaves.

### Files to modify

1. **`src/Plugins/ApiResources/Support/ResourceClassLocator.php`** — replace the line-75 unconditional
   bail with `return $this->locateFromReturnExpression($descriptor, silent: true);` (new branch,
   parallel to and after the `#[ResponseResource]` branch at L64-73, so attribute precedence is
   preserved and the typed base-Resource branches at L85-110 are untouched). Add the `bool $silent`
   parameter to `locateFromReturnExpression()`. Update the class docblock's resolution-order
   sentence to say the body scan also runs for **untyped** actions.
2. **`src/Plugins/ApiResources/Support/ReturnExpressionResourceReader.php`** — thread `bool $silent`
   from `read()` → `resolve()` → all `note()` call sites (and the existing `$log` flags) so the
   untyped-path fall-through can suppress the refusal NOTICE without affecting the typed call sites.
   See "Noise consideration" below for why the planned one-line `log: false` mirror is insufficient.

### Noise consideration — silencing mechanism (plan-review correction)

Broadening means the reader now runs on **every untyped controller action**, logging one NOTICE per
action it can't resolve (memoised once per run). On Koel/AureusERP that is hundreds of non-resource
methods (`destroy` → `response()->json([...])`, redirects, scalar returns). Decision (maintainer):
**silent fallback on the untyped path only** — keep notices on the typed base-Resource paths (#108's
diagnostic value), suppress them when the fall-through is from an absent/builtin return type.

**Plan-review correction — the "mirror `paginatedFromBody` `log: false`" framing is wrong and the
coder must not follow it as written.** Verified against `ReturnExpressionResourceReader`: the `log`
flag exists **only** on `canonicalReturnExpression()` and `expressionAssignedTo()`. The refusal
notices that fire for the high-volume untyped non-resource case live in `targetFromExpression()`
(L509 "is chained into ->json()…", L520 "does not match a recognised resource-returning shape"),
`targetFromStaticCall()`, `targetFromNewExpression()`, and `targetFromTransformCall()` — **none of
which take a `log` flag**. So threading `log: false` through the two methods that have it would
**still** flood NOTICEs for `destroy()`/scalars/redirects, and the planned negative test
(`destroy()` "must not emit a refusal notice") would be **unsatisfiable**.

Correct mechanism: thread a `bool $silent` (defaulting false) from a new
`ResourceClassLocator::locateFromReturnExpression($descriptor, silent: true)` on the **untyped
branch only** → `ReturnExpressionResourceReader::read($method, silent: true)` → `resolve()` → into
`note()` (and the existing `$log` flags). When `$silent`, `note()` either no-ops or drops to
`debug`. The typed-path call sites (L99, L105) pass `silent: false`, so #108's notices are
unchanged. This is a few more touched lines than the plan implies; keep the threaded flag minimal
and defaulted so the typed paths and lint-rule call sites are untouched by default.

### Lint-rule impact (plan-review addition)

`ResourceClassLocator` is shared by `ResourceResponseResolver` **and four lint rules**
(`ResourceResponseAmbiguous`, `ResourceResponseEmpty`, `ResourceFieldTypeMissing`,
`ResourceFieldsUndeclared`). Broadening `locate()` changes what these rules see for untyped actions:
an untyped action that now resolves to `ResourceTarget(resourceClass: null, modelClass: …)` (the
`new JsonResource($model)` wrapped-model shape) or to a concrete target will newly trigger / change
`resource.*` findings where it previously produced none. The coder must confirm the **lint** corpus
delta (run `openapi:lint` on the fixtures and ideally a corpus app), not just the spec delta, and
add a unit assertion that an untyped non-resource action still yields `null` (no spurious
`resource.response-ambiguous`). Note: the `silent` flag is for the **NOTICE log only** and must not
change `locate()`'s return value, so lint behaviour is driven purely by resolution, not by silencing.

## Tests (TDD — failing first)

**Feature** — new `tests/Feature/Plugins/ApiResources/UntypedReturnResolutionTest.php` (parse-only
fixture controller, mirroring `ReturnExpressionController` but with **no return types**):
- `index()` → `return FooResource::collection($q->paginate());` (no return type) → paginated
  `{data,links,meta}` collection schema with `data.items.$ref` → `FooResource`, fields inferred.
- `indexUnpaginated()` → `FooResource::collection($models)` (no paginate, no return type) → plain
  `{data}` envelope.
- `show()` → `return new FooResource($model);` (no return type) → single `{data}` schema with
  resolved fields.
- `make()` → `FooResource::make($model)` (no return type) → single resource schema.
- **Negative / degrade:** `destroy()` → `return response()->json(['ok' => true]);` (no return type)
  → must stay **no content / no spurious resource schema** and must **not** emit a refusal notice
  (asserts the silent-fallback decision); `untypedConditional()` with two differing returns → refuses
  cleanly; `untypedScalar()` → `return 1;` → no resource.

**Unit** — extend `tests/Unit/Plugins/ApiResources/ResourceClassLocatorTest.php`:
- An untyped action returning a recognised resource shape resolves to the right `ResourceTarget`.
- An untyped action returning a non-resource yields `null` (no resolution).
- `#[ResponseResource]` and the typed paths are unchanged (existing cases stay green).

### Minimised regression fixture

A new resource `tests/Fixtures/Resources/UntypedReturnResource.php` (clean literal `toArray()` with
a couple of `$this->resource->field` / literal fields) plus the parse-only controller above. This
reproduces the AureusERP/Koel pattern (untyped method → resource body) in one file.

## Acceptance criteria mapping

- Root cause identified ✔ (locator gate, not synthesis) and minimised regression fixture ✔.
- Restores emission for untyped-return resource actions; Vito/Lychee already emit and are untouched
  by this path (they hit the typed branches) → **survey gate must confirm AureusERP/Koel json-2xx up,
  Vito/Lychee unchanged**.
- CHANGELOG `[Unreleased]` → `Fixed`: "ApiResources: actions that return an API Resource without
  declaring a return type (relying on convention or third-party doc attributes) now resolve their
  response schema via the return-expression reader, instead of emitting no content." (Closes #391.)
- Docs: `docs/plugins.md` ApiResources resolution-order note — mention untyped actions are now scanned.

## Controversy / escalation flag — NEEDS MAINTAINER DECISION

Two things make this **not** a clean auto-merge:

1. **The issue's symptom is stale.** It describes a `{}` synthesis bug; the actual current defect is
   no-content from a resolution gate. The fix lives in a different file than the issue points to.
   The maintainer should confirm the re-scope (fix under-emission via the locator) is acceptable, and
   the issue body/title should be updated to match — or the issue closed as "symptom changed" and a
   fresh one filed. **I did not silently plan a different feature; flagging for the lead.**
2. **It broadens when the body scan runs to every untyped action** — a behavioural expansion that
   touches resolution for the whole corpus, with a stderr-noise tradeoff (the silent-vs-notice
   decision above). Generation-affecting, `area:responses`/`area:plugins` → **survey gate applies**,
   and the noise/severity decision is a judgement call worth a human glance.

No `src/Contracts/**` change, no stage-order change, no config key, no public-signature change —
but the scope-divergence (1) is the reason this should pass through plan-review and likely a
maintainer ack before coding.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -60,6 +60,7 @@ All notable changes to this project are documented here.
 - `#[AllowedFilter(nullable: true)]` widens the generated `filter[…]` schema to a nullable type array.
 - `--path` / `--diff` no longer leak findings for out-of-scope routes (#50).
 - `ValidationRulesToSchema` maps additional rules (`multiple_of`, Email/Exists/Unique/NotIn objects, wildcard keys) it previously dropped (#83).
+- ApiResources: actions that return an API Resource without declaring a return type (relying on convention or a third-party doc attribute) now resolve their response schema via the return-expression reader, instead of emitting no content; the refusal notice is suppressed on the untyped path to avoid a notice per non-resource action. (#391)
 
 ### Documentation
 - README refreshed for the current feature set (Tier-1 method-body inference, SwaggerPhp plugin) and trimmed to a gentle overview that defers detail to `docs/`.
--- a/docs/plugins.md
+++ b/docs/plugins.md
@@ -144,9 +144,10 @@ degrades to a permissive `items: { type: object }`.
 
 Many actions type their return as a **base** resource class — `JsonResource`, a
 bare `ResourceCollection`, or `AnonymousResourceCollection` — and name the
-concrete resource only in the method body. When the signature yields nothing
-concrete, the generator reads the method's **return expression** (a bounded scan
-of the first 10 statements; no dataflow):
+concrete resource only in the method body. Others declare **no return type at
+all** (relying on convention or a third-party doc attribute). In both cases the
+signature yields nothing concrete, so the generator reads the method's **return
+expression** (a bounded scan of the first 10 statements; no dataflow):
 
 ```php
 public function index(): AnonymousResourceCollection
@@ -192,7 +193,9 @@ since attributes carry no body context.
 Anything else — a conditional return, a variable of unknown origin, an
 unrecognised chained call, a receiver that would need dataflow — degrades to the
 previous behaviour with a generation-log note; `#[ResponseResource]` is the
-escape hatch and always wins.
+escape hatch and always wins. On the **untyped** path that note is suppressed:
+the scan runs on every untyped action and most are not resources, so a notice
+per non-resource would be pure noise. The base-resource paths keep their notes.
 
 ### Collection endpoints
 
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -99,11 +99,15 @@ parameters:
                 - tests/Fixtures/Lint/
         # The ReturnTypeNudgeController fixture intentionally omits return types — the missing
         # return type is exactly what the `operation.return-type-missing` lint rule detects, so
-        # adding types would defeat the fixture.
+        # adding types would defeat the fixture. The untyped-return resolution fixtures reproduce
+        # the AureusERP/Koel pattern (resource returned from an untyped action), where the missing
+        # return type is the input under test.
         -
             identifier: missingType.return
             paths:
                 - tests/Fixtures/Lint/ReturnTypeNudgeController.php
+                - tests/Feature/Plugins/ApiResources/UntypedReturnResolutionTest.php
+                - tests/Unit/Plugins/ApiResources/ResourceClassLocatorTest.php
         # Attribute-misuse fixtures intentionally trip the bundled PHPStan rules so the linter
         # / generator tests have something to consume. The static rules report the misuse for
         # everyone else; here the misuse is the test input. Identifiers use camelCase because
--- a/src/Plugins/ApiResources/Support/ResourceClassLocator.php
+++ b/src/Plugins/ApiResources/Support/ResourceClassLocator.php
@@ -29,9 +29,10 @@ use function is_string;
  * {@see ResourceResponseResolver} and the ApiResources lint rules so resolution is defined once.
  *
  * Resolution order: `#[ResponseResource]`, then concrete return type (or `#[Collects]` /
- * `$collects`). Only when the signature names a base resource type does
- * {@see ReturnExpressionResourceReader} inspect the return expression; refusal leaves the
- * endpoint ambiguous.
+ * `$collects`). When the signature names a base resource type, or declares no resource return type
+ * at all (absent/builtin), {@see ReturnExpressionResourceReader} inspects the return expression;
+ * refusal leaves the endpoint ambiguous. On the untyped path the refusal notice is suppressed,
+ * since the reader then runs on every action and most are not resources.
  */
 final readonly class ResourceClassLocator implements ResourceTargetLocator
 {
@@ -73,7 +74,10 @@ final readonly class ResourceClassLocator implements ResourceTargetLocator
         }
 
         if (!$returnType instanceof ReflectionNamedType || $returnType->isBuiltin()) {
-            return null;
+            // Untyped actions (no resource return type) still commonly return a resource by
+            // convention. Consult the body scan; a non-resource shape reads back as null and
+            // keeps today's no-content behaviour, silently to avoid a notice per non-resource.
+            return $this->locateFromReturnExpression($descriptor, silent: true);
         }
 
         $name = $returnType->getName();
@@ -194,12 +198,14 @@ final readonly class ResourceClassLocator implements ResourceTargetLocator
     /**
      * @throws ReflectionException
      */
-    private function locateFromReturnExpression(ActionDescriptor $descriptor): ?ResourceTarget
-    {
+    private function locateFromReturnExpression(
+        ActionDescriptor $descriptor,
+        bool $silent = false,
+    ): ?ResourceTarget {
         if ($descriptor->method === null) {
             return null;
         }
 
-        return $this->returnExpressionReader->read($descriptor->method);
+        return $this->returnExpressionReader->read($descriptor->method, silent: $silent);
     }
 }
--- a/src/Plugins/ApiResources/Support/ReturnExpressionResourceReader.php
+++ b/src/Plugins/ApiResources/Support/ReturnExpressionResourceReader.php
@@ -117,9 +117,12 @@ final class ReturnExpressionResourceReader
     }
 
     /**
+     * @param bool $silent Suppress the refusal notice (untyped path: the scan runs on every action,
+     *                     most of which are not resources). Never affects the resolved target.
+     *
      * @throws ReflectionException
      */
-    public function read(ReflectionMethod $method): ?ResourceTarget
+    public function read(ReflectionMethod $method, bool $silent = false): ?ResourceTarget
     {
         $key = $method->getDeclaringClass()->getName() . '::' . $method->getName();
 
@@ -127,13 +130,13 @@ final class ReturnExpressionResourceReader
             return $this->cache[$key];
         }
 
-        return $this->cache[$key] = $this->resolve($method);
+        return $this->cache[$key] = $this->resolve($method, $silent);
     }
 
     /**
      * @throws ReflectionException
      */
-    private function resolve(ReflectionMethod $method): ?ResourceTarget
+    private function resolve(ReflectionMethod $method, bool $silent): ?ResourceTarget
     {
         $documented = $this->targetFromReturnTag($method);
 
@@ -142,21 +145,21 @@ final class ReturnExpressionResourceReader
         }
 
         $statements = $this->scanner->firstStatements($method, self::STATEMENT_LIMIT);
-        $returnExpression = $this->canonicalReturnExpression($statements, $method);
+        $returnExpression = $this->canonicalReturnExpression($statements, $method, log: !$silent);
 
         if ($returnExpression === null) {
             return null;
         }
 
         $expression = $returnExpression instanceof Variable
-            ? $this->expressionAssignedTo($returnExpression, $statements, $method)
+            ? $this->expressionAssignedTo($returnExpression, $statements, $method, log: !$silent)
             : $returnExpression;
 
         if ($expression === null) {
             return null;
         }
 
-        return $this->targetFromExpression($expression, $method);
+        return $this->targetFromExpression($expression, $method, $silent);
     }
 
     // region Return-expression location
@@ -315,8 +318,12 @@ final class ReturnExpressionResourceReader
         return $topLevelReturn->expr;
     }
 
-    private function note(ReflectionMethod $method, string $reason): void
+    private function note(ReflectionMethod $method, string $reason, bool $silent = false): void
     {
+        if ($silent) {
+            return;
+        }
+
         $this->logger->notice(
             sprintf(
                 'The return expression of %s::%s %s; the concrete resource stays unresolved.'
@@ -473,16 +480,17 @@ final class ReturnExpressionResourceReader
     private function targetFromExpression(
         Expr $expression,
         ReflectionMethod $method,
+        bool $silent,
     ): ?ResourceTarget {
         $current = $expression;
 
         while (true) {
             if ($current instanceof StaticCall) {
-                return $this->targetFromStaticCall($current, $method);
+                return $this->targetFromStaticCall($current, $method, $silent);
             }
 
             if ($current instanceof NewExpression) {
-                return $this->targetFromNewExpression($current, $method);
+                return $this->targetFromNewExpression($current, $method, $silent);
             }
 
             if ($current instanceof MethodCall && $current->name instanceof Identifier) {
@@ -503,6 +511,7 @@ final class ReturnExpressionResourceReader
                         $current,
                         $methodName,
                         $method,
+                        $silent,
                     );
                 }
 
@@ -512,6 +521,7 @@ final class ReturnExpressionResourceReader
                         'is chained into ->%s(), which the scan does not recognise',
                         $current->name->toString(),
                     ),
+                    $silent,
                 );
 
                 return null;
@@ -522,6 +532,7 @@ final class ReturnExpressionResourceReader
                 'does not match a recognised resource-returning shape '
                 . '(X::collection(...), X::collect(...), X::make(...), new X(...), '
                 . '$model->toResource())',
+                $silent,
             );
 
             return null;
@@ -538,6 +549,7 @@ final class ReturnExpressionResourceReader
     private function targetFromStaticCall(
         StaticCall $call,
         ReflectionMethod $method,
+        bool $silent,
     ): ?ResourceTarget {
         $resourceClass = $call->class instanceof Name
             ? $this->concreteResourceClass($call->class->toString())
@@ -551,6 +563,7 @@ final class ReturnExpressionResourceReader
             $this->note(
                 $method,
                 'is a static call that does not name a concrete JsonResource subclass',
+                $silent,
             );
 
             return null;
@@ -576,9 +589,10 @@ final class ReturnExpressionResourceReader
     private function targetFromNewExpression(
         NewExpression $new,
         ReflectionMethod $method,
+        bool $silent,
     ): ?ResourceTarget {
         if (!$new->class instanceof Name) {
-            $this->note($method, 'instantiates a dynamically-resolved class');
+            $this->note($method, 'instantiates a dynamically-resolved class', $silent);
 
             return null;
         }
@@ -595,6 +609,7 @@ final class ReturnExpressionResourceReader
                 $this->note(
                     $method,
                     'wraps a value of unknown model type in the base JsonResource',
+                    $silent,
                 );
 
                 return null;
@@ -613,6 +628,7 @@ final class ReturnExpressionResourceReader
             $this->note(
                 $method,
                 'instantiates a class that is not a concrete JsonResource subclass',
+                $silent,
             );
 
             return null;
@@ -662,6 +678,7 @@ final class ReturnExpressionResourceReader
         MethodCall $call,
         string $methodName,
         ReflectionMethod $method,
+        bool $silent,
     ): ?ResourceTarget {
         $arguments = $call->getArgs();
 
@@ -675,6 +692,7 @@ final class ReturnExpressionResourceReader
                         'passes a non-literal or non-resource class to ->%s()',
                         $call->name instanceof Identifier ? $call->name->toString() : $methodName,
                     ),
+                    $silent,
                 );
 
                 return null;
@@ -692,7 +710,11 @@ final class ReturnExpressionResourceReader
         }
 
         if ($methodName === 'toresourcecollection') {
-            $this->note($method, 'calls ->toResourceCollection() without naming the resource class');
+            $this->note(
+                $method,
+                'calls ->toResourceCollection() without naming the resource class',
+                $silent,
+            );
 
             return null;
         }
@@ -702,7 +724,11 @@ final class ReturnExpressionResourceReader
             : null;
 
         if ($modelClass === null) {
-            $this->note($method, 'calls ->toResource() on a receiver whose model type is not statically knowable');
+            $this->note(
+                $method,
+                'calls ->toResource() on a receiver whose model type is not statically knowable',
+                $silent,
+            );
 
             return null;
         }
@@ -716,6 +742,7 @@ final class ReturnExpressionResourceReader
                     'calls ->toResource() on %s, but no conventional resource class could be resolved for it',
                     $modelClass,
                 ),
+                $silent,
             );
 
             return null;
--- a/tests/Feature/Plugins/ApiResources/UntypedReturnResolutionTest.php
+++ b/tests/Feature/Plugins/ApiResources/UntypedReturnResolutionTest.php
@@ -1,0 +1,229 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Tests\Feature\Plugins\ApiResources;
+
+use Illuminate\Routing\Controller;
+use Illuminate\Support\Facades\Route;
+use Psr\Log\LoggerInterface;
+use Radiergummi\OpenApi\Plugins\ApiResources\Lint\Rules\ResourceResponseAmbiguous;
+use Radiergummi\OpenApi\Plugins\ApiResources\Support\ResourceClassLocator;
+use Radiergummi\OpenApi\Tests\Fixtures\Models\Author;
+use Radiergummi\OpenApi\Tests\Fixtures\Resources\UntypedReturnResource;
+use Radiergummi\OpenApi\Tests\Support\ActionDescriptorFactory;
+use Radiergummi\OpenApi\Tests\Support\OperationNodeFactory;
+
+use function array_any;
+use function iterator_to_array;
+use function random_int;
+use function str_contains;
+
+uses()->group('openapi', 'plugin:api-resources');
+
+/**
+ * Reproduces the AureusERP/Koel pattern: actions return an API Resource but declare no return type
+ * (relying on convention or a third-party doc attribute the library does not read). The concrete
+ * resource is only recoverable from the method body. Actions are never invoked, only parsed.
+ */
+class UntypedReturnController extends Controller
+{
+    public function index()
+    {
+        return UntypedReturnResource::collection(Author::query()->paginate());
+    }
+
+    public function indexUnpaginated()
+    {
+        return UntypedReturnResource::collection(Author::all());
+    }
+
+    public function show()
+    {
+        return new UntypedReturnResource(Author::query()->firstOrFail());
+    }
+
+    public function make()
+    {
+        return UntypedReturnResource::make(Author::query()->firstOrFail());
+    }
+
+    public function destroy()
+    {
+        return response()->json(['ok' => true]);
+    }
+
+    public function untypedConditional(bool $flag)
+    {
+        if ($flag) {
+            return UntypedReturnResource::make(Author::query()->firstOrFail());
+        }
+
+        return new UntypedReturnResource(Author::query()->firstOrFail());
+    }
+
+    public function untypedScalar()
+    {
+        return random_int(0, 1);
+    }
+}
+
+/**
+ * @param array<string, mixed> $spec
+ *
+ * @return array<string, mixed>
+ */
+function untypedSuccessSchema(array $spec, string $path): array
+{
+    $schema = $spec['paths'][$path]['get']['responses']['200']['content']['application/json']['schema'] ?? null;
+
+    expect($schema)->not->toBeNull();
+
+    return $schema;
+}
+
+// region Resolved shapes
+
+it('resolves a paginated collection from an untyped index() action', function (): void {
+    Route::get('/untyped-authors', [UntypedReturnController::class, 'index']);
+
+    $spec = generateSpec();
+    $schema = untypedSuccessSchema($spec, '/untyped-authors');
+
+    expect($schema['properties'])->toHaveKeys(['data', 'links', 'meta'])
+        ->and($schema['properties']['data']['type'])->toBe('array')
+        ->and($schema['properties']['data']['items']['$ref'])->toBe('#/components/schemas/UntypedReturnResource')
+        ->and($spec['components']['schemas']['UntypedReturnResource']['properties'])->toHaveKeys(['id', 'title']);
+});
+
+it('resolves a plain {data} envelope from an untyped unpaginated collection', function (): void {
+    Route::get('/untyped-authors-unpaginated', [UntypedReturnController::class, 'indexUnpaginated']);
+
+    $schema = untypedSuccessSchema(generateSpec(), '/untyped-authors-unpaginated');
+
+    expect($schema['properties'])->toHaveKey('data')
+        ->and($schema['properties'])->not->toHaveKeys(['links', 'meta'])
+        ->and($schema['properties']['data']['items']['$ref'])->toBe('#/components/schemas/UntypedReturnResource');
+});
+
+it('resolves a single resource from an untyped show() returning new X()', function (): void {
+    Route::get('/untyped-author', [UntypedReturnController::class, 'show']);
+
+    $schema = untypedSuccessSchema(generateSpec(), '/untyped-author');
+
+    expect($schema['properties']['data']['$ref'])->toBe('#/components/schemas/UntypedReturnResource');
+});
+
+it('resolves a single resource from an untyped X::make() action', function (): void {
+    Route::get('/untyped-author-make', [UntypedReturnController::class, 'make']);
+
+    $schema = untypedSuccessSchema(generateSpec(), '/untyped-author-make');
+
+    expect($schema['properties']['data']['$ref'])->toBe('#/components/schemas/UntypedReturnResource');
+});
+
+// endregion
+
+// region Negatives & silent degradation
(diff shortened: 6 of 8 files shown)


Radiergummi:
🤖 **[planner]** root-cause investigation done — plan posted. **Flagging for maintainer scope decision before plan-review.**

**The `{}` symptom is not reproducible on current `main`.** Freshly regenerated all 11 corpus specs against the live `$LIB` symlink (HEAD 3791a751, php@8.4): **zero** empty-object 2xx schemas anywhere. The 99 `{}` routes from the #360 snapshot now emit **no content** (bare 200), not `{}` — intervening response-fidelity work changed the symptom.

**Real current root cause:** AureusERP (0/277 methods typed) and Koel API controllers don't declare return types and rely on Scribe's `#[ResponseFromApiResource]` (unread by us). `ResourceClassLocator::locate()` bails at `:75` when there's no return type + no `#[ResponseResource]`, so the body-scan reader — which already resolves `new X()`, `X::make()`, `X::collection()` — is never consulted. Vito/Lychee emit fine precisely because they *type* their returns. Fix is a Tier-1 broadening of the locator gate, **not** in `ResourceToArrayReader` synthesis.

Two items need a human call (detailed in PR body): (1) the issue's stated symptom/file is stale — confirm the re-scope or refile; (2) broadening runs the scan on every untyped action (stderr-noise tradeoff; I recommend silent fallback on the untyped path). Generation-affecting → survey gate applies.

Radiergummi:
🤖 **[lead]** root-cause investigation is strong but surfaces a scope decision (issue symptom stale; fix broadens the #108/#253 locator gate). Holding for maintainer scope ack before coding.

Radiergummi:
🤖 **[planner]** independent corroboration of this plan's root cause.

A second planner instance was (re)dispatched #391 and investigated from scratch — different method (static read of `ResourceToArrayReader` + reading the actual AureusERP/Koel/Vito/Lychee resource sources in the corpus). I independently re-ran the empirical check on the freshly-generated specs and confirm this PR's central claim:

| app | empty `{}` 2xx | substantive json 2xx | no-json 2xx |
|-----|----:|----:|----:|
| AureusERP | **0** | 14 | 146 |
| Koel | **0** | 17 | 277 |
| Vito | 0 | 151 | 306 |
| Lychee | 0 | 87 | 188 |

Zero empty-`{}` schemas; the AureusERP/Koel deficit is **no-content**, not `{}`. So the issue's premise ("resource resolves but the shape is dropped") is stale, and a `ResourceToArrayReader` synthesis fix would be dead code on this corpus — the resource never resolves. The locator untyped-return gate (`ResourceClassLocator::locate()`) is the real lever, exactly as this plan says.

One nuance worth recording for the coder: AureusERP's empty-cluster resources also use the `array_merge(parent::toArray($request), [literal])` idiom, which `SingleReturnArrayLiteralFinder` can't read — so even once the locator broadens, those *specific* subclasses will hit the wrapped-model fallback or a sparse schema unless `array_merge(parent::toArray())` is also taught. That's a **separate, smaller follow-up** (not in scope here), but the survey-gate delta for AureusERP may come in lower than the route count suggests because of it. Flagging so the survey result isn't misread as a regression.

Endorse this plan as-is; the maintainer-decision flag (stale issue symptom) is correct and should gate.

Radiergummi:
🤖 **[lead]** Maintainer scope decision (Moritz, via the team lead): **re-scope & implement**, per the planner's investigation.

Confirmed the original gate's rationale by reading #108 / PR #253: the body-scan reader was deliberately consulted **only when the signature declares a *base* Resource type** (`AnonymousResourceCollection` / bare `ResourceCollection` / exactly-base `JsonResource`) — #106 sized that at the 107 ops whose concrete type is one `::collection()` call away. **Untyped methods (no return type at all) were explicitly outside #108's scope.** So this is a real, deliberate expansion of that seam, not just a noise toggle.

Decision: broaden `ResourceClassLocator` to also consult the reader on the **untyped / builtin-return** path (new branch parallel to the base-Resource branch), with **silent fallback on the untyped path** (no NOTICE flood across hundreds of non-resource untyped actions), per the planner's recommendation. Issue #391 retitled to match the real defect.

→ plan-review (reviewer1) before coding — Full tier, generation-affecting, survey-gated.

Radiergummi:
🤖 **[reviewer]** plan-review: changes-requested (approach sound; amended the plan body). **The broadening cannot regress #108 — verified structurally — but the silencing mechanism as written is wrong and the negative test would be unsatisfiable.** Fixed both in the PR body; one item still needs the coder's empirical confirmation.

Pressure-tested the four axes the lead flagged against the actual code:

**1. Branch placement — CORRECT, #108 contract structurally safe.** Read `ResourceClassLocator::locate()`. The `#[ResponseResource]` branch (L64-73) sits *before* the line-75 bail, so attribute precedence is preserved. The new branch is a strictly-new early return on the previously-`null` untyped/builtin path (L75) — it returns before reaching the base-Resource branches (collection L85-101, bare `JsonResource` L103-110), which only execute when there *is* a non-builtin `JsonResource` return type. So #108/#253's deliberate base-Resource gate is untouched; this is a parallel branch, not an entanglement. ✔

**2. Refusal robustness — SOUND; planned negatives sufficient.** Read `ReturnExpressionResourceReader`. Refusal-by-shape is exhaustive: `canonicalReturnExpression` refuses on no-return or >1 method-level return (covers `untypedConditional`); `targetFromExpression` resolves only `StaticCall`/`New_`/`toResource(Collection)` and routes everything else to `note()`+`null`. `destroy()`→`response()->json([...])` is an unrecognised `MethodCall` → refused (L509); `untypedScalar`→`return 1;` is a non-call expr → refused (L520); non-resource `StaticCall`/`new` refused in their handlers. The three planned negatives map to the three structurally-distinct rejection paths — no missing shape that could mis-resolve into a bogus schema. ✔

**3. Noise / silent-fallback — MATERIAL PLAN ERROR, corrected in the body.** The plan says to make the refusal note silent "mirroring how `paginatedFromBody` passes `log: false`." That is **not implementable as described**: the `log` flag exists only on `canonicalReturnExpression()` and `expressionAssignedTo()`. The shape-rejection notices that fire for the *high-volume* untyped non-resource case (`targetFromExpression` L509/L520, `targetFromStaticCall`, `targetFromNewExpression`, `targetFromTransformCall`) carry **no** `log` flag. Threading `log: false` through only the two methods that have it would still flood NOTICEs for every `destroy`/scalar/redirect — and the planned negative test ("`destroy()` must not emit a refusal notice") would **fail**. Corrected: thread a `bool $silent` from a new `locateFromReturnExpression($descriptor, silent: true)` on the **untyped branch only** → `read()` → `resolve()` → `note()`, leaving the typed call sites (L99/L105) at `silent: false` so #108's diagnostics are unchanged. Added the reader to "Files to modify."

**Additional finding — lint-rule impact (added to the body).** `ResourceClassLocator` is shared by `ResourceResponseResolver` **and four lint rules** (`ResourceResponseAmbiguous/Empty/FieldTypeMissing/FieldsUndeclared`). Broadening `locate()` changes what those rules see for untyped actions (e.g. a `new JsonResource($model)` wrapped-model shape now resolves to `resourceClass: null` → could trigger `resource.response-ambiguous`). Coder must confirm the **lint** corpus delta, not just the spec delta, and add a unit assertion that an untyped non-resource still yields `null`. The `silent` flag must affect the NOTICE log only, never `locate()`'s return value.

**4. Survey expectation — endorsed, with the planner's own caveat.** Vito/Lychee type their returns (`index(): Response`, `json(): ResourceCollection`) so they hit the typed branches and are untouched by the new untyped branch — claim holds. The second planner's note stands: AureusERP's `array_merge(parent::toArray($request), [...])` resources won't gain full shape until that idiom is taught (separate follow-up), so the AureusERP json-2xx uplift may come in below the route count — must not be misread as a regression at the survey gate.

Tier-1 broadening, surgical, names proving tests + docs (`docs/plugins.md`) + CHANGELOG. Approach approved; please apply the corrected silencing mechanism and the lint-delta confirmation before/while coding. Recommend the lead proceed to `agent:coding` with the amended body.

Radiergummi:
🤖 **[lead]** plan-review: approach approved with two corrections now in the PR body (thread a `silent` flag through the reader on the untyped path — the `log:false` mirror was insufficient; confirm the lint-corpus delta since the locator feeds 4 lint rules). → coding.

Radiergummi:
🤖 **[coder]** implemented; local gates green (test/pint/phpstan). Pushed 6c11c3e9. Ready for review.

**Fix (Tier-1, per the amended plan):**
- `ResourceClassLocator::locate()`: new early branch on the untyped/builtin return path — `return $this->locateFromReturnExpression($descriptor, silent: true)` — placed after the `#[ResponseResource]` check (L64-73) and before the base-Resource branches (L85+), so attribute precedence and the #108/#253 base-Resource gate are untouched.
- **Silencing (correction #1 applied):** added `bool $silent` threaded from the new `locateFromReturnExpression(..., silent: true)` → `ReturnExpressionResourceReader::read()` → `resolve()` → `targetFromExpression()`/`targetFromStaticCall()`/`targetFromNewExpression()`/`targetFromTransformCall()` → `note()`. `note()` no-ops when `$silent`. The two methods that already had a `log` flag (`canonicalReturnExpression`, `expressionAssignedTo`) get `log: !$silent`. Typed call sites (L99/L105) keep the default `silent: false`, so #108 diagnostics are unchanged. The flag gates the NOTICE log only, never `locate()`’s return value.

**Lint-rule delta (correction #2):** confirmed in tests rather than only by spec delta. The locator feeds 4 `resource.*` rules; broadening only changes untyped actions. Added:
- Unit: untyped-resource resolves to the right target; untyped-non-resource → `null` (no resolution → no spurious `resource.response-ambiguous`).
- Feature: `ResourceResponseAmbiguous` yields `[]` for both an untyped non-resource (`destroy()`) and an untyped body-resolved collection (`index()`).
- Negative `destroy()` test note: Core’s inline `response()->json()` reader still types the literal `{ok: boolean}` body (pre-existing, unrelated to the locator); asserted the schema is **not** a resource `$ref` and no `UntypedReturnResource` component is emitted, plus no refusal notice.

**Tests:** new `tests/Feature/Plugins/ApiResources/UntypedReturnResolutionTest.php` (parse-only untyped controller: paginated `index()`, `indexUnpaginated()`, `show()`→`new X()`, `make()`; negatives `destroy()`→`response()->json()`, conditional, scalar — all silent) + minimised fixture `tests/Fixtures/Resources/UntypedReturnResource.php`; unit cases in `ResourceClassLocatorTest`. Full suite 2783 pass.

**Docs/CHANGELOG:** `docs/plugins.md` resolution note now covers untyped actions + the silent-on-untyped-path behaviour; CHANGELOG `[Unreleased]` Fixed entry appended at end of section.

**phpstan.neon:** extended the existing `missingType.return` ignore (the same one used for `ReturnTypeNudgeController`) to the two new parse-only untyped fixtures — the missing return type is the input under test.

**Survey note (for the gate):** per the second planner, AureusERP’s `array_merge(parent::toArray($request), [...])` resources won’t gain full shape until that idiom is taught (separate follow-up), so the AureusERP json-2xx uplift may come in below the route count — not a regression. I did not run the full corpus lint (survey phase / surveyor owns that; the dogfood vendor symlink points at `main`, not this worktree).

Radiergummi:
🤖 **[lead]** implemented (6c11c3e9), CI green, diff matches the amended plan (untyped-path silent body-scan; silent flag threaded; phpstan missingType.return ignore extended to the 2 new untyped fixtures, well-justified). Production diff ~52 lines, all @internal/private surface. → review (2 independent approvals).

Radiergummi:
🤖 **[reviewer]** review (2/2): approved — no blocking concerns. I plan-reviewed this; both corrections I flagged landed correctly and completely. Ran `composer check` myself in the worktree: **green** (Pint `passed`, PHPStan `[OK] No errors`, Pest **2783 passed / 6871 assertions**); the 19 new/touched ApiResources tests pass in isolation.

**Correction 1 — silent flag threading: VERIFIED end-to-end, gates the log only.** The flag is plumbed `locate()` (untyped/builtin bail, L80 `silent: true`) → `locateFromReturnExpression($descriptor, bool $silent = false)` → `read($method, silent:)` → `resolve($method, $silent)` → both `canonicalReturnExpression`/`expressionAssignedTo` (via `log: !$silent`) **and** `targetFromExpression($expression, $method, $silent)` → `targetFromStaticCall`/`targetFromNewExpression`/`targetFromTransformCall` → every inline `note(..., $silent)`, with `note()` itself early-returning when `$silent`. So suppression reaches the **high-volume shape-rejection notes** (the `->json()` "unrecognised chain" L518 and the catch-all L530, plus the static-call/new-expr/transform-call notes) — not just `canonicalReturnExpression`, which was the exact gap I flagged. The variable/return-shape notes (L298/309/406/430) are `if ($log)`-gated and `log = !$silent`, so they're suppressed too. `$silent` is threaded **only** into the logging path; it never touches the resolved `ResourceTarget` — resolution is identical with or without it. The two typed call sites stay noisy: L103 (`?? new ResourceTarget(..., isCollection: true)`) and L109 (bare-`JsonResource` path) both call `locateFromReturnExpression($descriptor)` with no `silent:` → defaults `false`. #108's diagnostics intact.

**The `destroy()` negative asserts BOTH** (a) no resource schema — `$schema` has `ok`, no `$ref`, no `UntypedReturnResource` component — **and** (b) no refusal notice (`$noted` false). Nice correctness nuance the coder got right: Core's inline `response()->json()` reader still types the literal body, so the action isn't no-content; the ApiResources locator returning null just declines to *override* it with a bogus resource schema. Conditional (`>1 return`) and scalar (`random_int`) negatives likewise assert no-json-content + no notice.

**Correction 2 — lint-delta: VERIFIED.** Two `ResourceResponseAmbiguous` feature tests (untyped non-resource → `[]`, untyped body-resolved collection → `[]`) plus the unit `returns null for an untyped action that does not return a resource`. No spurious `resource.*` on untyped non-resources; the `silent` flag is log-only and never changes `locate()`'s return, so lint behaviour is driven purely by resolution as required.

**Branch placement / #108:** the new branch is a clean parallel early-return — `#[ResponseResource]` attribute branch (precedence) → untyped/builtin `locateFromReturnExpression(silent: true)` → unchanged typed base-Resource branches. The pre-existing locator unit tests (return-type, bare-collection-ambiguous, `#[ResponseResource]`, `#[Collects]`, `$collects`) all still pass — base-Resource path demonstrably undisturbed. **Refusal robustness:** the three negatives map to the three structurally-distinct refusal paths (unrecognised MethodCall, >1 return, non-call expr) — exhaustive for untyped actions.

**phpstan.neon `missingType.return`:** extends the existing scoped block (identifier + specific paths) to the two new untyped fixtures, comment updated to explain the AureusERP/Koel pattern (missing return type is the input under test). `reportUnmatchedIgnoredErrors` defaults on, so green PHPStan proves the ignores are matched, not blanket.

**Scope/conventions:** ~52 production lines across the two ApiResources support files; no Contracts/stage/config/Plugin/public-signature change (the new `$silent` params are `private`/defaulted on the `@internal` reader + locator). strict-types intact, comments concise and on the why. Fixtures parse-only.

Net: approve. Generation-affecting (`area:responses`/`area:plugins`) → survey gate applies before merge (per the planner's note, the AureusERP uplift may read below route count because `array_merge(parent::toArray())` resources aren't taught yet — a separate follow-up, not a regression).

Radiergummi:
🤖 **[reviewer]** review (1/2): approved, no concerns (one non-blocking rebase note for the lead).

Reviewed independently against the maintainer-approved re-scope (HEAD 6c11c3e9).

- **Precedence preserved.** The new untyped branch (`ResourceClassLocator.php:76-81`) sits *after* the `#[ResponseResource]` branch (L65-74) and *before* the typed base-Resource branches (L83-117), replacing the old unconditional `return null`. A null read returns null → today’s no-content behaviour. The typed paths (L103, L109) call `locateFromReturnExpression($descriptor)` with `silent` defaulting **false**, so #108’s notices stay noisy there.
- **Silent-flag correctness (the key plan-review fix).** `silent` is threaded `locate → locateFromReturnExpression(silent) → read(silent) → resolve → note()` and the existing `log:` flags. It gates the NOTICE **only**, never the resolved target. I walked **all 14** `note()` call sites: 10 take `$silent` directly; the other 4 (`:298/:309` in `canonicalReturnExpression`, `:406/:430` in `expressionAssignedTo`) are guarded by `if ($log)` and `resolve()` passes `log: !$silent` to both. No leak on the untyped path; the planned `log: false`-only mirror would have leaked, the coder used the correct mechanism.
- **Refusal robustness.** Non-resource untyped shapes read back null: `destroy()`→`response()->json()`, `untypedConditional()` (two returns → `canonicalReturnExpression` count>1 → null), `untypedScalar()`. All three feature negatives assert no `$ref`, no `UntypedReturnResource` component, and **no refusal notice**.
- **Lint-delta.** `isAmbiguous = resourceClass===null && modelClass===null`. The reader never constructs a both-null `ResourceTarget` (wrapped-model sets `modelClass`, L617; everything else returns null), so broadening cannot make `resource.response-ambiguous` fire spuriously on untyped actions. Unit assertions cover untyped non-resource → null and untyped collection → no ambiguous finding. The intended delta (untyped resource with empty/untyped `toArray()` now reachable by `response-empty`/field rules) is correct and desired.
- **phpstan.neon.** `missingType.return` ignore extended to the two new fixtures only, scoped by identifier + paths, with a justified comment (untyped return is the input under test). Not blanket.
- **Scope.** Production change is the two ApiResources files (~52 lines), all private/`@internal`; no Contracts, public-signature, stage-order, or config change.
- **Gate.** `composer check` in the worktree: **2783 passed (6871 assertions)**, Pint clean, PHPStan clean.

**Non-blocking (lead):** the branch is one commit behind `main` (#394 landed the GFM `MarkdownFormatter`). Both touch `CHANGELOG.md` / `docs/plugins.md` / `phpstan.neon`, so a rebase before squash-merge is advisable to confirm no textual conflict. No overlap in the `src/Plugins/ApiResources` files themselves.

Generation-affecting (`area:responses`/`area:plugins`) → survey gate applies (AureusERP/Koel json-2xx up, Vito/Lychee unchanged).

Radiergummi:
🤖 **[lead]** two independent approvals (silent-flag verified log-only end-to-end; lint-delta clean; refusal-robust). → survey (queued behind #392 — single surveyor). Pre-merge TODO: rebase onto main (#394 landed; shared CHANGELOG/docs/plugins.md/phpstan.neon — ApiResources src has no overlap with #394, but #392 will overlap ReturnExpressionResourceReader if it merges first).

Radiergummi:
🤖 **[surveyor]** Layer A **PASS** — 0 regressions across 11 apps; **AureusERP + Koel uplift confirmed real `JsonResource` resolution** (not noise). Baseline main@`f531087a` vs candidate `6c11c3e9` (fix/391), full 11-app corpus, php@8.4, both `CORPUS_EXIT=0`, all 11 `gen_exit=0`.

### Deterministic metrics — nothing regressed

| app | paths | ops | componentSchemas | responseSchemas | compl% |
|-----|------:|----:|----:|----:|----:|
| AureusERP | 111 | 160 | **21→36 ↑** | **14→66 ↑** | 91.1→91.1 |
| Koel | 181 | 294 | **85→106 ↑** | **8→42 ↑** | 82.4→72.9 † |
| Vito | 329 | 456 | 84→87 ↑ | 114→114 | 87.5→87.5 |
| Lychee | 213 | 273 | 280 | 82 | 97.6 |
| *(other 7 apps)* | — | — | byte-identical to baseline | — | — |

8 of 11 apps are unchanged — the change is surgical, touching only apps with untyped controller actions that return resources. No `paths`/`ops`/`componentSchemas`/`responseSchemas`/`documentedResponses`/`requestBodies`/`apiOperations` count dropped on any app.

### Class-type verification (the #392 check) — uplift is real, not Data-noise

Reflected every API-prefix action whose return expression resolves via the library's own `ReturnExpressionResourceReader`, on the **untyped path #393 newly enables**:

| app | resolved-resource routes (untyped path) | `JsonResource` | SpatieData | other |
|-----|----:|----:|----:|----:|
| AureusERP | 40 | **40** | 0 | 0 |
| Koel | 31 | **31** | 0 | 0 |

**100% `JsonResource`, zero Spatie Data, zero noise.** Resolved classes are exactly the expected ones (AureusERP `TitleResource`/`TagResource`/`BankResource`/…; Koel `AlbumResource`/`ArtistResource`/`SongResource`/…), all on actions with **no return type** — the precise gap #391 identified. Component schemas now materialise: AureusERP **15 resource schemas, all rich** (0→11+ props each); Koel **22 resource schemas, 18 rich + 4 empty**.

### † The Koel `compl%` dip is a metric artifact, NOT content loss — verified

16 Koel GET endpoints (`/api/songs/{song}`, `/api/podcasts`, `/api/songs/favorite`, …) flipped "complete→incomplete" in `completeness.php`. I diffed the baseline vs candidate spec for each: **all 16 had a bare `200 OK` with no schema in baseline; the candidate replaced them with a real `{data: {$ref: …Resource}}` envelope.** 0 of the 16 had a real response schema in baseline → **zero content was lost** (confirmed programmatically). `completeness.php` scores a bare-200 as "complete" (no-content rule) but a thin-`$ref` envelope as "incomplete" — so adding a `$ref` to an as-yet-unshaped resource *lowers* the metric while *adding* information. This is a quirk of the metric, not a regression.

The dip traces to **4 Koel resources** (`SongResource`, `PodcastResource`, `SongFileResource`, `LiveEventResource`) whose `toArray()` is a **build-then-conditionally-mutate-then-`return $var`** shape (e.g. `PodcastResource`: `$data = [...]; if ($this->withSubscriptionData) { $data[...] = ...; } return $data;`). That's a **Tier-2 dataflow `toArray()` read** — correctly refused — so they emit an empty schema. This is a **pre-existing `toArray()`-shaping limitation, independent of #393**: before the fix these routes didn't resolve at all (bare-200); #393 makes them resolve to a named `$ref`, and the empties are downstream of the separate Tier-2 boundary. AureusERP's resources are direct array literals → all shape cleanly (0 empty).

### Lint delta — expected, not spurious
Only Koel moves: `resource.fields-undeclared` **1→19** (+18). This is the lint rule **doing its job** — the 4 newly-resolved-but-unshaped resources now correctly trigger "resolved this resource, couldn't read its fields; annotate it." No other rule changed on any app; AureusERP gained 0 findings.

### Verdict
**PASS — clear to merge.** Real, verified `JsonResource` uplift on both target apps (AureusERP responseSchemas 14→66, Koel 8→42; 100% JsonResource), Vito/Lychee unchanged as predicted, no count regressions, no content loss, no spurious lint. The 4 empty Koel resources are a separate pre-existing Tier-2 `toArray()` limitation (build-then-return-`$var`), surfaced — not caused — by this fix and honestly flagged by `resource.fields-undeclared`; worth a follow-up note but not a blocker for #393.

<sub>Tooling (uncommitted, `$WS`): `gate393.py` (metric diff), `verify-resource-types.php` (class-type reflection). Baseline `baseline-main-f531087a.*`, candidate `candidate-393-6c11c3e9.*` cached.</sub>

Radiergummi:
🤖 **[lead]** Survey Layer A **PASS** — and a *real* uplift this time: AureusERP responseSchemas 14→66 (40/40 JsonResource), Koel 8→42 (31/31 JsonResource), zero Spatie-Data noise, Vito/Lychee unchanged, no count regressions. Merge gate fully satisfied (2 approvals + CI green + survey pass + non-controversial + maintainer-approved re-scope). Merging after #392 lands (both touch ReturnExpressionResourceReader → #393 rebases onto the result, keeping both changes).

**Follow-up (not a blocker, surfacing since issue-filing is blocked here):** the Koel completeness% metric dipped 82.4→72.9 — a *metric artifact*, not content loss (16 routes went bare-200 → thin {data:$ref}; completeness.php scores bare-200 as 'complete' but a thin-$ref envelope as 'incomplete'). Traced to 4 Koel resources (SongResource/PodcastResource/SongFileResource/LiveEventResource) with a build-then-conditionally-mutate-then-return-$var toArray() — a Tier-2 dataflow shape correctly refused → empty schema, flagged by resource.fields-undeclared (+18 Koel). Pre-existing toArray()-shaping limitation, independent of #393. Candidate for a future Tier-2-boundary issue + possibly a completeness.php scoring refinement (thin-$ref shouldn't score below bare-200).

Radiergummi:
🤖 **[coder]** rebased onto main (after #392 landed aefa2a9f); pushed d07f2316. Full CI matrix green.

Conflict was only in `ReturnExpressionResourceReader.php` — one refusal-note message string. Kept BOTH: #392’s updated whitelist wording (`X::collect(...)` added) AND my `$silent` argument on that `note()` call. #392’s `collect` whitelist entries and my `silent`-flag threading are orthogonal and both retained throughout the file. CHANGELOG / docs/plugins.md / phpstan.neon auto-merged (different blocks; all entries kept). `composer check` green locally (2796 tests) and the union verified: ApiResources suite 109 pass, the `collect` filter 78 pass, the untyped-return tests pass.

Radiergummi:
🤖 **[lead]** all gates green → squash-merged. Two independent reviewer approvals (silent-flag verified log-only end-to-end; lint-delta clean), full CI green on rebased head d07f2316 (PHP 8.4/8.5 × Laravel 12/13 + swagger-php 5.8 + Snapshot), survey Layer A pass with REAL JsonResource uplift (AureusERP responseSchemas 14→66, Koel 8→42; 40/40 + 31/31 JsonResource, zero Spatie-Data noise; Vito/Lychee unchanged; no regressions). Non-controversial (~52 src lines, @internal/private surface); maintainer pre-approved the untyped-return re-scope. Closes #391.

Issue #21, opened 2026-06-01
`lint --fix` Phase 2: Tier B corrections + Tier C stubs


Radiergummi:
## Summary

Second phase of `openapi:lint --fix`: teach the lint rules that emit *correctable* and *additive* findings how to repair themselves in source. Phase 1 (#20) lands the fixer infrastructure and the simple removal fixers (Tier A); this phase adds the two harder fixer families — Tier B corrections (rewrite an existing attribute argument) and Tier C additive stubs (insert a placeholder where the author left something out).

## Motivation

Tier A (#20) only deletes no-op attributes — the easy wins. The findings engineers actually trip over most are the *missing description / missing example / wrong type* class, which today require hunting down the right `#[ResponseField]` or `#[QueryParam]` and editing it by hand. Closing these under `--fix` is what turns the flag from a curiosity into a real time-saver: a freshly-scaffolded API can be brought to lint-clean (modulo genuinely ambiguous findings) in one command. The full fixable inventory and the rationale for which findings are auto-fixable versus left for humans live in the parent epic — see #7.

## Proposed approach

Tier 1 (bounded AST whitelist), building on the `nikic/php-parser` v5 format-preserving printer and the `FixableRule` / `Fixer` / `Fix` / `FixApplicator` contracts introduced in #20 (none of these exist on `main` yet — they arrive with Phase 1). All work lands under `src/Lint/`. A rule becomes fixable by also implementing `FixableRule` and returning a `Fixer` from `fixer()`.

The canonical rule IDs below are taken verbatim from the parent epic's fixable inventory (#7). Note the split between rules that already exist and findings that need a new owning rule first:

- **Tier B — corrections** via a `ModifyAttributeArgumentFixer`. Both rules already exist:
  - `apifield.conflicting-type` — owned by `FieldConflictingType` (`src/Lint/Rules/FieldConflictingType.php`); set the attribute's `type:` argument to the PHP property's declared type. (The class currently returns the ID `field.conflicting-type`; align it to the epic's `apifield.conflicting-type` as part of this work, or keep the existing ID and update the epic — flag the chosen direction in the PR.)
  - `apifield.invalid-format` — owned by `FieldInvalidFormat` (`src/Lint/Rules/FieldInvalidFormat.php`, currently returns `field.invalid-format`); remove the invalid `format:` argument. Same ID-alignment note applies.
[...]

PR #395, opened 2026-06-17
feat(#382): operationId fixers + AddAttribute AST primitive (lint-fix catalog slice 1)

### LINKED ISSUE #382: Build the additive/modification lint-fix catalog on the AST backend
## Context

#368 (Closes #367) swapped the lint Fix application backend from byte-addressed `SourceEdit`
splices to **AST mutation on a cloned tree + one `printFormatPreserving()` per file**. A `Fix` now
carries a single `AstOperation`; the backend ships three operation primitives:

- `RemoveAttribute` — drop attribute(s) by flat index.
- `SetAttributeArgument` — add / modify / remove a **named argument** on an *existing* attribute.
- `SetDocComment` — set or replace a docblock.

That work proved the substrate (the additive `SetAttributeArgument` op has a passing test), but
**every wired fixer today is still a removal / de-duplication**: `RemoveAttributeFixer`,
`TagDuplicate`, `QueryParamDuplicate`, `ResponseDuplicateStatus`, `LinkDuplicateName`,
`FieldNoEffect`, and the swagger-php OA removers. The additive / modification path has a primitive
and a proof, but no rule fixers on top of it.

This issue is the explicit follow-on that #367 deferred ("Building out the full insert/modify fixer
catalog — that follows on this backend") and that #368's phase 5 only sampled ("Build *one*
insertion/modification fixer … to serve as the template").

## Goal

For every finding that is **mechanically fixable** — the corrected value can be derived
deterministically from code/spec with no human judgement — ship a `FixableRule::fixer()` built on
the AST-mutation backend. Leave only the genuinely human work (prose summaries / descriptions,
sunset dates, replacement targets) as diagnostics.

**Litmus test for catalog membership:** a rule belongs here only if the fix value is synthesizable
without inventing intent. "Add a description" fails it (prose); "rename `user_id` → `userId` to match
the configured convention" and "set `required: true` on a path parameter" pass it.

## Prerequisite primitive(s)

The current ops only *modify an existing attribute* or *set a docblock*. Several additive fixes
target a member that has **no relevant attribute yet** (e.g. synthesize `#[OperationId]` where none
exists). Before/within this work we need:

- [ ] An **`AddAttribute`** (insert-new-attribute) AST operation — synthesize an `AttributeGroup`
      onto a class/method/property/param, format-preserving. (`SetAttributeArgument` cannot create
      the attribute it targets.)
- [ ] Confirm the docblock substrate can **insert a tag into an existing block** (vs. only
      replacing whole-block text via `SetDocComment`) for any docblock-tag fixers that land here;
      add the phpdoc-parser path if a concrete fixer needs it.

## Candidate fixable rules (triage + build)

Grouped by fix shape; each needs per-rule confirmation against the litmus test before building.
Several naming rules target route/path/class-name sources rather than attributes — those may be
out of reach of the attribute/docblock substrate and should be marked **deferred** rather than forced.

**Value synthesis (additive) — `SetAttributeArgument` / `AddAttribute` + existing synthesisers**
- [ ] `operation.id-missing` — synthesize an operationId from the action/route (needs `AddAttribute`).
- [ ] `parameter.example-missing` — Faker-backed value via `FakerExampleSynthesiser` → `example:` arg.
- [ ] `response.example-missing`
- [ ] `request-body.example-missing`
- [ ] `schema.example-missing`

**Deterministic modification — `SetAttributeArgument`**
- [ ] `operation.id-invalid-chars` — sanitize to the allowed character set.
- [ ] `parameter.path-must-be-required` — set `required: true`.
- [ ] `parameter.example-conflict` — reconcile to the single canonical example.
- [ ] `schema.nullable-via-deprecated-keyword` — rewrite the deprecated `nullable` keyword to the 3.1 form.
- [ ] `header.invalid-name` — normalize the header name.

**Naming-convention rewrites — modification (confirm the source is attribute-addressable)**
- [ ] `field.name-naming-inconsistent`
- [ ] `parameter.name-naming-inconsistent`
- [ ] `header.name-naming-inconsistent`
- [ ] `tag.name-naming-inconsistent`
- [ ] `operation.id-naming-inconsistent`
- [ ] `component.name-naming-inconsistent` *(likely deferred — class-name addressed)*
- [ ] `path.segment-naming-inconsistent` / `path.trailing-slash-inconsistent` *(likely deferred — route source)*

(Triage will likely move some rows to "human-only" or "deferred"; record the disposition of each
rejected candidate so the catalog is provably exhaustive, not silently truncated.)

## Scope / acceptance criteria

- [ ] `AddAttribute` primitive (+ visitor support) with byte-stable golden coverage.
- [ ] Each accepted candidate implements `FixableRule` and returns a `Fixer` emitting AST operations.
- [ ] Golden tests per fixer: idempotent (re-running produces no change) and Pint-clean output.
- [ ] Every candidate above is dispositioned — built, deferred (with reason), or human-only (with reason).
- [ ] `docs/linting.md` catalog reflects the now-fixable rules; `openapi:lint --fix` / `--list`
      surface them. CHANGELOG `[Unreleased]` entry.
- [ ] `composer lint` (PHPStan L8), `vendor/bin/pint --test`, full Pest suite green.

## Non-goals

- The byte-splice backend (already retired in #368) — not revisited here.
- Prose synthesis (summaries, descriptions, sunset dates, deprecation replacements) — stays diagnostic.
- New lint *rules* — this is fixers for existing findings only.

## Notes

Internal Fix-layer architecture; no tier label (not an inference measure). Depends on #368 landing.
Implementation plan lives in the draft PR.

Radiergummi:
Closes #382

## Implementation plan — #382 additive/modification lint-fix catalog (slice 1)

`<type>=feat` · area:lint · **no tier label** (Fix-layer infra, not an inference measure).
Depends on #368 (merged: AST-mutation Fix backend).

### TL;DR

This issue is too large for one auto-mergeable PR (AddAttribute primitive + ~15 fixers + per-rule
triage + goldens + docs). This PR ships a **genuinely-mergeable first slice**: the missing
`AddAttribute` AST primitive plus the **two operation-level fixers that are cleanly
attribute-addressable today** — `operation.id-missing` and `operation.id-invalid-chars`. Every
remaining candidate from the issue is **dispositioned below** (built / deferred-with-reason /
human-only) so the catalog is provably exhaustive, not silently truncated.

### Why this slice, and not the example/parameter set

The exploration surfaced a hard architectural constraint that reshapes the issue's candidate list:

1. **A `Finding` carries no byte location; a fixer needs a *structural address* (FQCN + member).**
   The only tree node that exposes a source back-reference is `OperationNode`, via its
   `descriptor` (`ActionDescriptor` → controller + method reflection). This is exactly what the
   existing `RemoveAttributeFixer::contextForOperation()` consumes. So **operation-level rules are
   the only ones whose findings can be made attribute-addressable without new plumbing.**

2. **`ParameterNode`/`ResponseNode`/header/schema-property nodes have no descriptor and no path
   back to a source attribute.** They are reconstructed from the generated `OA\*` objects. A spec
   parameter may originate from a `#[QueryParam]`/`#[PathParameter]` attribute, from
   route-model-binding inference, or be synthesized from the route URI with **no attribute at all**
   (the common case for path params). A fixer cannot mutate an attribute that does not exist and
   that the node cannot point to. Making these fixable requires a separate, larger plumbing change
   (thread a source-member + attribute-index back-reference through `SpecTreeBuilder`), which is
   its own slice.

3. **`SetAttributeArgument` is scalar-only** (`string|int|float|bool|null`). Faker example
   synthesis (`FakerExampleSynthesiser::synthesise()`) and example values are frequently
   arrays/objects, which the current op cannot express. The `*.example-missing` set therefore needs
   *both* (2) the source back-reference *and* a structured-value op — out of scope here.

Building `operation.id-missing` first also exercises the new `AddAttribute` primitive end-to-end
(synthesize `#[Operation(operationId: ...)]` where no `#[Operation]` exists), which is the issue's
named prerequisite.

### What this PR builds

**0. Extract the operationId-derivation logic into a shared, callable surface** (REQUIRED by plan-review)
- Today `PathsStage::buildOperationId()` / `routeNameOperationId()` / `methodPathOperationId()` /
  `sanitiseOperationId()` are **all `private`** on `Support\Generator\Stages\PathsStage`
  (PathsStage.php:156-203) — not callable from the Fix layer. Both new fixers must produce *exactly*
  what the generator emits, so the logic must be shared, not reinvented.
- New `src/Support/Generator/OperationIdDeriver.php` — a stateless, plugin-agnostic `final` helper in
  `Support/` (correct home: it is config-driven, plugin-agnostic infrastructure):
  - `public function derive(ActionDescriptor $descriptor, HttpMethod $method): string` — the body of
    today's `buildOperationId()`, including the `config('openapi.operation_id_strategy')` read and
    the `route-name` / `method-path` branch (moves `routeNameOperationId()` + `methodPathOperationId()`
    in as private methods).
  - `public function sanitise(string $operationId): string` — the body of today's
    `sanitiseOperationId()` (`preg_replace('/[^A-Za-z0-9._-]+/', '_')` then strip leading non-letters,
    preserving dots). This is exactly what `operation.id-invalid-chars` needs.
- **Pure extraction, no behaviour change.** `PathsStage` keeps its current call sites but delegates to
  the helper (construct/inject it; it has no state). Stays in `Support/`, no `Contracts/` change, no
  config-key change — the config read moves verbatim. The existing PathsStage tests pin the
  no-behaviour-change guarantee; add a direct unit test for the helper (`OperationIdDeriverTest`)
  covering both strategies + sanitisation.
- The `AddOperationIdFixer` calls `derive(...)`; the `SanitizeOperationIdFixer` calls `sanitise(...)`.
  Both inputs (`ActionDescriptor`, `HttpMethod`) are available on the finding's `OperationNode`
  (`descriptor` + `method`), so the fixers can reconstruct them from the stamped context.

**1. `AddAttribute` AST primitive** (the issue's checklist prerequisite)
- `src/Lint/Fix/Ast/AddAttribute.php` — `final readonly class AddAttribute extends AstOperation`.
  Carries the attribute FQCN (`class-string`) and an ordered map of named arguments
  (`array<string, string|int|float|bool|null>` — scalar-only, matching `SetAttributeArgument`'s
  value domain). Targets a class/method/property/param via the inherited `TargetSelector`.
- Wire it into `src/Lint/Fix/Ast/FixOperationVisitor.php`: in `enterNode()`, add an
  `$this->operation instanceof AddAttribute` branch that synthesizes a new
  `PhpParser\Node\AttributeGroup` (one `Attribute` with a resolved `Name` + `Arg` list built from
  the scalar args, reusing the same scalar→expr construction `SetAttributeArgument` uses) and
  **prepends** it to the located member's `attrGroups`. Prepend (not append) gives a deterministic,
  byte-stable position and keeps the attribute above any existing ones.
- Edge handling: if the member already carries the target attribute class, `AddAttribute` is a
  no-op (yield/apply nothing) — the *fixer* decides add-vs-modify, the op just inserts. Name
  rendering uses the short class name with a `use` assumption only if already imported; otherwise
  emit the FQ name (leading `\`) to stay correct without touching the `use` block. **Decision:**
  emit FQCN form to avoid editing imports in slice 1; a later slice can add import management.
  (Record this as a PR comment if the reviewer pushes on aesthetics.)

**2. Fixer: `operation.id-missing`** (additive, uses `AddAttribute`)
- `OperationIdMissing` (`src/Lint/Rules/OperationIdMissing.php`) gains `implements FixableRule` +
  `fixer()`. In `checkOperation()`, stamp finding `context` with the source class/member via
  `RemoveAttributeFixer::contextForOperation($operation->descriptor)` (reuse the existing
  context-builder; no discriminator) plus the synthesized operationId value under a new context key.
- New `src/Lint/Fix/AddOperationIdFixer.php implements Fixer`: reads the source class/method from
  the finding context, synthesizes the operationId by calling
  `OperationIdDeriver::derive($descriptor, $method)` (the shared surface from step 0) so the fix
  equals exactly what inference would emit, then:
  - if a `#[Operation]` attribute exists on the method → emit `SetAttributeArgument(operationId: …)`;
  - else → emit `AddAttribute(Operation::class, ['operationId' => …])`.
  Degrade to nothing when the descriptor/reflection is unavailable.

**3. Fixer: `operation.id-invalid-chars`** (deterministic modification, uses `SetAttributeArgument`)
- `OperationIdInvalidChars` gains `FixableRule` + `fixer()`, stamping the same operation context.
- `src/Lint/Fix/SanitizeOperationIdFixer.php`: sanitize the existing id by calling
  `OperationIdDeriver::sanitise($operationId)` (the shared surface from step 0 — identical to the
  generator's `sanitiseOperationId`, which already maps to the allowed set `/^[A-Za-z][A-Za-z0-9._-]*$/`),
  then emit `SetAttributeArgument` setting `operationId` on the existing `#[Operation]` attribute.
  (An invalid id can only have come from an explicit attribute, so the attribute always exists →
  no `AddAttribute` branch needed here.)

### Disposition of every issue candidate (acceptance: exhaustive)

| Rule | Disposition | Reason |
|---|---|---|
| `operation.id-missing` | **BUILT** | `OperationNode.descriptor` is addressable; needs/exercises `AddAttribute`. |
| `operation.id-invalid-chars` | **BUILT** | Addressable; value sanitizable deterministically. |
| `parameter.path-must-be-required` | **DEFERRED** | `ParameterNode` has no descriptor/attribute back-ref; path params are often synthesized from the route URI with no attribute to set `required:` on. Needs source back-reference plumbing (separate slice). |
| `parameter.example-missing` | **DEFERRED** | No source back-ref **and** `SetAttributeArgument` is scalar-only (Faker values may be arrays). Needs both plumbing + a structured-value op. |
| `response.example-missing` | **DEFERRED** | Same as above; response media-type examples are typically structured. |
| `request-body.example-missing` | **DEFERRED** | Same as above. |
| `schema.example-missing` | **DEFERRED** | Schema property nodes have no source-attribute back-ref; scalar-only op also insufficient. |
| `parameter.example-conflict` | **DEFERRED** | No back-ref; reconciliation also needs to pick which of N attribute sites is canonical (not expressible without the back-ref). |
| `schema.nullable-via-deprecated-keyword` | **DEFERRED** | Operates on generated schema shape, not on a source attribute the node points to; the deprecated `nullable` keyword has no 1:1 attribute argument to rewrite. |
| `header.invalid-name` | **DEFERRED** | Header nodes have no source back-ref to the `#[ResponseHeader]` attribute/index. |
| `field.name-naming-inconsistent` | **DEFERRED** | Field name source is the property/`#[Field]` name; needs the field back-ref + a rename op (renaming a property name, not an attribute arg). Larger, distinct shape. |
| `parameter.name-naming-inconsistent` | **DEFERRED** | No source back-ref (as above). |
| `header.name-naming-inconsistent` | **DEFERRED** | No source back-ref (as above). |
| `tag.name-naming-inconsistent` | **DEFERRED** | Tag value is addressable in principle, but a *rename to the configured convention* must agree with the convention engine; bundle with the naming-convention slice for one consistent rename op. |
| `operation.id-naming-inconsistent` | **DEFERRED** | Operation id is addressable, but the rename target is convention-driven; group with the naming slice so all naming rewrites share one convention-rename path (avoids divergent per-rule logic). |
| `component.name-naming-inconsistent` | **DEFERRED (likely permanent)** | Class-name addressed; renaming a class touches every reference — out of reach of the attribute/docblock substrate. |
| `path.segment-naming-inconsistent` | **DEFERRED (likely permanent)** | Route-source addressed; the fix is a route definition change, not an attribute. |
| `path.trailing-slash-inconsistent` | **DEFERRED (likely permanent)** | Route-source addressed (as above). |
| `apifield.conflicting-type` (`FieldConflictingType`) | **DEFERRED** (slice 2) | Deterministic `SetAttributeArgument` set `type:` to the property's declared PHP type — passes the litmus — but needs the field/attribute back-ref (`FieldNode` has no descriptor). **ID-alignment flag:** the rule currently returns `field.conflicting-type`, not `apifield.conflicting-type`; the slice-2 PR must pick a direction (rename the rule ID vs. update the epic) — flagged here, not decided now. |
| `apifield.invalid-format` (`FieldInvalidFormat`) | **DEFERRED** (slice 2) | Remove the invalid `format:` arg (`SetAttributeArgument` remove) — passes the litmus — but same field back-ref dependency. Rule currently returns `field.invalid-format`; **same ID-alignment flag** as above. |
| `response.heuristic` / `request.heuristic` | **OUT OF SCOPE** | Neither rule exists on `main`. This catalog is *fixers for existing findings only* (issue non-goal: "New lint *rules* … not here"). Materialising an inferred heuristic as an explicit attribute is authoring-first work; triage whether it belongs to this cluster at all happens when/if the rules are authored. |
| Prose rows from #21: `summary.missing`, `field.description-missing`, `parameter.description-missing`, `queryparam.description-missing`, `meta.no-suppression-reason` | **HUMAN-ONLY** | #21 proposed hard-coded `'TODO'` stubs; per this issue's non-goals prose stays diagnostic — the value requires human judgement (fails the litmus test). |
| Prose (issue body): summaries/descriptions/sunset dates/deprecation replacements | **HUMAN-ONLY** | Per issue non-goals; value requires human judgement (fails the litmus test). |

**Follow-on slices the lead should surface to the user** (we cannot `gh issue create` here):
- *Slice 2 — source back-reference plumbing*: thread source-member + attribute-index identity
  through `SpecTreeBuilder` onto `ParameterNode`/`ResponseNode`/`FieldNode`/header/schema-property
  nodes, so parameter/header/field/example rules become addressable. Unblocks
  `parameter.path-must-be-required`, the example-conflict/header rows, and the `apifield.*` rows
  (`conflicting-type`, `invalid-format`) — the slice-2 PR also resolves the `field.*` vs `apifield.*`
  rule-ID-alignment question flagged above.
- *Slice 3 — structured-value AST op*: a `SetAttributeArgument` variant (or new op) accepting
  array/object values, to land the `*.example-missing` Faker-backed set.
- *Slice 4 — naming-convention rename op*: one convention-driven rename path shared by the
  `*.name-naming-inconsistent` family (field/parameter/header/tag/operation-id rows).

New unit tests under `tests/Unit/Lint/Fix/`, fixtures under `tests/Fixtures/Lint/Fix/`, using the
existing `AttributeFixFixture::run()` golden harness (copy fixture → synthetic Finding → apply →
assert exact `after` source).

- `AddAttributeOperationTest` (op-level, byte-stable golden):
  - inserts `#[Operation(operationId: 'x')]` on a method with **no** attributes → exact output.
  - inserts above an existing unrelated attribute (e.g. `#[Tag('users')]`) → ordering + byte-stable.
  - target already has the attribute class → no-op: the test asserts the output is **byte-identical
    to the input source** (source-unchanged), not merely that nothing throws, pinning the op contract
    independently of any fixer.
  - **idempotent**: re-running the produced op yields no further change.
- `AddOperationIdFixerTest`:
  - method with no `#[Operation]` → `AddAttribute` path, exact synthesized id, Pint-clean output.
  - method with existing `#[Operation(summary: …)]` but no `operationId` → `SetAttributeArgument`
    add-arg path (no duplicate attribute).
  - descriptor/reflection unavailable → yields nothing (degrade).
  - idempotent re-run.
- `SanitizeOperationIdFixerTest`:
  - id with spaces/illegal chars → sanitized to allowed set, starts with a letter.
  - id starting with a digit → letter-prefixed/repaired branch.
  - already-valid id (defensive) → yields nothing.
  - idempotent re-run.
- Every golden asserts the full `after` string (byte-stable) and the suite already runs in the
  `snapshot`-excluded path; ensure outputs are Pint-clean by construction (printer + format-preserve).

### Docs & changelog (observable surface)

- `docs/linting.md` — between the `lint-rule-catalog` markers, the catalog is severity-only and
  hand-maintained; add/adjust a "Fixable" indication for `operation.id-missing` and
  `operation.id-invalid-chars` consistent with how existing fixable rules (`tag.duplicate`, etc.)
  are shown. Confirm during coding whether the catalog encodes fixability; if not, note the two
  rules as `--fix`-able in the surrounding prose. `openapi:lint --list` is the live source and will
  reflect them automatically once they implement `FixableRule`.
- `CHANGELOG.md` `[Unreleased] → Added`: one line — `openapi:lint --fix` now repairs missing and
  codegen-unsafe operationIds (synthesizes/sanitizes the `#[Operation(operationId: …)]`), via the
  new `AddAttribute` fix primitive. (#382)

### Controversy flags for the lead (expect a human merge gate)

- **Touches the Fix-layer substrate**: new `AddAttribute` AST op + `FixOperationVisitor` branch.
  This is the same substrate #368 introduced; new primitive = architectural surface.
- **Likely > ~150 src/ LOC** across the op, visitor branch, two fixers, and the two rule
  `FixableRule` wirings (plus tests/fixtures). Above the fast-path mechanical threshold.
- **No `src/Contracts/**` change, no config key, no public-signature change** — confirmed: rules
  gain `FixableRule` (already a public-but-internal Fix contract from #368), fixers are new
  `Fixer` implementations, the op extends the existing internal `AstOperation`. Nothing in the
  documented stable extension surface moves.
- `area:lint` ⇒ **survey gate skipped** (non-generation-affecting).
- Several candidates are **deferred without a follow-up issue** (runtime can't `gh issue create`) —
  the lead must surface slices 2–4 to the user.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,6 +38,7 @@ All notable changes to this project are documented here.
 - `#[ResponseExampleFile('path.json')]` authoring attribute attaches a JSON file's contents as the singular `example` on a response (the auto-derived primary by default, `status:` targets a declared response). Skipped silently for a conventionally bodyless status (204/205/304) or a status with no matching response, and yields to named `#[ResponseExample]`s already on the same media type (`example`/`examples` are mutually exclusive). (#40)
 - `openapi:lint --format=markdown` now renders a real GitHub-Flavored Markdown body (a coverage summary line and a `Severity | Rule | Location | Message` findings table) instead of aliasing the CLI tree dump, so the CI coverage-comment recipe can post it verbatim. (#389)
 - ApiResources: the return-expression reader now recognises the `X::collect(...)` static resource-collection factory, emitting the same collection response schema as `X::collection(...)`. (#390)
+- `openapi:lint --fix` now repairs missing and codegen-unsafe operationIds (synthesizes/sanitizes the `#[Operation(operationId: …)]`), via the new `AddAttribute` fix primitive. (#382)
 
 ### Changed
 - `spatie/laravel-data` is now a soft dependency (moved from `require` to `require-dev`); Fractal and query-builder packages are opt-in.
--- a/docs/linting.md
+++ b/docs/linting.md
@@ -144,9 +144,14 @@ run your formatter over them (e.g., `vendor/bin/pint --dirty`) to match your
 project style. A finding is only auto-fixable when its owning rule opts in
 (by implementing `Lint\Fix\FixableRule`); everything else is reported as before.
 
-Currently fixable (all removals of a redundant or no-op attribute):
+Currently fixable. Removals of a redundant or no-op attribute:
 `tag.duplicate`, `queryparam.duplicate`, `response.duplicate-status`,
-`link.duplicate-name`, and `field.no-effect`.
+`link.duplicate-name`, and `field.no-effect`. Additive/modification fixes on
+the `#[Operation]` attribute: `operation.id-missing` (synthesizes
+`#[Operation(operationId: …)]`, or adds the argument when the attribute already
+exists) and `operation.id-invalid-chars` (rewrites the operationId to the
+codegen-safe character set). Both produce exactly the operationId inference
+would have emitted.
 
 ## Documentation-coverage gate
 
--- a/src/Lint/Fix/AddOperationIdFixer.php
+++ b/src/Lint/Fix/AddOperationIdFixer.php
@@ -1,0 +1,102 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Lint\Fix;
+
+use Override;
+use Radiergummi\OpenApi\Attributes\Operation;
+use Radiergummi\OpenApi\Lint\Finding;
+use Radiergummi\OpenApi\Lint\Fix\Ast\AddAttribute;
+use Radiergummi\OpenApi\Lint\Fix\Ast\SetAttributeArgument;
+use Radiergummi\OpenApi\Lint\Fix\Ast\TargetKind;
+use Radiergummi\OpenApi\Lint\Fix\Ast\TargetSelector;
+use ReflectionException;
+use ReflectionMethod;
+
+use function is_a;
+use function is_string;
+use function sprintf;
+
+/**
+ * Adds a missing operationId to the source `#[Operation]` attribute.
+ *
+ * The rule stamps the operationId inference would have emitted (via {@see \Radiergummi\OpenApi\Support\Generator\OperationIdDeriver})
+ * onto the finding, so the fix equals exactly what the generator produces. If the action method
+ * already carries `#[Operation]`, the fixer sets its `operationId:` argument; otherwise it
+ * synthesises a new `#[Operation(operationId: …)]`. Degrades to nothing when the source member or
+ * the stamped operationId is unavailable.
+ *
+ * @internal
+ */
+final readonly class AddOperationIdFixer implements Fixer
+{
+    public const string CONTEXT_OPERATION_ID = 'fixOperationId';
+
+    /**
+     * @return iterable<Fix>
+     */
+    #[Override]
+    public function fix(Finding $finding, FixContext $context): iterable
+    {
+        $class = $finding->context[Finding::CONTEXT_SOURCE_CLASS] ?? null;
+        $member = $finding->context[Finding::CONTEXT_SOURCE_MEMBER] ?? null;
+        $operationId = $finding->context[self::CONTEXT_OPERATION_ID] ?? null;
+
+        if (!is_string($class) || !is_string($member) || !is_string($operationId) || $operationId === '') {
+            return [];
+        }
+
+        $target = new TargetSelector($class, TargetKind::Method, $member);
+        $existingIndex = $this->existingOperationAttributeIndex($class, $member);
+
+        $operation = $existingIndex === null
+            ? new AddAttribute($target, Operation::class, ['operationId' => $operationId])
+            : new SetAttributeArgument($target, $existingIndex, 'operationId', $operationId);
+
+        $file = $this->declaringFile($class, $member);
+
+        if ($file === null) {
+            return [];
+        }
+
+        return [new Fix(
+            file: $file,
+            description: sprintf('Set operationId "%s" on %s::%s', $operationId, $class, $member),
+            ruleId: $finding->ruleId,
+            operation: $operation,
+        )];
+    }
+
+    /**
+     * The flat, source-order position of an existing `#[Operation]` attribute on the method, or null
+     * when the method carries none (so the fixer adds one instead of mutating).
+     */
+    private function existingOperationAttributeIndex(string $class, string $member): ?int
+    {
+        try {
+            $reflector = new ReflectionMethod($class, $member);
+        } catch (ReflectionException) {
+            return null;
+        }
+
+        foreach ($reflector->getAttributes() as $index => $attribute) {
+            if (is_a($attribute->getName(), Operation::class, true)) {
+                return $index;
+            }
+        }
+
+        return null;
+    }
+
+    private function declaringFile(string $class, string $member): ?string
+    {
+        try {
+            $file = new ReflectionMethod($class, $member)->getFileName();
+        } catch (ReflectionException) {
+            return null;
+        }
+
+        return $file ?: null;
+    }
+}
--- a/src/Lint/Fix/Ast/AddAttribute.php
+++ b/src/Lint/Fix/Ast/AddAttribute.php
@@ -19,10 +19,10 @@ namespace Radiergummi\OpenApi\Lint\Fix\Ast;
 final readonly class AddAttribute extends AstOperation
 {
     /**
-     * @param class-string                                  $attributeClass The attribute to add, e.g.
-     *                                                                      `Operation::class`.
-     * @param array<string, string|int|float|bool|null>     $arguments      Named arguments in the
-     *                                                                      order they should render.
+     * @param class-string                              $attributeClass The attribute to add, e.g.
+     *                                                                  `Operation::class`.
+     * @param array<string, null|bool|float|int|string> $arguments      Named arguments in the
+     *                                                                  order they should render.
      */
     public function __construct(
         TargetSelector $target,
--- a/src/Lint/Fix/Ast/AddAttribute.php
+++ b/src/Lint/Fix/Ast/AddAttribute.php
@@ -1,0 +1,34 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Lint\Fix\Ast;
+
+/**
+ * Adds a new attribute application to the target member (a class, method, property, or promoted
+ * parameter), synthesised from a fully-qualified attribute class and an ordered map of named scalar
+ * arguments. The new attribute is prepended as its own {@see \PhpParser\Node\AttributeGroup} above
+ * any existing groups, giving a deterministic, byte-stable position.
+ *
+ * If the member already carries the attribute class, the operation is a no-op (nothing is applied):
+ * the *fixer* decides add-vs-modify, the op only inserts. The argument values share
+ * {@see SetAttributeArgument}'s scalar domain; a structured-value variant is a separate concern.
+ *
+ * @internal
+ */
+final readonly class AddAttribute extends AstOperation
+{
+    /**
+     * @param class-string                                  $attributeClass The attribute to add, e.g.
+     *                                                                      `Operation::class`.
+     * @param array<string, string|int|float|bool|null>     $arguments      Named arguments in the
+     *                                                                      order they should render.
+     */
+    public function __construct(
+        TargetSelector $target,
+        public string $attributeClass,
+        public array $arguments = [],
+    ) {
+        parent::__construct($target);
+    }
+}
--- a/src/Lint/Fix/Ast/FixOperationVisitor.php
+++ b/src/Lint/Fix/Ast/FixOperationVisitor.php
@@ -8,7 +8,10 @@ use PhpParser\Comment\Doc;
 use PhpParser\Node;
 use PhpParser\Node\Arg;
 use PhpParser\Node\Attribute;
+use PhpParser\Node\AttributeGroup;
 use PhpParser\Node\Identifier;
+use PhpParser\Node\Name;
+use PhpParser\Node\Name\FullyQualified;
 use PhpParser\Node\Param;
 use PhpParser\Node\PropertyItem;
 use PhpParser\Node\Scalar;
@@ -17,10 +20,15 @@ use PhpParser\Node\Stmt\ClassMethod;
 use PhpParser\Node\Stmt\Property;
 use PhpParser\NodeVisitorAbstract;
 
+use function array_any;
+use function array_keys;
+use function array_map;
+use function array_unshift;
 use function array_values;
 use function is_float;
 use function is_int;
 use function krsort;
+use function ltrim;
 use function rsort;
 use function strcasecmp;
 
@@ -65,6 +73,8 @@ final class FixOperationVisitor extends NodeVisitorAbstract
             $this->setDocComment($member, $this->operation);
         } elseif ($this->operation instanceof SetAttributeArgument) {
             $this->setAttributeArgument($member, $this->operation);
+        } elseif ($this->operation instanceof AddAttribute) {
+            $this->addAttribute($member, $this->operation);
         }
 
         return null;
@@ -312,6 +322,62 @@ final class FixOperationVisitor extends NodeVisitorAbstract
         }
 
         return true;
+    }
+
+    private function addAttribute(ClassLike|ClassMethod|Property|Param $member, AddAttribute $operation): void
+    {
+        // The fixer decides add-vs-modify; the op only inserts. Adding a second application of an
+        // attribute the member already carries is never what a fixer wants, so refuse it here, which
+        // also makes the op idempotent against its own output.
+        if ($this->hasAttributeClass($member, $operation->attributeClass)) {
+            return;
+        }
+
+        $arguments = array_map(
+            fn(string|int|float|bool|null $value, string $name): Arg => new Arg(
+                value: $this->literalToNode($value),
+                name: new Identifier($name),
+            ),
+            array_values($operation->arguments),
+            array_keys($operation->arguments),
+        );
+
+        // Emit the fully-qualified attribute name so the rewrite stays correct without editing the
+        // file's `use` block.
+        $group = new AttributeGroup([
+            new Attribute(new FullyQualified(ltrim($operation->attributeClass, '\\')), $arguments),
+        ]);
+
+        array_unshift($member->attrGroups, $group);
+        $this->applied = true;
+    }
+
+    /**
+     * Whether the member already carries an application of the given attribute class, comparing by
+     * the name {@see \PhpParser\NodeVisitor\NameResolver} resolved (it stamps `resolvedName` on
+     * statically-resolvable names) so a short, `use`-imported reference matches its FQCN.
+     */
+    private function hasAttributeClass(ClassLike|ClassMethod|Property|Param $member, string $attributeClass): bool
+    {
+        $target = ltrim($attributeClass, '\\');
+
+        foreach ($member->attrGroups as $group) {
+            $matches = array_any(
+                $group->attrs,
+                static function (Attribute $attribute) use ($target): bool {
+                    $resolved = $attribute->name->getAttribute('resolvedName');
+                    $name = $resolved instanceof Name ? $resolved : $attribute->name;
+
+                    return ltrim($name->toString(), '\\') === $target;
+                },
+            );
+
+            if ($matches) {
+                return true;
+            }
+        }
+
+        return false;
     }
 
     private function literalToNode(string|int|float|bool|null $value): Node\Expr
--- a/src/Lint/Fix/Ast/FixOperationVisitor.php
+++ b/src/Lint/Fix/Ast/FixOperationVisitor.php
@@ -383,9 +383,9 @@ final class FixOperationVisitor extends NodeVisitorAbstract
     private function literalToNode(string|int|float|bool|null $value): Node\Expr
     {
         return match (true) {
-            $value === null  => new Node\Expr\ConstFetch(new Node\Name('null')),
-            $value === true  => new Node\Expr\ConstFetch(new Node\Name('true')),
-            $value === false => new Node\Expr\ConstFetch(new Node\Name('false')),
+            $value === null  => new Node\Expr\ConstFetch(new Name('null')),
+            $value === true  => new Node\Expr\ConstFetch(new Name('true')),
+            $value === false => new Node\Expr\ConstFetch(new Name('false')),
             is_int($value)   => new Scalar\Int_($value),
             is_float($value) => new Scalar\Float_($value),
             default          => new Scalar\String_($value),
--- a/src/Lint/Fix/SanitizeOperationIdFixer.php
+++ b/src/Lint/Fix/SanitizeOperationIdFixer.php
@@ -1,0 +1,91 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Lint\Fix;
+
+use Override;
+use Radiergummi\OpenApi\Attributes\Operation;
+use Radiergummi\OpenApi\Lint\Finding;
+use Radiergummi\OpenApi\Lint\Fix\Ast\SetAttributeArgument;
+use Radiergummi\OpenApi\Lint\Fix\Ast\TargetKind;
+use Radiergummi\OpenApi\Lint\Fix\Ast\TargetSelector;
+use ReflectionException;
+use ReflectionMethod;
+
+use function is_a;
+use function is_string;
+use function sprintf;
+
+/**
+ * Rewrites a codegen-unsafe operationId to the sanitised form on the source `#[Operation]` attribute.
+ *
+ * The rule stamps the sanitised operationId (via {@see \Radiergummi\OpenApi\Support\Generator\OperationIdDeriver::sanitise()})
+ * onto the finding. An invalid operationId can only have come from an explicit attribute, so the
+ * fixer always targets the existing `#[Operation]` attribute and sets its `operationId:` argument.
+ * Degrades to nothing when the source member, the stamped value, or the attribute is unavailable.
+ *
+ * @internal
+ */
+final readonly class SanitizeOperationIdFixer implements Fixer
+{
+    public const string CONTEXT_OPERATION_ID = 'fixOperationId';
+
+    /**
+     * @return iterable<Fix>
+     */
+    #[Override]
+    public function fix(Finding $finding, FixContext $context): iterable
+    {
+        $class = $finding->context[Finding::CONTEXT_SOURCE_CLASS] ?? null;
+        $member = $finding->context[Finding::CONTEXT_SOURCE_MEMBER] ?? null;
+        $operationId = $finding->context[self::CONTEXT_OPERATION_ID] ?? null;
+
+        if (!is_string($class) || !is_string($member) || !is_string($operationId) || $operationId === '') {
+            return [];
+        }
+
+        [$file, $index] = $this->locateOperationAttribute($class, $member);
+
+        if ($file === null || $index === null) {
+            return [];
+        }
+
+        return [new Fix(
+            file: $file,
+            description: sprintf('Sanitise operationId to "%s" on %s::%s', $operationId, $class, $member),
+            ruleId: $finding->ruleId,
+            operation: new SetAttributeArgument(
+                target: new TargetSelector($class, TargetKind::Method, $member),
+                attributeIndex: $index,
+                argumentName: 'operationId',
+                value: $operationId,
+            ),
+        )];
+    }
+
+    /**
+     * The declaring file and flat, source-order position of the method's `#[Operation]` attribute.
+     * Returns `[null, null]` when the method cannot be reflected or carries no `#[Operation]`.
+     *
+     * @return array{0: ?string, 1: ?int}
+     */
+    private function locateOperationAttribute(string $class, string $member): array
+    {
+        try {
+            $reflector = new ReflectionMethod($class, $member);
+        } catch (ReflectionException) {
+            return [null, null];
+        }
+
+        $file = $reflector->getFileName() ?: null;
+
+        foreach ($reflector->getAttributes() as $index => $attribute) {
+            if (is_a($attribute->getName(), Operation::class, true)) {
+                return [$file, $index];
+            }
+        }
+
+        return [$file, null];
+    }
+}
--- a/src/Lint/Rules/OperationIdInvalidChars.php
+++ b/src/Lint/Rules/OperationIdInvalidChars.php
@@ -5,12 +5,16 @@ declare(strict_types=1);
 namespace Radiergummi\OpenApi\Lint\Rules;
 
 use Override;
-use Radiergummi\OpenApi\Contracts\Lint\Rule;
 use Radiergummi\OpenApi\Contracts\Lint\Severity;
 use Radiergummi\OpenApi\Lint\Finding;
+use Radiergummi\OpenApi\Lint\Fix\FixableRule;
+use Radiergummi\OpenApi\Lint\Fix\Fixer;
+use Radiergummi\OpenApi\Lint\Fix\RemoveAttributeFixer;
+use Radiergummi\OpenApi\Lint\Fix\SanitizeOperationIdFixer;
 use Radiergummi\OpenApi\Lint\LintContext;
 use Radiergummi\OpenApi\Lint\Tree\OperationNode;
 use Radiergummi\OpenApi\Lint\Visitors\OperationRule as OperationRuleVisitor;
+use Radiergummi\OpenApi\Support\Generator\OperationIdDeriver;
 
 use function preg_match;
 use function sprintf;
@@ -19,9 +23,14 @@ use function sprintf;
  * Reports operationIds that don't match `/^[A-Za-z][A-Za-z0-9._-]*$/` (not codegen-safe).
  * Operations without an operationId are skipped; see `operation.id-missing`.
  */
-final class OperationIdInvalidChars implements Rule, OperationRuleVisitor
+final class OperationIdInvalidChars implements FixableRule, OperationRuleVisitor
 {
-    private const string PATTERN = '/^[A-Za-z][A-Za-z0-9._-]*$/';
+    /** The codegen-safe identifier shape; {@see OperationIdDeriver::sanitise()} maps to this set. */
+    public const string PATTERN = '/^[A-Za-z][A-Za-z0-9._-]*$/';
+
+    public function __construct(
+        private readonly OperationIdDeriver $operationIdDeriver = new OperationIdDeriver(),
+    ) {}
 
     /**
      * @return iterable<Finding>
@@ -37,6 +46,19 @@ final class OperationIdInvalidChars implements Rule, OperationRuleVisitor
             return;
         }
 
+        // An invalid operationId can only have come from an explicit attribute, so the fixer always
+        // has an #[Operation] to rewrite; stamp the sanitised value and source member for it.
+        $fixContext = [];
+
+        if ($operation->descriptor !== null) {
+            $fixContext = [
+                ...RemoveAttributeFixer::contextForOperation($operation->descriptor),
+                SanitizeOperationIdFixer::CONTEXT_OPERATION_ID => $this->operationIdDeriver->sanitise(
+                    $operation->operationId,
+                ),
+            ];
+        }
+
         yield new Finding(
             ruleId: $this->id(),
             severity: $this->severity(),
@@ -47,7 +69,14 @@ final class OperationIdInvalidChars implements Rule, OperationRuleVisitor
                 $operation->pathUri,
             ),
             fixHint: 'Use only letters, digits, dots, hyphens and underscores in operationId, starting with a letter.',
+            context: $fixContext,
         );
+    }
+
+    #[Override]
+    public function fixer(): Fixer
+    {
+        return new SanitizeOperationIdFixer();
     }
 
     #[Override]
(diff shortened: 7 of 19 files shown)


Radiergummi:

Scoped to a mergeable **slice 1**: the `AddAttribute` AST primitive + the two operation-level fixers that are attribute-addressable today (`operation.id-missing`, `operation.id-invalid-chars`). All other issue candidates are dispositioned in the PR body (DEFERRED with reason / HUMAN-ONLY) so the catalog is provably exhaustive.

Key constraint driving the slice: only `OperationNode` carries a source back-reference (`descriptor`); `ParameterNode`/`ResponseNode`/header/schema nodes do not, and `SetAttributeArgument` is scalar-only — so the parameter/example/header/naming rows need separate plumbing slices (2–4, listed in the body for the lead to surface; we cannot `gh issue create` here).

Radiergummi:
🤖 **[lead]** plan posted (PR #395) → plan-review. Reviewer: validate the slice-1 boundary (AddAttribute + operation.id-missing/-invalid-chars), the code-constraint rationale (only OperationNode carries a source back-ref), and that every other #382 candidate is dispositioned. area:lint ⇒ no survey.

Radiergummi:
🤖 **[reviewer]** plan-review: changes-needed — 2 required, 1 nit. The slice boundary and litmus discipline are sound; two concrete gaps must be closed before coding.

**Load-bearing claim VERIFIED.** I checked the nodes against `src/Lint/Tree/`:
- `OperationNode` carries `public readonly ?ActionDescriptor $descriptor` (OperationNode.php:48) — the only source back-reference, and `RemoveAttributeFixer::contextForOperation(?ActionDescriptor, ?string)` consumes exactly it (RemoveAttributeFixer.php:71).
- `ParameterNode` (name + `?OA\Parameter $raw` only), `HeaderNode` (name + `?OA\Header $raw`), `FieldNode` (name + `?OA\Property $raw`), `ResponseNode` (field/example/header/link lists + raw) carry **no descriptor and no attribute back-reference**.

So the deferral justification holds: only operation-level rules are attribute-addressable today. Slice 1 = `operation.id-missing` + `operation.id-invalid-chars` is correctly drawn, and the allowed set in the plan (`/^[A-Za-z][A-Za-z0-9._-]*$/`) matches `OperationIdInvalidChars::PATTERN` exactly (OperationIdInvalidChars.php:24). `#[Operation(operationId: …)]` is the right target — `Operation.php:28` has the `?string $operationId` arg; no separate `#[OperationId]` attribute exists, and the plan does not assume one. Good.

**REQUIRED 1 — the operationId-derivation helper the fixer must reuse is PRIVATE to a generation stage; the plan must name the extraction.**
The plan says "match the generator's own operationId derivation … locate that helper during coding and call it, do not reinvent." That logic is `PathsStage::buildOperationId()` / `routeNameOperationId()` / `methodPathOperationId()` / `sanitiseOperationId()` — all `private` on `Support\Generator\Stages\PathsStage` (PathsStage.php:156-203). They are **not callable from the Fix layer as-is**. `sanitiseOperationId()` (PathsStage.php:198) is already exactly what the `operation.id-invalid-chars` fixer needs (`preg_replace('/[^A-Za-z0-9._-]+/', '_')` then strip leading non-letters). The plan must commit to one mechanism: **extract these into a shared, callable surface** (e.g. a small `Support/` helper that both `PathsStage` and the two fixers call) rather than leaving "locate and call" open — as written a coder cannot call a private method and may silently reinvent/diverge, defeating the "fix equals what inference would emit" requirement. Note `buildOperationId` reads `config('openapi.operation_id_strategy')`, so the extracted surface must preserve that config read. Name the extraction (file + signature) in the plan.

**REQUIRED 2 — disposition table is NOT exhaustive: the #21-folded candidates from the issue comment are missing.** The table covers the issue *body* rows but omits the four candidate groups the lead folded in via the 2026-06-17 reconciliation comment on #382:
- `apifield.conflicting-type` (`FieldConflictingType`) — deterministic `SetAttributeArgument` set `type:`; needs the field/attribute back-ref → **DEFERRED** (slice 2), and the ID-alignment question (`field.conflicting-type` vs `apifield.conflicting-type`) must be flagged per that comment.
- `apifield.invalid-format` (`FieldInvalidFormat`) — same back-ref dependency → **DEFERRED**, same ID note.
- `response.heuristic` / `request.heuristic` — neither rule exists on `main`; per #382 non-goals this catalog is "fixers for existing findings only," so these are **out of scope / authoring-first** (state it explicitly).
- The consciously-dropped prose rows (`summary.missing`, `field.description-missing`, `parameter.description-missing`, `queryparam.description-missing`, `meta.no-suppression-reason`) → fold under the existing **HUMAN-ONLY** row so the catalog is provably exhaustive against the *reconciled* issue, not just its body.

**NIT (no block) — `AddAttribute` no-op semantics.** The plan has the op no-op when the target attribute class already exists ("the fixer decides add-vs-modify"). Good and consistent with the `AddOperationIdFixerTest` "existing `#[Operation(summary:…)]` → `SetAttributeArgument`" case. The plan already lists "target already has the attribute class → no-op" under `AddAttributeOperationTest` — just confirm that test asserts **source-unchanged** (byte-stable), not merely "no throw," so the op contract is pinned independently of the fixer.

Everything else (TDD-first failing tests, idempotent + Pint-clean goldens, `docs/linting.md` fixable indication, CHANGELOG `[Unreleased]` line, the FQCN-form-to-avoid-import-editing decision, controversy flags, follow-on slices 2–4) is correctly specified. Close REQUIRED 1 and 2 and this is approvable.

Radiergummi:
🤖 **[planner]** plan-review changes addressed — amended PR body, ready for re-review.

**REQUIRED 1 (operationId helper extraction) — done.** Added step 0 to "What this PR builds": extract the four private `PathsStage` methods (PathsStage.php:156-203) into a new stateless `src/Support/Generator/OperationIdDeriver.php` with `derive(ActionDescriptor, HttpMethod): string` and `sanitise(string): string`. Pure extraction, no behaviour change; preserves the `config(openapi.operation_id_strategy)` read; `PathsStage` delegates. Both fixers now call the named helper (`derive()` for add, `sanitise()` for invalid-chars) instead of "locate and reinvent". Added `OperationIdDeriverTest` to the test list.

**REQUIRED 2 (exhaustive against reconciled #382) — done.** Disposition table now includes the four #21-folded groups: `apifield.conflicting-type` + `apifield.invalid-format` → DEFERRED slice 2 (with the `field.*` vs `apifield.*` ID-alignment flag carried into the slice-2 PR); `response.heuristic`/`request.heuristic` → OUT OF SCOPE (rules do not exist on `main`; catalog is fixers-for-existing-findings-only); the five dropped prose rows folded under HUMAN-ONLY. Slice-2 follow-on now also lists the `apifield.*` unblock + ID-alignment resolution.

**NIT — done.** `AddAttributeOperationTest` no-op case now explicitly asserts byte-identical output (source-unchanged), not just no-throw.

No new controversy: extraction stays in `Support/`, no `Contracts/`/config/public-signature change.

Radiergummi:
🤖 **[reviewer]** plan-review round 2: approved.

Both required changes are correctly resolved and the nit is closed:

- **REQUIRED 1 (extraction) — resolved, correctly scoped.** Step 0 lifts the four private `PathsStage` methods (PathsStage.php:156-203) into a stateless `final` `src/Support/Generator/OperationIdDeriver.php` exposing `derive(ActionDescriptor, HttpMethod): string` and `sanitise(string): string`. It is a **pure extraction, no behaviour change**: the `config('openapi.operation_id_strategy')` read moves verbatim, `PathsStage` delegates, both fixers call the shared surface so the fix equals exactly what inference emits. Home is right — `Support/`, plugin-agnostic, no `Contracts/` change, no config-key change. `OperationIdDeriverTest` covers both strategies + sanitisation, and the existing PathsStage tests pin the no-behaviour-change guarantee.
- **REQUIRED 2 (exhaustiveness) — resolved.** The disposition table now covers the #21-reconciliation groups: `apifield.conflicting-type` / `apifield.invalid-format` → DEFERRED (slice 2), each carrying the `field.*` vs `apifield.*` rule-ID-alignment flag and routing the decision to the slice-2 PR; `response.heuristic` / `request.heuristic` → OUT OF SCOPE, correctly grounded in the issue's "no new lint rules" non-goal; the five #21 prose rows (`summary.missing`, `field.description-missing`, `parameter.description-missing`, `queryparam.description-missing`, `meta.no-suppression-reason`) → HUMAN-ONLY. The catalog is now provably exhaustive against the reconciled issue.
- **NIT — closed.** `AddAttributeOperationTest`'s already-has-attribute case now asserts the output is **byte-identical to the input source**, pinning the op's no-op contract independently of the fixer.

Slice boundary, litmus discipline, TDD-first goldens (idempotent + Pint-clean), `docs/linting.md` fixable rows, CHANGELOG `[Unreleased]` line, and the controversy flags / follow-on slices 2–4 are all in order. Ready for coding.

Radiergummi:
🤖 **[lead]** plan-review round 2 APPROVED by reviewer1 → coding. Implement per the amended PR-body plan: Step 0 = extract OperationIdDeriver into Support/ (pure, PathsStage delegates); then AddAttribute AST op (+ FixOperationVisitor branch, byte-stable goldens incl no-op source-unchanged); then the two fixers (operation.id-missing via AddAttribute→#[Operation(operationId:)], operation.id-invalid-chars via sanitise). TDD: idempotent + Pint-clean goldens. docs/linting.md catalog rows + CHANGELOG. area:lint ⇒ no survey.

Radiergummi:
🤖 **[coder]** deviation: where the OperationIdDeriver is called.

The approved plan said the *fixers* call `OperationIdDeriver::derive()`/`sanitise()`, reconstructing `ActionDescriptor`+`HttpMethod` from the finding's stamped context. In practice a `Finding` only carries scalar `context` (`array<string,mixed>`), and the golden-fixture harness cannot supply a live `ActionDescriptor` — so reconstructing one inside the fixer is impractical and untestable.

Instead, the **rule** (which holds the live `OperationNode` with `descriptor`+`method`) calls the deriver/sanitiser at check time and stamps the resulting scalar operationId into the finding context (`fixOperationId`); the **fixer** reads that scalar and emits the AST op. This is identical to how `TagDuplicate` stamps its discriminator via `RemoveAttributeFixer::contextForOperation()`, and it honours the `Fixer` contract ("read source via \$context"). `OperationIdDeriver` is still the single source of truth — the fix equals exactly what inference emits — only the call site moved from fixer to rule.

Net effect on behaviour: none. The deriver is shared and byte-for-byte the generator's logic (Step 0 is a pure extraction; the Snapshot job proves no generated-output change).

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed 03577e3e. Ready for review.

**Slice 1 delivered** (per the approved disposition table — only the two BUILT rows):
- **Step 0** — `Support\Generator\OperationIdDeriver` extracted from `PathsStage` (pure, no behaviour change; the Snapshot job is green, proving no generated-output drift). `OperationIdDeriverTest` covers both strategies + sanitisation.
- **`AddAttribute` AST op** + `FixOperationVisitor` branch: synthesizes a new `AttributeGroup`, prepended for a byte-stable position, FQ-name form so the `use` block is never touched. No-op (source-identical, asserted) when the member already carries the attribute class; idempotent. Byte-stable goldens in `AddAttributeOperationTest`.
- **`operation.id-missing`** → `AddOperationIdFixer`: `AddAttribute(Operation, operationId:)` when no `#[Operation]` exists, else `SetAttributeArgument` add-arg. Degrades when descriptor/value unavailable.
- **`operation.id-invalid-chars`** → `SanitizeOperationIdFixer`: `SetAttributeArgument` setting the sanitised id on the existing `#[Operation]`.

Both rules now implement `FixableRule`; the rule stamps the deriver's scalar output onto the finding context and the fixer reads it (see the deviation comment above — net behaviour unchanged, deriver is the single source of truth).

**Bookkeeping:** `docs/linting.md` fixable-rules prose updated (the catalog is severity-only, so no per-row change; `--list` auto-reflects fixability via `FixableRule`). CHANGELOG `[Unreleased]` line appended.

**Gates:** `composer format` → `composer check` (PHPStan level 8, Pint, 2819 tests) all green locally. Full remote matrix green incl swagger-php 5.8 + the byte-exact Snapshot job (run 27712263003/27712263552 on HEAD `03577e3e`).

Radiergummi:
🤖 **[lead]** slice 1 implemented (HEAD 03577e3e); full matrix green incl swagger-php 5.8 + Snapshot → review. Two independent reviewers. Scrutinize the documented deviation (deriver called in the rule, scalar id stamped into Finding context for the fixer; mirrors TagDuplicate).

Radiergummi:
🤖 **[reviewer]** review round 1: approved, no concerns. (Independent pass A. HEAD `03577e3e`; verified in the `feat-382-lint-fix` worktree — read every changed file, ran the four new test files (23 pass), Pint-checked the touched sources, and proved the reflection/AST index invariant the deviation rests on.)

**1. The documented deviation (deriver in the rule, scalar stamped into Finding context) — sound.**
- **Single source of truth preserved.** `OperationIdDeriver` (Support/, `@internal`, stateless) is the *only* place operationId logic lives; both `PathsStage::buildOperation()` (deriver.php call at PathsStage.php:131) and the two rules call it. A fixer never reinvents derivation — it consumes the deriver's scalar output. Step 0 is a verbatim extraction (bodies byte-identical to the old private methods; orphaned imports `array_filter`/`config`/`count`/`preg_replace`/`strtolower` correctly dropped; Snapshot green proves zero generated-output drift).
- **Finding context is the right carrier; nothing leaks.** `Finding.context` is `array<string,mixed>` of scalars — the rule stamps `CONTEXT_SOURCE_CLASS`/`CONTEXT_SOURCE_MEMBER` (via the reused `RemoveAttributeFixer::contextForOperation()`) plus a plain operationId string under `fixOperationId`. No `ActionDescriptor` or live object is placed in context or serialized; the fixer reconstructs nothing — it reflects the named class/member on demand. Determinism: the fix is a pure function of three scalars + reflection.
- **Genuinely mirrors `TagDuplicate`.** Both fixers reuse `contextForOperation()` exactly as the duplicate-removal fixers do; this is the established discriminator-into-context pattern, not a new mechanism. The deviation only moves the *call site* of derivation from fixer to rule (because the rule holds the live `OperationNode->descriptor`/`->method` and the golden harness cannot supply a live descriptor). Behaviour is unchanged.

**2. `AddAttribute` AST op + visitor — correct, format-preserving, byte-stable.**
- Synthesises one `AttributeGroup` with a `FullyQualified` name (FixOperationVisitor.php:347-349) and `array_unshift`s it (prepend → deterministic position, `use` block untouched). Scalar→node via the shared `literalToNode`.
- **True no-op, not just no-throw.** `hasAttributeClass()` (resolving via `NameResolver`'s `resolvedName`, so a short `use`d reference matches the FQCN) short-circuits *before* `$this->applied = true`. The test asserts `->toBe($source)` AND `applied === []` AND `skipped` count 1 AND `modifiedFiles === []` — source-identical, accounted as skipped (AddAttributeOperationTest.php:118-145). Idempotency re-runs the op against its own output and asserts no change.
- Edge coverage present: no-attrs prepend (byte-exact), prepend-above-unrelated (existing attr left byte-identical), multi-arg declared order, idempotent. The `Property`/`Param` target kinds of `AddAttribute` aren't exercised, but neither shipped fixer uses them (both target `Method`); fine for slice 1 as generic-op surface (`locateMember` already handles those kinds for the other ops).

**3. The two fixers emit correct ops; tests would fail without the change.**
- `operation.id-missing` → `AddOperationIdFixer`: `AddAttribute(Operation, operationId:)` when no `#[Operation]` exists, else `SetAttributeArgument` to add the arg (no duplicate attribute — asserted via `substr_count(...'#[Operation')->toBe(1)`). Both degrade paths tested (absent stamped id; unknown member).
- `operation.id-invalid-chars` → `SanitizeOperationIdFixer`: `SetAttributeArgument` setting the sanitised id on the existing `#[Operation]`. Both degrade paths tested (absent value; method with no `#[Operation]`). The rule's premise ("invalid id ⇒ explicit attribute exists") holds — inference-derived ids are already sanitised by the deriver, so they can't trip the invalid-chars rule.
- **Index invariant verified.** The fixers compute the `#[Operation]` flat index via `ReflectionMethod::getAttributes()` enumeration; the visitor's `attributeAt()`/no-op walk counts position across `attrGroups`. I confirmed by reflection that `getAttributes()` returns flat source order across *separate* attribute groups (`#[Tag]`=0, `#[Operation]`=1 on distinct lines) — so reflection index == AST flat index. This is the same contract `RemoveAttributeFixer` relies on. No off-by-group bug.
- **Convergence:** a synthesized `{method}_{path}` id (e.g. `get_users_user_`, trailing `_`) matches `OperationIdInvalidChars::PATTERN`, so an id-missing fix never produces something the invalid-chars rule then flags. The two fixers don't fight.

**4. Step 0 purity + conventions + bookkeeping.** Pure extraction (Snapshot green is the proof; bodies identical). strict-types headers present; `@internal` on the op, deriver, and both fixers; no abbreviations; comments are concise and explain *why* (no issue numbers in code); 100-char width respected; modern PHP (readonly, promoted ctor props, `match` in `literalToNode`, first-class enums). `docs/linting.md` fixable-rules prose extended for both rules with the "exactly what inference emits" framing; `CHANGELOG [Unreleased] → Added` line present. `--list` auto-reflects fixability via `FixableRule`.

No findings. Correct, well-tested across the public surface and its edges, conventions-clean. (Size +455/-77 makes the merge a human gate regardless; the code itself is sound.)

Radiergummi:
🤖 **[reviewer]** review round 1 (pass B): approved — 1 minor coverage finding (non-blocking)

Independent second review. Ran in the worktree on HEAD `03577e3e`: `vendor/bin/pint --test` passed, `composer lint` (PHPStan L8) 0 errors, and the `fix` (49) / `lint` (961) / `generator` (25) groups all green.

**Adversarial focus, point by point:**

1. **The deviation (deriver in the rule, scalar stamped into `Finding` context) — sound.** It genuinely mirrors `TagDuplicate`: both call `RemoveAttributeFixer::contextForOperation($operation->descriptor)` to stamp `sourceClass`/`sourceMember` (+ a discriminator), and this PR adds one extra **scalar** key `fixOperationId`. `OperationIdDeriver` is the single source of truth — the fixer never re-derives, it reads the stamped scalar verbatim — and the fixer is fully deterministic from `(sourceClass, sourceMember, fixOperationId)` + reflection. **No new JSON leak:** `Finding::toArray()`/`jsonSerialize()` already emit `'context' => $this->context` for *every* finding (so `--format=json` already exposes `fixDiscriminator` etc. on `main`). The new scalar id is benign and arguably useful; this is pre-existing surface, not introduced here.

2. **`AddAttribute` op — well covered.** Format-preserving, FQ-name form (use-block untouched), prepended for byte-stable position. `AddAttributeOperationTest` asserts byte-exact output for: no-attribute, prepend-above-an-unrelated-attribute (neighbour byte-identical), **multiple named args in declared order**, **already-has-attribute = source-identical + `skipped` count 1 + `modifiedFiles` empty** (true no-op, not just no-throw), and **idempotent re-run**. The already-has case uses the short `use`-imported `Operation` name, exercising the `resolvedName` match path in `hasAttributeClass()`. Solid.

3. **Both fixers + rules now `FixableRule`.** Add-vs-modify branch (`AddAttribute` when no `#[Operation]`, `SetAttributeArgument` when present) is asserted; degrade paths (missing stamped id, unknown member, no `#[Operation]` for sanitise) covered; flat reflection attribute index aligns with the visitor's flat `attrGroups`/`attrs` walk. Output is Pint-clean by construction.

4. **Step 0 extraction is a verbatim no-op.** Diffed `5d2f0639:PathsStage` against the new `OperationIdDeriver`: `buildOperationId`/`methodPathOperationId`/`routeNameOperationId`/`sanitiseOperationId` moved byte-for-byte, including the `config('openapi.operation_id_strategy')` read; PathsStage delegates via an injected stateless helper and correctly drops the now-orphaned `array_filter`/`config`/`count`/`preg_replace`/`strtolower` imports. Snapshot green + the pre-existing `OperationIdSanitisationTest` (generator end-to-end) confirm zero drift.

5. **Conventions/bookkeeping clean.** strict-types throughout, modern 8.4, no abbreviations, `@internal` present, concise comments, no issue numbers in code. `docs/linting.md` fixable-rules prose updated for both rules; CHANGELOG `[Unreleased]` line added.

**Finding 1 (minor, coverage — not merge-blocking):** the rule-side stamping path introduced by the deviation is untested. `OperationIdMissingTest` / `OperationIdInvalidCharsTest` build operations via `OperationNodeFactory::makeOperation(...)`, which defaults `descriptor: null`, so the `if ($operation->descriptor !== null)` branch — where `OperationIdDeriver::derive()`/`sanitise()` is called and `fixOperationId` is stamped — never executes in any test. The fixer tests inject `fixOperationId` directly via `extraContext`, so nothing asserts the **rule actually stamps the right key with the derived value**. A `forDescriptor(...)`-based case (a real `ActionDescriptor` → assert `$finding->context['fixOperationId']` equals the deriver's output, and `sourceClass`/`sourceMember` are present) would close the one seam the deviation added. The constituents (`derive`/`sanitise`, `contextForOperation`, the fixers) are each independently well-tested, so risk is low — flagging for the lead/coder to add, not to block.

Size (+455/-77 src) means a human merge gate regardless. Correctness + coverage confirmed; recommend merge after the human gate, ideally with Finding 1's test added.

Radiergummi:
🤖 **[coder]** addressed pass-B coverage finding + rebased onto main (#387); pushed e29ed680; CI green.

**Coverage finding closed.** The deviation's rule-side stamping path (the `if ($descriptor !== null)` branch that calls `derive()`/`sanitise()` and stamps `fixOperationId` into `Finding.context`) is now directly tested. Added to both rule tests, via `OperationNodeFactory::forDescriptor(...)`:
- `OperationIdMissingTest`: asserts the finding context carries `sourceClass`/`sourceMember` + the **derived** id (`get_x` for the factory's unnamed `GET /x` route under the route-name strategy), plus a no-descriptor case asserting `context === []`.
- `OperationIdInvalidCharsTest`: asserts the **sanitised** id (`list users!` → `list_users_`) + source member are stamped, plus the no-descriptor degrade.

**Rebased onto `main`** (now at #387 / `ec1cb491`). All six commits replayed cleanly — no conflict; my CHANGELOG line sits at the end of `Added` while #387's landed under `Documentation`, so they didn't collide. Both entries present.

**Gates re-run post-rebase:** `vendor/bin/pint --test` clean, PHPStan L8 no errors, 2837 tests pass. **Full remote matrix green** on HEAD `e29ed680` incl swagger-php 5.8 + byte-exact Snapshot (runs 27713079472 + 27713080146). Merge-ready for the human gate.

Radiergummi:
🤖 **[lead]** Moritz-authorized merge on green. Lint-fix catalog slice 1: AddAttribute AST op + OperationIdDeriver extraction (pure, Snapshot-verified) + operation.id-missing/operation.id-invalid-chars fixers. Two independent reviewer approvals (deviation sound — scalar-only Finding.context, mirrors TagDuplicate; AddAttribute no-op is source-identical; reflection-index==AST-flat-index invariant proven); pass-B coverage finding closed; full matrix green incl swagger-php 5.8 + byte-exact Snapshot. Deferred slices (2: source back-ref plumbing, 3: structured-value AST op for example fixers, 4: convention-rename op) surfaced to Moritz.

Issue #383, opened 2026-06-17
Lint --fix: opt-in level for destructive fixes (--fix=safe|dangerous)


Radiergummi:
## Context

`openapi:lint --fix` applies every fixable finding it can resolve, and `--check` previews the same.
Each `FixableRule` returns a `Fixer` whose `fix()` yields `Fix` value objects (`file`, `description`,
`ruleId`, `operation`); the runner applies them. Today **all** fixes are treated as equally safe —
they edit annotation/attribute sites the developer already owns at the finding's location.

Some planned fixes are different: they **mutate files the developer hand-curates or that live far
from the finding** — e.g. #199 wants to write document-level `@OA\Info` / `Server` /
`SecurityScheme` into `config/openapi.php`. Applying that silently under a plain `--fix` is the kind
of destructive edit other dev tools (codemods, Rector's risky rules, `eslint --fix` vs unsafe fixes)
deliberately gate behind an explicit opt-in.

## Problem

`--fix` is "safe to run on a whim." If destructive fixes ride the same flag, a routine `--fix`
silently rewrites a central config file — surprising, and hard to notice in a large diff. We need a
way for a fixer to declare a fix **potentially destructive**, hold it back from the default `--fix`,
and require a conscious opt-in (plus a safety net) to apply it.

## Proposal

**1. Mark destructiveness at the `Fix` level.** Add a `FixSafety` enum (`Safe`, `Destructive`),
carried on the `Fix` VO and defaulting to `Safe` so every existing fixer is unaffected and its
goldens stay byte-identical. Fix-level (not fixer/rule-level) because a single rule can emit both —
#199's *scaffold* output is `Safe`, its *config auto-write* is `Destructive`.
*(Alternative considered: a marker interface / capability on `Fixer`/`FixableRule`. Coarser, but
a rule-level "has destructive fixes" view can still be derived from the emitted fixes for
`openapi:lint --list` if we want it surfaced there.)*

**2. Partition in the runner.** `--fix` applies only `Safe` fixes. `Destructive` fixes are held
back and counted.

**3. `--fix` takes an optional level `safe | dangerous`, defaulting to `safe`.** So:
- `--fix` (bare) or `--fix=safe` → applies only `Safe` fixes (today's behavior, unchanged).
- `--fix=dangerous` → applies both `Safe` and `Destructive` fixes.

`--check` accepts the same optional value symmetrically (`--check[=safe|dangerous]`): default reports
[...]

PR #396, opened 2026-06-17
feat(#22): AST-era fix-conflict semantics + frozen --check JSON fix-run schema

### LINKED ISSUE #22: `lint --fix` Phase 3: polish
## Summary

Final hardening phase for `openapi:lint --fix`: get the conflict-resolution edge cases right, make the fixed-vs-remaining reporting trustworthy across every output format, lock down the `--check --format=json` schema, and ship the user-facing docs.

## Motivation

Phases 1 and 2 (#20, #21) build the fixer infrastructure and the actual fixers. Once engineers run `--fix` on real codebases, the rough edges show up where multiple fixes land in the same file or on the same line — exactly the cases unit-level golden fixtures tend to miss. If line-number arithmetic drifts after one fix is applied, later fixes in the same file corrupt unrelated source, which is far worse than the lint finding they were meant to resolve. This phase exists to make `--fix` safe to run unattended and to make `--check` a dependable CI gate, with the behaviour documented so users know what to expect.

## Proposed approach

Tier 1. This phase polishes, not extends — no new fixable rules.

- Conflict resolution: harden `FixApplicator::apply()` (per the epic, fixes are grouped by file and applied **bottom-to-top** so earlier edits never invalidate the line numbers of later ones). Cover the edge cases the epic calls out: overlapping fixes, and same-line attribute-group removals (e.g. two findings targeting attributes in one `#[...]` group). Overlapping later fixes are skipped, not applied, and the skip is surfaced in the report. The fixer code lives under `src/Lint` (the `FixApplicator` itself is introduced by Phase 1, #20).
- Reporting: extend each formatter in `src/Console` (`LintCommand` resolves `cli|json|github|markdown` — see `src/Console/LintCommand.php:61`) to show a fixed-vs-remaining summary, so the output distinguishes findings that were auto-resolved from those still needing a human, plus any fixes skipped due to conflict.
- JSON schema: define and freeze the `--fix --check --format=json` output shape (fixed count, remaining `Finding`s, skipped-conflict entries) and assert it in tests. `Finding` fields are `ruleId,level,message,location,fixHint,context,spec` (`src/Lint/Finding.php`), collected via `FindingsCollector` (`src/Lint/FindingsCollector.php`).
- Docs: add a user-facing `--fix` / `--check` page under `docs/` (see `docs/README.md` for the page index).

## Acceptance criteria

- Overlapping fixes are skipped and reported; bottom-to-top ordering never corrupts line numbers, verified by golden before/after fixtures exercising same-file multi-fix and same-line attribute-group cases under `tests/Unit/Lint/`.
- The `--fix --check --format=json` schema is documented and covered by a test asserting the exact output shape.
- A docs page for `--fix` / `--check` is added.
- All modified source still passes `vendor/bin/pint --test`.

## References

Parent: #7 (the `FixApplicator` conflict-resolution model, the `--fix`/`--check` command surface and exit codes, and the fixable-rules inventory all live there). Related phases: #20 (Phase 1: infra + Tier A), #21 (Phase 2: Tier B/C). Verify: `vendor/bin/pest tests/Unit/Lint/`

Radiergummi:
Closes #22

## Implementation plan — #22 `lint --fix` Phase 3: AST-era conflict semantics + fix-run reporting

`<type>=feat` · area:lint + area:cli · tier-1 label on the issue, but this is **Fix-layer polish**,
not an inference measure — no body-scanning. Depends on #382 (merged @ `f13d1abd`: `AddAttribute` +
the two operationId fixers now exist).

> **Read the issue THROUGH its 2026-06-17 lead-reconciliation comment.** The body's central mechanic
> — "group fixes per file, apply **bottom-to-top** so earlier edits don't drift later line numbers" —
> is **DEAD** under #368's AST-mutation + `printFormatPreserving()` backend. There is no shared mutable
> offset space; each file is parsed once, cloned, mutated, and reprinted once. **No line arithmetic is
> planned.** Surviving, backend-agnostic scope only.

### What the code already does (verified on `main` @ f13d1abd)

So the plan targets real gaps, not work already shipped by #368/#382:

- `FixApplicator::apply()` already groups fixes by file, applies one `FixOperationVisitor` per fix on
  a cloned tree, reprints once, and writes atomically (FixApplicator.php:52-88).
- `FixResult` already carries `applied` / `skipped` / `modifiedFiles` (FixResult.php). Today a fix
  lands in `skipped` **only** when its visitor never matched a node (`$visitor->applied === false`) or
  the FP-print threw — i.e. *mechanical* failure, **not** a detected same-node conflict.
- `FixRunResult` already carries `fixResult` + `remainingFindings` + the `--check` dry-run exit-code
  logic (FixRunResult.php). `FixRunner` already maps applied fixes → resolved findings (FixRunner.php).
- `LintCommand::renderFixSummary()` already prints a human "N fix(es) skipped due to overlapping
  edits" line — but **only on the human `info`/`warn` channel**, never through the four formatters,
  and the remaining findings are rendered through formatters with **no fix-run envelope** at all
  (LintCommand.php:420-459, 253-261).
- A `--fix`/`--check` section **already exists** in `docs/linting.md` (lines 124-226).

### The two real gaps this PR closes

**Gap 1 — same-node conflict is not detected (the correctness gap).** `applyToFile()` registers every
fix's visitor on one traverser and runs them in registration order against the same cloned nodes
(FixApplicator.php:113-124). Two ops on the *same* node therefore mutate it in sequence with no guard:
- two `SetAttributeArgument` on the same `attributeIndex`+`argumentName` → silent last-wins;
- a `RemoveAttribute` dropping an index that a `SetAttributeArgument`/`AddAttribute`-modify on the same
  member depends on → the surviving op mutates a shifted/absent attribute.
Both are reported as `applied` today. Per the reconciliation comment: **detect, apply the safe subset,
skip the conflicting remainder, surface the skip.**

**Gap 2 — fix-run reporting/JSON schema.** The skipped dimension and the applied/remaining/skipped
envelope never reach the formatters; `--check --format=json` emits only the remaining `LintResult`
(JsonFormatter.php:42-57), so CI cannot read "what would be fixed / what was skipped" as data.

### Design

**1. Conflict detector — `src/Lint/Fix/Ast/FixConflictDetector.php` (new, `@internal`).**
A pure pre-apply pass, run by `FixApplicator::apply()` before `applyToFile()` (or inside it, per
file — confirm placement during coding; per-file is enough since cross-file ops never share a node).
- Group the file's fixes by their op's `TargetSelector` identity (`className` + `kind` + `memberName`
  — `TargetSelector.php`). Different nodes never conflict.
- Within a target group, partition into a **kept (safe)** subset and a **skipped (conflicting)** subset
  with a **deterministic, stable rule** (first-in-list wins, so re-running `--fix` makes progress):
  - Provably independent (all KEPT): multiple `RemoveAttribute` whose `attributeIndices` are pairwise
    disjoint (the visitor already removes descending and order-independently). A single op of any kind.
  - Conflicting (KEEP first, SKIP rest): two ops whose effects on the same node are not provably
    independent — same `attributeIndex` touched by more than one of {`SetAttributeArgument`,
    `RemoveAttribute`}; two `SetAttributeArgument` on the same `(attributeIndex, argumentName)`;
    an `AddAttribute` on a target another op also mutates (index space changes under insertion).
  - **The exact pair matrix is small and finite (4 op types); enumerate it explicitly in the detector
    with a comment, and pin it with unit tests.** Bias toward *correctness over coverage*: when in
    doubt, skip rather than risk a wrong mutation (skips are recoverable by re-running; corruption is
    not). Flag any pair the coder finds ambiguous for the reviewer rather than guessing.
- Return kept + skipped(with a reason string). `applyToFile()` then only registers kept visitors; the
  skipped fixes join `FixResult::$skipped`. **The `skipped` set thus gains a second source (detected
  conflict) alongside the existing mechanical one** — both already flow to the same field, so
  `FixRunner`'s remaining-findings logic needs no change (a finding whose fix was skipped stays
  unresolved, which is already the behaviour via `wasFixed()`).

**2. Carry a skip *reason*.** Today `skipped` is `list<Fix>` with no reason. Add a lightweight
`list<SkippedFix>` (`Fix` + `string $reason` enum/const: `node-not-found`, `print-failed`, `conflict`)
on `FixResult`, or keep `list<Fix>` and add a parallel reasons map — **decision:** introduce a small
`SkippedFix` readonly VO (`Fix $fix`, `FixSkipReason $reason` enum) for a typed, testable surface;
update the three existing skip sites in `FixApplicator` to tag their reason. Confirm this is the only
`FixResult` shape change. *(This is an internal `@internal` Fix-layer VO, not a public contract.)*

**3. Reporting across formatters.**
- **JSON (`--check`/`--fix --format=json`) — freeze the schema.** The fix run must emit a stable
  object, not just the remaining findings. Add a fix-run payload to `JsonFormatter` (the surgical
  route: `JsonFormatter` gains an optional fix-run section, or — cleaner — the command builds a
  dedicated fix-run JSON when `--fix/--check` + json, reusing `Finding`'s existing `JsonSerializable`).
  Frozen shape (bump `JsonFormatter::SCHEMA_VERSION` only if the lint shape itself changes; the fix
  envelope is additive):
  ```json
  {
    "schema_version": "1",
    "mode": "check" | "fix",
    "applied": <int>,
    "skipped": [ { "rule_id": "...", "file": "...", "reason": "conflict|node-not-found|print-failed" } ],
    "modified_files": [ "..." ],
    "remaining": [ <Finding>, ... ],
    "exit_code": <int>
  }
  ```
  Exact key names/casing to match the existing JSON formatter's snake_case convention (it uses
  `schema_version`, `exit_code`). **Decision to flag:** whether the fix-run JSON lives in
  `JsonFormatter` (fix-aware) or is assembled in the command. Recommend a small
  `FixRunResult::toArray()` / dedicated serializer so the shape is owned in one place and unit-testable
  without the console — flag for the reviewer.
- **Human formats (cli / github / markdown).** Extend the existing summary so all three surface
  fixed / remaining / **skipped(+reason)** consistently. The remaining findings already render through
  each formatter; add a one-line fix-run summary line per format (CLI already has the prose at
  LintCommand.php:443-456 — move/share it so github + markdown get the same line; github should emit
  the skip as a workflow notice/warning, markdown as a bullet). Keep it minimal — no per-format
  redesign.

Under `tests/Unit/Lint/Fix/` (fixtures under `tests/Fixtures/Lint/Fix/`):

- **`FixConflictDetectorTest`** (the heart):
  - two `SetAttributeArgument` on the same `(attributeIndex, argumentName)` on one node → first KEPT,
    second SKIPPED(`conflict`).
  - `RemoveAttribute[0]` + `SetAttributeArgument` on index 0, same member → the dependent op SKIPPED.
  - two `RemoveAttribute` with **disjoint** indices on one member → both KEPT (no false positive).
  - ops on **different** members of the same class → both KEPT.
  - ops in **different** files → untouched (detector is per-target).
  - determinism: the kept/skipped partition is stable across runs (first-wins).
- **`FixApplicatorConflictTest`** (golden, byte-stable): a fixture file with two findings that produce
  a same-node conflict → after `apply()`, the safe subset is written, output is **byte-stable** and
  Pint-clean, the skipped fix is returned with `reason: conflict`, and a **re-run applies the
  previously-skipped fix** (idempotent progress).
- **`--check --format=json` shape assertion** (`tests/Feature/Console/` or the existing lint-command
  test home — locate it): run `openapi:lint --check --format=json` against a fixture app with a mix of
  applied + skipped + remaining, assert the **exact** decoded array shape (all keys present, types,
  `skipped[].reason`, `exit_code`). Edge cases as separate assertions: **zero fixes** (empty applied,
  empty skipped, `exit_code` 0), **all-skipped** (applied 0, skipped non-empty), **mixed**.
- **Per-format summary**: assert cli/github/markdown each emit the fixed/remaining/skipped line
  (expectsOutput-style), including the github notice form and markdown bullet form.

### Docs & changelog

- `docs/linting.md` — the `--fix`/`--check` section already exists (lines 124-226). **Extend it**
  (do not add a new page): document the skipped-conflict behaviour ("conflicting fixes on the same
  node are skipped and reported; re-run `--fix` to apply them") and add the frozen
  `--check --format=json` schema as a fenced block. `docs/README.md` already indexes linting — no new
  index row needed (confirm during coding; add one only if a new page is created, which it should not
  be).
- `CHANGELOG.md` `[Unreleased]`: one line — `--fix`/`--check` now detect same-node fix conflicts
  (apply the safe subset, skip + report the rest) and emit a frozen `--check --format=json` fix-run
  schema. (#22)

### Controversy flags for the lead

- **area:lint + area:cli ⇒ no survey** (non-generation-affecting).
- **No `src/Contracts/**` change, no config key, no public CLI-signature change** — confirmed: the
  `--fix`/`--check`/`--format` flags already exist; new code is the `@internal` `FixConflictDetector`,
  an `@internal` `SkippedFix`/`FixSkipReason` VO, a `FixResult` field-shape change (internal), and a
  formatter/JSON additive section. **One judgement call to confirm at review:** the
  `--check --format=json` shape is a *de facto public contract for CI consumers* once frozen — that is
  the point of the issue, but it means the JSON key set is now load-bearing. Flag so the lead/reviewer
  treats the schema block + its test as the stable surface.
- **Size risk:** the conflict detector + skip-reason VO + per-formatter wiring + JSON envelope is
  likely **> ~150 src/ LOC**, above the fast-path mechanical threshold → expect a human merge gate.
  If the conflict pair-matrix turns out larger than the four-op enumeration suggests, it grows further.
- **Out-of-scope discoveries to surface (cannot `gh issue create` here):**
  - The github/markdown summary-line sharing may want a small shared "fix-run summary" renderer; if it
    becomes more than a few lines, that is a candidate refactor, not part of this issue.
  - `FixRunResult::toArray()` (if chosen) is the natural seam the next cluster item **#383**
    (`--fix=safe|dangerous`) will extend; note the boundary so #383 doesn't re-cut it.

---

### Reviewer addendum (plan-review, reviewer1)

Verified every load-bearing claim against `main` @ `f13d1abd`; all hold. Three tightenings folded in
(none change the design):

1. **Grouping key is provably correct — affirmed, not just flagged.** The lead's question "can two
   *different* `TargetSelector`s alias the same node?" → **no.** A `Method` target resolves to the
   `ClassMethod` (its own `attrGroups`); a `Property` target resolves to the `Property` statement
   **or** a promoted constructor `Param` (`FixOperationVisitor::findProperty`, properties-first, so it
   is deterministic). The promoted-param node is reachable **only** via `(class, Property, name)`,
   never via a `Method` selector — so *same attribute-bearing node ⟺ same `(className, kind,
   memberName)` identity*. Grouping by `TargetSelector` identity is sound; no aliasing gap. (Existing
   emitters confirm only `ClassNode`/`Method`/`Property` selectors are ever produced.)

2. **Reword the existing skip message — it still describes the dead byte-overlap model.**
   `LintCommand::renderFixSummary()` (LintCommand.php:449-456) prints *"N fix(es) skipped due to
   **overlapping edits**"* and `FixResult`'s docblock says *"dropped due to overlap with an
   already-applied edit"* (FixResult.php:11). Both predate the rescope and are now inaccurate under the
   AST-reprint backend. The reporting work **must** reword these to AST-era "same-node conflict"
   semantics (and the `FixResult` docblock once `SkippedFix`/`FixSkipReason` lands), so the message
   matches what actually happened. Add this to the per-formatter summary task.

3. **Pin the fix-run JSON as a distinct envelope; the lint `findings`/`summary` shape stays
   untouched.** Today `JsonFormatter` emits `{schema_version, level, exit_code, findings, summary
   (+coverage?)}` (JsonFormatter.php:42-57). The frozen fix-run object uses a top-level `remaining`
   key (not `findings`) — so it is a **separate envelope**, not the lint payload plus keys. State
   explicitly that existing lint-JSON consumers (reading `findings`/`summary`/`exit_code`) are
   unaffected, and that the fix-run `remaining[]` entries reuse `Finding`'s existing
   `JsonSerializable` shape verbatim. This resolves the planner's "lives in JsonFormatter vs assembled
   in command" decision in favour of the recommended one-place serializer (`FixRunResult::toArray()`),
   which is also the #383 seam.

**Gap 1 (correctness) confirmed at the code:** `applyToFile()` registers every fix's visitor on one
`NodeTraverser` and runs them in a single pass over the shared cloned tree (FixApplicator.php:113-124);
two ops on the same node both set `$visitor->applied = true` and both land in `accepted` (lines
129-135) with no guard. The detector is the minimal correct fix; there is no existing seam to reuse
(the per-fix visitor isolation that would otherwise catch this does not exist — they share one
traversal). First-wins is deterministic and a re-run converges because the skipped op is re-emitted by
its rule on the next lint pass and now has no surviving conflictor.

**Verdict: approved** with tightenings 2 and 3 to be honoured during coding (both are small, in-scope
wording/seam clarifications, not design changes). Proving tests, docs-extension (not a new page),
CHANGELOG, and the #383 seam note are all correctly specified.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,7 @@ All notable changes to this project are documented here.
 - `openapi:lint --format=markdown` now renders a real GitHub-Flavored Markdown body (a coverage summary line and a `Severity | Rule | Location | Message` findings table) instead of aliasing the CLI tree dump, so the CI coverage-comment recipe can post it verbatim. (#389)
 - ApiResources: the return-expression reader now recognises the `X::collect(...)` static resource-collection factory, emitting the same collection response schema as `X::collection(...)`. (#390)
 - `openapi:lint --fix` now repairs missing and codegen-unsafe operationIds (synthesizes/sanitizes the `#[Operation(operationId: …)]`), via the new `AddAttribute` fix primitive. (#382)
+- `--fix`/`--check` now detect same-node fix conflicts (apply the safe subset, skip + report the rest with a typed reason) and emit a frozen `--check --format=json` fix-run envelope for CI. (#22)
 
 ### Changed
 - `spatie/laravel-data` is now a soft dependency (moved from `require` to `require-dev`); Fractal and query-builder packages are opt-in.
--- a/docs/linting.md
+++ b/docs/linting.md
@@ -156,12 +156,13 @@ would have emitted.
 ### Conflicting fixes
 
 When two fixes would touch the **same** code node (for example two rules that
-both rewrite the same attribute argument, or a removal an edit on the same
-attribute depends on), applying them together is unsafe. The fixer keeps the
-first of the conflicting set and **skips** the rest, reporting them rather than
-guessing — a skip is recoverable, a wrong edit is not. Re-run `--fix`: the
-skipped fix is re-emitted on the next pass and, with its conflictor already
-resolved, applies cleanly. The summary line reports `N skipped (conflict: N)`.
+both rewrite the same attribute argument, or a removal alongside another edit on
+the same member, where removing one attribute shifts the others), applying them
+together is unsafe. The fixer keeps the first of the conflicting set and
+**skips** the rest, reporting them rather than guessing — a skip is recoverable,
+a wrong edit is not. Re-run `--fix`: the skipped fix is re-emitted on the next
+pass and, with its conflictor already resolved, applies cleanly. The summary
+line reports `N skipped (conflict: N)`.
 
 ### Machine-readable fix runs (`--check --format=json`)
 
--- a/docs/linting.md
+++ b/docs/linting.md
@@ -153,6 +153,41 @@ exists) and `operation.id-invalid-chars` (rewrites the operationId to the
 codegen-safe character set). Both produce exactly the operationId inference
 would have emitted.
 
+### Conflicting fixes
+
+When two fixes would touch the **same** code node (for example two rules that
+both rewrite the same attribute argument, or a removal an edit on the same
+attribute depends on), applying them together is unsafe. The fixer keeps the
+first of the conflicting set and **skips** the rest, reporting them rather than
+guessing — a skip is recoverable, a wrong edit is not. Re-run `--fix`: the
+skipped fix is re-emitted on the next pass and, with its conflictor already
+resolved, applies cleanly. The summary line reports `N skipped (conflict: N)`.
+
+### Machine-readable fix runs (`--check --format=json`)
+
+`--fix` / `--check` with `--format=json` emit a **stable fix-run envelope** for
+CI, distinct from the lint JSON (note the top-level `remaining`, not
+`findings`, so existing lint-JSON consumers are unaffected):
+
+```json
+{
+  "schema_version": "1",
+  "mode": "check",
+  "applied": 3,
+  "skipped": [
+    { "rule_id": "operation.id-invalid-chars", "file": "app/Http/Controllers/UserController.php", "reason": "conflict" }
+  ],
+  "modified_files": ["app/Http/Controllers/UserController.php"],
+  "remaining": [],
+  "exit_code": 1
+}
+```
+
+`mode` is `check` (dry run) or `fix`; `applied` counts the fixes applied (or, in
+`check` mode, that would apply); each `skipped` entry carries a `reason`
+(`conflict`, `node-not-found`, or `print-failed`); `remaining` holds the lint
+findings still unresolved, in the same shape the lint JSON uses.
+
 ## Documentation-coverage gate
 
 `openapi:lint` derives a **documentation-coverage** metric from its findings: the share of API
--- a/src/Console/LintCommand.php
+++ b/src/Console/LintCommand.php
@@ -44,6 +44,7 @@ use Symfony\Component\TypeInfo\Exception\UnsupportedException;
 use UnexpectedValueException;
 
 use function array_filter;
+use function array_keys;
 use function array_map;
 use function array_values;
 use function count;
@@ -51,11 +52,18 @@ use function explode;
 use function fclose;
 use function fopen;
 use function getenv;
+use function implode;
 use function is_array;
 use function is_numeric;
 use function is_resource;
 use function is_string;
+use function json_encode;
+use function ksort;
 use function sprintf;
+
+use const JSON_PRETTY_PRINT;
+use const JSON_THROW_ON_ERROR;
+use const JSON_UNESCAPED_SLASHES;
 
 /**
  * Lints OpenAPI documentation gaps across the API surface.
@@ -248,19 +256,59 @@ class LintCommand extends Command
 
         $outcome = $fixRunner->run($options, $dryRun);
 
-        $this->renderFixSummary($outcome, $dryRun);
-
-        if ($outcome->remainingFindings !== []) {
-            $this->renderToTargets(
-                new LintResult(
-                    findings: $outcome->remainingFindings,
-                    level: $outcome->level,
-                    exitCode: $outcome->exitCode(),
-                ),
-            );
-        }
+        $this->renderFixOutcome($outcome);
 
         return $outcome->exitCode();
+    }
+
+    /**
+     * Renders the fix run to every requested target. A JSON target emits the frozen fix-run envelope
+     * ({@see FixRunResult::toArray()}); the human formats emit a one-line fixed/remaining/skipped
+     * summary plus the remaining findings through their normal formatter.
+     *
+     * @throws BindingResolutionException
+     * @throws InvalidArgumentException
+     * @throws JsonException
+     * @throws RuntimeException
+     */
+    private function renderFixOutcome(FixRunResult $outcome): void
+    {
+        $remaining = new LintResult(
+            findings: $outcome->remainingFindings,
+            level: $outcome->level,
+            exitCode: $outcome->exitCode(),
+        );
+
+        foreach ($this->resolveTargets() as $target) {
+            $output = $this->openOutput($target->target);
+
+            try {
+                if ($target->format === LinterOutputFormat::Json) {
+                    $output->writeln($this->encodeFixRun($outcome));
+
+                    continue;
+                }
+
+                $this->renderFixSummary($outcome, $target->format, $output);
+
+                if ($outcome->remainingFindings !== []) {
+                    $this->formatterFor($target->format)->render($remaining, $output);
+                }
+            } finally {
+                $this->closeOutput($output);
+            }
+        }
+    }
+
+    /**
+     * @throws JsonException
+     */
+    private function encodeFixRun(FixRunResult $outcome): string
+    {
+        return json_encode(
+            value: $outcome->toArray(),
+            flags: JSON_THROW_ON_ERROR | JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES,
+        );
     }
 
     /**
@@ -417,45 +465,47 @@ class LintCommand extends Command
         return (int) $raw;
     }
 
-    private function renderFixSummary(FixRunResult $outcome, bool $dryRun): void
+    /**
+     * Writes a one-line fixed/remaining/skipped summary in the target's format. GitHub emits a
+     * workflow notice, Markdown a bullet, CLI plain prose; the skipped count is surfaced (with the
+     * conflict reason) so a re-run hint is actionable.
+     */
+    private function renderFixSummary(FixRunResult $outcome, LinterOutputFormat $format, OutputInterface $output): void
     {
-        $files = $outcome->fixResult->modifiedFiles;
-        $count = count($outcome->fixResult->applied);
+        $applied = count($outcome->fixResult->applied);
+        $remaining = count($outcome->remainingFindings);
+        $skipped = count($outcome->fixResult->skipped);
+        $verb = $outcome->dryRun ? 'pending' : 'fixed';
 
-        if ($count === 0) {
-            $this->info('openapi:lint --fix: nothing to fix.');
+        $summary = sprintf('openapi:lint --fix: %d %s, %d remaining, %d skipped', $applied, $verb, $remaining, $skipped);
 
-            return;
+        if ($skipped > 0) {
+            $summary .= sprintf(' (%s); re-run --fix to resolve conflicting fixes', $this->skipReasonBreakdown($outcome));
         }
 
-        if ($dryRun) {
-            $this->warn(
-                sprintf(
-                    '%d fixable finding(s) pending across %d file(s). Run `php artisan openapi:lint --fix` to apply them.',
-                    $count,
-                    count($files),
-                ),
-            );
+        $output->writeln(match ($format) {
+            LinterOutputFormat::GitHub => '::notice title=OpenAPI fix::' . $summary,
+            LinterOutputFormat::Markdown => '- ' . $summary,
+            default => $summary,
+        });
+    }
 
-            return;
+    /** A compact `reason: count` breakdown of skipped fixes, e.g. `conflict: 2, node-not-found: 1`. */
+    private function skipReasonBreakdown(FixRunResult $outcome): string
+    {
+        $counts = [];
+
+        foreach ($outcome->fixResult->skipped as $skip) {
+            $counts[$skip->reason->value] = ($counts[$skip->reason->value] ?? 0) + 1;
         }
 
-        $this->info(sprintf('Fixed %d finding(s) across %d file(s):', $count, count($files)));
+        ksort($counts);
 
-        foreach ($files as $file) {
-            $this->line('  ' . $file);
-        }
-
-        if ($outcome->fixResult->skipped !== []) {
-            $this->warn(
-                sprintf(
-                    '%d fix(es) skipped due to overlapping edits; re-run --fix to resolve the rest.',
-                    count($outcome->fixResult->skipped),
-                ),
-            );
-        }
-
-        $this->line('Run your formatter on the changes, e.g., `vendor/bin/pint --dirty`.');
+        return implode(', ', array_map(
+            static fn(string $reason, int $count): string => sprintf('%s: %d', $reason, $count),
+            array_keys($counts),
+            array_values($counts),
+        ));
     }
 
     /**
--- a/src/Lint/Fix/Ast/FixConflictDetector.php
+++ b/src/Lint/Fix/Ast/FixConflictDetector.php
@@ -1,0 +1,131 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Lint\Fix\Ast;
+
+use Radiergummi\OpenApi\Lint\Fix\Fix;
+use Radiergummi\OpenApi\Lint\Fix\FixSkipReason;
+use Radiergummi\OpenApi\Lint\Fix\SkippedFix;
+
+use function array_intersect;
+use function in_array;
+
+/**
+ * Partitions a batch of {@see Fix}es into a safe subset to apply and a conflicting subset to skip.
+ *
+ * Every fix's visitor runs in a single pass over one shared cloned tree, so two operations on the
+ * *same* node mutate it in sequence with no guard (silent last-wins, or one op acting on an index
+ * another shifted). This detector finds those same-node conflicts up front and keeps only a subset
+ * whose effects are provably independent.
+ *
+ * Grouping is by {@see TargetSelector} identity (`className` + `kind` + `memberName`): two different
+ * selectors can never address the same attribute-bearing node, so fixes in different groups never
+ * conflict. Within a group the rule is **first-in-list wins**: a fix is kept only when it is provably
+ * independent of every already-kept fix; otherwise it is skipped with {@see FixSkipReason::Conflict}.
+ * First-wins is deterministic and lets a re-run converge, the skipped op is re-emitted on the next
+ * lint pass with no surviving conflictor.
+ *
+ * @internal
+ */
+final readonly class FixConflictDetector
+{
+    /**
+     * @param list<Fix> $fixes
+     *
+     * @return array{kept: list<Fix>, skipped: list<SkippedFix>}
+     */
+    public function partition(array $fixes): array
+    {
+        /** @var array<string, list<Fix>> $byTarget */
+        $byTarget = [];
+
+        foreach ($fixes as $fix) {
+            $byTarget[$this->targetKey($fix)][] = $fix;
+        }
+
+        $kept = [];
+        $skipped = [];
+
+        foreach ($byTarget as $group) {
+            /** @var list<Fix> $groupKept */
+            $groupKept = [];
+
+            foreach ($group as $fix) {
+                $conflicts = false;
+
+                foreach ($groupKept as $keptFix) {
+                    if ($this->conflicts($keptFix, $fix)) {
+                        $conflicts = true;
+
+                        break;
+                    }
+                }
+
+                if ($conflicts) {
+                    $skipped[] = new SkippedFix($fix, FixSkipReason::Conflict);
+
+                    continue;
+                }
+
+                $groupKept[] = $fix;
+                $kept[] = $fix;
+            }
+        }
+
+        return ['kept' => $kept, 'skipped' => $skipped];
+    }
+
+    private function targetKey(Fix $fix): string
+    {
+        $target = $fix->operation->target;
+
+        return $target->className . "\0" . $target->kind->name . "\0" . ($target->memberName ?? '');
+    }
+
+    /**
+     * Whether two fixes on the *same* node cannot be safely applied together in one traversal.
+     *
+     * The op set is small and finite, so the matrix is enumerated explicitly. Bias toward
+     * correctness: when independence is not provable, treat the pair as conflicting (a skip is
+     * recoverable by re-running; a wrong mutation is not).
+     */
+    private function conflicts(Fix $a, Fix $b): bool
+    {
+        $first = $a->operation;
+        $second = $b->operation;
+
+        // An insertion prepends a new attribute group, shifting the flat index space every other
+        // attribute op resolves against, so it is never provably independent of another op here.
+        if ($first instanceof AddAttribute || $second instanceof AddAttribute) {
+            return true;
+        }
+
+        // Doc-comment edits touch a different node aspect (comments, not attrGroups), so they are
+        // independent of attribute ops; two doc-comment edits on one node both rewrite it.
+        if ($first instanceof SetDocComment || $second instanceof SetDocComment) {
+            return $first instanceof SetDocComment && $second instanceof SetDocComment;
+        }
+
+        if ($first instanceof RemoveAttribute && $second instanceof RemoveAttribute) {
+            return array_intersect($first->attributeIndices, $second->attributeIndices) !== [];
+        }
+
+        if ($first instanceof SetAttributeArgument && $second instanceof SetAttributeArgument) {
+            // Distinct attributes never collide; the same attribute collides only on the same argument
+            // (last-wins), distinct arguments add independent named args to that attribute.
+            return $first->attributeIndex === $second->attributeIndex
+                && $first->argumentName === $second->argumentName;
+        }
+
+        // The remaining pairs mix RemoveAttribute with SetAttributeArgument: a conflict iff the set
+        // operates on an index the remove drops (the survivor would mutate a shifted/absent attribute).
+        [$remove, $set] = $first instanceof RemoveAttribute
+            ? [$first, $second]
+            : [$second, $first];
+
+        return $remove instanceof RemoveAttribute
+            && $set instanceof SetAttributeArgument
+            && in_array($set->attributeIndex, $remove->attributeIndices, true);
+    }
+}
--- a/src/Lint/Fix/Ast/FixConflictDetector.php
+++ b/src/Lint/Fix/Ast/FixConflictDetector.php
@@ -8,9 +8,6 @@ use Radiergummi\OpenApi\Lint\Fix\Fix;
 use Radiergummi\OpenApi\Lint\Fix\FixSkipReason;
 use Radiergummi\OpenApi\Lint\Fix\SkippedFix;
 
-use function array_intersect;
-use function in_array;
-
 /**
  * Partitions a batch of {@see Fix}es into a safe subset to apply and a conflicting subset to skip.
  *
@@ -89,15 +86,20 @@ final readonly class FixConflictDetector
      * The op set is small and finite, so the matrix is enumerated explicitly. Bias toward
      * correctness: when independence is not provable, treat the pair as conflicting (a skip is
      * recoverable by re-running; a wrong mutation is not).
+     *
+     * Every kept op runs against the *same* progressively-mutated tree in one traversal, so an op
+     * that changes the member's attribute layout, inserting or removing an attribute group (which
+     * re-indexes the flat attribute list), invalidates the indices every later op resolves against.
+     * Both {@see AddAttribute} and {@see RemoveAttribute} are therefore treated as conflicting with
+     * any other op on the same node; a re-run converges because the skipped op is re-emitted next
+     * pass with fresh indices and no surviving layout-changer.
      */
     private function conflicts(Fix $a, Fix $b): bool
     {
         $first = $a->operation;
         $second = $b->operation;
 
-        // An insertion prepends a new attribute group, shifting the flat index space every other
-        // attribute op resolves against, so it is never provably independent of another op here.
-        if ($first instanceof AddAttribute || $second instanceof AddAttribute) {
+        if ($this->changesLayout($first) || $this->changesLayout($second)) {
             return true;
         }
 
@@ -107,25 +109,18 @@ final readonly class FixConflictDetector
             return $first instanceof SetDocComment && $second instanceof SetDocComment;
         }
 
-        if ($first instanceof RemoveAttribute && $second instanceof RemoveAttribute) {
-            return array_intersect($first->attributeIndices, $second->attributeIndices) !== [];
-        }
+        // The only remaining pair is SetAttributeArgument vs SetAttributeArgument (neither changes
+        // the layout): distinct attributes never collide; the same attribute collides only on the
+        // same argument (last-wins), distinct arguments add independent named args to it.
+        return $first instanceof SetAttributeArgument
+            && $second instanceof SetAttributeArgument
+            && $first->attributeIndex === $second->attributeIndex
+            && $first->argumentName === $second->argumentName;
+    }
 
-        if ($first instanceof SetAttributeArgument && $second instanceof SetAttributeArgument) {
-            // Distinct attributes never collide; the same attribute collides only on the same argument
-            // (last-wins), distinct arguments add independent named args to that attribute.
-            return $first->attributeIndex === $second->attributeIndex
-                && $first->argumentName === $second->argumentName;
-        }
-
-        // The remaining pairs mix RemoveAttribute with SetAttributeArgument: a conflict iff the set
-        // operates on an index the remove drops (the survivor would mutate a shifted/absent attribute).
-        [$remove, $set] = $first instanceof RemoveAttribute
-            ? [$first, $second]
-            : [$second, $first];
-
-        return $remove instanceof RemoveAttribute
-            && $set instanceof SetAttributeArgument
-            && in_array($set->attributeIndex, $remove->attributeIndices, true);
+    /** Whether the op inserts or removes an attribute group, re-indexing the flat attribute list. */
+    private function changesLayout(AstOperation $operation): bool
+    {
+        return $operation instanceof AddAttribute || $operation instanceof RemoveAttribute;
     }
 }
--- a/src/Lint/Fix/FixApplicator.php
+++ b/src/Lint/Fix/FixApplicator.php
@@ -11,10 +11,12 @@ use PhpParser\NodeVisitor\NameResolver;
 use PhpParser\Parser;
 use PhpParser\ParserFactory;
 use PhpParser\PrettyPrinter\Standard;
+use Radiergummi\OpenApi\Lint\Fix\Ast\FixConflictDetector;
 use Radiergummi\OpenApi\Lint\Fix\Ast\FixOperationVisitor;
 use RuntimeException;
 use Throwable;
 
+use function array_map;
 use function dirname;
 use function file_get_contents;
 use function file_put_contents;
@@ -38,10 +40,13 @@ final readonly class FixApplicator
 
     private Standard $printer;
 
+    private FixConflictDetector $conflictDetector;
+
     public function __construct()
     {
         $this->parser = new ParserFactory()->createForNewestSupportedVersion();
         $this->printer = new Standard();
+        $this->conflictDetector = new FixConflictDetector();
     }
 
     /**
@@ -63,16 +68,24 @@ final readonly class FixApplicator
         $modifiedFiles = [];
 
         foreach ($byFile as $file => $fileFixes) {
+            // Keep only fixes whose effects on a shared node are provably independent; the rest are
+            // skipped as conflicts before any mutation so a same-node clash never lands silently.
+            ['kept' => $kept, 'skipped' => $conflicts] = $this->conflictDetector->partition($fileFixes);
+
+            foreach ($conflicts as $conflict) {
+                $skipped[] = $conflict;
+            }
+
             $source = file_get_contents($file) ?: '';
 
-            [$accepted, $rejected, $newSource] = $this->applyToFile($source, $fileFixes);
+            [$accepted, $rejected, $newSource] = $this->applyToFile($source, $kept);
 
             foreach ($accepted as $fix) {
                 $applied[] = $fix;
             }
 
-            foreach ($rejected as $fix) {
-                $skipped[] = $fix;
+            foreach ($rejected as $skip) {
+                $skipped[] = $skip;
             }
 
             if ($accepted !== [] && $newSource !== $source) {
@@ -90,14 +103,14 @@ final readonly class FixApplicator
     /**
      * @param list<Fix> $fileFixes
      *
-     * @return array{list<Fix>, list<Fix>, string}
+     * @return array{list<Fix>, list<SkippedFix>, string}
      */
     private function applyToFile(string $source, array $fileFixes): array
     {
         $oldStatements = $this->parser->parse($source);
 
         if ($oldStatements === null) {
-            return [[], $fileFixes, $source];
+            return [[], $this->skipAll($fileFixes, FixSkipReason::NodeNotFound), $source];
         }
 
         $oldTokens = $this->parser->getTokens();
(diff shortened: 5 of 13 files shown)


Radiergummi:

Planned THROUGH the 2026-06-17 rescoping comment: the bottom-to-top line-arithmetic mechanic is dead under the #368 AST-reprint backend — **no line arithmetic planned**. Verified the current state on `main` @ f13d1abd before scoping, which trims the issue to two real gaps:

1. **Same-node conflict detection (the correctness gap).** `FixApplicator` already tracks `skipped`, but only for *mechanical* misses (node-not-found / FP-print-throw). Two ops on the **same node** are applied in sequence with no guard (silent last-wins / index-shift). New `@internal FixConflictDetector`: group by `TargetSelector`, keep a deterministic safe subset (first-wins, so re-run makes progress), skip the rest with a typed reason (`SkippedFix`/`FixSkipReason` VO).
2. **Fix-run reporting / frozen `--check --format=json` schema.** Skipped+applied+remaining never reach the formatters; JSON emits only remaining findings. Add the frozen fix-run envelope + a one-line summary across cli/github/markdown.

Scope corrections vs the issue body: the `--fix`/`--check` **docs section already exists** (linting.md:124-226) → extend it + add the JSON schema block, no new page. `FixResult` already carries applied/skipped/modifiedFiles → the change is adding a skip *reason*.

Controversy flags: >150 src/ LOC likely (human merge gate); the frozen JSON schema becomes a de-facto CI contract (intended); no Contracts/config/CLI-signature change. area:lint+cli ⇒ no survey. Noted #383 (`--fix=safe|dangerous`) will extend the `FixRunResult` serialization seam — flagged so it does not re-cut it.

Radiergummi:
🤖 **[lead]** plan posted (PR #396, issue #22) → plan-review. Reviewer: validate (a) the same-node conflict-detection design (FixConflictDetector groups by TargetSelector, first-wins safe subset so re-run progresses, typed SkippedFix/FixSkipReason) is correct + minimal; (b) the frozen --check --format=json schema is treated as the stable/load-bearing surface (schema block + its assertion test); (c) scope corrections hold (docs section exists → extend; FixResult already carries applied/skipped). area:lint+cli ⇒ no survey. Likely >150 src LOC → human merge gate.

Radiergummi:
🤖 **[reviewer]** plan-review: approved (amended the PR body with a reviewer addendum — 1 affirmation + 2 small in-scope tightenings to honour during coding).

I verified every load-bearing claim against `main` @ `f13d1abd`; the plan is correctly scoped through the 2026-06-17 rescoping comment (no line arithmetic), and all four "what the code already does" claims check out.

**Grouping key — affirmed correct (the lead's specific question).** Two *different* `TargetSelector`s cannot alias the same attribute-bearing node: `Method` resolves to the `ClassMethod`, `Property` resolves to the `Property` statement or a promoted `Param` (`FixOperationVisitor::findProperty`, properties-first → deterministic), and the promoted-param node is reachable only via `(class, Property, name)`, never via a `Method` selector. So *same node ⟺ same `(className, kind, memberName)` identity* — grouping by `TargetSelector` identity is sound. (Confirmed only `ClassNode`/`Method`/`Property` selectors are ever emitted.)

**Gap 1 (correctness) confirmed at the code.** `applyToFile()` registers every fix's visitor on one `NodeTraverser` and runs them in a single pass over the shared cloned tree (FixApplicator.php:113-124); two ops on the same node both set `$visitor->applied = true` and both land in `accepted` (129-135) with no guard — exactly the silent last-wins / shifted-index corruption the issue targets. The detector is the minimal correct fix; there is no existing per-fix isolation seam to reuse (they share one traversal). First-wins is deterministic, and a re-run converges (the skipped op is re-emitted by its rule next pass with no surviving conflictor).

**Scope corrections all verified against code:** `FixResult` carries `applied`/`skipped`/`modifiedFiles` (skipped is `list<Fix>`, no reason — the real gap); `FixRunResult` carries `fixResult` + `remainingFindings` + the `--check` dry-run `exitCode()`; the `--fix`/`--check` docs section exists (docs/linting.md:121-142 — extend, don't add a page); `JsonFormatter` emits `{schema_version, level, exit_code, findings, summary(+coverage?)}` with no fix-run envelope (JsonFormatter.php:42-57). No duplication risk.

**Two tightenings folded into the body (small, in-scope, not design changes):**
1. **Reword the existing skip message** — `LintCommand::renderFixSummary()` (LintCommand.php:449-456) still says *"skipped due to overlapping edits"* and `FixResult`'s docblock says *"overlap with an already-applied edit"* (FixResult.php:11); both describe the **dead byte-overlap model**. The reporting work must reword to AST-era "same-node conflict" so the message is honest.
2. **Pin the fix-run JSON as a distinct envelope** (top-level `remaining`, not `findings`) so existing lint-JSON consumers reading `findings`/`summary`/`exit_code` are provably unaffected; `remaining[]` reuses `Finding`'s `JsonSerializable` shape. This resolves the planner's "JsonFormatter vs command" decision toward the one-place `FixRunResult::toArray()` serializer — which is also the **#383 seam**, cleanly cut so #383 (`--fix=safe|dangerous`) extends rather than re-cuts it.

Litmus, surgical scope, named proving tests (FixConflictDetectorTest: same-arg conflict / remove-vs-modify / disjoint-removes-both-kept / different-members / different-files / determinism; FixApplicatorConflictTest byte-stable golden + re-run-converges; the exact `--check --format=json` shape assertion with zero/all-skipped/mixed edges; per-format cli/github/markdown summary), docs extension, and CHANGELOG are all correctly specified. Size + the frozen-JSON-as-CI-contract make this a human merge gate regardless — correctly flagged.

Radiergummi:
🤖 **[lead]** plan-review APPROVED by reviewer1 (+2 in-scope tightenings folded into body) → coding. Implement per amended PR body: FixConflictDetector (group by TargetSelector, first-wins safe subset, typed SkippedFix/FixSkipReason) wired into FixApplicator's single-traverser apply path (FixApplicator.php:113-124, the unguarded same-node gap); skipped-reason → FixResult; one-place FixRunResult::toArray() serializer with a DISTINCT fix-run JSON envelope (top-level remaining, not findings) so existing lint-JSON consumers are unaffected — this is also the #383 seam, cut cleanly. Reword the stale 'overlapping edits' message (LintCommand.php:449-456 + FixResult docblock) to AST-era 'same-node conflict'. Extend the existing --fix docs section (linting.md:121-142), don't add a page. area:lint+cli ⇒ no survey.

Radiergummi:
🤖 **[coder]** decision note: where the fix-run summary/JSON is assembled.

Per the plan + reviewer1 addendum #3, the fix-run JSON envelope is owned in **one place** — `FixRunResult::toArray()` — and is a **distinct envelope** (top-level `remaining`, not `findings`), so existing lint-JSON consumers reading `findings`/`summary`/`exit_code` are unaffected. `remaining[]` reuses `Finding`'s `JsonSerializable` shape verbatim. `FixRunResult::SCHEMA_VERSION = '1'`.

The `Formatter` interface takes a `LintResult` only, so I did **not** widen it. Instead `LintCommand` assembles the fix-run output per resolved target: a JSON target writes `toArray()` (via `encodeFixRun`); cli/github/markdown get a one-line `renderFixSummary` (github → `::notice title=OpenAPI fix::…`, markdown → `- …`, cli → plain) then the remaining findings through their normal formatter. This keeps the public-ish `Formatter` surface untouched while still surfacing fixed/remaining/skipped(+reason) in every human format. If a reviewer prefers a shared fix-run renderer object, that's a small extraction, flagged as a candidate, not built here (the plan called it out-of-scope).

Tightenings honoured: (2) the stale "overlapping edits" wording is gone from both `LintCommand` and the `FixResult` docblock, replaced by AST-era "same-node conflict"; (3) the JSON envelope is the frozen, distinct `remaining` shape, pinned by `FixRunResultTest` + the command-level `--check --format=json` test.

**#383 seam note:** `FixRunResult::toArray()` is the natural seam `--fix=safe|dangerous` (#383) will extend (e.g. a `safety`/`mode` dimension on skipped entries). It is cut cleanly and owned in one place; #383 should extend it, not re-cut it.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed 7a58c505. Ready for review.

**Both real gaps closed:**
- **Gap 1 (correctness):** new `@internal FixConflictDetector` groups a file's fixes by `TargetSelector` identity and keeps a provably-independent **first-wins** subset, skipping the conflicting remainder with `FixSkipReason::Conflict`. Wired into `FixApplicator::apply()` **before** mutation (the gap at FixApplicator.php:113-124). The op pair-matrix is enumerated explicitly with a why-comment and pinned by `FixConflictDetectorTest` (same arg+index, Set-vs-Remove on a shared index, disjoint removes, AddAttribute-as-conflict, different members/classes, determinism).
- **Gap 2 (reporting):** `SkippedFix` VO + `FixSkipReason` enum (`conflict`/`node-not-found`/`print-failed`); `FixResult::$skipped` now carries typed reasons (the three skip sites tagged). `FixRunResult::toArray()` is the one-place frozen fix-run JSON envelope (`schema_version`/`mode`/`applied`/`skipped`/`modified_files`/`remaining`/`exit_code`); `--format=json` emits it, cli/github/markdown get a per-format fixed/remaining/skipped(+reason) summary.

**Tightenings honoured:** stale "overlapping edits" wording removed from `LintCommand` + the `FixResult` docblock (now "same-node conflict"); the JSON is a distinct `remaining` envelope (not `findings`), so lint-JSON consumers are unaffected. Decision/seam note in a separate comment (Formatter interface left untouched; #383 should extend `toArray()`).

**Tests (TDD):** `FixConflictDetectorTest` (10), `FixApplicatorConflictTest` (golden byte-stable + re-run progress + node-not-found distinct), `FixRunResultTest` (exact shape + zero/all-skipped/mixed edges), `LintFixCommandTest` (+3: json envelope, github notice, markdown bullet). Docs: `--fix`/`--check` section extended (conflicting-fixes + frozen JSON schema), no new page. CHANGELOG `[Unreleased]` line.

**Gates:** local `composer check` (PHPStan L8, Pint, 2857 tests) green. **Full remote matrix green** incl swagger-php 5.8 + byte-exact Snapshot (runs 27714576994 + 27714574679 on HEAD `7a58c505`).

Radiergummi:
🤖 **[lead]** implemented (HEAD 7a58c505); full matrix green incl swagger-php 5.8 + Snapshot → review (two independent reviewers). Moritz pre-authorized merge-on-green; reviews are the quality gate before I merge. Scrutinize: FixConflictDetector first-wins subset correctness + the op pair-matrix; the frozen FixRunResult::toArray() envelope (distinct top-level remaining); re-run-progress invariant.

Radiergummi:
🤖 **[reviewer]** review round 1 — 1 blocking finding (correctness), 0 nits. (Independent pass A, HEAD `7a58c505`, verified in the `feat-22-conflict` worktree with two end-to-end reproductions through the real `FixApplicator`.)

## BLOCKER 1 — `FixConflictDetector`'s independence model is wrong: it classifies two index-shifting ops on the same node as *safe*, and the applicator then silently corrupts the source.

**Root cause.** The detector reasons as if each op resolves its attribute indices against the *original* attribute layout. But `FixApplicator::applyToFile()` registers every fix's visitor on **one** `NodeTraverser` and runs them in a **single pass over the same, progressively-mutated cloned node** (FixApplicator.php:113-124). I confirmed with php-parser directly that within one `traverse()`, a later visitor's `enterNode` on the *same* node sees the earlier visitor's mutation: after one visitor removes attribute group 0, the next visitor sees the **compacted** `attrGroups` (indices shifted down by one). So any `RemoveAttribute` (or `AddAttribute`) on a node shifts the flat-index space that every *later* attribute op on that node resolves against (`FixOperationVisitor::attributeAt()` counts flat position across the **current** groups, FixOperationVisitor.php:241-256). The "provably independent" premise in `conflicts()` does not hold under this execution model.

Two of the detector's own *blessed-as-safe* cases corrupt. Both reproduced E2E through `FixConflictDetector::partition()` → `FixApplicator::apply($kept)`:

**(a) `RemoveAttribute([0])` + `SetAttributeArgument(index: 1, operationId: …)` on the same method** — classified `kept=2, skipped=0`. Input:
```php
#[\…\Tag('dup')]                              // flat 0
#[\…\Operation(operationId: 'bad id')]        // flat 1
public function m(): void {}
```
After applying the "safe" subset:
```php
#[\…\Operation(operationId: 'bad id')]        // Tag removed — but operationId STILL 'bad id'
public function m(): void {}
```
The `SetAttributeArgument(index: 1)` was reported **applied**, but because the Tag removal compacted the array first, flat index 1 no longer addressed the `Operation` attribute (now at 0) — so it set nothing. A fix counted as applied that mutated the wrong/no node: exactly the silent corruption this issue exists to prevent. **This pair is reachable with shipped fixers today**: a method carrying a duplicate `#[Tag]` (→ `tag.duplicate` `RemoveAttribute([0])`) and `#[Operation(operationId: 'bad id')]` (→ `operation.id-invalid-chars` `SetAttributeArgument(index: 2/1)`) triggers both fixers in one run.

**(b) Two `RemoveAttribute` with *disjoint* indices `[0]` and `[2]`** — classified `kept=2, skipped=0` (and `FixConflictDetectorTest` line 70-78 asserts this as correct). Input `#[Tag('a')] #[Tag('b')] #[Tag('c')]`, intent "drop a and c, keep b". Actual result: `#[Tag('b')] #[Tag('c')]` — **`c` survived**, because the first removal compacted the array and the second op's flat index 2 no longer pointed at `c`. Disjoint indices are **not** independent across *separate* ops in one traversal. (A *single* `RemoveAttribute` Fix carrying `[0,2]` is safe — it removes highest-first internally — but two separate Fixes are not.)

**Why the suite is green.** The unit tests encode the broken independence model (the disjoint-removes-both-kept and the does-not-touch-keeps cases assert the wrong classification), and `FixApplicatorConflictTest` has no golden that *removes/sets across a shifted index* — so no test exercises the corruption. This is a test-suite gap masking the defect, not evidence of correctness.

**Fix direction (coder's call on the cleanest form):** the safe condition must account for index shifting, not just index *identity/overlap*. Two robust options:
1. **Treat any node-shape-changing op as conflicting with every other op on the same node** — i.e. once a `RemoveAttribute` or `AddAttribute` is kept for a node, skip every subsequent op on that node (`SetAttributeArgument` *and* other `RemoveAttribute`). This generalises the existing `AddAttribute`-conflicts-with-everything rule (line 100) to `RemoveAttribute` too. Simplest, provably correct, and re-run converges (the skipped op applies cleanly once the shape-changer is gone). Per-node at most one structural op + independent same-attribute-distinct-arg `SetAttributeArgument`s.
2. Keep finer granularity only if you also make the *applicator* apply each node's fixes against a stable index space (e.g. one visitor per node that batches that node's ops and resolves all indices against the pre-mutation layout). That's a larger applicator change; option 1 is the surgical fix for this PR.

**Tests to add (would fail today):**
- `FixConflictDetectorTest`: `RemoveAttribute([0])` + `SetAttributeArgument(index: 1)` on one member → second SKIPPED (not kept). Two `RemoveAttribute([0])` + `[2]` on one member → second SKIPPED. (Replace/repair the current line-70 disjoint-both-kept assertion accordingly.)
- `FixApplicatorConflictTest` (golden, byte-stable): the reproduction (a) fixture above → after applying the kept subset, assert the **operationId is unchanged from the original** when the SetArg is (correctly) skipped, and a **re-run** then applies it cleanly. And a remove-a-and-c-keep-b fixture → after kept subset, `b` and the un-conflicting removal land, the conflicting removal is skipped + reported, re-run finishes. These pin the corruption shut.

---

## Everything else is correct — and the reporting half is genuinely well done.

I verified the rest adversarially and found no other issues:
- **Grouping key** (`className\0kind\0memberName`, line 79-84) is sound; two different `TargetSelector`s never alias the same node (confirmed in plan-review). NUL-separated key avoids delimiter collisions. Determinism holds: insertion-ordered `$byTarget` + first-in-list-wins, no hashing/iteration-order dependence (the determinism test passes for the right reason).
- **Op-pair matrix is otherwise complete and correctly conservative** for the cases that *don't* involve cross-op index shift: AddAttribute×any = conflict; SetDocComment×SetDocComment = conflict, SetDocComment×attr-op = independent (different node aspect — correct); SetArg×SetArg = conflict only on same `(index, argumentName)`. Only the Remove-shift dimension is missing.
- **Frozen `FixRunResult::toArray()` envelope** (FixRunResult.php:44-78) is exactly right: distinct top-level `remaining` (not `findings`) so existing lint-JSON consumers are unaffected; `schema_version`/`exit_code` present; `skipped[]` flattened to `{rule_id, file, reason}` with **no internal VO leak** (`SkippedFix`/`FixSkipReason` never serialize raw); `remaining` reuses `Finding`'s `JsonSerializable` verbatim; frozen shape pinned in the `@return`. This is the clean #383 seam.
- **Plan-review tightenings honoured:** the stale "overlapping edits" wording is **gone** everywhere (`grep` clean in src/ + docs); `FixResult`'s docblock now describes the typed AST-era skip reasons.
- **`SkippedFix` / `FixSkipReason`** are tidy `@internal` VOs; three skip sites tagged. Conventions clean (strict-types, no abbreviations, why-comments, 100-char, modern PHP). Pint passes on touched files; docs extend the existing section (no new page); CHANGELOG present.

The reporting/JSON/wording work could merge as-is. The conflict detector is the issue's core deliverable and currently corrupts source on two of its own safe-classified, shipped-fixer-reachable cases — so this is a hard blocker. Reproductions above are copy-pasteable. Back to the coder.

Radiergummi:
🤖 **[lead]** review round 1: BLOCKING correctness bug (reviewer1) → back to coding (round 1/3). Detector's independence model is wrong: FixApplicator runs all visitors in ONE traversal over a progressively-mutated tree (FixApplicator.php:113-124), so a RemoveAttribute shifts the flat-index space every later same-node op resolves against. Two blessed-as-safe cases corrupt source (one reachable with shipped fixers: tag.duplicate Remove[0] + operation.id-invalid-chars SetArg on one method). Fix: treat RemoveAttribute as conflicting with every other op on the same node (first-wins, skip rest, re-run converges); fix the 2 broken unit tests + add applicator-level goldens. Reporting half is clean — leave it.

Radiergummi:
🤖 **[reviewer]** review round 1 (pass B): **1 blocking correctness finding** (+ the reporting/JSON/conventions all clean)

Independent second review on HEAD `7a58c505` in worktree `feat-22-conflict`. Gates I ran myself: `vendor/bin/pint --test` passed, `composer lint` (PHPStan L8) 0 errors, `--group=fix` (69) + `--group=lint` (985) green.

**Finding 1 — BLOCKING (correctness): the detector misclassifies two disjoint `RemoveAttribute` ops on the same node as "provably independent" and keeps both, but they corrupt source when applied together.** `src/Lint/Fix/Ast/FixConflictDetector.php:110-112` treats a `RemoveAttribute`/`RemoveAttribute` pair as conflicting **only** when their `attributeIndices` intersect; disjoint indices ⇒ both KEPT (and `FixConflictDetectorTest.php:70` pins exactly this as the intended "no false positive" case). But kept fixes are not applied independently: `FixApplicator::applyToFile()` registers every kept visitor on **one** `NodeTraverser` and runs them in a single pass over the shared cloned tree (FixApplicator.php:126-137). `FixOperationVisitor::removeAttributes()` splices and **re-indexes** (`array_values`) the member's `attrGroups`, so the *second* RemoveAttribute's indices — which `RemoveAttributeFixer` computed against the **original** source order via reflection — now address shifted attributes. The detector's "provably independent" claim (its own docblock) is false for this pair.

I proved it with an uncommitted PoC against this worktree: a method `#[A] #[B] #[C] #[D]`, two fixes `RemoveAttribute([0])` (A) + `RemoveAttribute([2])` (C), applied in one batch → result keeps `#[B] #[C]` (**A and D removed**), reported as `applied=2 skipped=0`. The intent was to remove A and C; it silently removed A and D. This is the exact "wrong mutation, not recoverable" outcome the detector's stated correctness-bias exists to prevent — and it ships as `applied`, not `skipped`.

Reachability: a single `RemoveAttributeFixer` call emits **one** op carrying all indices, so this needs **two findings on the same member** each emitting a removal — e.g. `tag.duplicate` + `response.duplicate-status` (or `field.no-effect`) both firing on one method. `FixRunner::run()` collects every finding's fixes into one `$allFixes` batch (FixRunner.php:50-64), so they land together and the detector keeps both. Plausible, not purely theoretical.

Fix (minimal, matches the stated bias): treat **any** two `RemoveAttribute` on the same node as conflicting — keep first, skip rest. Strictly more conservative; re-run converges because the skipped removal is re-emitted next pass against fresh indices. One line at FixConflictDetector.php:110-112 (`return true`) + flip the `FixApplicatorConflictTest` to add an end-to-end **two-disjoint-removes-on-one-member** golden (this is the missing applicator-level case — the unit test at line 70 asserts the *wrong* contract, and no golden exercises two removes together, which is why the corruption isn't caught). Alternatively (larger) make the applicator apply each kept op against a fresh re-parse, but the conservative skip is the right call here.

Everything else holds:
- **Pair matrix otherwise correct.** AddAttribute-vs-anything → conflict (index-space shift) ✓; SetDocComment independent of attr ops but two doc edits conflict ✓; Set/Set same `(index,arg)` → conflict, different args kept ✓; Remove-vs-Set conflicts iff the set's index is removed ✓. Determinism: grouping key is `className\0kind->name\0memberName`, first-wins, no hash/iteration-order dependence ✓ (and reviewer1's plan-review affirmation that distinct TargetSelectors can't alias one node holds).
- **Re-run progress** proven by `FixApplicatorConflictTest` (skipped Set re-applies on the next run).
- **Wiring applies the detector BEFORE mutation** — `partition()` runs in `apply()` before `applyToFile()`, conflicts join `skipped` with no op bypassing the guard (FixApplicator.php:70-98) ✓.
- **Frozen JSON envelope** is excellent: one-place `FixRunResult::toArray()`, distinct top-level `remaining` (not `findings`), `reason->value` scalar (no enum/VO leak), `SCHEMA_VERSION='1'`; `FixRunResultTest` + `LintFixCommandTest` assert the **exact** key set (`array_keys(...)->toBe([...])`) and `not->toHaveKey('findings')`, with zero-fix / all-skipped / mixed edges and JSON-encodability covered. Existing lint-`--format=json` consumers provably unaffected.
- **Reporting** across cli/github/markdown is wired without widening the `Formatter` interface; deterministic `skipReasonBreakdown` (ksort); stale "overlapping edits" wording removed from both `LintCommand` and the `FixResult` docblock (tightening #2 honoured).
- **Conventions/bookkeeping:** strict-types, PHP 8.4 (property hook on the non-readonly `FixResult`, enums, readonly VOs), `@internal` on detector/SkippedFix/FixSkipReason, no abbreviations, concise comments, 100-char. docs/linting.md `--fix` section EXTENDED (conflict behaviour + frozen schema block), CHANGELOG `[Unreleased]` line present.

Net: the reporting/JSON/docs half is merge-ready; Finding 1 is a real same-node corruption that the design explicitly set out to prevent. Recommend fixing before merge (one-line conservative skip + an end-to-end golden). Holding my approval on that one change.

Radiergummi:
🤖 **[coder]** addressed round 1: fixed the same-node corruption blocker; pushed 69c287bc.

**Root cause confirmed** (reproduced both reviewers' repros through the real `FixApplicator` before fixing): all kept visitors run in ONE traversal over the same progressively-mutated tree, so `RemoveAttribute`'s `array_values` compaction re-indexes the flat attribute list every later op resolves against. My 'disjoint indices are independent' model was wrong.

**Fix** (reviewer1's option 1, the conservative one): generalised the existing `AddAttribute`-conflicts-with-everything rule to **any layout-changing op** — a new `changesLayout()` returns true for both `AddAttribute` and `RemoveAttribute`, so once one is kept for a node, every other op on that node is skipped (`conflict`). Strictly more conservative; a re-run converges because the skipped op is re-emitted next pass against fresh indices. The detector's only remaining same-node pair is Set-vs-Set (neither changes layout), kept iff different `(attributeIndex, argumentName)`.

**Tests:**
- Repaired the two unit tests that encoded the broken model: `SetArg` after a `Remove` (even on an index the remove does not name) → SKIPPED; two `RemoveAttribute` with disjoint indices on one member → second SKIPPED. The old line-70 'disjoint-both-kept' assertion is gone.
- Added **applicator-level end-to-end goldens** (the gap that masked this) that drive `FixApplicator` across a shifted index and assert byte-exact output + correct applied/skipped accounting + re-run convergence:
  - `#[Tag] + #[Operation(operationId)]` (the shipped-fixer-reachable `tag.duplicate` + `operation.id-invalid-chars` case): the set is skipped, the operationId is left untouched, and a re-run (set now at flat 0) applies it.
  - `#[Tag('a'..'d')]` with two disjoint removes [0] and [2]: only the first removal lands, the second is skipped + reported, and a re-run (C now at flat 1) finishes — the exact `b,c` corruption is pinned shut.
  Both goldens fail on the pre-fix detector.

Docs prose tweaked to describe removal-shifts-others. Reporting/JSON/wording half untouched (both reviewers passed it).

**Gates:** local `composer check` (PHPStan L8, Pint, 2858 tests) green. **Full remote matrix green** incl swagger-php 5.8 + byte-exact Snapshot (runs 27715365157 + 27715366314 on HEAD `69c287bc`).

Radiergummi:
🤖 **[lead]** round 1 fix pushed (69c287bc); full matrix green → round-2 re-review (both reviewers). Fix: changesLayout() helper → any Add/Remove conflicts with every other same-node op; only Set-vs-Set (distinct index+arg) co-applies. New E2E applicator goldens (remove+set index-shift; two-disjoint-removes) that fail pre-fix. Reporting half untouched (already approved by both).

Radiergummi:
🤖 **[reviewer]** review round 2: approved, blocker resolved. (Pass A, fix delta `7a58c505..69c287bc`, verified in the worktree: re-ran my two original reproductions against the fixed detector, ran the conflict + full fix-group suites, Pint, PHPStan.)

The fix is exactly the surgical option I recommended and it closes the corruption cleanly.

1. **Both corrupting cases now correctly SKIP, proven byte-exact.** I re-ran my two original PoCs against the fixed `FixConflictDetector`:
   - `RemoveAttribute([0])` + `SetAttributeArgument(index: 1, operationId)` → now `kept=1, skipped=1, reason=conflict` (was `kept=2, skipped=0` + silent corruption).
   - two disjoint `RemoveAttribute([0])` + `([2])` → now `kept=1, skipped=1`.
   The new applicator goldens are literally these reproductions: golden (a) asserts the Tag is removed while **operationId stays `'bad id'`** (the Set is skipped, not silently misapplied) with `applied===[$remove]` + `skipped[0].reason===Conflict`, then a re-run resolves the Set against the compacted layout (index 0) → `operationId: 'fixed'`. Golden (b) (`a..d`) asserts only `a` is dropped, `b,c,d` remain, second removal skipped, then re-run drops `c` (now index 1) → `b,d`. Byte-exact `->toBe(...)` on full source + accurate applied/skipped accounting throughout. Both fail on the pre-fix detector.

2. **`changesLayout()` covers every index-shifting op — no gap left.** Of the four op types, exactly `AddAttribute` and `RemoveAttribute` re-index the flat attribute list, and `changesLayout()` returns true for both, short-circuiting to conflict against any other same-node op. `SetDocComment` touches comments (not `attrGroups`), so it shifts no attribute index — correctly handled by the separate DocComment branch (DocComment×DocComment conflict; DocComment×attr-op independent), not a layout changer. `SetAttributeArgument` mutates an existing attribute's args without adding/removing a group, so it's the **only** co-applyable same-node pair, kept iff distinct `(attributeIndex, argumentName)`. That is genuinely safe: two Sets never change the layout the other resolves against. No index-shifting op is left classified safe. Verified no false positive: two Sets on the same attribute, different args → still both kept.

3. **Re-run convergence holds for the new skips.** Proven directly by both goldens (the re-emitted op resolves against the now-compacted layout and lands) and sound by design: a skip never re-skips itself, because next pass the kept layout-changer is already applied and gone, so the re-emitted op has fresh indices and no surviving conflictor.

4. **No regression to the reporting/JSON half.** The delta touches only the detector, its unit tests, the two new applicator goldens, and one docs sentence. The frozen `FixRunResult::toArray()` envelope, per-formatter summary, and AST-era wording I approved last round are untouched. Full **fix-group suite: 70 passed (207 assertions)**; conflict files 14 passed; **Pint clean**, **PHPStan L8 clean** on the touched file. The two unit assertions I flagged as encoding the wrong model were correctly inverted to assert SKIP, and the now-subsumed "overlapping index" remove×remove test was removed (all remove×remove pairs conflict now). Docs sentence updated to describe the removal-shifts-others case.

Clean. This is an approval — the second independent pass.

Radiergummi:
🤖 **[reviewer]** review round 2 (pass B): approved, no concerns.

Re-reviewed the fix delta `7a58c505..69c287bc` (independent of reviewer1). Gates I re-ran in the worktree on HEAD `69c287bc`: Pint `--test` passed, PHPStan L8 0 errors, `--group=fix` (70) + `--group=lint` (986) green.

**My round-1 blocker is pinned shut.** The new `changesLayout()` gate (FixConflictDetector.php:102, 121-125) treats **both** `AddAttribute` and `RemoveAttribute` as conflicting with any other op on the same node, so any index-shifting op forces keep-first/skip-rest. My exact corruption case (`#[Tag('a'..'d')]`, `RemoveAttribute([0])` + `RemoveAttribute([2])`) is now an end-to-end golden (FixApplicatorConflictTest.php) asserting **byte-exact** output where only `#[Tag('a')]` is removed (b, c, d survive), the second remove is `skipped`/`Conflict`, and a re-run finishes the job. That golden fails against the pre-fix detector (which kept both → removed a and d, `applied=2`), so TDD discipline holds.

**Classification is exhaustive across the 4 op types.** After the `changesLayout` gate, neither op is a layout-changer, leaving only `SetDocComment` (two doc edits conflict; doc-vs-Set independent — different node aspect, correct) and `SetAttributeArgument`/`SetAttributeArgument` (conflict iff same `attributeIndex` AND `argumentName`). No index-shifting pair can slip through as safe. **Set-vs-Set on distinct index-or-arg is correctly the only co-applyable same-node pair**, and it is genuinely safe because neither op changes the group count.

**Re-run converges** — proven by both new goldens (the skipped Set/Remove re-emits next pass against the compacted layout and lands) and the pre-existing same-arg case.

**No regression in the reporting/JSON half** — the delta is confined to `FixConflictDetector.php` + the two test files + a docs wording tweak; the frozen `FixRunResult::toArray()` envelope, formatter wiring, and their tests are untouched.

**Tests updated correctly, not just patched green:** the two unit tests that previously asserted the *wrong* contract (`keeps two disjoint removes`, `keeps a Set on an untouched index`) are flipped to assert the conservative skip; the now-subsumed overlapping-index Remove/Remove test is removed. **Docs** (`docs/linting.md`) reworded to describe the actual semantics ("a removal alongside another edit on the same member, where removing one attribute shifts the others") — honest and consistent with the code. Conventions clean (the `changesLayout` helper has a concise why-doc, no abbreviations, `@internal` intact).

Both my findings resolved; this is my approval.

Radiergummi:
🤖 **[lead]** Moritz-authorized merge on green. #22 AST-era fix conflict semantics + fix-run reporting: FixConflictDetector (changesLayout() gate — any Add/Remove conflicts with every same-node op, first-wins, re-run converges), typed SkippedFix/FixSkipReason, frozen FixRunResult::toArray() JSON envelope (distinct top-level remaining; clean #383 seam), per-formatter fixed/remaining/skipped summary. Caught + fixed a source-corruption blocker in review (both reviewers independently; remove+set and disjoint-removes index-shift cases) before merge — pinned shut with E2E applicator goldens. Two independent reviewer approvals (round 2); full matrix green incl swagger-php 5.8 + Snapshot.

PR #398, opened 2026-06-17
feat(#199): migration.* rule — document-level @OA\* → config/openapi.php (report + scaffold)

Radiergummi:
Closes #199

## Implementation plan — #199 migration: document-level `@OA\*` → `config/openapi.php`

`<type>=feat` · area:lint · tier-1 label on the issue, but this is a **SwaggerPhp-plugin migration
rule** (read-only detection of authored annotations), no body-scanning inference. Depends on the
full lint-fix cluster (#382/#22/#383, merged; `main` @ `0c6c8b6f`). Parent: #122.

### THE SCOPE DECISION (resolved with code evidence; flagged for the lead)

The issue lists three ambition levels — (1) report-only, (2) report + scaffold snippet, (3) auto-write
`config/openapi.php`. The lead asked me to plan (1)+(2) and **evaluate (3) as the first
`FixSafety::Destructive` consumer** the #383 mechanism was built for.

**Recommendation: ship (1)+(2) now; DEFER (3) to a dedicated follow-up — for an architectural reason,
not timidity.** The exploration surfaced a hard blocker the issue's framing predates:

> **The entire AST-mutation Fix backend is PHP-class-only.** Every `AstOperation`
> (`RemoveAttribute`/`SetAttributeArgument`/`AddAttribute`/`SetDocComment`) addresses a target via
> `TargetSelector(className, kind, memberName)`, and `FixOperationVisitor::enterNode()` only acts on
> `ClassLike` nodes matched by FQCN. `config/openapi.php` is a plain `return [ … ];` file with **no
> class** — there is no operation, no target kind, and no visitor path that can edit it. `grep` for
> any config-file/array-writing mechanism returns nothing.

So (3) is **not** a thin consumer of #383's gate. #383 gives us the *safety rail*
(`FixSafety::Destructive` + `WorkingTreeGuard::assertClean(files, allowDirty)`) — that part is ready —
but the rail gates a write the backend **cannot currently perform**. Auto-write needs a whole new
non-class primitive: a config-file/array-AST operation, a visitor branch for top-level `Return`/array
nodes, deep-merge-into-existing-array logic that preserves the hand-tuned file's formatting and
comments, and conflict handling when a key already exists. That is a substantial slice in its own
right (comparable to the AddAttribute work), and it edits a single shared hand-curated file where a
merge mistake is high-blast-radius. Bundling it here would blow scope and risk the whole PR.

**The #383 mechanism is still validated** — by being *ready and correct* for the first Destructive
[...]

PR #399, opened 2026-06-18
feat(#195): migration.oa-replaceable-by-attribute rule + fixer (Phase 1: OA\Property + query OA\Parameter)

### LINKED ISSUE #195: Migration lint rule + fixer: replaceable `#[OA]` / `@OA` → our authoring attributes (oa-replaceable-by-attribute)
Parent: #122. Deferred from PR #185 (which shipped the removal rule). This is the **replacement** half of the migration story.

## Summary

`migration.oa-replaceable-by-attribute`: where inference *cannot* recover a hand-authored `#[OA\*]` / `@OA` annotation, but one of laravel-openapi's authoring attributes expresses the same thing more concisely, **rewrite** the swagger-php annotation into the equivalent attribute. It complements `migration.oa-redundant-with-inference` (removal): removal deletes what inference already produces; this converts what inference *can't* produce into a first-class, shorter attribute.

## Replacement whitelist (the full mapping)

| OA annotation | → laravel-openapi attribute |
|---|---|
| `OA\Property` | `#[ResponseField]` / `#[RequestField]` / `#[ResourceField]` — covers description, type, format, example, enum, nullable, default, min/maxLength, pattern, min/maxItems, uniqueItems, read/writeOnly |
| `OA\Parameter`(query) / `QueryParameter` / `Query` | `#[QueryParam]` |
| `OA\Parameter`(path) / `PathParameter` | `#[PathParam]` |
| `OA\Parameter`(header) / `HeaderParameter` | `#[Header]` |
| `OA\Response` | `#[Response]` |
| `OA\Header` (on a response) | `#[ResponseHeader]` |
| `OA\Link` | `#[Link]` |
| `OA\Examples` (request) | `#[Example]` |
| `OA\Examples` (response) | `#[ResponseExample]` |
| `OA\RequestBody` | `#[RequestBody]` (+ `#[RequestField]`) |
| `OA\Discriminator` | `#[Discriminator]` (+ `#[RequestVariant]`) |
| `OA\Webhook` | `#[Webhook]` |
| operation `summary` / `description` / `operationId` | `#[Summary]` / `#[Description]` / `#[Operation]` |
| operation `tags` / operation-level `OA\Tag` | `#[Tag]` |
| operation `deprecated` | `#[Deprecated]` |
| operation `OA\ExternalDocumentation` | `#[ExternalDocs]` |
| operation `security` | `#[Security]` / `#[PublicEndpoint]` |

(Document-level annotations — `OA\Info`/`OA\Server`/`OA\SecurityScheme`/… — are **not** here; they map to `config/openapi.php`, tracked separately. Constructs with no equivalent at all — cookie params, XML, `additionalProperties`, etc. — are the irreducible residual, also tracked separately.)

## Suggested phasing

- **Phase 1** (highest value, lowest risk — single-line, well-bounded): `OA\Property` → field attribute; `OA\Parameter`(query) → `#[QueryParam]`.
- **Phase 2**: responses + headers + links + examples (`OA\Response`→`#[Response]`, response `OA\Header`→`#[ResponseHeader]`, `OA\Link`→`#[Link]`, `OA\Examples`→`#[Example]`/`#[ResponseExample]`).
- **Phase 3**: operation metadata (`summary`/`description`/`operationId`/`tags`/`deprecated`/`externalDocs`/`security`), request bodies, discriminators, webhooks.

## Soundness

Same discipline as the removal rule: rewrite only when the target attribute expresses *exactly* what the OA annotation carried — no field dropped, no semantics changed. Verify by reusing the `SchemaEquivalence` / `InferenceControlDocument` machinery from #185: the attribute's contribution must reproduce the annotation's. `log()` / document any OA shape intentionally not covered so a partial migration never reads as complete.

## Fixer shape

`InsertBefore` (the new attribute) + remove the OA block (reuse `RedundantOaAnnotationFixer` / `DocblockAnnotationRemover` from #185). Placement matches where the OA annotation sat — controller method for operation-level, property / promoted param for field-level.

## Acceptance criteria

- `migration.oa-replaceable-by-attribute` registered (SwaggerPhp plugin, `migration.*` family, level 4).
- Each whitelisted OA shape rewrites to the documented attribute, preserving comments/whitespace; the generated document is unchanged afterward.
- Before/after golden fixtures per shape + a same-document feature flavor.
- Uncovered shapes logged; `docs/linting.md` row + `CHANGELOG.md` entry.

## References

Follows #122; builds on #185 infrastructure. The attributes it rewrites into live in `src/Attributes/`.

Radiergummi:
## Plan — `migration.oa-replaceable-by-attribute` (Phase 1)

Closes #195

The **replacement** half of the migration story. Where inference *cannot* recover a hand-authored
`#[OA\*]` / `@OA` annotation, but one of our authoring attributes expresses the same thing more
concisely, flag it and offer a fixer that rewrites the OA annotation into the equivalent attribute.
Complements `migration.oa-redundant-with-inference` (removal): removal deletes what inference
already reproduces; this converts what inference *can't* into a first-class attribute.

**Scope: Phase 1 only.** Two shapes — `OA\Property` → `#[ResponseField]` / `#[RequestField]` /
`#[ResourceField]`, and `OA\Parameter(in: query)` → `#[QueryParam]`. Phases 2/3 (responses,
headers, links, examples; operation metadata, request bodies, discriminators, webhooks) are
follow-ups, noted as such in the issue. Bounded `src/` diff, no new config key, no public-signature
change, no stage-order change.

### Tier

**Tier 0/1 boundary, on the same footing as the existing migration rules.** The verdict reads
swagger-php's *already-parsed* annotation objects (harvested by `HarvestAuthoredAnnotationsStage`)
and reflection — no method-body parsing. The fixer mutates source via the existing
format-preserving php-parser AST backend (`AddAttribute` + the `OaAttributeRemover` /
`DocblockAnnotationRemover` from #185). No Tier-2 dataflow. Nothing here needs escalation on tier
grounds.

### The core design tension (read before implementing)

1. **Granularity.** The existing rules fire at *whole-schema* / *whole-operation* level and key the
   scanner by class / `class::method`. This rule is **finer**: one finding per *replaceable
   `OA\Property`* (addressed to a PHP property / promoted param) and one per *replaceable
   `OA\Parameter(query)`* (addressed to the controller method). `AuthoredAnnotationScanner` must
   expose the per-property and per-parameter annotation objects; today it only exposes whole `Schema`
   / `Operation`. See "Files" for the two new accessors.

2. **The fire condition is the *inverse* of the redundancy rule.** Redundancy fires when inference
   ⊇ authored. This fires when inference does **not** carry the field/parameter but the attribute
   *can* express it. So we do **not** reuse `OaRedundancyEngine`/subsumption as the trigger. The
   trigger is: "authored annotation X exists, and every key X carries is representable as a scalar
   argument on the target attribute." (Soundness is verified the other way — see #3.)

3. **`AddAttribute` is scalar-only.** `FixOperationVisitor::addAttribute()` only renders
   `null|bool|float|int|string` named arguments (`literalToNode`). `OA\Property` / `OA\Parameter`
   carry array-valued keys (`enum`, `example`-as-array, `x`). **Phase 1 rewrites only annotations
   whose every carried key maps to a scalar attribute argument.** An annotation carrying `enum`,
   array `example`, or `x` is **not flagged** (logged as an uncovered shape) — never silently
   dropped. This is the load-bearing boundary that keeps Phase 1 sound *and* small; widening the
   `AddAttribute` value domain to arrays is explicitly deferred to a later phase.

4. **Soundness ("no field dropped, no semantics changed").** Before flagging, build the candidate
   attribute's contribution and assert it reproduces exactly the authored annotation's schema/parameter
   — reuse `SchemaEquivalence` for *bidirectional* containment (`subsumes(authored, attr) &&
   subsumes(attr, authored)` = equal). If any authored key has no attribute counterpart, or the
   attribute would add/alter a key, do not flag (log). This guards against, e.g., an authored
   `OA\Property` key our `FieldAttribute` maps differently.

### Mapping table (Phase 1, scalar keys only)

| OA\Property / OA\Parameter key | attribute argument | notes |
|---|---|---|
| `description`, `title`, `format`, `pattern` | same | string |
| `type` | `type` | scalar OA type only; array/union `type` not flagged |
| `example` (scalar) | `example` | array example → not flagged |
| `nullable`, `readOnly`, `writeOnly`, `uniqueItems`, `deprecated` | same | bool |
| `default` (scalar) | `default` | array/object default → not flagged |
| `minimum`/`maximum`/`exclusiveMinimum`/`exclusiveMaximum`/`multipleOf` | same | numeric |
| `minLength`/`maxLength`/`minItems`/`maxItems` | same | int |
| `enum`, `x`, array `example`/`default` | — | **not flagged**, logged |
| (parameter) `name`, `required` | `#[QueryParam]` `name`/`required` | `in` must be `query` |

**Which field attribute** (`ResponseField` vs `RequestField` vs `ResourceField`): keyed off where
the `OA\Property` declares — a property/param on a Spatie Data class with no request-context marker
defaults to `#[ResponseField]`; on an API-Resource class → `#[ResourceField]`. Phase 1 uses the
simplest sound rule: **map by the class the property lives on** (Data class → `ResponseField`,
Resource class → `ResourceField`), and **skip + log** an `OA\Property` whose host class is neither
(can't pick an attribute soundly). Document this in the rule docblock. `RequestField` selection
(request DTOs) is deferred to a follow-up to keep the host-class classification trivial.

### Files

**Create**
- `src/Plugins/SwaggerPhp/Lint/OaReplaceableByAttribute.php` — the rule. Implements `Rule`,
  `FixableRule`, the relevant tree-visitor interface(s) (`ComponentSchemaRule` for properties;
  an operation/parameter visitor for query params — match what the existing operation rule uses).
  `excludedStages()` returns `[HarvestAuthoredAnnotationsStage::class]` (same as siblings). `id()` =
  `migration.oa-replaceable-by-attribute`; `severity()` = `Improvable` (level 4); `description()`
  one-liner. Emits one `Finding` per replaceable property/parameter with context: source class,
  member (property/method) name, OA shape (docblock vs attribute), target attribute class-string,
  and the scalar argument map. Logs (via injected logger) every authored shape it *examined but did
  not flag* with the reason (array-valued key, host class neither Data nor Resource, attribute
  wouldn't reproduce).
- `src/Plugins/SwaggerPhp/Lint/Fix/OaReplaceableByAttributeFixer.php` — the fixer. Per finding,
  emits **two** `Fix`es on the same file: (a) `AddAttribute(target, attrClass, scalarArgs)` placing
  the new attribute on the PHP property/param (field shape) or the controller method (query shape);
  (b) remove the source OA block via the existing `OaAttributeRemover` (attribute shape) or
  `DocblockAnnotationRemover` (docblock shape) — reused from #185, addressed by the same
  `TargetSelector`. Degrades to nothing when the target node can't be located. Mirror
  `RedundantOaAnnotationFixer`'s structure and the `ResolvesDeclaringFile` trait.
- `src/Plugins/SwaggerPhp/Lint/Support/OaAttributeArgumentMapper.php` (small, internal) — pure
  mapper: `(OA\Property|OA\Parameter) -> ?array<string, scalar>`. Returns null when any carried key
  is non-scalar or unmappable (the "not flagged / logged" signal). Single source of truth for the
  mapping table; unit-tested directly. Keeps the rule thin and the soundness check honest.

**Modify**
- `src/Plugins/SwaggerPhp/Support/AuthoredAnnotationScanner.php` — add two accessors:
  `propertiesForClass(string $class): array<string, OA\Property>` (keyed by PHP property name; OA
  `property` field → declaring property) and `queryParametersForMethod(string $class, string
  $method): list<OA\Parameter>` (filtered to `in === 'query'`). Index them in
  `indexSchemas()` / `indexOperations()` (walk `$schema->properties` / `$operation->parameters`).
  Reuse the existing ancestry/trait resolution for the method case.
- `src/Plugins/SwaggerPhp/SwaggerPhpPlugin.php` — one `$registry->addRule(OaReplaceableByAttribute::class)`.
- `docs/linting.md` — one catalog row inside the `lint-rule-catalog` markers; a paragraph in the
  Migration-rules section explaining replacement vs removal and the Phase-1 scalar-only boundary.

**Unit — `tests/Unit/Plugins/SwaggerPhp/OaAttributeArgumentMapperTest.php`** (the mapping core)
- scalar keys (description/type/format/min/max/bool flags) → expected arg map.
- array `enum` → null (not flagged). array `example` → null. `x` present → null.
- array/union `type` → null.
- `OA\Parameter` with `in: query` → name + scalar args; `in: path`/`header`/`cookie` → null.
- empty annotation (only `property`/`name` set) → still mappable (no-op-ish attribute), or null per
  decision — pick "null" (nothing worth replacing) and assert it; document the choice.

**Unit — `tests/Unit/Plugins/SwaggerPhp/OaReplaceableByAttributeFixerTest.php`** (golden before→after)
One fixture file per shape under `tests/Fixtures/SwaggerPhp/` (new fixtures, do not perturb the
removal-rule fixtures). For each: build the finding, run the fixer, apply via `FixApplicator`,
assert the resulting source **byte-for-byte** against an expected string (comments/whitespace
preserved, new attribute fully-qualified and prepended, OA block gone). Shapes:
- `OA\Property` attribute on a Data property → `#[ResponseField]`.
- `@OA\Property` docblock on a Data property → `#[ResponseField]` + docblock annotation removed.
- `OA\Property` attribute on a promoted constructor param → attribute on the param.
- `OA\Property` on an API-Resource key → `#[ResourceField]`.
- `OA\Parameter(query)` attribute on a controller method → `#[QueryParam]`.
- Branch: target node not locatable → fixer yields nothing.

**Feature — `tests/Feature/SwaggerPhpReplaceableRuleTest.php`** (mirror `SwaggerPhpMigrationRuleTest`)
- flags a replaceable `OA\Property` (attribute + docblock shapes) under `--only migration.*`.
- flags a replaceable `OA\Parameter(query)`.
- does **not** flag an annotation carrying `enum` (uncovered → logged, asserted via a spy logger).
- does **not** flag a `path`/`header` parameter.
- does **not** flag an `OA\Property` on a class that is neither Data nor Resource.
- stays off a default-level run; disabled by `--skip migration.*`; expanded by
  `openapi.lint.enabled_rules: ['migration.*']` (copy the three governance tests verbatim).
- **unchanged-or-improved invariant:** the rewrite reproduces (bidirectional `SchemaEquivalence`)
  exactly what the OA annotation *declared*, so the generated document is **unchanged or improved**,
  never degraded. (Unlike a removal, a rewrite can *improve* the document: where inference left a
  hand-authored `#[OA\Property]` inert, the equivalent `#[ResponseField]` is one the generator reads,
  so a field the annotation only declared may now actually appear.) Verified by the rejection +
  generation-level reproduction tests, not a byte-identical document diff.
- `exposes its fixer and human-readable description` (mirror the existing rule's last test).

- **Not controversial by precedent:** run 5 auto-merged a new SwaggerPhp `migration.*` rule (#398,
  `DocumentAnnotationMigration`). A new rule in this family, off by default and `class_exists`-guarded
  by the plugin being enabled, is the established pattern.
- **No new config key**, no change to any existing rule's default severity, no `src/Contracts/**`
  touch, no stage-order change, no public-signature change. The two new `AuthoredAnnotationScanner`
  accessors are `@internal`.
- **Watch item for the reviewer:** the field-attribute selection rule (Data→`ResponseField`,
  Resource→`ResourceField`, else skip) is a deliberate Phase-1 simplification. If the reviewer wants
  `RequestField` covered now, that widens host-class classification — raise it rather than guessing.
- Keep the `src/` diff under ~150 LOC; if soundness forces more, stop and flag.

---

## 🤖 [reviewer1] plan-review amendments (verified against `src/`)

The plan is **approved with the corrections below folded in** — apply these when coding.

**Required corrections (factual):**

1. **`ResourceField` is NOT in `src/Attributes/`.** It lives at
   `src/Plugins/ApiResources/Attributes/ResourceField.php`
   (`Radiergummi\OpenApi\Plugins\ApiResources\Attributes\ResourceField`). `ResponseField`,
   `RequestField`, `QueryParam` are in `src/Attributes/`. The fixer must emit the *correct FQCN* for
   the Resource case. `AddAttribute` prints the class-string verbatim as a `FullyQualified` name, so
   this is a string-only reference — **no runtime/boundary dependency** is introduced from SwaggerPhp
   onto ApiResources. Fine to keep, just use the right namespace. (The issue's "the attributes it
   rewrites into live in `src/Attributes/`" is inaccurate for ResourceField.)

2. **Do not reuse `RedundantOaAnnotationFixer` itself** for removal — it removes **all** `#[OA\*]` on
   the class *and every property/param* (`classAttributeTargets()`). The new fixer must compose
   `OaAttributeRemover` / `DocblockAnnotationRemover` (the removers) at **single-property /
   single-param / single-method granularity**, addressed by `TargetSelector(class, Property, name)`
   (or `Method`). The plan's prose already says "reuse the removers"; make sure the code does not
   reach for the whole-schema fixer.

3. **`DocblockAnnotationRemover::removeBlocks` is all-or-nothing per docblock** — it strips *every*
   `@OA\…` block in the targeted docblock. This is sound for the field-docblock shape only because a
   property carries exactly one `@OA\Property` in its own docblock. Add a fixer test asserting that a
   property docblock carrying a `@OA\Property` **plus prose** keeps the prose (the remover's
   `docblockHasOtherContent` path) — and do **not** attempt the docblock field shape on a
   class/method docblock that may hold multiple `@OA\Property` blocks (out of Phase-1 scope; skip+log).

**Precision notes (non-blocking):**

- For the query-param shape, prefer the existing **`QueryParameterRule`** visitor
  (`src/Lint/Visitors/QueryParameterRule.php`) — it yields one finding per query parameter, matching
  this rule's per-parameter granularity better than `OperationRule`. (Plan left this open; this is the
  cleaner choice.)
- New mapper path is `src/Plugins/SwaggerPhp/Lint/Support/OaAttributeArgumentMapper.php` — confirmed
  `Lint/Support/` is where `SchemaEquivalence` already lives. Good.
- `QueryParam` declares `string $name` positionally, but `AddAttribute` emits **named** args
  (`name: '…'`) — PHP accepts this. Mapper emits `name` as a key. Confirmed consistent.
- Soundness check uses `SchemaEquivalence::subsumes($broader, $narrower)` (confirmed signature).
  Compare the attribute's **rendered `descriptor()` contribution** (post type-inference + description-
  directive parsing, see `FieldAttribute::descriptor()`), not its raw ctor args, against the authored
  OA — the plan already says "build the candidate attribute's contribution", which is correct and
  load-bearing. Keep that.

**Confirmed against precedent (#398 / #185):** `Severity::Improvable` (= level 4),
`excludedStages() = [HarvestAuthoredAnnotationsStage::class]`, `migration.*` id, off by default,
plugin-`class_exists`-guarded. No new config key, no Contracts/stage-order/public-signature change.
The two new `AuthoredAnnotationScanner` accessors are `@internal`. Tier-1 confirmed (parsed-annotation
+ reflection reads only; no body parsing, no Tier-2 dataflow).

**Three pressure-tests — verdict:** (1) scalar-only fire condition is **sound and the right
load-bearing limit** — `AddAttribute` is genuinely scalar-typed; logged-not-dropped = no silent loss.
(2) host-class selection sound with correction #1; deferring `RequestField` is acceptable, not a gap.
(3) `migration.*` precedent followed exactly.


Radiergummi:

Key decisions flagged for review:
- **Scalar-only boundary**: `AddAttribute` renders only scalar args, so Phase 1 flags an OA annotation *only* when every carried key maps to a scalar attribute argument. `enum` / array `example` / `x` → not flagged, logged (never silently dropped). This is what keeps the diff small and sound.
- **Field-attribute selection**: Data class → `#[ResponseField]`, Resource class → `#[ResourceField]`, else skip+log. `#[RequestField]` deferred to a follow-up (would widen host-class classification). Watch item for the reviewer.
- Reuses #185 infra (`OaAttributeRemover` / `DocblockAnnotationRemover`, `SchemaEquivalence` for bidirectional-equality soundness) + the `AddAttribute` AST primitive. No new config key, no stage-order change, no Contracts touch.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Reviewer: pressure-test the planner's 3 boundary decisions — (1) scalar-only fire condition (enum/array-example/x logged-not-flagged), (2) Data→ResourceField/Resource→ResourceField host-class selection with RequestField deferred, (3) migration.* family precedent (#398). Confirm Phase-1 scoping keeps it sound + non-controversial, or call a deliberate widen.

Radiergummi:

Verified the plan against `src/`. Tier-1 confirmed (reads already-parsed annotations + reflection; no body parsing, no Tier-2 dataflow smuggled in). Scope is surgical; follows the `migration.*` precedent (#398/#185) exactly — `Severity::Improvable` (level 4), `excludedStages()=[HarvestAuthoredAnnotationsStage::class]`, off by default, plugin-guarded; no new config key, no Contracts/stage-order/public-signature change.

**Three boundary pressure-tests:**
1. **Scalar-only fire condition — sound, correct load-bearing limit.** `AddAttribute::$arguments` is genuinely typed `array<string, null|bool|float|int|string>` and `FixOperationVisitor::addAttribute()` renders only `literalToNode` scalars. Firing only when *every* carried key is scalar, with `enum`/array-`example`/`x` **logged not dropped**, means no silent information loss. Widening `AddAttribute` to structured values is rightly deferred.
2. **Host-class field-attribute selection — sound, with one factual fix.** `ResourceField` is NOT in `src/Attributes/`; it is `Radiergummi\OpenApi\Plugins\ApiResources\Attributes\ResourceField`. The fixer must emit that FQCN (string-only via `AddAttribute`, so **no boundary dependency** is created). Deferring `RequestField` is acceptable — request-DTO classification would need request-context markers the plan correctly avoids. Not a gap to widen now.
3. **`migration.*` precedent — followed exactly.**

**Required corrections (in PR body):**
- Use the correct `ResourceField` namespace (above).
- Do **not** reuse `RedundantOaAnnotationFixer` itself (it removes *all* `#[OA\*]` on the class + every property via `classAttributeTargets()`); compose `OaAttributeRemover`/`DocblockAnnotationRemover` at single-property/param/method granularity via `TargetSelector(class, Property, name)`.
- `DocblockAnnotationRemover::removeBlocks` is all-or-nothing per docblock — sound only because a property holds one `@OA\Property` in its own docblock; add a test that a property docblock with `@OA\Property` + prose keeps the prose, and keep the docblock field shape off class/method docblocks (skip+log).

**Precision (non-blocking):** prefer the existing `QueryParameterRule` visitor for per-parameter granularity; compare the attribute’s rendered `descriptor()` contribution (not raw ctor args) in the `SchemaEquivalence::subsumes` bidirectional soundness check.

No controversy flags. Cleared to move to coding.

Radiergummi:
🤖 **[lead]** plan-review approved (amendments folded into PR body) → coding. Coder: implement per the amended plan. Key reviewer corrections: (1) ResourceField is Plugins\ApiResources\Attributes\ResourceField — emit as verbatim FQCN via AddAttribute, no boundary dep; (2) do NOT reuse RedundantOaAnnotationFixer wholesale (it strips all class+property #[OA] blocks) — compose removers at single-property granularity; (3) test a property docblock carrying @OA\Property + prose keeps the prose; (4) skip+log multi-property class/method docblocks. TDD: failing golden fixtures first.

Radiergummi:
🤖 **[coder]** implemented; local gates green (`composer format` clean, PHPStan `[OK] No errors`, full Pest suite 2929 passed); pushed `86712e24`. Ready for review.

**Deviations from the plan (all verified against `src/`):**

1. **New AST primitive `RewriteToAttribute`** (`src/Lint/Fix/Ast/`) instead of two composable fixes per finding. The plan's "AddAttribute + remove via OaAttributeRemover/DocblockAnnotationRemover" pair is **incompatible with the applicator's conflict model**: `FixConflictDetector` treats *both* `AddAttribute` and `RemoveAttribute` as layout-changers and, per its first-in-list-wins rule, skips the second of any two layout-changing ops on the *same node* — and `FixRunner` does **no re-lint pass** ("fixers are trusted and golden-tested"), so a finding whose second fix is skipped lands half-applied. The new op removes the source annotation (OA attrs by flat index, or the `@OA` docblock block) **and** prepends the new attribute in one node visit. Removal data is still computed at fixer-time with source access (reusing `OaAttributeRemover` / `DocblockAnnotationRemover`); the visitor methods are extracted from the existing `removeAttributesByIndex` / `replaceDocComment` so no logic is duplicated.

2. **Query params use `OperationRule`, not `QueryParameterRule`.** `QueryParameterNode` carries no easy controller class::method handle; `OperationRule` gives `descriptor->controller/method` directly (matching how `OaRedundantOperationWithInference` resolves the method) and I emit one finding per query parameter from there. Same per-parameter granularity, simpler wiring. The OA\Parameter removal is matched by `name:` so `#[OA\Get]` and sibling parameters are left intact (single-annotation granularity, per reviewer amendment #2).

3. **Phase-1 host-class scope narrowed to Spatie Data → `#[ResponseField]` only.** An API-Resource key's `#[OA\Property]` lives *inside* the class-level `#[OA\Schema(properties: [...])]`, which member-level removal (`TargetSelector(class, Property, name)`) cannot address — so `#[ResourceField]` selection is **deferred** alongside `#[RequestField]`, rather than emitting findings the fixer can't soundly apply. Reviewer amendment #1's namespace note is moot for Phase 1 as a result.

4. **Query-parameter docblock shape deferred (skip+log).** `DocblockAnnotationRemover` is all-or-nothing per docblock; a *method* docblock routinely carries sibling `@OA\Get`/other `@OA\Parameter` blocks, so a partial strip is unsound. Only the attribute shape rewrites a query param. The property docblock shape *is* supported (a property carries exactly one `@OA\Property`), incl. the prose-survival case (amendment #5).

5. **No `NeedsInferenceDocument`.** The rule reads authored annotations (scanner) + reflection only; it never consumes `context->inference`, so it carries no `excludedStages()`. It does **not** gate on inference-absence — it fires on any cleanly-mappable authored annotation on a Data class. (Rationale: rewriting to a first-class attribute is a strict improvement regardless, and the soundness check guarantees the attribute reproduces what the annotation *declared*.)

**Finding worth surfacing (not a blocker):** the issue's "generated document is unchanged afterward" framing does **not** literally hold for the `#[OA\Property]`-on-Data shape. With the SpatieData plugin active, an orphan `#[OA\Property]` on a Data property is currently **inert** — swagger-php does not merge it into the inferred Data-class schema (verified: the generated `ReplaceableAttributeData` schema carries only `type`, not the authored `format`/`description`). The rewrite to `#[ResponseField]` makes that metadata *effective* (and inference additionally synthesises a Faker example). So the rule's real value is *activating* documented-but-dead annotations, and soundness = the attribute reproduces the schema the annotation **declared** (asserted by the unit `reproduces()` bidirectional `SchemaEquivalence` check and a generation-level feature test on the rewritten twin), not byte-identical document preservation. Filed nothing separately; flagging here for the reviewer.

**`src/` diff** is larger than the plan's ~150 LOC target (rule 389 / fixer 284 / mapper 180), driven by the soundness machinery, the per-parameter OA matching, and the `RewriteToAttribute` primitive. Surfacing per the "stop and flag if soundness forces more" guidance.

Tests: mapper unit (14), fixer golden unit (6, incl. byte-exact attribute + docblock + prose + query + degrade), feature (10, incl. the three governance globs, enum/non-Data/path skips, and the rewrite-reproduces-schema invariant). Existing fix-group suite (106) still green after the `FixOperationVisitor` refactor.

Radiergummi:
🤖 **[lead]** local+remote CI green (pushed 86712e24) → code-review (two independent reviewers). Heads-up for reviewers: this deviated from the agreed plan on three axes — (1) src diff ~850 LOC (rule/fixer/mapper) vs the ~150 plan target; (2) a NEW AST primitive RewriteToAttribute (atomic remove+add) added to the fix backend because FixConflictDetector rejects a same-node Add+Remove pair and FixRunner does no re-lint pass; (3) the issue's 'document unchanged after rewrite' criterion does NOT hold for OA\Property-on-Data — coder redefined soundness as bidirectional SchemaEquivalence (attribute reproduces what the annotation declared) and narrowed Phase-1 host scope to Data→ResponseField only (Resource/RequestField deferred). All documented in the coder comment. This PR is already a needs-human candidate on the LOC + new-primitive + acceptance-criterion-reinterpretation triggers; review vets it before I escalate.

Radiergummi:
🤖 **[reviewer1]** review round 1 — 1 required finding (+ 2 minor):

I checked out the worktree and ran the suites: new feature **10 passed**, mapper+fixer unit **20 passed**, existing **fix group 106 passed** (the `FixOperationVisitor` refactor did not regress). The three deviations are documented, justified, and verified against `src/`. I focused hardest on the `RewriteToAttribute` / applicator-corruption surface the lead flagged.

**Verdict on the corruption-class risk: structurally sound, but the proving test is missing.**

`RewriteToAttribute` does remove+prepend in a single node visit, so within one finding there is no second op resolving a stale flat index. And `FixConflictDetector::changesLayout()` (src/Lint/Fix/Ast/FixConflictDetector.php:124-126) **does** include `RewriteToAttribute`, so two of them on the same node are detected as conflicting → first-wins kept, second skipped with `FixSkipReason::Conflict`, re-emitted next pass. That is the correct convergence model and it closes the run-5 corruption class. Good.

**1. [REQUIRED] No test exercises two `RewriteToAttribute` ops on the same node through the real `FixApplicator`.** This is the exact scenario the lead demanded verified, and it is reachable: `OaReplaceableByAttribute::checkOperation()` (src/Plugins/SwaggerPhp/Lint/OaReplaceableByAttribute.php:120-126) emits **one finding per query parameter**, so a controller method with **two** replaceable query `OA\Parameter`s yields two findings → two `RewriteToAttribute` Fixes on the *same* `TargetKind::Method` node. The single-fix golden (OaReplaceableByAttributeFixerTest) runs the real applicator, but never the multi-fix-same-node path. Add a test:
   - Fixture: a controller method carrying `#[OA\Get]` + two scalar-only `#[OA\Parameter(in: query, name: ...)]` (say `q` and `page`).
   - Build both findings, run the fixer for both, pass *both* Fixes to one `FixApplicator()->apply()` (mirror the `runReplaceableFixer` harness but with two fixes on one temp file).
   - Assert: exactly **one** rewrite lands this pass (kept), the other is skipped (not corrupted) — `#[OA\Get]` intact, one `#[QueryParam]` added, the other `#[OA\Parameter]` still present and well-formed (no half-spliced attribute, no dropped `#[OA\Get]`). A second apply pass then converges (both rewritten). This proves the layout-changer classification holds end-to-end, not just in the detector unit.

**2. [MINOR] The same-document soundness invariant is asserted against a hand-written twin, not the fixer's own output.** The feature test "rewrites to an attribute that reproduces…" (SwaggerPhpReplaceableRuleTest.php:162) generates from a separately authored `RewrittenResponseFieldData` fixture rather than running lint→fix→regenerate on the original. It is a reasonable proxy and the unit `reproduces()` bidirectional `SchemaEquivalence` check guards the real soundness, but the twin can drift from what the fixer actually emits. Consider (non-blocking) asserting the post-`--fix` source generates the same `name` schema as the documented keys, closing the loop the issue's acceptance criterion intended.

**3. [MINOR] Unlogged skip path.** `evaluateProperty()` returns null without `logSkip` when `propertyShape()` is null (OaReplaceableByAttribute.php:159). It is effectively unreachable (the scanner found an `OA\Property`, so a shape exists), but the "never silently skip" discipline would have it logged. Trivial.

Everything else is clean: scalar-only fire condition correctly drops to `null` on any non-whitelisted/non-scalar key in `OaAttributeArgumentMapper::mapSchema()` (enum/array-example/x/union-type all force null, logged by the rule — no silent loss); host scope honestly narrowed to Data→ResponseField with the deferral documented in the rule docblock and CHANGELOG; conventions OK (strict_types, @internal, no abbreviations, concise comments, no issue numbers in code); boundaries OK (`RewriteToAttribute` lives in the general `src/Lint/Fix/Ast/`, the ApiResources `ResourceField` is moot for Phase 1); CHANGELOG entry + docs/linting.md catalog row + Migration-rules paragraph all present and honest about Phase-1 scope.

Blocking on finding #1 only. Add the multi-fix-same-node applicator test and I approve.

Radiergummi:
🤖 **[reviewer]** review round 1 — 4 findings (0 blocking; approve-with-nits)

Independent review (did not write this). Verified the diff against `src/`, ran the new + conflict suites (35 passed), PHPStan clean on all changed files, and pressure-tested the `RewriteToAttribute` mechanism with two throwaway PoCs. The mechanism is **correct**: a `RewriteToAttribute` composes correctly with another `RewriteToAttribute` on a *neighboring* node (two sibling promoted params on one Data class, one applicator pass → both applied, byte-preserving, no corruption) and conflicts correctly with any op on the *same* node (Rewrite + SetAttributeArgument same property → Rewrite applies, Set skipped `Conflict`, no corruption). Deviations 1–5 and the inert-orphan finding are all sound and documented. Boundaries clean (no SwaggerPhp leak into Support/Core/Contracts; ResourceField never emitted in Phase 1). Verdicts on the three flagged axes:

- **New `RewriteToAttribute` primitive** — sound. The single-pass-over-mutated-tree corruption class from the prior run does **not** recur: each op only mutates its own located member node, `array_unshift` on `attrGroups` rebuilds only that member's array, and `changesLayout()` (`FixConflictDetector:122`) now treats Rewrite as a layout-changer so any same-node pairing skips the second op. Verified empirically. **But see finding 1** — this is asserted only by a single-finding golden; the multi-fix-per-file path (which this rule routinely produces) has no checked-in test, and that is precisely the regression class that bit before.
- **Acceptance-criterion reinterpretation (doc-unchanged → bidirectional `SchemaEquivalence`)** — the redefinition is **sound**. An orphan `#[OA\Property]` on a Data property is genuinely inert under the SpatieData plugin, so "document byte-identical" was never the right invariant; "the emitted attribute reproduces the schema the annotation *declared*" is the correct one, and it is enforced by the `reproduces()` bidirectional containment guard (`OaReplaceableByAttribute:261`) + the generation-level twin test. Activating documented-but-dead metadata is a correctness *improvement*, not a surprising change — the author wrote that metadata intending it to apply. **But see finding 2** — the load-bearing guard itself is untested.
- **Scope narrowing to Data→`#[ResponseField]` only** — **acceptable, not a gap.** Resource keys live inside the class-level `#[OA\Schema(properties:[…])]`, which single-member removal cannot soundly address; deferring rather than emitting un-appliable findings is the right call. Documented in the coder comment and the rule docblock.

**Findings:**

1. **Test gap — `RewriteToAttribute` has zero conflict-detector / multi-fix-per-file coverage** (`tests/Unit/Lint/Fix/FixApplicatorConflictTest.php`; `src/Lint/Fix/Ast/FixConflictDetector.php:122-127`). `RewriteToAttribute` is wired into `changesLayout()` and runs through the same single-pass applicator that previously mis-mutated sibling nodes, yet no test exercises it in combination — neither the **neighboring-node** case (two Rewrites on sibling properties of one file in one `apply()`, which this rule emits whenever a Data class has ≥2 replaceable properties) nor the **same-node conflict** case (Rewrite + Set → second skipped `Conflict`). Both behaviours are correct (I verified via PoC) but unguarded. Add two `FixApplicatorConflictTest` cases: (a) two `RewriteToAttribute` ops on `name` and `age` promoted params of one class → `applied=2, skipped=0`, byte-exact expected source; (b) a `RewriteToAttribute` + a `SetAttributeArgument` on the same property → `applied=[rewrite], skipped[0].reason=Conflict`, source carries only the rewrite. This is the highest-value addition given the prior corruption regression lived here.

2. **Test gap — the `reproduces()` soundness-rejection branch is untested** (`src/Plugins/SwaggerPhp/Lint/OaReplaceableByAttribute.php:151-155`). This is the deepest correctness guard: the mapper succeeds (all keys scalar) yet the candidate attribute's `descriptor()->toSchema()` does **not** bidirectionally equal the authored schema → skip + log. The whole soundness story of deviation 2 rests on it, and codecov shows 16 missing lines in this file (this branch among them). Add a feature/unit case with a fixture whose `#[OA\Property]` maps cleanly but whose `#[ResponseField]` contribution diverges (e.g. an authored key the attribute renders differently, or where inference adds a key into `toSchema()`), asserting the property is **not** flagged. Without it, a future `FieldAttribute` mapping change could silently start mis-rewriting and no test would catch it.

3. **Test gap — the "logged, not dropped" guarantee is never asserted** (`tests/Feature/SwaggerPhpReplaceableRuleTest.php:111-131`). The plan called for the uncovered-shape skips to be "asserted via a spy logger." The enum-skip and non-Data-skip tests assert only `not->toContain(...)` — they prove the rule doesn't *flag*, not that it *logs* the refusal. Since "never silently dropped" is the explicit Phase-1 soundness promise, bind a spy `LoggerInterface` in `replaceableRuleSetup()` and assert `logSkip()` fired with the enum reason for `EnumPropertyData::status`. Low effort, closes the promise.

4. **Nit — stale class docblock contradicts the implementation** (`src/Plugins/SwaggerPhp/Lint/OaReplaceableByAttribute.php:46,56`). The docblock states the rule maps "an Eloquent API-Resource class to `#[ResourceField]`" as current behaviour, but `fieldAttributeFor()` (`:231-238`) only ever returns `ResponseField` or null — Resource is deferred (correctly, per finding-3 of the coder comment). A reader of this `@internal` class would believe Resource classes are flagged. Reword lines 46/56 to state ResourceField/RequestField are deferred, matching `:220-227` and `fieldAttributeFor()`.

**Non-blocking observations:** (a) A few lines exceed the 100-char soft limit, notably `:164` (148 chars, the `sprintf` message arg); Pint is green so not a gate, but `:164` reads better wrapped. (b) The scanner keys `OA\Property` by its `property` field assuming no key-remapping; a remapped `OA\Property(property:'other')` degrades safely to no-finding (reflection finds no PHP property) but is untested — fine to leave for a follow-up.

None of these block correctness. Recommend the coder add findings 1–2 (and ideally 3) before merge given this PR is already a needs-human candidate.

Radiergummi:
🤖 **[coder]** addressed round 1; local gates green (Pint passed, PHPStan `[OK] No errors`, full suite 2933 passed); CI green on `eec06239`.

**Required 1 — E2E `RewriteToAttribute` conflict cases (the corruption-class insurance), `FixApplicatorConflictTest`:**
- *Same-node multi-fix:* a method with two query `#[OA\Parameter]`s → two Rewrites on one Method node. Asserts the first lands, the second is skipped `FixSkipReason::Conflict`, the `#[OA\Get]` is intact, the output is byte-exact, and a **second pass converges** (re-emitted Rewrite for `b` lands).
- *Neighboring-node:* two Rewrites on sibling promoted params of one Data class → both apply in one pass, byte-exact output.
- **Behaviour fix surfaced by the same-node convergence test:** I dropped the `hasAttributeClass` guard from `rewriteToAttribute()` (kept it on plain `addAttribute()`). The package's field attributes are `IS_REPEATABLE`, so a sibling rewrite legitimately adds a second `#[QueryParam]` with different args; the guard wrongly blocked the re-run. Re-emit idempotency is preserved the right way: a stale `attributeIndices` (source already gone) makes `removeAttributesByIndex` return false and the rewrite aborts without adding. Existing 7-case conflict suite + the fixer golden suite stay green.

**Required 2 — `reproduces()` soundness-rejection branch** (was uncovered): new feature test with a `#[OA\Property]` whose description carries an `@example` directive on its own line. The field attribute parses that out of its rendered description, so `#[ResponseField]` would not reproduce the authored schema verbatim → bidirectional `SchemaEquivalence` fails → not flagged + logged. (Verified the divergence: candidate emits `example: 'Ada'` + stripped description, authored has neither.)

**Required 3 — logged-not-dropped:** spy `LoggerInterface` (capturing `AbstractLogger`) bound into the rule; asserts the reproduces-rejection path *and* the enum-carrying skip path emit their `migration.oa-replaceable-by-attribute: left … in place (…)` log line.

**Minors:** corrected the stale class docblock (Data→`#[ResponseField]` only; `#[ResourceField]`/`#[RequestField]` deferred, with the why); logged the previously-unlogged "no member-level OA annotation found" skip path in `evaluateProperty`; wrapped the 148-char `sprintf`.

**#7 (optional, skipped):** strengthening the same-document invariant to run the fixer's own lint→fix→regenerate is not cheap — the rewritten class can't be re-`require`d under its own name in-process, so it'd need a subprocess generate (flaky). The current twin-fixture generation invariant + the new `reproduces()`-rejection coverage already pin the soundness contract. Flagging the trade-off per the lead's "skip if it balloons".

Radiergummi:
🤖 **[reviewer1]** review round 2: **approved, no concerns.**

Re-ran the suites on `eec06239`: new `FixApplicatorConflictTest` **7 passed** (incl. the two `RewriteToAttribute` cases), feature **12 passed** (was 10, +2), fix group **108 passed** (was 106, no regression). All three round-1 findings satisfied.

**Finding #1 (the corruption-class insurance) — fully closed.** `FixApplicatorConflictTest` now drives the real `FixApplicator`:
- *Same-node:* two `RewriteToAttribute` on one Method node assert byte-exact that the first lands (`#[QueryParam(name: 0a0)]`), `#[OA\Get]` survives intact, the second `#[OA\Parameter(name: 0b0)]` is left **untouched** (no half-splice), `applied == [first]`, `skipped[0].reason == Conflict`, and a **second pass converges** (both QueryParams, no `#[OA\Parameter]`, `#[OA\Get]` intact).
- *Neighboring-node:* two Rewrites on sibling promoted params (distinct selectors) both apply in one pass, byte-exact.

This proves the `changesLayout()` classification end-to-end, not just in the detector unit.

**The guard-drop mechanism change — verified correct on all three axes the lead raised:**
- **(a) Fix is correct.** `rewriteToAttribute()` (FixOperationVisitor.php:398) now gates the early return on removal success: `if (attributeIndices !== null && !removeAttributesByIndex(...)) return;`. `removeAttributesByIndex` returns false on any stale flat index **before** mutating (it builds the per-group map and aborts on the first missing index, ahead of the splice loop), so a stale rewrite aborts cleanly without a partial removal and without adding.
- **(b) No genuinely-redundant re-rewrite slips through.** Doubly protected: the scanner only harvests `OA\*` annotations, so an already-emitted `#[QueryParam]` is invisible to the rule (no re-fire), and the fixer recomputes indices from the *live* AST each pass. The visitor-level stale-index abort is correct defense-in-depth for the batch-skip→re-emit timing only. Dropping the class guard is *required* here: the field attributes are `IS_REPEATABLE`, so a legitimate sibling rewrite must be able to add a second `#[QueryParam]` with different args — the old `hasAttributeClass` guard wrongly blocked exactly that (the same-node test is what surfaced it).
- **(c) Plain `addAttribute` idempotency untouched.** The `hasAttributeClass` guard is still in `addAttribute()` (line 379), and the fix-group suite (108) stays green.

**Findings #2 / #3 satisfied.** The `reproduces()` false branch is now covered by a real semantic-divergence fixture (an `@example` description directive the field attribute parses out, so the rendered schema diverges → bidirectional `SchemaEquivalence` rejects → skipped + logged). A spy `LoggerInterface` asserts both the reproduces-rejection and the enum-skip log lines (logged-not-dropped). The previously-unlogged `evaluateProperty` skip path is now logged; the stale class docblock and the 148-char line are corrected. The optional lint→fix→regenerate strengthening (#7) was skipped with a sound, documented rationale (in-process re-require of the rewritten class is infeasible; subprocess generate is flaky), and the existing twin-fixture generation invariant + the new reproduces-rejection coverage already pin the soundness contract.

No remaining concerns. The deviations (new `RewriteToAttribute` primitive, soundness-as-reproduces redefinition, Data→`#[ResponseField]` Phase-1 narrowing) are all documented, justified, and now test-covered. Approving on my merits.

Radiergummi:
🤖 **[reviewer]** review round 2 — **approved, no remaining concerns.**

Re-reviewed `eec06239` independently. Ran the fix + replaceable suites (**61 passed, 143 assertions**), PHPStan clean on both changed src files. Findings #1–#4 from round 1 are all satisfied with behaviour-pinning tests; I focused this round on the one real mechanism change — dropping the `hasAttributeClass` guard from `rewriteToAttribute()`.

**Mechanism change verdict — sound on all three axes the lead flagged:**

1. **The fix is correct.** The package field attributes are `IS_REPEATABLE`, so a sibling rewrite legitimately adds a *second* `#[QueryParam]`/`#[ResponseField]` with different args. The old leading `hasAttributeClass` guard aborted the second sibling's rewrite (seeing the first application already present) → the convergence re-run could never land parameter `b`. Removing it from the Rewrite path is the right fix; the new abort is `attributeIndices !== null && !removeAttributesByIndex(...)` — a stale/exhausted index aborts before adding. `FixOperationVisitor.php:391-407`.

2. **Dropping the guard does not admit a redundant re-rewrite in the real pipeline.** I traced the flow and pressure-tested with a throwaway PoC. `FixRunner::run()` (`src/Lint/Fix/FixRunner.php:53-94`) runs **once per process**: lint → collect fixes for *that* lint's findings (indices computed fresh from current source) → one `apply()` batch. Two Rewrites on the same node conflict → second skipped `Conflict`, left in `remaining`, **never replayed in-process**. A subsequent `--fix` re-lints the now-modified source, where the rewritten `#[OA\Parameter(name:'a')]` no longer exists, so the scanner emits no finding for it and no fix targets it. The convergence test (`FixApplicatorConflictTest.php:302-357`) proves exactly this end-to-end. This is the same fresh-index-per-pass contract `RemoveAttribute`/`SetAttributeArgument` have always relied on (their existing re-run cases recompute indices identically), so `RewriteToAttribute` is now consistent with the established model, not a special case.

   *Note (non-blocking, defense-in-depth only):* I confirmed via PoC that if a caller **manually replayed a stale Rewrite Fix** whose now-stale index happens to point at a *different still-removable* attribute, the op would mis-remove it and add a duplicate. That path is **unreachable** — no caller replays a stale `Fix` object against changed source; every re-emission goes through a fresh lint. The op simply trusts its caller to pass live indices, which `FixRunner` always does. The new doc comment states this contract accurately. Worth a one-line awareness if a future caller ever caches `Fix` objects across source mutations, but nothing to change here.

3. **Plain `addAttribute` idempotency is unaffected.** The `hasAttributeClass` guard was **kept** on `addAttribute()` (`:379`); only `rewriteToAttribute()` changed. The `AddAttribute`/`SetAttributeArgument` golden + conflict suites stay green (re-ran, all pass).

**Findings #1–#4 confirmed satisfied:**
- **#1** — `FixApplicatorConflictTest` adds both the same-node convergence case (two query Rewrites on one method → first lands byte-exact, `#[OA\Get]` intact, second skipped `Conflict`, second pass converges) and the neighboring-node case (two sibling promoted-param Rewrites → both apply in one pass, byte-exact). This is the corruption-class insurance, end-to-end through the real `FixApplicator`. ✔
- **#2** — `reproduces()`-rejection branch covered by `DirectiveDescriptionData` (an `@example` directive embedded in the OA description that `#[ResponseField]` parses into a separate `example` + stripped description → bidirectional `SchemaEquivalence` fails). Genuine divergence, not contrived; asserts not-flagged **and** logged. ✔
- **#3** — spy `LoggerInterface` asserts the logged-not-dropped line on both the reproduces-rejection and the enum-skip paths; plus the previously-unlogged `propertyShape === null` skip now logs. ✔
- **#4** — class docblock corrected to Data→`#[ResponseField]` only, with the deferral rationale; 148-char `sprintf` wrapped. ✔

No new concerns. Boundaries still clean, conventions OK, CHANGELOG + docs catalog row + Migration-rules paragraph honest about Phase-1 scope. **Approved** — this is my second independent approval.

Radiergummi:
🤖 **[lead]** **Escalated → needs-human.** Two independent reviewer approvals, all CI green (PHP 8.4/8.5 × Laravel 12/13 + swagger-php 5.8 + Snapshot/quality/Coverage), full suite green. **The code is sound and vetted — this is a design-acceptance gate, not a quality blocker.** Three decisions are yours:

**1. Acceptance-criterion reinterpretation (the substantive one).** Issue #195 says the generated document is *unchanged* after `--fix`. That does **not** hold for the `OA\Property`-on-a-Data-class shape: an orphan `#[OA\Property]` is left inert by swagger-php (never merged into the inferred Data schema), so rewriting it to `#[ResponseField]` *activates* previously-dead metadata — the document changes. The coder + both reviewers argue this is a **correctness improvement**, not a regression, and redefine soundness as a **bidirectional `SchemaEquivalence` check** (the emitted attribute reproduces exactly what the annotation *declared*), which is what's tested. **Accept this reinterpretation?** If yes, #195's acceptance criteria should be updated to match.

**2. New AST primitive `RewriteToAttribute`.** Added to the Fix backend (atomic remove+add in one node visit) because `FixConflictDetector` rejects a separate same-node Add+Remove pair and `FixRunner` does no re-lint pass — the plan's two-fix design would land half-applied. Reviewed hard against the prior source-corruption class: it's correctly in `changesLayout()`, E2E applicator goldens (same-node convergence + neighboring-node) drive the real mutation, and the round-1 goldens **caught a real bug** (an idempotency guard wrongly blocked a repeatable attribute's sibling rewrite — now fixed, re-verified). **Accept a new primitive in the fix-backend surface?**

**3. Scope narrowed to `Data → #[ResponseField]` only.** Planned `Resource → #[ResourceField]` and `RequestField` were deferred: Resource keys live inside the class-level `#[OA\Schema(properties:[…])]`, which member-granularity removal can't address; deferring beats emitting un-appliable findings. Query `#[OA\Parameter]` → `#[QueryParam]` shipped as planned. **Accept the narrowing** (and want the deferred shapes filed as Phase-2 follow-ups)?

**Why a human gate at all** (both reviewers approved on correctness): `src/` diff ≈ 850 LOC (rule 389 / fixer 284 / mapper 180) — well over the ~150 auto-merge threshold — *plus* the new primitive *plus* the criterion reinterpretation. Any one trips the gate.

**Non-blocking note for the record** (reviewer2 PoC): a *theoretical* corruption exists only if a caller replays a **stale `Fix` object** against already-mutated source — unreachable today (no caller caches Fixes across mutations; `FixRunner` recomputes indices fresh each pass). Worth a hardening note only if a future caller ever caches Fix objects. Branch + worktree left in place.

Radiergummi:
For your questions:
1. Change that to "unchanged or improved", so yes, accepted;
2. Yes, a new primitive is fine;
3. Fine by me

Radiergummi:
🤖 **[lead]** Moritz accepted all 3 design decisions — cleared to merge. One directive: reframe the soundness criterion from 'document unchanged' → **'unchanged or improved'** in the user-facing wording (docs/linting.md Migration-rules paragraph + CHANGELOG + PR body where stated). Coder applying that, then merging on green.

Radiergummi:
🤖 **[coder]** reframed the soundness criterion to "unchanged or improved" per Moritz's accepted phrasing — `docs/linting.md` Migration-rules paragraph, the `CHANGELOG.md` entry, and the PR body's invariant bullet. Wording only, no behavior or test changes; kept the "bidirectional `SchemaEquivalence` reproduces what was declared" detail alongside. Pushed `acb6531b` (docs-only diff: CHANGELOG + docs/linting.md). Local `composer check` green (Pint passed, PHPStan OK, 2933 tests). CI green; the substantive matrix's green on `eec06239` stands (this commit is path-filtered docs-only). Ready to merge.

Radiergummi:
🤖 **[lead]** All 3 design decisions accepted by Moritz; two independent reviewer approvals; full matrix green (incl. swagger-php 5.8 + PHP 8.5 + Snapshot). Soundness criterion reframed to 'unchanged or improved'. Squash-merged → closes #195. Phase-1 of the oa-replaceable-by-attribute migration (Data→#[ResponseField], query OA\Parameter→#[QueryParam]) + the new RewriteToAttribute AST primitive land.

Issue #199, opened 2026-06-08
Migration: move document-level `@OA\Info` / `@OA\Server` / `@OA\SecurityScheme` → config


Radiergummi:
Parent: #122. The third migration target, alongside removal (#122/#196/#197/#198) and attribute-rewrite (#195).

## Summary

Document-level swagger-php annotations have **no per-route attribute equivalent** — in laravel-openapi they live in `config/openapi.php`:

| OA annotation | → config key |
|---|---|
| `OA\Info` / `OA\Contact` / `OA\License` | `openapi.info` |
| `OA\Server` / `OA\ServerVariable` | `openapi.servers` |
| `OA\SecurityScheme` / `OA\Flow` | `openapi.security_schemes` |
| root-level `OA\Tag` (name + description) | `openapi.tags` |
| document-level `OA\ExternalDocumentation` | `openapi.external_docs` |

l5-swagger / swagger-php apps routinely park these (especially `@OA\Info` and `@OA\SecurityScheme`) on a base controller, so this is real annotation volume that neither the removal rule nor the rewrite rule touches.

## Open decision

What should the migration do here? Editing `config/openapi.php` is materially riskier than removing an annotation or inserting an attribute (it's a single shared file, often already hand-tuned). Options, roughly in increasing ambition:

1. **Report-only** — a `migration.*` lint rule that detects authored document-level annotations and prints the equivalent config key + value to set, without writing anything. Safest; leaves the edit to the user.
2. **Report + scaffold** — additionally emit a config diff / snippet the user can paste.
3. **Auto-write** — patch `config/openapi.php`. Probably too invasive; likely out of scope.

Leaning toward (1), possibly (2).

## Dependency

`AuthoredAnnotationScanner` currently indexes only schemas and operations. Detecting `OA\Info`/`OA\Server` / `OA\SecurityScheme` needs the scanner to also expose the document-level annotations its swagger-php pass already parses (they land on the analysed `OpenApi` root), or a small reflection pass over the scanned sources.

## Acceptance criteria

- A `migration.*` rule (or, if we decide against a rule, a documented manual step) covering the `OA\Info` / `OA\Server` / `OA\SecurityScheme` → config mapping above.
- If a rule: it reports each authored document-level annotation with the config key to set; golden/feature coverage; `docs/linting.md` row + `CHANGELOG.md` entry.

Radiergummi:
[...]

PR #400, opened 2026-06-18
feat(#197): migration.oa-redundant-property-with-inference — per-property OA\Property removal (member-level attr shape)

### LINKED ISSUE #197: Migration: per-property redundant `#[OA\Property]` / `@OA\Property` removal within still-needed schemas

Radiergummi:
## Plan — per-property `OA\Property` redundancy removal (#197)

Closes #197

`migration.oa-redundant-with-inference` (PR #185) is **whole-class, all-or-nothing**: it fires only
when inference subsumes the *entire* authored schema. A Data class where `name`/`count` merely
restate inferred types but `email` carries a load-bearing `description` gets **nothing** flagged.
This adds **per-property** granularity: flag an individual member-level `#[OA\Property]` inference
subsumes, even while the enclosing schema must stay.

### Tier

**Tier 0/1, same footing as the existing migration rules.** Reads swagger-php's already-parsed
annotation objects (harvested) + reflection; the fixer mutates source via the merged
format-preserving AST backend. No method-body parsing, no Tier-2 dataflow.

### The soundness boundary (the hard part — read first)

The issue's caveat: per-property removal is unsound when the authored schema "survives independently
under its own name," because deleting one of *its* properties silently mutates that emitted
component. The merged code already draws exactly the right line, and Phase 1 must stay inside it:

- An authored `#[OA\Property]` is **member-granular and safely removable only when it sits on a
  class member** (a Data property or promoted constructor parameter) — the `RemoveAttribute` /
  `OaAttributeRemover` infra removes one member's `#[OA\*]` attribute at member granularity, and the
  property merges into the *convention-derived* schema (not emitted as a standalone authored
  component). Fixture precedent: `RedundantAttributeData` (each `#[OA\Property]` on a promoted param).
- An `OA\Property` nested inside a class-level `#[OA\Schema(properties: [...])]` **attribute**, or
  inside a `@OA\Schema(... @OA\Property ...)` **docblock** (precedent: `RedundantDocblockData`), is
  **out of scope**: removing one entry needs intra-array-argument / intra-docblock-line surgery that
  no current AST op supports (`RemoveAttribute` is whole-attribute; `DocblockAnnotationRemover`'s
  paren-depth walk treats the outer `@OA\Schema(` as one block and would swallow the nested
  `@OA\Property`). And such a schema is typically `registerSchema`'d as a named component — the exact
  "survives independently" hazard. These shapes are **logged and left in place**, never partially
  edited.

So Phase 1 scope = **member-level `#[OA\Property]` attributes on Spatie Data classes**, mirroring the
host-class line `OaReplaceableByAttribute::fieldAttributeFor()` already established and documented.
This keeps the rule sound *and* small. Widening to the nested/docblock shapes requires a new
element-granular AST primitive and is a deferred follow-up (note it in the PR body).

**Belt-and-braces guard:** before flagging, reuse the harvester's name-supersession signal — skip
(log) any property whose declaring class also has an authored `#[OA\Schema]` registered as a named
component under its own name (`scanner->schemaForClass()` defined + that name registered), so we
never remove a property from a schema that *is* emitted authored-as-is. In practice a member-level
Data `#[OA\Property]` won't, but the guard makes the soundness explicit and survives future shapes.

### Fire condition

For each authored member-level `#[OA\Property]` on a Data class: the property is redundant when the
**inferred** schema for that class carries a property of the same name that **subsumes** the authored
one (`SchemaEquivalence::subsumes(inferredProperty, authoredProperty)` — authored ⊆ inferred, the
same direction as the whole-class rule). A property carrying a description/title/example inference
can't derive fails subsumption and is kept. This is the *redundancy* direction (one-way subsumption),
distinct from #195's *replaceability* bidirectional-equality check — do **not** confuse the two.

### Files

**Create**
- `src/Plugins/SwaggerPhp/Lint/OaRedundantPropertyWithInference.php` — the rule. Implements `Rule`,
  `ComponentSchemaRule`, `FixableRule`; `excludedStages()` = `[HarvestAuthoredAnnotationsStage::class]`.
  `id()` = `migration.oa-redundant-property-with-inference`; `severity()` = `Improvable` (level 4).
  Per Data-class schema node: enumerate `scanner->propertiesForClass($class)`, pull the inferred
  schema (`context->inference->schemaForClass($class)`), and for each authored property whose
  member-level shape is an `#[OA\*]` attribute (reuse `OaReplaceableByAttribute::propertyShape`-style
  reflection — extract that to a tiny shared helper if it stays private there) and which inference
  subsumes, emit a `Finding` addressed to the property/param. Apply the supersession guard. Log every
  property examined-but-kept with the reason (not subsumed; nested/docblock shape; schema survives
  independently).
- `src/Plugins/SwaggerPhp/Lint/Fix/RedundantPropertyAnnotationFixer.php` — the fixer. Per finding,
  locate the property/param node, build the member's `#[OA\*]` removal via `OaAttributeRemover`
  (it already removes a member's OA attributes at member granularity, dropping an emptied group),
  emit one `Fix` (or none if the node can't be located). Reuse `ResolvesDeclaringFile` +
  `FixContext::classNode`, mirroring `RedundantOaAnnotationFixer`'s attribute branch — this fixer is
  the attribute-shape-only, single-member subset of it.

**Reuse, do not duplicate**
- `SchemaEquivalence::subsumes` (one-way), `scanner->propertiesForClass`, `context->inference`,
  `OaAttributeRemover`, `RemoveAttribute`, `AuthoredAnnotationShape`, `copy_schema_fields`.

**Modify**
- `src/Plugins/SwaggerPhp/SwaggerPhpPlugin.php` — one `$registry->addRule(...)`.
- `docs/linting.md` — one catalog row inside the `lint-rule-catalog` markers + a sentence in the
  Migration-rules section distinguishing per-property from whole-class removal and stating the
  member-level-attribute-only Phase-1 boundary.

If a shared member-shape helper is the cleanest way to avoid copy-pasting `propertyShape` out of
`OaReplaceableByAttribute`, extract it to `Lint/Support/` and have both rules use it — but only if it
stays trivial; otherwise duplicate the few lines rather than over-abstract.

**Feature — `tests/Feature/SwaggerPhpRedundantPropertyRuleTest.php`** (mirror `SwaggerPhpMigrationRuleTest`)
New fixture `tests/Fixtures/SwaggerPhp/MixedPropertyData.php`: a Spatie Data class with
member-level `#[OA\Property]` on each promoted param — two redundant (`type` only, matching the
inferred PHP type) and one load-bearing (`description: 'human prose'`). Acceptance criterion:
- flags **only** the two redundant properties, by member name (assert finding `sourceMember`s).
- does **not** flag the description-carrying property.
- **same-document invariant:** lint → `--fix`, then assert the emitted component schema for the class
  is byte-identical before and after (the removed properties were pure restatements; the kept one and
  its description survive). Use the `inferenceOnly` / generator comparison helpers already in
  `SwaggerPhpMigrationRuleTest`.
- governance (copy verbatim): off on a default-level run; disabled by `--skip migration.*`; expanded
  by `openapi.lint.enabled_rules: ['migration.*']`.
- `exposes its fixer and human-readable description`.

**Edge/branch cases (feature + a focused fixer unit test)**
- **nested / docblock shape skipped:** point the rule at `RedundantDocblockData` (nested
  `@OA\Property` in a class docblock) → **no** per-property findings; assert the skip is logged
  (spy logger). Same for a class-level `#[OA\Schema(properties: [...])]` attribute fixture if one
  exists or add a minimal one.
- **survives-independently guard:** a Data class that *also* carries an authored
  `#[OA\Schema(schema: 'X')]` registered as a named component → its member properties are **not**
  flagged (log the reason). (If no such fixture exists, add a minimal one.)
- **fixer golden, member-attribute shape:** `tests/Unit/Plugins/SwaggerPhp/RedundantPropertyAnnotationFixerTest.php` —
  build the finding for one redundant `#[OA\Property]` on a promoted param, run the fixer, apply via
  `FixApplicator`, assert the source byte-for-byte: that one attribute (and its now-empty group) gone,
  the sibling `#[OA\Property]` and all whitespace/comments untouched.
- **fixer degrade:** member not locatable → fixer yields nothing.
- whole-class rule unaffected: a class where *all* properties are redundant is still flagged by
  `migration.oa-redundant-with-inference` (whole-class) — confirm the two rules don't double-fix the
  same source (they target different granularities; the per-property rule fires on member-level
  attributes, the whole-class rule on the class node — assert no overlap on the all-redundant fixture,
  or document that `--fix` resolves them in one pass without conflict via the existing
  `FixConflictDetector`).

- **Not controversial by precedent:** fourth rule in the auto-merged `migration.*` family (#185,
  #199/#398, #195/#399). Off by default, level 4, `class_exists`/plugin-gated.
- **No new config key**, no existing-rule severity change, no `src/Contracts/**` touch, no
  stage-order change, no public-signature change.
- **Watch item:** the member-level-attribute-only Phase-1 boundary is deliberate — the nested
  `#[OA\Schema(properties:[…])]` / `@OA\Schema` docblock shapes need an element-granular AST primitive
  and are deferred. If the reviewer wants those now, that's a scope widen (new AST op) — raise it,
  don't guess. Per the lead: **if the `src/` diff looks like it'll exceed ~150 LOC, stop and flag** so
  Moritz can gate it as he did #399.

---

## 🤖 [reviewer1] plan-review amendments (verified against merged `src/`)

**Approved with the corrections below.** Tier-1 confirmed (parsed-annotation + reflection reads, AST-mutation fixer; no body parsing, no Tier-2). Scope surgical, follows the `migration.*` precedent. Both soundness pillars hold. Apply these when coding:

**1. [REQUIRED] The rule MUST implement `NeedsInferenceDocument`.** The Files list names only `Rule, ComponentSchemaRule, FixableRule`. But `LintRunner::inferenceExcludedStages()` only consults `excludedStages()` for rules implementing `NeedsInferenceDocument`, and `LintContext::$inference` defaults to an **empty** `InferenceView` unless at least one active rule declares that interface (LintRunner.php:188-189,250). Without it, `context->inference->schemaForClass($class)` returns null, the fire condition never triggers, and the rule silently does nothing. The whole-class sibling `OaRedundantWithInference` implements it — mirror that exactly (interface + `excludedStages()` returning `[HarvestAuthoredAnnotationsStage::class]`).

**2. [LOAD-BEARING — make explicit] `propertyShape` reflection is what makes the member-attribute-only boundary real.** The scanner's `propertiesForClass()` returns the **merged** property set: swagger-php folds member-level `#[OA\Property]` *and* class-level `#[OA\Schema(properties:[…])]` entries into the same `$schema->properties` array (`AuthoredAnnotationScanner::indexProperties`), so they are **indistinguishable in scanner data**. The only way to filter to the safely-removable member-level-attribute shape is to re-read the actual `ReflectionProperty`/`ReflectionParameter` attributes (the `OaReplaceableByAttribute::propertyShape`-style check). The plan relies on this; just be sure the shared helper returns `Attribute` (not `Docblock`/null) and the rule skips+logs everything else. A nested `OA\Property` will appear in `propertiesForClass()` with no member-level `#[OA\*]` on its reflector → correctly skipped.

**3. [PRECISION on the guard] The harvester's named-component trigger is `is_defined($schema->schema)`, not merely `schemaForClass() !== null`.** `HarvestAuthoredAnnotationsStage::registerSchema()` registers a class schema as an independent component **only when it carries an explicit `schema:` name** (line 294 early-returns otherwise). A bare member-level `#[OA\Property]` on a Data class with no `#[OA\Schema(schema:'X')]` is never emitted authored-as-is — it merges into the convention-derived schema, which re-derives the same property, so removal is safe. Implementing the guard as "`schemaForClass()` non-null" alone is **sound but over-conservative** (it would also skip a class carrying a bare nameless `#[OA\Schema(properties:[…])]`, which is fine to skip in Phase 1 anyway). Prefer gating on `is_defined($schema->schema)` for precision; if you keep the looser form, note in the docblock that it is deliberately conservative. Either way it never produces an unsound removal.

**Soundness pillars — verdict:**
- **(1) member-attribute-only:** airtight given #2+#3. A member-level Data `#[OA\Property]` cannot belong to an independently-emitted component (no `schema:` name ⇒ not registered), and the guard covers the one case where it could (`#[OA\Schema(schema:'X')]` present). No harvester collision/supersession case slips through that the guard misses.
- **(2) one-way subsumption (authored ⊆ inferred):** correct for removal and **consistent** with the whole-class rule (`SchemaSubsumption::subsumes(inferred, authored)`). Must NOT be bidirectional here — removal should delete anything inference reproduces even when inference adds more (e.g. a synthesised example); #195's bidirectional equality is the *replaceability* test, a different question.

**Non-blocking:** the "no double-fix overlap with the whole-class rule on an all-redundant class" test the plan names is worth keeping — both rules target the same source (class node vs member nodes) and the `FixConflictDetector` (a member `RemoveAttribute` and the class-level removal are layout-changers on different `TargetSelector`s, so they don't conflict; both can land) — assert the all-redundant fixture converges cleanly. Confirmed no controversy: no config key, no Contracts/stage-order/signature change, level-4 `migration.*`, `@internal`.


Radiergummi:

Key decisions flagged for review:
- **Soundness boundary = member-level `#[OA\Property]` attributes on Spatie Data classes only.** These are member-granular (removable via the existing `OaAttributeRemover`/`RemoveAttribute`) and merge into the *convention-derived* schema, so removing one does not mutate any independently-emitted component. The nested `#[OA\Schema(properties:[…])]` and `@OA\Schema` docblock shapes are **deferred + logged** — per-element removal there needs an element-granular AST primitive (`RemoveAttribute` is whole-attribute; the docblock remover swallows the outer `@OA\Schema(`), and such schemas are the "survives independently under its own name" hazard the issue calls out.
- **Belt-and-braces guard:** skip+log any property whose class also carries an authored `#[OA\Schema]` registered as a named component.
- **Fire direction is one-way subsumption** (authored ⊆ inferred), same as the whole-class rule — *not* #195's bidirectional equality. New rule ID `migration.oa-redundant-property-with-inference`, level 4, off by default.
- No new config key, no Contracts touch, no stage-order change. If the src diff trends past ~150 LOC I've asked the coder to stop and flag for Moritz's gate.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. New rule migration.oa-redundant-property-with-inference (removal-only, member-level #[OA\Property] on Data classes, one-way subsumption authored⊆inferred). Reviewer watch items: (1) the member-attribute-only boundary soundness — is deferring nested-in-#[OA\Schema(properties:)] + @OA\Schema-docblock shapes correct, and is the 'skip any class with an authored named-component #[OA\Schema]' guard sufficient to guarantee removal never mutates an independently-emitted component? (2) one-way subsumption (vs #195's bidirectional) is the right oracle for removal.

Radiergummi:

Verified against merged `src/` (#195/#398 now on main). Tier-1 confirmed; scope surgical; follows the `migration.*` precedent. Both soundness pillars hold. Three corrections to apply when coding:

**1. [REQUIRED] The rule must implement `NeedsInferenceDocument`.** The Files list omits it. `LintRunner::inferenceExcludedStages()` only consults `excludedStages()` for rules with that interface, and `LintContext::$inference` defaults to an **empty** `InferenceView` unless an active rule declares it (LintRunner.php:188-189,250). Without it, `context->inference->schemaForClass()` returns null → the fire condition never triggers → the rule silently no-ops. Mirror the whole-class `OaRedundantWithInference` exactly.

**2. [LOAD-BEARING] `propertyShape` reflection is what makes the member-attribute-only boundary real.** `propertiesForClass()` returns the **merged** set — swagger-php folds member-level `#[OA\Property]` *and* nested `#[OA\Schema(properties:[…])]` entries into one `$schema->properties` array (`indexProperties`), indistinguishable in scanner data. Only re-reading the `ReflectionProperty`/`ReflectionParameter` attributes distinguishes the safely-removable shape. The plan relies on this; ensure the helper returns `Attribute` and the rule skips+logs `Docblock`/null (nested properties appear here with no member-level `#[OA\*]` on their reflector → correctly skipped).

**3. [PRECISION] The harvester's named-component trigger is `is_defined($schema->schema)`, not merely `schemaForClass() !== null`.** `registerSchema()` registers a class schema as an independent component only when it carries an explicit `schema:` name (line 294 early-returns otherwise). A bare member-level `#[OA\Property]` with no `#[OA\Schema(schema:X)]` is never emitted authored-as-is. Gate the guard on `is_defined($schema->schema)` for precision; the looser "schemaForClass() non-null" form is sound but over-conservative (document it if kept).

**Pillars:**
- **(1) member-attribute-only — airtight** given #2+#3. A member-level Data `#[OA\Property]` cannot belong to an independently-emitted component (no `schema:` name ⇒ not registered), and the guard covers the one case where it could. No harvester collision/supersession case the guard misses.
- **(2) one-way subsumption (authored ⊆ inferred) — correct and consistent** with the whole-class rule. Must NOT be bidirectional: removal deletes anything inference reproduces even when inference adds more (e.g. a synthesised example); #195's bidirectional equality answers the *replaceability* question, not redundancy.

Non-blocking: keep the no-double-fix-overlap test (member `RemoveAttribute` vs class-level removal are layout-changers on different `TargetSelector`s, so they do not conflict and both land — assert the all-redundant fixture converges). No controversy: no config key, no Contracts/stage-order/signature change, level-4 `migration.*`. Cleared to coding once #1 is in.

Radiergummi:
🤖 **[lead]** plan-review approved (amendments folded into PR body) → coding. Three corrections the coder MUST honor: (1) [REQUIRED] rule implements NeedsInferenceDocument — without it LintRunner never builds the inference view, schemaForClass() returns null, the rule is a silent no-op (mirror OaRedundantWithInference); (2) [LOAD-BEARING] use ReflectionProperty/Parameter attribute re-read to distinguish removable member-level #[OA\Property] from nested #[OA\Schema(properties:)] props (propertiesForClass returns the MERGED set) — nested props skip+log; (3) [PRECISION] the independent-component guard triggers on is_defined($schema->schema), not schemaForClass()!==null. TDD.

Radiergummi:
🤖 **[coder] feasibility blocker** — holding before tests; the plan's core premise does not hold for Spatie Data. Three findings, each reproduced against the merged harvester/generator (not theory):

**1. On a Spatie Data class, an authored `#[OA\Property]` never merges into the inferred schema — it is inert.** Generating `RedundantAttributeData` (which carries `#[OA\Schema(schema:'RedundantAttribute')]` + a member `#[OA\Property]` on each promoted param) emits **one** component — the SpatieData-inferred `RedundantAttributeData` (`{name:{type:string}, count:{type:integer,…}}`). There is **no** standalone `RedundantAttribute` component and the authored `#[OA\Property]` data is absent from output. The SpatieData plugin owns the component for that class and the harvested authored schema loses. (Same inert-annotation class as the #195/#399 finding, but here it is fatal to the rule's value, not just a framing nuance.)

**2. The scanner indexes member `#[OA\Property]` only when the class carries a class-level `#[OA\Schema]`** (swagger-php needs a hosting schema to attach them). The plan's acceptance fixture shape — member `#[OA\Property]` on each promoted param with **no** class `#[OA\Schema]` (my `MixedPropertyData`) — indexes **zero** authored properties (`propertiesForClass()` → `[]`), so the rule cannot see them and never fires.

**3. Any class-level `#[OA\Schema]`, even nameless, yields a defined `$schema->schema`.** swagger-php's `AugmentSchemas` auto-derives the schema name from the class, and the scanner keeps that processor (it only removes `OperationId`/`AugmentTags`). So reviewer correction #3's guard — skip when `is_defined($schema->schema)` — is **true for every class whose properties the scanner indexed**.

**Net:** the guard fires (skips) precisely on the set where the scanner has properties to evaluate, because "scanner has properties" ⟺ "class has an `#[OA\Schema]`" ⟹ "`$schema->schema` is defined" ⟹ "guard skips". The intersection of *rule-can-see-the-properties* and *guard-lets-them-through* is **empty**. With the reviewer-mandated corrections honored, the rule is a guaranteed **no-op** for Spatie Data — the only Phase-1 host. The `RedundantAttributeData`-style "each `#[OA\Property]` merges into the convention-derived schema" precedent the plan relies on is incorrect; those annotations are inert.

Scanner probe (merged code):
```
RedundantAttributeData: props=[name,count] schemaName=RedundantAttribute   (guard skips: named component)
MixedPropertyData:      props=[]            schemaName=(none)               (nothing to evaluate)
ReplaceableAttributeData: props=[name]      schemaName=ReplaceableAttribute  (guard skips: named component)
```

**Options (need a premise/scope call — not guessing):**

- **(a)** Loosen the guard to *explicit-name only* by reflecting the class's own `#[OA\Schema]` attribute arguments (distinguish an authored `schema:'X'` from a swagger-php auto-derived name). The rule would then fire on member `#[OA\Property]` under a *nameless* class `#[OA\Schema]`. But per finding #1 those are still **inert**, so removal changes nothing in the document — the rule becomes "remove dead annotations", the same honest-but-reframed basis as #195's replaceable rule, **not** "redundant-with-inference". Defensible and small, but a semantics reinterpretation that should be gated.
- **(b)** Pivot the Phase-1 host to a shape where member `#[OA\Property]` genuinely merges into the emitted schema — plain non-Data classes / API-Resource models (e.g. `AttributeServer`, whose harvested `#[OA\Schema]` **is** the emitted component). There per-property removal is real and sound. Widens past the plan's "Data only" line.
- **(c)** Mark #197 not-feasible-as-specified for Data and defer / redefine.

Nothing pushed; branch is still at the seed commit. The rule + fixer + a shared `MemberAnnotationShape` helper are implemented locally and ready to adapt once the premise/scope is decided.

Radiergummi:
🤖 **[reviewer1]** independent feasibility verification (reproduced empirically with throwaway generates against the merged scanner + the coder fixtures, then deleted). **All three claims are TRUE; the net no-op claim holds. This contradicts the pillar-1 reasoning I approved in plan-review — that was wrong on the mechanism, and I own it.**

**Claim 1 — authored member `#[OA\Property]` is inert on a Spatie Data class. TRUE.** Generated `RedundantAttributeData` (class-level `#[OA\Schema(schema: 0RedundantAttribute0)]` + member `#[OA\Property]`s) through the real package generator with SpatieData active. Emitted components: exactly **one**, `RedundantAttributeData` (the SpatieData-inferred name, `name`/`count` from PHP types, **no** descriptions). There is **no** standalone `RedundantAttribute` authored component. The authored `#[OA\Property]`s neither merge into the inferred schema nor emit their own. Same inert-annotation behavior #195 documented.

**Claim 2 — the scanner indexes member `#[OA\Property]` only when the class carries a class-level `#[OA\Schema]`. TRUE.** `scanner->propertiesForClass(MixedPropertyData::class)` (the plan acceptance fixture: member `#[OA\Property]`, **no** class `#[OA\Schema]`) returns **`[]`**. swagger-php itself warns `Unexpected @OA\Property(), expected to be inside @OA\Schema…` and harvests no schema for it. Control `RedundantAttributeData` (HAS class `#[OA\Schema]`) returns `[name, count]`. The scanner only reaches `indexProperties()` from `indexSchemas()`, which iterates `components->schemas` — no class-level schema ⇒ no entry ⇒ zero properties visible.

**Claim 3 — any class-level `#[OA\Schema]`, even nameless, yields a defined `$schema->schema`. TRUE.** `RedundantBareSchemaDocblockData` carries a **bare `@OA\Schema`** (no `schema:` arg) yet harvests as `schema=RedundantBareSchemaDocblockData` (swagger-php AugmentSchemas derives the name from the class). So `is_defined($schema->schema)` is true for the **entire** set of classes the scanner indexed. My plan-review correction #3 (gate on `is_defined($schema->schema)`) does **not** rescue the rule — it is true for every visible class.

**Net: the intersection of {rule can see the properties} (needs class-level `#[OA\Schema]`) and {guard lets them through} (needs no independently-emitted named component, but every class-level `#[OA\Schema]` has a defined name) is EMPTY.** The rule is a guaranteed no-op for Spatie Data, the only Phase-1 host. Worse, the existing **whole-class** `migration.oa-redundant-with-inference` already fires on exactly these classes (it flags `RedundantAttributeData` today) because the *entire* authored schema is inert — and a load-bearing authored `description` on a Data property is inert too, so the "mixed redundant + load-bearing" niche this rule was built for **does not exist for Data**: the whole-class rule already removes the whole inert annotation.

**Recommendation: (c) — defer/redefine #197 as not-feasible-as-specified for Spatie Data, and re-spec toward (b) as the real follow-up.** Reasoning:
- **(a) loosen-guard → "remove dead annotations"** is the weakest: it would flag inert member `#[OA\Property]`s, but those already fall under the whole-class rule on Data (which removes the inert annotation wholesale), so the per-property rule adds little and risks deleting annotations a user wrote deliberately. A semantics reinterpretation for near-zero marginal coverage.
- **(b) pivot host to plain / API-Resource model classes** is where the rule premise *actually* holds — there the harvested `#[OA\Schema]` IS the emitted component and member `#[OA\Property]` genuinely merges into inference, so per-property removal is real and sound. But this is past the plan's "Data only" line and needs its own feasibility PoC + plan (confirm the merge-into-inference behavior for those hosts, and that per-property removal leaves the emitted component byte-identical). That is a fresh spec, not an in-flight tweak.
- **(c)** is therefore the honest call now: the issue as written targets a niche that is empty for its declared host. Re-open #197 scoped to non-Data model hosts (essentially (b)), or fold it into a broader "authored-schema migration for model classes" issue. This goes to Moritz.

This is a real, structural blocker, not a coding miss. The coder correctly stopped and flagged it.

Radiergummi:
🤖 **[lead]** **Escalated → needs-human: feasibility blocker, vetted.** Issue #197 as specified (per-property `#[OA\Property]` removal on Spatie Data) targets an **empty niche** — independently confirmed by coder + reviewer1 (reviewer1 retracted its own plan-review pillar after reproducing).

**Verified findings (all TRUE, via real generator + scanner probes):**
1. Authored member `#[OA\Property]` on a Data class is **inert** (SpatieData-owned component wins; authored props absent from output — same root as #195).
2. The scanner indexes member `#[OA\Property]` **only when a class-level `#[OA\Schema]` exists**; without it, zero properties indexed (the plan's `MixedPropertyData` acceptance fixture sees nothing).
3. Any class-level `#[OA\Schema]`, even nameless, yields a **defined `$schema->schema`** (auto-derived name) → the independent-component guard skips the *entire* visible set.

**Net:** {rule can see properties} ∩ {guard lets them through} = ∅ → guaranteed no-op for Data. And the whole-class `migration.oa-redundant-with-inference` rule (#185) already fires on these classes, so the "mixed redundant + load-bearing description" niche doesn't exist for Data (the description is inert too).

**Team recommendation: (c) defer #197 as not-feasible-as-specified, re-spec toward (b).**
- **(a)** loosen guard → "remove dead annotations": weakest; overlaps the whole-class rule on Data, near-zero marginal value, risks deleting deliberately-authored (if inert) annotations.
- **(b)** pivot host to plain / API-Resource model classes (where the harvested `#[OA\Schema]` IS the emitted component and member `#[OA\Property]` genuinely merges into inference): this is where per-property removal is real + sound — but it's past the plan's "Data only" line and needs its own feasibility PoC + fresh plan.
- **(c)** honest call: the issue as written targets an empty niche for its declared host.

Draft PR #400 holds only the empty seed commit (no implementation pushed); worktree in place. Awaiting your call.

Radiergummi:
🤖 **[lead]** Closing — #197 deferred as not-feasible-as-specified for Data (Moritz's call). No implementation was pushed; the verdict + (b) re-spec direction are recorded on #197 and in this PR's comments.

PR #401, opened 2026-06-18
feat(#198): SchemaEquivalence follows $ref by name (cycle-guarded, conservative)

### LINKED ISSUE #198: Migration: follow `$ref` by name in `SchemaEquivalence` subsumption

## Summary

`SchemaEquivalence` compares `$ref` strings **literally** (the conservative first cut). An authored schema that references another component under a name differing from inference's name therefore fails subsumption and is kept, even when the referenced targets are themselves equivalent. Follow `$ref` by name through **both** the authored side (`AuthoredAnnotationScanner`) and the control document, so referencing schemas are recognized as redundant when their targets are equivalent.

## Notes

- Cycle-guard the traversal (a schema graph can be recursive).
- Stay conservative: an unresolvable ref on either side → not subsumed (keep), never a false redundancy verdict.

## Acceptance criteria

- A fixture where the authored schema `$ref`s a nested component under a name that differs from inference's name, and inference reproduces the whole graph, is correctly flagged redundant.
- A fixture where a referenced target genuinely differs is **not** flagged.
- Unit tests on `SchemaEquivalence` covering ref-following + cycles.

Radiergummi:
## Plan — follow `$ref` by name in `SchemaEquivalence` subsumption (#198)

Closes #198

### Feasibility probe (done first, per the #197 lesson) — scenario holds ✅

`SchemaEquivalence` compares `$ref` strings **literally** (`contains()` falls through to `==`). The
divergent-name scenario is **real and already present in the fixtures**:

- Convention names a component by `class_basename` — so inference emits
  `RefParentData.child → $ref: #/components/schemas/RefChildData`.
- The author chose a different name (`#[OA\Schema(schema: 'RefChild')]`) — so the authored
  `RefParentData.child → $ref: #/components/schemas/RefChild`.

Both point at the **same class** (`RefChildData`), whose schema inference and the author each
reproduce. But the ref strings differ, so subsumption fails today and `RefParentData` is kept even
though it is genuinely redundant. The fix lets the comparator follow both refs to their target
schemas and compare *those*. Acceptance is **unit tests on `SchemaEquivalence`** — self-contained,
exactly as the issue states. Not a #197-style empty premise. Proceeding.

### Tier

**Not a new rule — an algorithm refinement to the redundancy oracle.** Tier-0/1 territory (it reads
already-parsed annotation graphs; no body parsing). No `migration.*` ID added; the existing
`migration.oa-redundant-with-inference` simply stops under-reporting once the oracle can follow refs.

### Design

`SchemaEquivalence` must stay a **pure, independently-unit-testable comparator** — today every call
site is `new SchemaEquivalence()`. So ref-resolution is an **optional injected resolver**, defaulting
to "no resolution" (literal compare) so all existing call sites and the existing
`SchemaEquivalenceTest` cases are byte-for-byte unchanged.

- Introduce a tiny resolver seam, e.g. `SchemaEquivalence::__construct(?SchemaRefResolver $refs = null)`
  where `SchemaRefResolver::resolve(string $refName): ?OA\Schema`. **Two sides, two names:** the
  authored ref name resolves against the authored side, the broader ref name against the control
  side. Cleanest is a single resolver carrying both lookups, or pass a side-tagged resolver — pick
  one and keep it `@internal`. (Decision: one `SchemaRefResolver` with `resolveAuthored()` /
  `resolveInferred()`, since the two namespaces are distinct and must not be conflated.)
- In `contains()`, when **both** sides are schema-maps carrying a differing `$ref` (literal `==`
  already handles the equal-name case for free), resolve each name to its target schema and recurse
  into `contains(resolvedBroader, resolvedNarrower)`. Resolution applies only when literal compare
  fails, so the common path is untouched and cheap.
- **Cycle guard:** carry a visited-pair set (`{authoredName ⇒ inferredName}` already seen) down the
  recursion; on re-entry of a pair, return `true` (the cycle's equivalence is being established by the
  outer frame — standard coinductive bisimulation). A recursive schema graph must not loop.
- **Conservative fallback (load-bearing):** if **either** side's ref is unresolvable (no resolver, or
  the name isn't found), `contains` returns **false** for that branch (not subsumed → keep). Never a
  false redundancy verdict. This preserves today's behaviour exactly when no resolver is wired.

### Wiring the resolver where the rule runs

The rule (`OaRedundantWithInference`) constructs `SchemaSubsumption(new SchemaEquivalence(...))`.
Feed it a resolver backed by:
- **authored side:** `AuthoredAnnotationScanner::schemaForName($name)` (already exists).
- **inferred/control side:** a **by-name** component lookup the control document already contains but
  `InferenceView` does not yet expose. `InferenceOnlyGeneration.document->components->schemas` is keyed
  by component name; add `InferenceView::schemaForName(string $name): ?OA\Schema` (and thread the
  by-name map through `InferenceView::from()`), mirroring the existing `schemaForClass()`. Small,
  internal, additive.

`SchemaEquivalence` and the resolver are constructed in the rule/DI seam, not inside the comparator,
so the comparator stays dependency-light and unit-testable with a hand-built resolver stub.

### Files

**Create**
- `src/Plugins/SwaggerPhp/Lint/Support/SchemaRefResolver.php` — `@internal` interface (or small final
  class) with `resolveAuthored(string $name): ?OA\Schema` / `resolveInferred(string $name): ?OA\Schema`.
- (if the rule's concrete resolver needs a home) a small adapter wiring scanner + `InferenceView` to
  that interface — likely a private/anonymous impl in the rule or a one-file class under `Lint/Support/`.

**Modify**
- `src/Plugins/SwaggerPhp/Lint/Support/SchemaEquivalence.php` — optional resolver ctor arg; ref-follow
  branch in `contains()` with cycle guard + conservative fallback. Keep `@internal`. **This is the
  only behavioural change to the oracle.**
- `src/Lint/InferenceView.php` — add `schemaForName()` + thread the by-name component map through
  `from()`.
- `src/Support/Generator/InferenceOnlyGeneration.php` — if the by-name map isn't already derivable
  from `document->components->schemas` at `InferenceView::from()` time, expose it (prefer deriving in
  `InferenceView::from()` from `document` to avoid widening this VO).
- `src/Plugins/SwaggerPhp/Lint/OaRedundantWithInference.php` — construct `SchemaEquivalence` with the
  resolver. (Constructor already receives `SchemaEquivalence`; check whether to build the resolver here
  or in the DI factory — keep the rule's public constructor signature unchanged if possible; if the
  resolver must be injected, that's an `@internal` ctor change — flag it.)
- `CHANGELOG.md` `[Unreleased]` — Added/Changed entry (oracle now follows `$ref` by name).

**No `docs/linting.md` row** — internal oracle change, no new rule ID. Confirm with the reviewer but I
believe none is warranted; the existing `migration.oa-redundant-with-inference` row still describes it.

**Unit — extend `tests/Unit/Plugins/SwaggerPhp/SchemaEquivalenceTest.php`** (the issue's acceptance is
here). Build small `OA\Schema` graphs + a hand-written `SchemaRefResolver` stub:
- **(i) differing ref name, equivalent target → subsumed.** broader has `child → $ref: ChildA`,
  narrower has `child → $ref: ChildB`; resolver maps `ChildA`/`ChildB` to structurally equal schemas.
  `subsumes(broader, narrower)` is **true**. (This fails before the change.)
- **(ii) referenced target genuinely differs → not subsumed.** Same refs, but resolver maps them to
  schemas that differ (`type: string` vs `type: integer`). **false**.
- **(iii) unresolvable ref → not subsumed.** Resolver returns null for one side. **false** (and with
  **no** resolver at all, the pre-existing literal-mismatch behaviour: **false**).
- **(iv) equal ref names short-circuit** without consulting the resolver (assert via a resolver spy
  that throws if called) — the cheap common path stays cheap.
- **(v) cycle.** `A.self → $ref: A` on both sides (or `A→B→A`); resolver returns those schemas;
  `subsumes` **terminates** and returns **true**. A second cycle case where the cycle's non-ref
  content differs → **false**.
- Keep all existing `SchemaEquivalenceTest` cases green untouched (proves the default-no-resolver path
  is unchanged).

**Unit — `tests/Unit/Lint/InferenceViewTest.php`** (or extend an existing one): `schemaForName()`
returns the control component by name; null for an unknown name.

**Feature — extend `tests/Feature/SwaggerPhpMigrationRuleTest.php`**: with the resolver wired, the
existing divergent-name fixture pair (`RefParentData` → `RefChild`/`RefChildData`) is now **flagged
redundant** end-to-end (previously kept). Guard the inverse: a fixture whose referenced target
genuinely differs is **not** flagged. Reuse `RefParentData`/`RefChildData` if their current shape
already exhibits the divergence (it does: author `RefChild` vs convention `RefChildData`); confirm,
and add a "genuinely-differs" sibling fixture if none exists.
**Watch:** `RefParentData`/`RefChildData` currently back the *dangling-$ref guard* test ("does not
flag a schema another authored annotation still references by name"). Verify this change doesn't flip
that test's expectation — the guard is about RefChild being *kept* because RefParent still refs it;
RefParent itself becoming flaggable is orthogonal, but re-run that test and reconcile if it moves.

- **`@internal` infra only.** `SchemaEquivalence`, `InferenceView`, `SchemaRefResolver`,
  `InferenceOnlyGeneration` are all `@internal`. No public/`Contracts\` surface, no new config key, no
  stage-order change, no new lint rule/severity.
- **Constructor-signature watch:** the change adds an optional ctor arg to `SchemaEquivalence`
  (default null = today's behaviour) and possibly threads a resolver into `OaRedundantWithInference`.
  Both `@internal`; the optional-arg default keeps every existing call site compiling unchanged. Call
  it out so the reviewer confirms no external constructor relies on the old arity.
- **Soundness is the whole point:** the conservative-fallback + cycle-guard cases are the
  acceptance-critical tests; the reviewer should weight those. If the resolver wiring balloons past
  ~150 LOC of `src/`, stop and flag for Moritz's gate (the #399 precedent).

---

## 🤖 [reviewer1] plan-review amendments (verified empirically against `src/`)

**Approved.** I re-ran the feasibility probe end-to-end (throwaway generates, now deleted) and the scenario holds with one important nuance that *confirms* the side-tagged-resolver decision. Tier-0/1, surgical, `@internal`-only. Apply the precision note below.

**Empirical confirmation of the seam shape (pressure-test #1) — the side-tagged resolver is necessary, not speculative DI.** Verified for the `RefParentData` fixture:
- Inference-only doc: `RefParentData.child.ref = #/components/schemas/RefChildData` (class basename). `InferenceView::schemaForName('RefChildData')` will resolve; `scanner->schemaForName('RefChildData')` returns **null**.
- Authored (scanner): `RefParentData.child.ref = #/components/schemas/RefChild`. `scanner->schemaForName('RefChild')` resolves; the inferred side does **not** know `RefChild`.

So the broader-side ref and narrower-side ref live in **two genuinely distinct namespaces** — a single conflated lookup fails both ways. `resolveInferred()` (broader) + `resolveAuthored()` (narrower) is the correct, data-justified design. An **interface** seam is warranted here (not over-abstraction): the unit tests need a hand-built stub to keep `SchemaEquivalence` a pure, independently-testable comparator, which is the legitimate second implementation. Default-`null` preserves arity for all current call sites — verified: `new SchemaEquivalence()` (zero args) in `SchemaEquivalenceTest`, `SchemaSubsumptionTest`, `OperationSubsumptionTest`, plus the rule's wrappers — all keep compiling unchanged.

**[PRECISION — fix in code] The ref-follow branch must `normalize()` the resolved schemas before recursing.** `contains()` operates on **normalized** values (maps with `_`-internals + `UNDEFINED` + the `schema` name key already stripped), but the resolver returns raw `OA\Schema`. The plan's "recurse into `contains(resolvedBroader, resolvedNarrower)`" is imprecise — it must be `contains(normalize(resolvedBroader), normalize(resolvedNarrower))` (or route back through `subsumes()`), or the comparison will mis-handle raw annotation objects. Thread the cycle-visited set through whichever entry you choose.

**[PRECISION — trigger condition] The ref-follow branch fires inside the map-compare, keyed on the `ref` key.** There is no dedicated `$ref` handling in `contains()` today: a differing `ref` is just a map key whose two string values fail `==` (SchemaEquivalence.php:42,65-69). The new branch should trigger when, comparing two maps, **both** carry a `ref` key with differing string values — resolve each (broader via `resolveInferred`, narrower via `resolveAuthored`) and recurse. Equal `ref` strings already pass via `==` and must **not** consult the resolver (the spy-throws test pins this).

**Pressure-tests #2 / #3 — both satisfied.**
- Tests pin all four acceptance-critical cases (divergent-name-equivalent → subsumed; genuine-diff → not; conservative-unresolvable → keep; cycle terminates) plus the equal-name short-circuit spy and the unchanged existing cases. Add one explicit assertion: **no resolver at all** (default null) still returns the pre-existing literal-mismatch `false` for the divergent-ref pair — proves the zero-arg path is untouched.
- Fixture reconciliation is fine: the dangling-`$ref` guard test asserts the **child** (`RefChildData`) stays unflagged via `isSchemaReferencedByOtherAuthored('RefChild', …)`, which is independent of the **parent** (`RefParentData`) becoming flaggable. Orthogonal; re-run it but the expectation won't move.

**Controversy: none.** Optional ctor arg on an `@internal` comparator does not trip the public-signature gate (not `Contracts\`, not a documented extension point). No config key, no stage-order change, no new rule/severity. No `docs/linting.md` row (the existing `migration.oa-redundant-with-inference` row still describes it accurately — it just stops under-reporting). CHANGELOG entry is the only bookkeeping needed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -52,6 +52,7 @@ All notable changes to this project are documented here.
 - `PaginatorKind` recognises Spatie's `PaginatedDataCollection`; `DataResponseResolver` handles union return types as `oneOf`.
 - The inline `response()->noContent($status)` body-scan reads an explicit literal/named 2xx status (body-less); a non-literal or non-2xx status degrades. (#328)
 - A typed `UploadedFile` property on a Spatie `Data` class (transitively, through nested Data classes) emits a `multipart/form-data` body with a `format: binary` field, matching the FormRequest file-rule path; the `multipart.file-without-multipart` lint rule is reframed as a contradiction guard that fires only when a `#[RequestBody]` override forces a non-multipart media type onto a file-carrying payload. (#333)
+- The `migration.*` redundancy oracle (`SchemaEquivalence`) now follows a component `$ref` by name to its target schema on each side when the author and convention named the same component differently, so a parent schema whose only difference is a divergent child `$ref` name is no longer under-reported as non-redundant; resolution is conservative (an unresolvable ref keeps the annotation) and cycle-guarded. (#198)
 
 ### Fixed
 - Lint now surfaces top-level and field-level bare `oneOf` / `anyOf` schemas across components, responses, and request bodies. (#294, #318)
--- a/src/Lint/InferenceView.php
+++ b/src/Lint/InferenceView.php
@@ -10,6 +10,7 @@ use Radiergummi\OpenApi\Support\Generator\InferenceOnlyGeneration;
 use function is_array;
 use function is_string;
 use function ltrim;
+use function Radiergummi\OpenApi\is_defined;
 use function strtolower;
 
 /**
@@ -25,10 +26,12 @@ final readonly class InferenceView
     /**
      * @param array<class-string, OA\Schema> $schemasByClass
      * @param array<string, OA\Operation>    $operationsByKey
+     * @param array<string, OA\Schema>       $schemasByName   Control components keyed by component name.
      */
     public function __construct(
         private array $schemasByClass = [],
         private array $operationsByKey = [],
+        private array $schemasByName = [],
     ) {}
 
     /**
@@ -48,7 +51,31 @@ final readonly class InferenceView
             }
         }
 
-        return new self($generation->schemasByClass, $operationsByKey);
+        return new self($generation->schemasByClass, $operationsByKey, self::schemasByName($generation));
+    }
+
+    /**
+     * The control document's component schemas keyed by component name (the `$ref` target name).
+     *
+     * @return array<string, OA\Schema>
+     */
+    private static function schemasByName(InferenceOnlyGeneration $generation): array
+    {
+        $components = $generation->document->components;
+
+        if (!$components instanceof OA\Components || !is_array($components->schemas)) {
+            return [];
+        }
+
+        $byName = [];
+
+        foreach ($components->schemas as $schema) {
+            if ($schema instanceof OA\Schema && is_defined($schema->schema) && is_string($schema->schema)) {
+                $byName[$schema->schema] = $schema;
+            }
+        }
+
+        return $byName;
     }
 
     /** Lookup key: lowercased method, space, URI without leading slash. */
@@ -71,5 +98,13 @@ final readonly class InferenceView
     public function operationForRoute(string $method, string $uri): ?OA\Operation
     {
         return $this->operationsByKey[self::operationKey($method, $uri)] ?? null;
+    }
+
+    /**
+     * The control component schema with the given component name (a `$ref` target), or null.
+     */
+    public function schemaForName(string $name): ?OA\Schema
+    {
+        return $this->schemasByName[$name] ?? null;
     }
 }
--- a/src/Plugins/SwaggerPhp/Lint/OaRedundantWithInference.php
+++ b/src/Plugins/SwaggerPhp/Lint/OaRedundantWithInference.php
@@ -18,6 +18,7 @@ use Radiergummi\OpenApi\Lint\Visitors\ComponentSchemaRule;
 use Radiergummi\OpenApi\Plugins\SwaggerPhp\Lint\Fix\RedundantOaAnnotationFixer;
 use Radiergummi\OpenApi\Plugins\SwaggerPhp\Lint\Support\AuthoredAnnotationShape;
 use Radiergummi\OpenApi\Plugins\SwaggerPhp\Lint\Support\OaRedundancyEngine;
+use Radiergummi\OpenApi\Plugins\SwaggerPhp\Lint\Support\ScannerInferenceRefResolver;
 use Radiergummi\OpenApi\Plugins\SwaggerPhp\Lint\Support\SchemaEquivalence;
 use Radiergummi\OpenApi\Plugins\SwaggerPhp\Lint\Support\SchemaSubsumption;
 use Radiergummi\OpenApi\Plugins\SwaggerPhp\Stages\HarvestAuthoredAnnotationsStage;
@@ -43,14 +44,10 @@ final class OaRedundantWithInference implements Rule, ComponentSchemaRule, Fixab
 {
     private readonly OaRedundancyEngine $engine;
 
-    private readonly SchemaSubsumption $comparator;
-
     public function __construct(
         private readonly AuthoredAnnotationScanner $scanner,
-        SchemaEquivalence $equivalence,
     ) {
         $this->engine = new OaRedundancyEngine();
-        $this->comparator = new SchemaSubsumption($equivalence);
     }
 
     /**
@@ -75,6 +72,12 @@ final class OaRedundantWithInference implements Rule, ComponentSchemaRule, Fixab
 
         $inferred = $context->inference->schemaForClass($class);
 
+        // The oracle follows a $ref the author and convention named differently to its target schema
+        // on each side; built per run since the inference view is per spec run.
+        $comparator = new SchemaSubsumption(
+            new SchemaEquivalence(new ScannerInferenceRefResolver($this->scanner, $context->inference)),
+        );
+
         // Removing a schema still referenced by name by another authored annotation would dangle it.
         $isEssential = fn(): bool
             => is_defined($authored->schema)
@@ -83,7 +86,7 @@ final class OaRedundantWithInference implements Rule, ComponentSchemaRule, Fixab
         $finding = $this->engine->evaluate(
             $authored,
             $inferred,
-            $this->comparator,
+            $comparator,
             fn(): ReflectionClass => new ReflectionClass($class),
             fn(AuthoredAnnotationShape $shape): Finding => new Finding(
                 ruleId: $this->id(),
--- a/src/Plugins/SwaggerPhp/Lint/Support/ScannerInferenceRefResolver.php
+++ b/src/Plugins/SwaggerPhp/Lint/Support/ScannerInferenceRefResolver.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Plugins\SwaggerPhp\Lint\Support;
+
+use OpenApi\Annotations as OA;
+use Override;
+use Radiergummi\OpenApi\Lint\InferenceView;
+use Radiergummi\OpenApi\Plugins\SwaggerPhp\Support\AuthoredAnnotationScanner;
+
+/**
+ * Resolves `$ref` names against the two sides the redundancy oracle compares: the authored side via
+ * the harvested {@see AuthoredAnnotationScanner}, the inferred side via the per-run
+ * {@see InferenceView}. Built at lint time, since the inference view is per spec run.
+ *
+ * @internal
+ */
+final readonly class ScannerInferenceRefResolver implements SchemaRefResolver
+{
+    public function __construct(
+        private AuthoredAnnotationScanner $scanner,
+        private InferenceView $inference,
+    ) {}
+
+    #[Override]
+    public function resolveAuthored(string $name): ?OA\Schema
+    {
+        return $this->scanner->schemaForName($name);
+    }
+
+    #[Override]
+    public function resolveInferred(string $name): ?OA\Schema
+    {
+        return $this->inference->schemaForName($name);
+    }
+}
--- a/src/Plugins/SwaggerPhp/Lint/Support/SchemaEquivalence.php
+++ b/src/Plugins/SwaggerPhp/Lint/Support/SchemaEquivalence.php
@@ -10,8 +10,11 @@ use function array_is_list;
 use function array_key_exists;
 use function array_values;
 use function get_object_vars;
+use function is_string;
 use function Radiergummi\OpenApi\is_undefined;
 use function str_starts_with;
+use function strrpos;
+use function substr;
 
 /**
  * Tests whether one swagger-php annotation subtree is structurally contained in another.
@@ -21,22 +24,32 @@ use function str_starts_with;
  * (`additionalProperties: false`, etc.) fails containment and the annotation is kept.
  *
  * Canonical form drops `UNDEFINED` sentinels and `_`-prefixed internals; collections are
- * compared order-insensitively. `$ref` targets are compared by literal string (conservative).
+ * compared order-insensitively. Equal `$ref` strings match literally. When a {@see SchemaRefResolver}
+ * is injected, two *differing* `$ref` names are followed to their target schemas and those compared,
+ * so a component the author and convention named differently does not defeat the verdict; an
+ * unresolvable ref is conservatively treated as not contained (the annotation is kept).
  *
  * @internal
  */
 final readonly class SchemaEquivalence
 {
+    public function __construct(private ?SchemaRefResolver $refs = null) {}
+
     /**
      * Whether every value in `$narrower` is present and equal in `$broader` (which may carry more).
      */
     public function subsumes(?OA\AbstractAnnotation $broader, ?OA\AbstractAnnotation $narrower): bool
     {
-        return $this->contains($this->normalize($broader), $this->normalize($narrower));
+        return $this->contains($this->normalize($broader), $this->normalize($narrower), []);
     }
 
-    /** Recursive containment: maps check key-by-key, lists are order-insensitive, scalars use ==. */
-    private function contains(mixed $broader, mixed $narrower): bool
+    /**
+     * Recursive containment: maps check key-by-key, lists are order-insensitive, scalars use ==.
+     *
+     * @param array<string, string> $visited Ref-name pairs (`inferred => authored`) already being
+     *                                       compared up the stack, so a recursive graph terminates.
+     */
+    private function contains(mixed $broader, mixed $narrower, array $visited): bool
     {
         if (!is_array($narrower)) {
             return $broader == $narrower;
@@ -51,7 +64,7 @@ final readonly class SchemaEquivalence
         if (array_is_list($narrower) || array_is_list($broader)) {
             foreach ($narrower as $narrowerElement) {
                 foreach ($broader as $broaderElement) {
-                    if ($this->contains($broaderElement, $narrowerElement)) {
+                    if ($this->contains($broaderElement, $narrowerElement, $visited)) {
                         continue 2;
                     }
                 }
@@ -62,13 +75,67 @@ final readonly class SchemaEquivalence
             return true;
         }
 
+        if (($followed = $this->followDifferingRefs($broader, $narrower, $visited)) !== null) {
+            return $followed;
+        }
+
         foreach ($narrower as $key => $narrowerValue) {
-            if (!array_key_exists($key, $broader) || !$this->contains($broader[$key], $narrowerValue)) {
+            if (!array_key_exists($key, $broader) || !$this->contains($broader[$key], $narrowerValue, $visited)) {
                 return false;
             }
         }
 
         return true;
+    }
+
+    /**
+     * When both maps carry a *differing* `ref` name, resolves each side to its target schema and
+     * compares those instead. Equal refs already pass via `==` and never reach here. Returns null
+     * when the branch does not apply (no resolver, no/equal refs), so the caller falls back to the
+     * literal map comparison; a true/false result short-circuits it.
+     *
+     * @param array<array-key, mixed> $broader
+     * @param array<array-key, mixed> $narrower
+     * @param array<string, string>   $visited
+     */
+    private function followDifferingRefs(array $broader, array $narrower, array $visited): ?bool
+    {
+        $broaderRef = $broader['ref'] ?? null;
+        $narrowerRef = $narrower['ref'] ?? null;
+
+        if ($this->refs === null || !is_string($broaderRef) || !is_string($narrowerRef) || $broaderRef === $narrowerRef) {
+            return null;
+        }
+
+        $broaderName = self::refName($broaderRef);
+        $narrowerName = self::refName($narrowerRef);
+
+        // A cycle re-entering the same pair is established by the outer frame (coinductive).
+        if (($visited[$broaderName] ?? null) === $narrowerName) {
+            return true;
+        }
+
+        $broaderTarget = $this->refs->resolveInferred($broaderName);
+        $narrowerTarget = $this->refs->resolveAuthored($narrowerName);
+
+        // Conservative: an unresolvable target is never treated as contained.
+        if ($broaderTarget === null || $narrowerTarget === null) {
+            return false;
+        }
+
+        return $this->contains(
+            $this->normalize($broaderTarget),
+            $this->normalize($narrowerTarget),
+            [...$visited, $broaderName => $narrowerName],
+        );
+    }
+
+    /** The component name from a `#/components/schemas/Name` pointer (or the string as-is). */
+    private static function refName(string $ref): string
+    {
+        $position = strrpos($ref, '/');
+
+        return $position === false ? $ref : substr($ref, $position + 1);
     }
 
     /**
--- a/src/Plugins/SwaggerPhp/Lint/Support/SchemaRefResolver.php
+++ b/src/Plugins/SwaggerPhp/Lint/Support/SchemaRefResolver.php
@@ -1,0 +1,27 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Plugins\SwaggerPhp\Lint\Support;
+
+use OpenApi\Annotations as OA;
+
+/**
+ * Resolves a component `$ref` name to its target schema, on each of the two sides the redundancy
+ * oracle compares.
+ *
+ * The authored and inferred documents name the same class differently (convention uses
+ * `class_basename`, the author may pin an explicit `#[OA\Schema(schema: '…')]`), so a `$ref` string
+ * carries a side-specific name: the narrower (authored) ref must resolve against the authored
+ * components, the broader (inferred) ref against the inference-only control. Keeping the two lookups
+ * distinct avoids conflating the namespaces. An unknown name yields null, which the comparator treats
+ * conservatively (not subsumed).
+ *
+ * @internal
+ */
+interface SchemaRefResolver
+{
+    public function resolveAuthored(string $name): ?OA\Schema;
+
+    public function resolveInferred(string $name): ?OA\Schema;
+}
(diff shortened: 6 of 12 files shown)


Radiergummi:

**Feasibility probe first (the #197 lesson): scenario HOLDS — proceeding.** Unlike #197 this is not an empty premise. The divergent-ref-name case is already present in the fixtures:
- Convention names the child component by `class_basename` → inference emits `RefParentData.child → $ref #/components/schemas/RefChildData`.
- The author chose `#[OA\Schema(schema: 'RefChild')]` → authored `RefParentData.child → $ref #/components/schemas/RefChild`.
Both target the same class; both sides reproduce that schema; but the ref strings differ, so `SchemaEquivalence` (literal `==`) keeps `RefParentData` even though it's genuinely redundant. Self-contained and unit-testable on `SchemaEquivalence`, exactly as the issue's acceptance states.

Design highlights for review:
- **`SchemaEquivalence` stays a pure comparator.** Ref-resolution is an *optional injected resolver* (`SchemaRefResolver`), default null = today's literal behaviour, so every existing call site + unit test is unchanged. Resolution only kicks in when literal compare fails.
- **Two sides, two namespaces:** authored refs resolve via `AuthoredAnnotationScanner::schemaForName`; inferred refs need a new `InferenceView::schemaForName` (the control doc has the by-name map; `InferenceView` only exposes by-class today — small additive plumbing).
- **Cycle-guarded** (visited-pair set, coinductive: re-entry → true) and **conservative** (unresolvable ref on either side → not subsumed → keep; never a false redundancy verdict).
- No new rule/ID, no `docs/linting.md` row, `@internal` infra only. One watch item: optional ctor arg on `SchemaEquivalence` (default preserves arity) — flagged for the reviewer to confirm no external constructor relies on the old shape. Stop-and-flag if src exceeds ~150 LOC.

One reconciliation note for the coder: `RefParentData`/`RefChildData` also back the dangling-$ref-guard test — re-run that and confirm this change doesn't move its expectation (orthogonal, but verify).

Radiergummi:
🤖 **[lead]** feasibility holds (divergent-ref-name case live in fixtures) → plan-review. Oracle refinement, no new rule. Reviewer watch items: (1) optional ctor arg on SchemaEquivalence + injected SchemaRefResolver seam — does default-null genuinely preserve today's behavior + all call-site arity, and is the injected-resolver the right seam vs a method param? (2) conservative + cycle-guard are acceptance-critical — confirm tests pin 'unresolvable ref → keep, never false redundancy' + recursion; (3) reconcile the shared RefParentData/RefChildData fixtures with the existing dangling-$ref-guard test (don't move its expectation); (4) new InferenceView::schemaForName additive plumbing is minimal.

Radiergummi:

I re-ran the feasibility probe end-to-end (throwaway generates against the real generator + scanner, now deleted). The scenario holds, and the empirical data **confirms** the side-tagged-resolver decision is necessary, not speculative.

**Pressure-test #1 (the seam) — sound.** Verified for `RefParentData`:
- Inference-only doc names the child ref `#/components/schemas/RefChildData`; the **authored** scanner names it `#/components/schemas/RefChild`. `scanner->schemaForName(0RefChild0)` resolves but `schemaForName(0RefChildData0)` returns null, and vice-versa on the inferred side. The broader and narrower refs live in **two distinct namespaces** — a conflated lookup fails both ways. `resolveInferred()` (broader) + `resolveAuthored()` (narrower) is data-justified. An **interface** seam is warranted: the unit tests need a stub to keep `SchemaEquivalence` pure + independently testable (the legitimate second impl). Default-null preserves arity — verified all three `new SchemaEquivalence()` zero-arg call sites (`SchemaEquivalenceTest`, `SchemaSubsumptionTest`, `OperationSubsumptionTest`) + the rule wrappers stay unchanged. This is the right seam, not over-DI.

**Two precision corrections for the coder (in the PR body):**
1. The ref-follow branch must `normalize()` the resolved schemas before recursing — `contains()` works on normalized maps (the resolver returns raw `OA\Schema`); recurse via `contains(normalize(b), normalize(n))` or back through `subsumes()`, threading the cycle-visited set.
2. There is no `$ref` branch in `contains()` today (a differing `ref` is just a map key failing `==`). Trigger the follow when **both** maps carry a differing `ref` string; equal refs already pass via `==` and must not consult the resolver (the spy test pins this).

**Pressure-test #2 (tests) — satisfied.** Plan pins all four acceptance-critical cases (divergent-equivalent → subsumed; genuine-diff → not; conservative-unresolvable → keep; cycle terminates) + equal-name short-circuit spy + unchanged existing cases. Add one explicit assertion: default-null (no resolver) still returns literal-mismatch `false` for the divergent pair, proving the zero-arg path is byte-unchanged.

**Pressure-test #3 (fixture reconciliation) — fine.** The dangling-`$ref` guard test asserts the **child** (`RefChildData`) stays unflagged via `isSchemaReferencedByOtherAuthored(0RefChild0, …)`, independent of the **parent** becoming flaggable. Orthogonal; expectation will not move (re-run to confirm).

**Controversy: none.** Optional ctor arg on an `@internal` comparator does not trip the public-signature gate. No config/Contracts/stage-order/new-rule. No `docs/linting.md` row warranted (the existing rule row still describes it). CHANGELOG only. Cleared to coding.

Radiergummi:
🤖 **[lead]** plan-review approved (amendments in PR body) → coding. Seam justified (two distinct namespaces; injected SchemaRefResolver, default-null preserves arity on all 3 zero-arg call sites). Two precision corrections the coder MUST apply: (1) the ref-follow branch normalize()s the resolved raw OA\Schema before recursing (contains() works on normalized maps), threading the cycle-visited set; (2) there's no $ref branch in contains() today — trigger the follow only when both maps carry a DIFFERING ref string; equal refs already pass via == and must NOT hit the resolver (spy-test pins it). Also assert default-null still returns literal-mismatch=false for the divergent pair (zero-arg path unchanged).

Radiergummi:
🤖 **[coder]** implemented; local gates green (Pint passed, PHPStan `[OK] No errors`, full suite 2945 passed); CI green on `cfad2d9f`. Ready for review.

**What shipped (both reviewer precision corrections applied):**
- `SchemaEquivalence::__construct(?SchemaRefResolver $refs = null)` — default null preserves every existing zero-arg call site byte-for-byte (the 10 existing `SchemaEquivalenceTest` cases stay green untouched).
- Ref-follow branch fires **only when both maps carry a *differing* `ref` string** (correction #2): equal refs pass via `==` and never reach the resolver — pinned by a resolver-spy test that throws if consulted. The resolved raw `OA\Schema` is **`normalize()`d before recursing** (correction #1), with the cycle-visited pair set threaded through.
- Cycle guard: a re-entered `inferred⇒authored` ref pair returns true (coinductive); conservative fallback: an unresolvable ref on **either** side → not subsumed → keep (never a false redundancy verdict).
- Side-tagged `SchemaRefResolver` interface (`resolveAuthored`/`resolveInferred`) + `ScannerInferenceRefResolver` adapter (scanner = authored side, `InferenceView` = inferred side). `InferenceView` gains `schemaForName()` (by-name map derived in `from()` from the control `document->components->schemas`).

**Wiring / ctor note (flagging per plan):** `OaRedundantWithInference` no longer takes `SchemaEquivalence` in its constructor — it builds a resolver-backed comparator **per check** from `$context->inference` (the inference view is per-run, so it can't be injected at DI/construction time). The rule is only ever constructed via `addRule(::class)` (container auto-wired); no test or `src/` site constructs it directly (verified), so the `@internal` ctor change breaks nothing. The injected `SchemaEquivalence` would otherwise have been dead.

**Tests:** the four acceptance-critical unit cases (divergent-name-equivalent → subsumed; genuine-diff → not; conservative-unresolvable → keep; cycle terminates + a cycle-with-differing-content → not), plus the equal-ref short-circuit spy and the explicit **default-null still returns literal-mismatch false** assertion. `InferenceViewTest` covers `schemaForName` (hit / unknown / no-components). Feature: `RefParentData` is now flagged end-to-end (was kept) and a new `DivergentRefParentData` (authored child target carries a description inference can't derive) is **not** flagged. The dangling-`$ref` guard test's expectation **did not move** — verified `RefChildData` still stays unflagged (orthogonal: it's about the child being kept, not the parent becoming flaggable).

**LOC flag:** `src/` diff is +181/-12 — modestly over the ~150 soft gate. The breakdown is `SchemaEquivalence` +73 (the ref-follow + cycle guard + mandated `@internal` docblocks), the reviewer-justified seam (`SchemaRefResolver` interface +27, `ScannerInferenceRefResolver` adapter +37), and `InferenceView` by-name plumbing +36; the rule itself is +8/-5. No padding — the overage is the interface+adapter the plan-review explicitly blessed as non-speculative (the unit-test stub is the legitimate second impl) plus docblocks. Surfacing it for the Moritz gate per the #399 precedent; happy to inline the adapter into the rule if you'd rather trade the seam for fewer lines.

Radiergummi:
🤖 **[lead]** CI green (pushed cfad2d9f) → code-review (two independent reviewers). Reviewer focus: (1) the @internal ctor refactor of the EXISTING OaRedundantWithInference rule — it drops the injected SchemaEquivalence and builds a resolver-backed comparator per-check from $context->inference; VERIFY no direct construction anywhere + that the literal-match (non-divergent-ref) path behaves identically to before (regression risk on a shipped rule); (2) conservative + cycle-guard + differing-ref-only trigger correctness, equal-ref short-circuit spy; (3) the divergent-equivalent→subsumed / genuine-diff→not / unresolvable→keep / cycle-terminates tests. NOTE: src diff +181/-12 modestly over the ~150 gate — the seam (SchemaRefResolver iface + ScannerInferenceRefResolver adapter) was plan-review-blessed as non-speculative; do NOT demand it be inlined to game LOC. I decide the merge-gate LOC call after your verdicts.

Radiergummi:
🤖 **[reviewer1]** review round 1: approved, no concerns.

Reviewed `cfad2d9f` in the worktree; matches the plan + my plan-review amendments exactly. Ran the suites: unit `SchemaEquivalence|InferenceView` 20 passed, feature migration rule 13 passed, the full `Migration|Subsumption|Redundant|Equivalence|InferenceView` set 75 passed, PHPStan `[OK] No errors`, Pint passed.

**Highest-priority — regression on the shipped `OaRedundantWithInference` rule: clean.**
- **(a) No direct construction.** Zero `new OaRedundantWithInference(...)` in `src/` or `tests/`; it is registered via `addRule(OaRedundantWithInference::class)` and container-resolved, with `AuthoredAnnotationScanner` its only dependency. Dropping the `SchemaEquivalence` ctor arg breaks nothing — the test files that do `new SchemaEquivalence()` construct the *comparator*, not the rule.
- **(b) Literal/non-divergent-ref path identical to before.** `followDifferingRefs()` returns null when there is no resolver, no `ref`, or **equal** refs, so it falls straight through to the original `==` / map comparison. The 75 passing migration/subsumption tests (predominantly non-divergent classes) confirm no existing verdict moved. The only behaviour change is the intended divergent-ref case (previously under-reported → now flagged).
- **(c) Per-check comparator construction.** `new SchemaSubsumption(new SchemaEquivalence(new ScannerInferenceRefResolver($scanner, $context->inference)))` per `checkComponentSchema` is sound: all stateless `readonly` collaborators, one trivial allocation per schema node, resolver wraps the already-built per-run scanner + inference view. No mutable state, no perf footgun. Building per-check is correct since the inference view is per-run and cannot be injected at construction.

**Correctness — all pillars hold.**
- Conservative: unresolvable target on either side → `false` (kept), never a false redundancy.
- Differing-ref-only trigger: equal refs short-circuit via `==` and the resolver is provably not consulted (spy throws on call, test green).
- Cycle guard: visited-pair set (`inferred => authored`), coinductive re-entry returns true; my amendment that the resolved raw `OA\Schema` be `normalize()`d before recursing is honored (SchemaEquivalence.php:126-129).

**Tests — acceptance pinned, including the edges I asked for.**
- Unit: divergent-equivalent → subsumed; genuine-diff → not; unresolvable → keep; **no-resolver-at-all → literal-mismatch false (zero-arg path unchanged)**; equal-ref short-circuit spy; cycle terminates+subsumes; cycle-content-differs → false.
- Feature: `RefParentData` now flagged; `DivergentRefParentData` not flagged. Clean controlled pair — the two fixtures are structurally identical and differ *only* by the divergent child carrying a load-bearing `description`, so the not-flagged case fails for the right reason (subsumption defeated by the description, not an unresolvable ref).
- The dangling-`$ref` guard test (`RefChildData` not flagged) still passes unchanged — orthogonal as predicted.

**Conventions / boundaries / bookkeeping.** `SchemaRefResolver` (interface) + `ScannerInferenceRefResolver` (adapter) both `@internal` + strict-types; the plugin depending on `InferenceView` (general `src/Lint/`) is the allowed direction. Optional `?SchemaRefResolver` ctor arg and the optional `InferenceView` `$schemasByName` arg both default-preserve arity. CHANGELOG `[Unreleased]` entry present and honest; no `docs/linting.md` row needed (no new rule). No `Contracts`/config/stage-order touch.

Approving on my merits.

Radiergummi:
🤖 **[reviewer]** review round 1 — **approved, 1 minor finding (non-blocking).**

Independent review (did not write this). Read issue #198, the plan-reviewed PR body + reviewer1 amendments, and the coder comment. Ran the equivalence/subsumption/migration/InferenceView suites (**38 passed**), the full migration feature file (**13 passed**), PHPStan clean on all five changed src files, and pressure-tested the cycle guard + conservative fallback with a throwaway PoC. The change is correct, surgical, and boundary-clean.

**Adversarial focus 1 — regression on the existing `OaRedundantWithInference` rule: clear.**
- **No direct construction anywhere.** `grep` for `new OaRedundantWithInference` in `src/` + `tests/` → **zero hits**. It is registered only via `$registry->addRule(OaRedundantWithInference::class)` (`SwaggerPhpPlugin.php:32`, container auto-wired); the one test reference (`SwaggerPhpMigrationRuleTest.php:240`) resolves via `app(...)`, not `new`. Dropping the `SchemaEquivalence` ctor arg breaks no caller — the `@internal` ctor change is safe. The previously-injected `SchemaEquivalence` would have been dead given the per-run resolver requirement.
- **Literal / non-divergent path is byte-identical — traced, not taken on faith.** `followDifferingRefs()` (`SchemaEquivalence.php:101-108`) returns `null` (no short-circuit, falls through to the unchanged map comparison) whenever `refs === null`, either side's `ref` is absent/non-string, or `broaderRef === narrowerRef`. It is only reached inside the keyed-map branch; lists and scalars are untouched. Every other call site still uses the zero-arg `new SchemaEquivalence()` (verified: `SchemaSubsumptionTest`, `OperationSubsumptionTest`, the operation-level rule) → default null → literal behavior. The 10 pre-existing `SchemaEquivalenceTest` cases stay green untouched, and a new explicit case pins "differing $ref with no resolver → false (today path)". `SchemaSubsumption` / `OperationSubsumption` / `OaRedundantOperationWithInference` are not in the diff.

**Adversarial focus 2 — correctness: verified empirically.** PoC results:
- **Conservative fallback:** unresolvable on either side → `false` (kept); a *one-sided* ref (broader `$ref` vs narrower inline schema) correctly does **not** follow → `false` (no false-redundancy). Pinned by the unit "unresolvable → false" + "no resolver → false" cases.
- **Equal refs short-circuit:** the spy-throws resolver is never consulted on `subsumes(parent('Same'), parent('Same'))` → the cheap common path stays cheap. Pinned.
- **Cycle guard terminates:** self-referential graph → `true` in <1ms (the test asserting `true` *is* the termination proof; an unguarded recursion would stack-overflow). I additionally verified a **mutual A↔B cycle** (the visited-pair set accumulating two distinct `(inferred ⇒ authored)` pairs) terminates → `true`, and a mutual cycle whose deep authored content carries an extra property → `false`. The coinductive guard is sound for both self- and mutual-recursion.
- **Normalization:** the resolved raw `OA\Schema` targets are `normalize()`d before recursing (reviewer1 precision #1 applied), and `normalizeAnnotation()` strips the `schema` name key, so two divergently-*named* targets compare on **content**, not name — exactly the intended behavior.

**Adversarial focus 3 — tests/gaps: strong.** `RefParentData` (divergent name, equivalent target) now flagged end-to-end (was kept); `DivergentRefParentData` (divergent name, child target carries a hand-written `description` inference can't derive) **not** flagged — a *genuine* difference, not a contrived one. Unresolvable → keep; cycle terminates; default-null literal path pinned; `InferenceView::schemaForName` covered (hit / unknown / no-components). The dangling-`$ref` guard test (`does not flag a schema another authored annotation still references by name`) is green with its **expectation unmoved** — confirmed orthogonal (it's about the *child* `RefChildData` staying kept, not the *parent* becoming flaggable).

**Finding 1 [MINOR, non-blocking] — the cycle test pins only self-reference, not a mutual A↔B cycle** (`tests/Unit/Plugins/SwaggerPhp/SchemaEquivalenceTest.php:213-244`). Both "terminates" cases use a single self-referential pair (`NodeData ⇒ Node`). The mutual-cycle path — where the visited set must accumulate two distinct pairs before re-entry catches the loop — is a meaningfully different traversal and is the more likely place a future refactor of the visited-set threading would silently break (a self-ref cycle catches on the very first re-entry; a mutual cycle only on the second). It works today (I verified `A↔B` terminates → `true` and a content-divergent `A↔B` → `false` via PoC), but isn't guarded. Add one unit case: `inferred AData.b→$ref BData, BData.a→$ref AData` vs `authored A.b→$ref B, B.a→$ref A`, resolver wired both sides, assert `subsumes` terminates `true`; ideally a sibling where authored `B` carries an extra prop → `false`.

**Conventions / boundaries / bookkeeping:** `@internal` on all four touched src files; **boundary clean** — `src/Lint/InferenceView.php` (general package) imports no SwaggerPhp code; the resolver adapter lives in the plugin and depends on `InferenceView` (correct direction). CHANGELOG `[Unreleased]` entry present and honest about the conservative + cycle-guarded semantics. No `docs/linting.md` row needed (no new rule; the existing `migration.oa-redundant-with-inference` row still describes it — it just stops under-reporting). The few >100-char lines are docblock alignment / signatures; Pint is green.

No blocking concerns. The +181/-12 LOC overage is the plan-review-blessed interface+adapter seam (the unit-test stub is the legitimate second impl) — a merge-gate call for the lead, not a review finding. Recommend merge; Finding 1 is a nice-to-have the coder can fold in or defer.

Radiergummi:
🤖 **[lead]** Two independent reviewer approvals; full matrix green (PHP 8.4/8.5 × Laravel 12/13 + swagger-php 5.8 + Snapshot). $ref-by-name following in SchemaEquivalence subsumption (optional injected SchemaRefResolver, default-null = today's literal behavior; cycle-guarded + conservative). **LOC-gate note:** src +181/-12 is modestly over the ~150 soft threshold, cleared as lead judgment — the overage is the plan-review-blessed SchemaRefResolver interface + adapter seam, the change is @internal infra with NO Contracts/config/public-signature/stage-order surface, conservative, strongly tested (incl. self + mutual-cycle termination). Squash-merged → closes #198.

PR #402, opened 2026-06-18
feat(#201): harvest @OA\Response / @OA\Parameter component definitions + resolve operation refs

### LINKED ISSUE #201: SwaggerPhp harvester: harvest `@OA\Response` / `@OA\Parameter` component definitions
Extends the swagger-php harvester (#17).

## Summary

The harvester currently merges operation-level `@OA\Response`s and registers authored `@OA\Schema` model definitions, but it does **not** harvest reusable **response** / **parameter** component definitions — authored `@OA\Response(response="X", …)` and `@OA\Parameter(parameter="Y", …)` blocks that live in `#/components/responses/*` and `#/components/parameters/*`. Today an operation that points at one (`@OA\Response(ref="#/components/responses/NotFound")`) is **skipped and logged** rather than resolved (see `HarvestAuthoredAnnotationsStage::mergeAuthoredOperation()`), so the reference would otherwise dangle.

## Proposed work

- Index authored `@OA\Response` / `@OA\Parameter` **component** definitions in `AuthoredAnnotationScanner` (alongside the existing schema/operation indexes), keyed by their component name.
- Register them into `$document->components->responses` / `->parameters` through the shared registries so the post-plugin `ComponentsStage` flush picks them up (same pattern as harvested schemas), with transitive schema pull-in.
- Resolve operation `@OA\Response(ref="#/components/responses/X")` against the harvested pool instead of skipping; only skip + log when the target is genuinely absent.

## Acceptance criteria

- Authored response/parameter component definitions land in the generated document under their authored names.
- An operation referencing a harvested response/parameter component resolves rather than being dropped.
- Fixtures covering a `@OA\Response` component + an operation `$ref`ing it; serializes + validates as OAS 3.1.

## Related

Unblocks the migration counterpart (removing such definitions once inference reproduces them) — see the linked migration issue.

Radiergummi:
## Plan — harvest `@OA\Response` / `@OA\Parameter` component definitions (#201)

Closes #201

The harvester merges operation-level `@OA\Response`s and registers authored `@OA\Schema` models, but
does **not** harvest reusable **response/parameter component** definitions. An operation pointing at
one (`@OA\Response(ref="#/components/responses/NotFound")`) is currently **skipped + logged** in
`HarvestAuthoredAnnotationsStage::mergeAuthoredOperation()`, so the ref would dangle. This harvests
authored `@OA\Response(response="X")` / `@OA\Parameter(parameter="Y")` component defs, registers them,
and resolves operation refs against them.

### Feasibility probe (done first, per the #197 lesson) — holds, with one asymmetry ✅

Ran the real `OpenApi\Generator::generate(validate: false)` path (exactly what
`AuthoredAnnotationScanner` uses) over an autoloadable fixture with a `@OA\Response(response="NotFound")`,
`@OA\Parameter(parameter="PageParam")`, and a transitively-referenced `@OA\Schema(schema="ErrorBody")`:

- `$document->components->responses` and `->parameters` are **both populated**, each carrying its
  component name (`->response` / `->parameter`) and declaring class (`_context->class`). The
  transitive schema is in `->schemas`. So the scanner *can* index them the same way it indexes
  schemas. **(a) confirmed.**
- The operation's `@OA\Response(ref=…)` / `@OA\Parameter(ref=…)` are preserved verbatim, ready to
  resolve. **Resolution target confirmed.**

**The asymmetry (the scope flag the lead asked for):**
- **Responses — already plumbed.** `ComponentSchemaRegistry` has `registerNamedResponse()` /
  `allResponses()`, and `ComponentsStage` already writes `components->responses`. `ErrorResponseInferenceStage`
  already drives this path, so response components serialize cleanly (incl. on 5.8). **Minimal new code.**
- **Parameters — NOT plumbed.** No `registerNamedParameter`, no `allParameters`, and `ComponentsStage`
  does **not** write `components->parameters`. This is the "deeper plumbing" caveat: the parameter
  component path needs a new registry store + accessor **and** a `ComponentsStage` write. Bounded and
  mechanical (mirror the response path), but it is net-new infra in `Support/` (not plugin-local).

Verdict: proceed. Both halves are feasible; flagging that parameters carry a small `Support/`-side
plumbing addition (registry + stage), which is generation-infra, not SwaggerPhp-plugin-only.

### Tier

**Tier 0/1** — reflection + swagger-php annotation parse of already-authored blocks; no method-body
parsing, no dataflow. Same footing as the existing harvester.

### Generation-affecting → survey gate

This **changes the emitted document** (authored response/parameter components now appear;
previously-dropped operation refs now resolve). Per the pipeline this requires the **Layer A survey
gate (11 apps)** before merge. Most consumer apps don't author `@OA` component blocks, so the
expected delta is *additive and localized* to apps that do; the survey confirms no regression
elsewhere. **Coder/reviewer: run the survey gate before marking mergeable.**

### Design

**1. Scanner — index the component defs** (`AuthoredAnnotationScanner`). Mirror the existing
`schemasByName` index:
- `private array $responseComponentsByName = []` keyed by `$response->response`.
- `private array $parameterComponentsByName = []` keyed by `$parameter->parameter`.
- New `indexResponseComponents()` / `indexParameterComponents()` walking
  `$document->components->responses` / `->parameters` (filter to entries with a defined name and **no**
  `ref` — a ref entry is a usage, not a definition). Call them from `scan()`.
- Accessors: `responseComponentForName(string $name): ?OA\Response`,
  `parameterComponentForName(string $name): ?OA\Parameter`. `@internal`.

**2. Registry — add the parameter store** (`ComponentSchemaRegistry`, `Support/`). Mirror the response
store exactly:
- `private array $parameters = []`; `registerNamedParameter(string $key, OA\Parameter $parameter): void`
  (idempotent, first-wins, same as `registerNamedResponse`); `hasParameter(string $key): bool`;
  `allParameters(): list<OA\Parameter>`.

**3. ComponentsStage — flush parameters** (`Support/`). Add the `parameters` branch alongside
`schemas`/`responses` (`allParameters()` → `components->parameters`), guarding the empty case as the
others do.

**4. Harvester — register + resolve** (`HarvestAuthoredAnnotationsStage`):
- A new step (run from the stage's `apply`, alongside the existing schema/operation harvest) that
  registers every authored response/parameter component def into the registry, with **transitive
  schema pull-in**: reuse the existing `registerSchema()` recursion for any `$ref` a response's content
  schema / a parameter's schema points at (same `collectRefNames` walk already used for schemas).
- In `mergeAuthoredOperation()`, replace the current **skip-on-`@OA\Response(ref=…)`** branch: when the
  ref names a response component now present in the harvested pool, **register it (if not already) and
  keep the ref** (the flushed `components->responses` makes it resolvable); only **skip + log when the
  target is genuinely absent** from the pool. Do the analogous thing for authored operation
  `@OA\Parameter(ref=…)`.

**Conservative boundary:** only **named component definitions** are harvested (defined name, no own
`ref`). An operation ref to a name the pool doesn't contain stays skipped + logged (today's behaviour
for the genuinely-dangling case) — never emit a dangling `$ref`.

### Files

**Modify**
- `src/Plugins/SwaggerPhp/Support/AuthoredAnnotationScanner.php` — two indexes, two index methods, two
  accessors.
- `src/Support/Generator/ComponentSchemaRegistry.php` — parameter store + 3 methods (mirror responses).
- `src/Support/Generator/Stages/ComponentsStage.php` — flush `parameters`; update the class docblock
  ("schemas, responses, and parameters").
- `src/Plugins/SwaggerPhp/Stages/HarvestAuthoredAnnotationsStage.php` — register component defs +
  transitive schemas; resolve operation response/parameter refs against the pool instead of skipping.

**No new public/Contracts surface.** All touched classes are `@internal` generation infra.

**Feature — `tests/Feature/SwaggerPhpComponentHarvestTest.php`** (mirror `SwaggerPhpHarvesterTest`).
New fixtures under `tests/Fixtures/SwaggerPhp/` (or a dedicated subdir): a class declaring a
`@OA\Response(response="NotFound", @OA\JsonContent(ref="#/components/schemas/ErrorBody"))` component, a
`@OA\Parameter(parameter="PageParam", name="page", in="query", @OA\Schema(type="integer"))` component,
the referenced `@OA\Schema(schema="ErrorBody")`, and a controller whose operation `@OA\Response(ref=…)`
+ `@OA\Parameter(ref=…)` point at them.
- **lands under authored names:** generated `components->responses['NotFound']` and
  `components->parameters['PageParam']` exist with the authored content.
- **operation ref resolves (not dropped):** the operation keeps its `$ref` to the harvested component;
  the previously-emitted skip warning is **gone** for the present case.
- **transitive schema pull-in:** `components->schemas['ErrorBody']` is present because the response
  references it.
- **genuinely-absent target still skip+logged:** an operation `@OA\Response(ref="#/components/responses/Missing")`
  with no matching component → ref dropped + warning logged (assert via spy logger), no dangling ref in
  output.
- **OAS 3.1 serialize + validate:** generate to YAML/JSON and assert it validates (reuse the existing
  harvester test's validation helper). **Add/confirm a 5.8-safe assertion** — the response/parameter
  component shapes must serialize on swagger-php 5.8 (the byte-exact `snapshot` group is excluded from
  the 5.x job, but this serialize+validate assertion must pass there). Keep schemas as `OA\*` objects,
  not raw nested arrays, to avoid the 5.8 nested-array rejection.

**Unit**
- `tests/Unit/Plugins/SwaggerPhp/AuthoredAnnotationScannerTest.php` (extend): `responseComponentForName`
  / `parameterComponentForName` return the indexed defs; null for unknown / for a ref-only (usage)
  entry.
- `tests/Unit/Support/Generator/ComponentSchemaRegistryTest.php` (extend or create):
  `registerNamedParameter` is idempotent/first-wins; `allParameters` returns them; empty by default.
- `tests/Unit/Support/Generator/ComponentsStageTest.php` (if exists; else fold into the feature test):
  flushes `components->parameters` only when non-empty, coexisting with schemas/responses/securitySchemes.

### Docs / CHANGELOG

- `docs/` — observable behaviour change (harvester now picks up response/parameter components). Update
  the swagger-php harvester page (the page index is in `docs/README.md`; the harvester doc that
  currently states response components are merged but reusable component defs are not). Note the
  conservative skip-when-absent boundary.

- **Generation-affecting → survey gate (Layer A, 11 apps)** before merge. Primary human/process gate.
- **Parameter-component plumbing touches `Support/`** (`ComponentSchemaRegistry`, `ComponentsStage`) —
  not just the SwaggerPhp plugin. It's additive, mirrors the existing response path, and adds no
  Contracts/public surface, but call it out: a generic registry/stage gains a parameter store driven
  (today) only by the SwaggerPhp harvester. This is correct per the architecture (component registries
  are shared `Support/` infra, plugins contribute to them) and matches how responses already work.
- **No new config key, no stage-order change, no Contracts touch, no public-signature change.**
- **5.8 watch:** the new component shapes must serialize on the swagger-php 5.8 CI job — keep nested
  schemas as `OA\*` objects. Coder: verify against the 5.8 path, not just local 6.x.
- If the `src/` diff trends past ~150 LOC (the parameter plumbing could push it), stop and flag for
  Moritz's gate, per the #399 precedent.

---

## 🤖 [reviewer1] plan-review amendments (verified against `src/`)

**Approved.** Feasibility probe holds, the response/parameter asymmetry is real (verified: `registerNamedResponse`/`allResponses` + the `ComponentsStage` response flush exist; no parameter equivalents anywhere). Tier-0/1, surgical, `@internal`-only. Generation-affecting → the Layer A survey gate is the right process gate. Apply the one correctness note.

**[REQUIRED — correctness] Extend the `ComponentsStage` empty-guard to include parameters.** Today the guard is `if ($schemas === [] && $responses === []) return;` (ComponentsStage.php). If a run harvests **only** parameter components (no schemas, no responses), the unextended guard returns early and **silently drops the parameters**. The new branch must be `if ($schemas === [] && $responses === [] && $parameters === []) return;`, with `$parameters = $this->schemaRegistry->allParameters();` read up front and `if ($parameters !== []) { $components->parameters = $parameters; }` mirroring the others. (The plan says "guarding the empty case as the others do" — this is the concrete shape; a parameters-only fixture in the tests would catch a regression here.)

**Pressure-tests — verdicts:**
1. **Parameter plumbing — right minimal mirror; no new seam, no stage-order change.** `ComponentSchemaRegistry` is the correct home (it already carries `schemas` + `responses`; parameters are the third OAS reusable-component type — a separate registry would fragment the single `ComponentsStage` flush). Both touched classes are `@internal` Support infra; adding `registerNamedParameter`/`hasParameterKey`/`allParameters` (faithful copies of the response methods, idempotent first-wins) is **not** a Contracts/public-surface change. The flush is an additive branch in the **existing** `ComponentsStage` at its existing pipeline position — no new stage, no reorder. The registry is `#[Scoped]` (per-run), so the new store carries no cross-run leakage. **Correct per the shared-Support-infra architecture; matches how responses already work.**
2. **Transitive schema pull-in — correct and cycle-safe.** Reusing `registerSchema()` is sound: it `registerNamed()`s the schema **before** recursing into its `collectRefNames()` (HarvestAuthoredAnnotationsStage.php:310-312), so a cyclic graph terminates via the registry's own `schemaForKey()` early-return (the visited-set is the registry itself). No extra cycle guard needed for the response/parameter content schemas.
3. **Resolve-or-skip+log — sound.** Replaces the unconditional skip-on-`@OA\Response(ref=…)` (HarvestAuthoredAnnotationsStage.php:125-135) with a pool lookup: keep the ref only when the named component is present (now resolvable via the flushed `components->responses`/`->parameters`), skip+log when genuinely absent. **Never a dangling `$ref`** — matches the issue's conservative boundary. Assert both: present-target ref survives with **no** warning; absent-target ref dropped **with** a logged warning (spy logger).
4. **5.8 serialization — correctly mitigated.** Keeping nested schemas as `OA\*` objects (not raw nested arrays) is the right call given 5.8's nested-array rejection. Keep the serialize+validate assertion in the non-snapshot path so it runs on the 5.8 CI job, not just local 6.x.

**Controversy:** the `Support/`-side parameter plumbing is the one thing to weigh, but it is additive, mirrors the shipped response path exactly, and adds no Contracts/public/stage-order surface — it does **not** trip a blocking registry gate; it is the established pattern. The genuine gate here is **generation-affecting → Layer A survey (11 apps)**, which the plan already flags. Docs: update the harvester page (observable change) + CHANGELOG. No `docs/linting.md` (not a lint rule).


Radiergummi:

**Feasibility probe first (the #197 lesson): HOLDS, with one asymmetry to know about.** I ran the real `OpenApi\Generator::generate(validate:false)` path (what the scanner uses) over an autoloadable fixture:
- `$document->components->responses` **and** `->parameters` are both populated from authored `@OA\Response(response=…)` / `@OA\Parameter(parameter=…)`, each carrying its component name + declaring class, plus the transitive `@OA\Schema`. So the scanner can index them like schemas. **(a) confirmed.**
- **Asymmetry (b):** the **response** component path is already fully plumbed — `ComponentSchemaRegistry::registerNamedResponse`/`allResponses` + `ComponentsStage` flush + `ErrorResponseInferenceStage` already drives it (so responses serialize cleanly incl. 5.8). **Parameters are NOT plumbed** — no `registerNamedParameter`, no `allParameters`, `ComponentsStage` does not write `components->parameters`. So this PR adds a small parameter store to the shared registry + a `ComponentsStage` write (mirrors responses exactly). That's net-new `Support/`-side infra, not plugin-local — flagging per your instruction.

Plan highlights:
- Scanner: two name-keyed indexes (`responseComponentsByName`/`parameterComponentsByName`) + accessors; harvest only *definitions* (named, no own `ref`).
- Harvester: register defs with transitive schema pull-in (reuse existing `registerSchema` recursion); in `mergeAuthoredOperation`, resolve `@OA\Response(ref=…)`/`@OA\Parameter(ref=…)` against the pool instead of skipping — **only skip+log when genuinely absent** (never a dangling ref).
- **Generation-affecting → Layer A survey gate (11 apps) before merge** — noted in the body for coder/reviewer.
- **5.8 watch:** keep nested component schemas as `OA\*` objects (not raw arrays) so they serialize on the 5.8 job.
- No Contracts/config/stage-order change. If the parameter plumbing pushes src past ~150 LOC, stop-and-flag for Moritz's gate.

Radiergummi:
🤖 **[lead]** feasibility holds (responses+parameters both populate from authored component defs) → plan-review. Key asymmetry: responses already plumbed (registerNamedResponse/allResponses + ComponentsStage flush), but PARAMETERS are not — plan adds a parameter store to shared ComponentSchemaRegistry + a ComponentsStage write, mirroring responses (net-new Support/-side gen infra, additive). Reviewer watch items: (1) is the parameter-store addition the right minimal mirror of responses, no new Contracts/public seam, no stage-order change, ComponentSchemaRegistry the right home? (2) transitive schema pull-in via registerSchema recursion correct? (3) resolve-or-skip+log never leaves a dangling ref; (4) survey gate (11 apps) noted as generation-affecting; (5) 5.8 serialization — nested component schemas stay OA\* objects not raw arrays.

Radiergummi:
🤖 **[reviewer1]** plan-review: **approved with one correctness amendment** (folded into the PR body).

Verified the asymmetry against `src/`: `registerNamedResponse`/`hasResponseKey`/`allResponses` + the `ComponentsStage` response flush exist; no parameter equivalents anywhere. The plan is accurate. Tier-0/1, surgical, `@internal`-only. Generation-affecting → Layer A survey gate is the right process gate.

**[REQUIRED — correctness] Extend the `ComponentsStage` empty-guard to include parameters.** Today it is `if ($schemas === [] && $responses === []) return;`. A run that harvests **only** parameter components (no schemas/responses) would hit that early-return and **silently drop the parameters**. Make it `... && $parameters === []`. Add a parameters-only fixture so a regression here is caught.

**Pressure-tests:**
1. **Plumbing — right minimal mirror, no new seam, no stage-order change.** `ComponentSchemaRegistry` is the correct home (already holds schemas+responses; parameters are the third reusable-component type — a separate registry fragments the flush). `@internal` Support infra; adding the three parameter methods (faithful copies of the idempotent first-wins response methods) is not a Contracts/public change. Flush is an additive branch in the existing stage at its existing position. Registry is `#[Scoped]` (per-run) → no cross-run leakage. Correct per the shared-Support-infra architecture; matches responses.
2. **Transitive pull-in — correct + cycle-safe.** `registerSchema()` registers before recursing (HarvestAuthoredAnnotationsStage.php:310-312), so a cyclic graph terminates via the registry’s own `schemaForKey()` dedup. No extra guard needed.
3. **Resolve-or-skip+log — sound.** Replaces the unconditional skip (line 125-135) with a pool lookup: keep the ref when present (resolvable via the flush), skip+log when absent. Never a dangling `$ref`. Assert both: present → ref survives, no warning; absent → ref dropped, warning logged (spy).
4. **5.8 serialization — correctly mitigated.** Keep nested schemas as `OA\*` objects; keep the serialize+validate assertion in the non-snapshot path so it runs on the 5.8 job.

**Controversy:** the `Support/` parameter plumbing is additive, mirrors the shipped response path, adds no Contracts/public/stage-order surface — does not trip a blocking registry gate; it is the established pattern. The real gate is generation-affecting → Layer A survey (11 apps), already flagged. Docs: harvester page + CHANGELOG; no `docs/linting.md` (not a lint rule). Cleared to coding; run the survey before mergeable.

Radiergummi:
🤖 **[lead]** plan-review approved (amendment in PR body) → coding. REQUIRED correctness fix: extend ComponentsStage empty-guard to '$schemas===[] && $responses===[] && $parameters===[]' (a parameters-only harvest must NOT early-return + silently drop them) + a parameters-only fixture. Other asks: assert resolve present→survives/no-warning + absent→dropped/warning-logged (spy); keep nested schemas as OA\* objects + serialize/validate in the non-snapshot path (5.8 job). Generation-affecting → SURVEY GATE (Layer A, 11 apps) runs after code-review, before mergeable.

Radiergummi:
🤖 **[coder]** implemented; local gates green (Pint passed, PHPStan `[OK] No errors`, full suite 2961 passed); CI green on `ad06d347` — **including the swagger-php 5.8 job** (the response/parameter component shapes serialize + `validate()` there, in the non-snapshot path). Ready for review.

**What shipped (all four reviewer-confirmed pieces + the required correctness fix):**
- **Scanner:** `responseComponentsByName` / `parameterComponentsByName` indexes + `indexResponseComponents()` / `indexParameterComponents()` (walking `components->responses`/`->parameters`, skipping `ref`-only usage entries) + `responseComponentForName()` / `parameterComponentForName()` accessors.
- **Registry (Support):** `registerNamedParameter` / `hasParameterKey` / `allParameters` — faithful idempotent first-wins copies of the response methods.
- **ComponentsStage (Support):** flushes `components->parameters`; **empty-guard extended to `&& $parameters === []`** (the required correctness fix) — proven by a parameters-only unit test.
- **Harvester:** resolves operation `@OA\Response(ref=…)` / `@OA\Parameter(ref=…)` against the harvested pool — registers the named component + its transitive schemas (reusing `registerSchema()` recursion) and **keeps the ref**; skip+logs only when the component is genuinely absent. Never a dangling `$ref`.

**Two gotchas worth recording (the second cost me the most):**
1. **`ComponentReference::name()` is schemas-only** (returns null for `responses/`/`parameters/`). Used `ComponentReference::parse()` gated on the expected `ComponentType` instead — added a small `componentRefName($ref, $type)` helper.
2. **`OpenApi\Generator::UNDEFINED` is a non-empty *string*** (`is_string(UNDEFINED) === true`). My first cut gated the ref branch on `is_string($response->ref)` alone, which treated *every* authored non-ref response as a ref-response and dropped it — 8 existing tests went red. Fixed by pairing `is_defined($x->ref) && is_string($x->ref)` (`is_defined` for the sentinel, `is_string` for PHPStan's `argument.type` narrowing).

**Fixtures isolation:** the new fixtures live in a dedicated `tests/Fixtures/SwaggerPhpComponents/` dir (not the shared `SwaggerPhp/`), so the existing harvester/snapshot tests' scan set is unchanged.

**LOC flag (per the #399 precedent):** `src/` diff is +218/-12, over the ~150 soft gate. The overage is the parameter plumbing the lead pre-flagged as the risk: scanner indexing (+72, two indexes/methods/accessors + mandated docblocks), the harvester resolve/register logic (+110), the registry parameter store (+29, three mirror methods), the stage flush (+7). All additive, no Contracts/public/stage-order surface, mirroring the shipped response path. Surfacing for Moritz's gate.

**Generation-affecting → survey gate:** this changes the emitted document (authored response/parameter components now appear; previously-dropped operation refs now resolve). Per the plan this needs the **Layer A survey (11 apps)** before mergeable — flagging so the surveyor runs it once code-review approves.

Radiergummi:
🤖 **[lead]** CI green incl 5.8 (pushed ad06d347) → code-review (two independent reviewers). Reviewer focus: (1) the new parameter store on ComponentSchemaRegistry + ComponentsStage write faithfully mirror the shipped response path (idempotent first-wins, #[Scoped] per-run, no cross-run leak), and the Support/-side boundary matches how responses already store OA\* objects; (2) empty-guard fix + parameters-only regression test; (3) resolve-or-skip+log never leaves a dangling $ref — and VERIFY the is_defined()+is_string() ref-detection fix (Generator::UNDEFINED is a non-empty string; coder hit 8 red tests) doesn't misclassify authored non-ref responses; (4) transitive pull-in cycle-safe; (5) the serialize+validate assertion runs in the NON-snapshot path (5.8 job). NOTE: src +218/-12, more over the ~150 gate than #198 — I'll make the merge-gate call after your verdicts + the survey; do NOT push inlining to game LOC. Generation-affecting → survey (Layer A, 11 apps) runs after you approve.

Radiergummi:
🤖 **[reviewer1]** review round 1 — 2 findings (both test-suite gaps; the shipped code is correct):

I verified the full diff + changed source/tests in the worktree. Local gates green: **Pint passed**, **PHPStan `[OK] No errors`**, the four changed-area test files **42 passed**, the full SwaggerPhp/harvester set **158 passed**. All lead watch-items check out: the parameter store on `ComponentSchemaRegistry` is a faithful idempotent first-wins mirror of the response store (`registerNamedParameter`/`hasParameterKey`/`allParameters`), `#[Scoped]` per-run so no cross-run leak; the empty-guard fix `$schemas===[] && $responses===[] && $parameters===[]` is in place **with** a parameters-only regression test (`ComponentsStageTest`); resolve-or-skip+log never leaves a dangling `$ref` (present→ref kept, absent→dropped+logged); the transitive schema pull-in reuses `registerSchema()` which registers-before-recursing, so it terminates on cycles via the registry’s own `schemaForKey()` dedup; the serialize+`validate()` assertion is in the **non-snapshot** path (runs on the 5.8 job). The `is_defined($x->ref) && is_string($x->ref)` ref-detection does **not** misclassify authored non-ref responses — I confirmed the 158 existing harvester tests (which author plenty of non-ref responses) stay green, and a non-ref `->ref` is `Generator::UNDEFINED` so `is_defined()` short-circuits false. Support/Contracts boundary is intact (registry stores plain `OA\*` objects; no plugin/third-party leak). CHANGELOG + `docs/plugins.md` are accurate. No Contracts/config/stage-order change.

The two gaps are the **parameter-side negative paths** — the response side has both, the parameter side has neither (this is the codecov-flagged ~8 uncovered lines, and I confirmed each path is reachable via a standalone probe):

1. **`src/Plugins/SwaggerPhp/Stages/HarvestAuthoredAnnotationsStage.php:221-227` (`registerParameterComponentRef` absent branch) — untested skip+log.** The feature test `drops and logs an operation response $ref whose target component is absent` covers the **response** case only; the symmetric **parameter** case (an operation `#[OA\Parameter(ref: "#/components/parameters/Missing")]` whose target is not in the harvested pool) has no test. I verified `parameterComponentForName("Missing")` returns null, so this branch fires in practice. **Add:** a fixture op with `#[OA\Parameter(ref: "#/components/parameters/Missing")]` → assert the parameter `$ref` is dropped from `operation->parameters` **and** the warning `SwaggerPhp harvester: authored parameter $ref "#/components/parameters/Missing" targets an unknown parameter component; skipping.` is logged (spy logger).

2. **`src/Plugins/SwaggerPhp/Support/AuthoredAnnotationScanner.php:179` & `:196` — the `!is_defined(...->ref)` ref-only filter is never exercised (minor).** The plan named a unit assertion that a ref-only **usage** entry in `components.responses`/`.parameters` is **not** indexed as a definition (`responseComponentForName`/`parameterComponentForName` → null for it). I probed the current fixtures: their `components.responses`/`.parameters` contain only definition entries (no ref-only usage), so the filter branch is dead in the suite — a regression that dropped the guard would go uncaught. Lower severity (reachability of a class-level `#[OA\Response(response: "X", ref: "...")]` definition+ref entry in practice is unclear), but the plan called for it. **Add:** a fixture authoring a ref-only component entry and assert it is not returned as a definition.

Neither blocks correctness — the behavior is proven right by probe — but per the reviewer test-gap duty these are concrete missing tests on the changed code’s degrade paths. Recommend the coder add at least #1 before merge; #2 is nice-to-have.

Separately noted (not a finding): `mergeAuthoredParameterRefs` appends resolved parameter refs to `operation->parameters` with no (name,in) dedup, so an inferred `page` query param + an authored `PageParam` ref for the same field would emit both. Narrow authored-vs-inferred-overlap edge, not a regression; flagging for awareness only.

Radiergummi:
🤖 **[reviewer2]** review round 1 — 3 findings (2 test gaps, 1 minor robustness note; no correctness bug on any supported shape):

I reviewed independently of reviewer1. Gates pass in the worktree: `pint --test` clean, PHPStan `[OK] No errors`, the new feature/unit tests + the existing `SwaggerPhpHarvesterTest` all green. I verified the load-bearing claims directly:

- **Ref-detection does NOT misclassify authored non-ref responses/parameters.** `is_defined()` routes through `Generator::isDefault()`, so a non-ref response/parameter (`ref === UNDEFINED`) returns `is_defined === false` and never enters the ref branch; the paired `is_string` is only PHPStan narrowing. Correct.
- **Transitive pull-in works on the *generated* document.** `OA\JsonContent` extends `OA\Schema` (not `MediaType`), but after `Generator::generate()` the response content is normalized to a real `OA\MediaType` whose `->schema` carries the ref — so `walkRefs(OA\Response → content → MediaType → schema)` resolves `ErrorBody`. Verified by probe + the passing test.
- **No dangling `$ref`.** `componentRefName()` type-gates via `ComponentReference::parse()`, so a mismatched-type ref (`responses/` in a parameter slot), a malformed ref, or an empty name all return null → component null → logged + dropped. Robust.
- **5.8 serialize+validate runs in the non-snapshot path** (`uses()->group(openapi)`, not `snapshot`) → exercised on the 5.8 CI job. Good.

**Findings:**

1. **Test gap — absent *parameter* `$ref` branch is untested** (`src/Plugins/SwaggerPhp/Stages/HarvestAuthoredAnnotationsStage.php:221-227`). The response side has the `gadgets`/`Missing` test asserting drop + warning; the symmetric parameter null-branch (`registerParameterComponentRef` → unknown component → `warning(...unknown parameter component...)` → drop) has **no** test. Add an operation fixture with `@OA\Parameter(ref: #/components/parameters/Missing)` and assert: the ref is absent from `operation->parameters`, and the message `SwaggerPhp harvester: authored parameter $ref "#/components/parameters/Missing" targets an unknown parameter component; skipping.` is logged (spy logger). This is one of the codecov-flagged uncovered lines.

2. **Test gap — *parameter* transitive schema pull-in is untested** (`HarvestAuthoredAnnotationsStage.php:230-238`). The `PageParam` fixture uses a plain `type: integer` schema with no `$ref`, so the `collectRefNames($component->schema)` loop never executes against a real ref. Add a parameter component whose `schema` is `new OA\Schema(ref: #/components/schemas/SomeBody)` (or contains a `$ref`) and assert `SomeBody` lands in `components.schemas`. (The response transitive path *is* covered via `ErrorBody`; the parameter one is not.)

3. **Minor robustness (non-blocking) — appended ref-parameters are not deduped against inferred params** (`HarvestAuthoredAnnotationsStage.php:175-178`). `mergeAuthoredParameterRefs` spreads resolved ref-parameters onto `operation->parameters` with no `(name, in)` dedup, and this stage runs *after* `OperationBuilder`’s dedup. If an inferred query param and a refd `@OA\Parameter` component target the same `(name, in)`, the operation emits two parameters (invalid OAS duplicate). Its only caught downstream by the `parameter.duplicate-name` lint rule, not prevented in generation. The `(name, in)` of a `$ref` parameter is opaque at merge time, so a full guard is hard; this mirrors the late-append response pattern, so Im flagging it as a known edge to document (a short `// why no dedup` comment) rather than block on. No test required.

Findings 1 and 2 are test-completeness on the parameter half of an otherwise well-tested change — the code paths are correct, just unexercised. Recommend the coder add the two missing parameter-side tests; finding 3 is optional. Once those land Im happy to approve.

Radiergummi:
🤖 **[lead]** review round 1: both reviewers converge — code is correct (all watch-items verified, gates green), but the **parameter half** of the change is under-tested → back to coder (round 1/3). Required: (1) test the absent-parameter-$ref skip+log branch (HarvestAuthoredAnnotationsStage.php:221-227) — op with @OA\Parameter(ref:'.../Missing') → assert ref dropped + 'unknown parameter component' warning logged (spy). (2) test parameter transitive schema pull-in (:230-238) — a param component whose schema carries a $ref → assert that schema lands in components.schemas. Nice-to-have: (3) test a ref-only usage entry is NOT indexed as a definition (AuthoredAnnotationScanner :179/:196). Non-blocking: (4) add a short '// why no dedup' comment on mergeAuthoredParameterRefs noting the no-(name,in) dedup edge. Test-only additions + one comment — NO production logic change. Re-review (round 2) after push.

Radiergummi:
🤖 **[coder]** addressed round 1; added 2 parameter-side feature tests + scanner skip-note + dedup comment; local gates green (test 2963 passed / pint passed / phpstan [OK] No errors); pushed `fba149d6`. Ready for re-review.

Findings closed:
- **#1 (REQUIRED) — absent parameter-`$ref` skip+log:** new feature test `drops and logs an operation parameter $ref whose target component is absent` (gadgets op now also carries `#[OA\Parameter(ref: "#/components/parameters/Missing")]`); asserts the ref is dropped from `operation->parameters` and the exact warning `…authored parameter $ref "#/components/parameters/Missing" targets an unknown parameter component; skipping.` is logged (spy). Red-verified by flipping the absent branch to `return true`.
- **#2 (REQUIRED) — parameter transitive schema pull-in:** added `BodyParam` (schema `new OA\Schema(ref: "#/components/schemas/SomeBody")`) + `SomeBodySchema` fixtures, ref'd from the widgets op; new test `pulls in a schema the harvested parameter references transitively through its schema` asserts `SomeBody` lands in `components.schemas`. Red-verified by disabling the `$component->schema` walk.
- **#4 (NON-BLOCKING) — dedup comment:** added a 3-line "why no (name, in) dedup" note on `mergeAuthoredParameterRefs` (opaque `$ref` identity at merge time → left to `parameter.duplicate-name` lint).

**#3 (NICE-TO-HAVE) skipped, documented why:** I probed an autoloadable class carrying a class-level ref-only `@OA\Response`/`@OA\Parameter` — swagger-php discards it (`Skipping unknown <Class>`), so no realistic fixture lands a ref-only *usage* entry in `components.responses`/`.parameters` for the scanner to index. The `!is_defined(...->ref)` filter stays defensive against an upstream change but is unreachable via normal generation; recorded as a one-line comment in `AuthoredAnnotationScannerTest`.

No production logic changed (src diff is the dedup comment only).

Radiergummi:
🤖 **[lead]** round 1 addressed (pushed fba149d6, test-only + 3-line dedup comment, gates green: 2963 passed / pint / phpstan OK) → re-review (round 2). Coder closed both REQUIRED findings (absent-param-$ref skip+log, param transitive pull-in — both red-verified); #3 skipped with sound rationale (swagger-php discards class-level ref-only @OA entries, so the defensive filter is unreachable via normal generation — documented in test); #4 dedup comment added. Reviewers: confirm the two new tests genuinely exercise their branches and approve if clean.

Radiergummi:
🤖 **[reviewer2]** review round 2: approved, no concerns.

Both round-1 test gaps are closed and I red-verified each new test independently in the worktree (not just trusting the green run):

- **Finding #1 (absent parameter `$ref` skip+log)** — `tests/Feature/SwaggerPhpComponentHarvestTest.php` now drives `@OA\Parameter(ref: #/components/parameters/Missing)` on `gadgets` and asserts the ref is dropped + the exact `...unknown parameter component; skipping.` warning is logged. I mutated the warning message in `registerParameterComponentRef` and the test went **red** → it genuinely exercises that branch.
- **Finding #2 (parameter transitive schema pull-in)** — the new `BodyParam` component schema carries `ref: #/components/schemas/SomeBody` (+ `SomeBodySchema` fixture) and the test asserts `SomeBody` lands in `components.schemas`. I gutted the `$component->schema` walk in `registerParameterComponentRef` and the test went **red** → the transitive pull-in is what makes it pass.
- **Finding #3 (dedup robustness, non-blocking)** — the 3-line comment on `mergeAuthoredParameterRefs` accurately states the `(name, in)` identity is opaque at merge time and the overlap is left to the `parameter.duplicate-name` lint rule. Exactly what I asked for.

The scanner-test comment honestly documents that the `!is_defined(...->ref)` defensive filter cannot be reached through normal generation (swagger-php discards a class-level ref-only `@OA\Response`/`@OA\Parameter`), so it is left as upstream-change insurance rather than chasing an unreachable test — fair call, not a hidden gap.

Gates green in the worktree: `pint --test` clean, PHPStan `[OK] No errors`, the component-harvest feature + scanner unit suites pass (20 passed locally; coder reports full 2963). The production change since round 1 is comment-only — no logic moved. Cleared from my side; remaining gate is the generation-affecting Layer A survey (11 apps).

Radiergummi:
🤖 **[reviewer1]** review round 2 — 1 finding (round-1 #1 & #2 resolved; new finding on round-1 #3’s skip rationale):

**Round-1 findings #1 and #2 are genuinely resolved — I red-verified both by mutation, not just by reading:**
- **#1 (absent parameter-$ref skip+log):** mutating `registerParameterComponentRef` to `return true` (neutralizing the absent branch) makes `it drops and logs an operation parameter $ref whose target component is absent` **fail** (the `Missing` ref leaks into `operation->parameters` and the expected warning is absent); the symmetric response-absent test stays green. The test exercises the intended branch.
- **#2 (parameter transitive schema pull-in):** replacing the `if ($component->schema instanceof OA\Schema)` walk-guard with `if (false)` makes `it pulls in a schema the harvested parameter references transitively through its schema` **fail** (`SomeBody` no longer in `components.schemas`). The new `BodyParam → SomeBody` fixtures correctly drive the parameter-side `collectRefNames` recursion.
- Worktree restored clean after both mutations; the changed-area suite is green (20 passed in the two touched files). The 3-line `(name, in)` dedup comment on `mergeAuthoredParameterRefs` is accurate and points at the right lint rule.

**[FINDING — round-1 #3 skip rationale is factually wrong; the branch is reachable and IS testable].** The new comment in `AuthoredAnnotationScannerTest.php` claims swagger-php discards a class carrying only a class-level ref-only `@OA\Response`/`@OA\Parameter` ("logs `Skipping unknown <Class>`"), so no realistic fixture lands a ref-only entry for the `!is_defined(...->ref)` filter to skip. **I empirically disproved this on the installed swagger-php 6.2.0.** Generating a fixture with class-level `#[OA\Response(response: "AliasNotFound", ref: "#/components/responses/NotFound")]` + `#[OA\Parameter(parameter: "AliasPage", ref: "#/components/parameters/PageParam")]` yields `components.responses["AliasNotFound"]` and `components.parameters["AliasPage"]` **with both `response`/`parameter` (a defined name) AND `ref` defined** — no `Skipping unknown` log, the entry is retained. So:
  1. The `!is_defined(...->ref)` filter in `indexResponseComponents`/`indexParameterComponents` (`AuthoredAnnotationScanner.php:179` & `:196`) is **not** dead/defensive-only — it is reachable via normal generation and actively prevents a ref-only **alias/usage** entry from being indexed and re-registered as a **definition** (which would emit `components.responses.AliasNotFound = {$ref: …}`, a usage masquerading as a definition). The branch has real protective value.
  2. It **is** testable, contrary to the recorded comment. **Add:** a fixture authoring a class-level ref-only `@OA\Response(response: "AliasNotFound", ref: …)` (and the parameter analogue) → assert `responseComponentForName("AliasNotFound")` / `parameterComponentForName("AliasPage")` return **null** (filtered as a usage, not indexed as a definition). This also closes round-1 #2 fully.
  3. The misleading comment should be corrected or replaced — as written it would invite a future maintainer to delete a filter that is actually load-bearing.

Severity: **not a correctness regression** — the production filter does the right thing today; the problem is an inaccurate "unreachable/defensive-only" justification guarding an untested but genuinely-reachable branch. I’d want the rationale corrected and (cheaply) the filter test added before merge, since the fixture is trivial. Your call on routing back vs. accepting with the comment fixed. (Note: I verified on 6.2.0; the 5.8 job may behave differently, but reachability on the supported 6.x line is enough to make the branch real.)

Radiergummi:
🤖 **[lead]** round 2: reviewer2 APPROVED (mutation-verified #1/#2); reviewer1 has 1 finding → back to coder (round 2/3). reviewer1 empirically disproved the #3 skip rationale on swagger-php 6.2.0: a class-level ref-only @OA\Response(response:'X', ref:'...') DOES land in components.responses with both name+ref (no 'Skipping unknown'), so the !is_defined(...->ref) filter is REACHABLE, load-bearing, and testable — the current 'unreachable/defensive-only' comment is misleading and would invite deleting a real filter. Coder: settle this EMPIRICALLY (build reviewer1's exact fixture: class-level ref-only @OA\Response + @OA\Parameter). If it indexes (reviewer1 right) → add the filter test (assert responseComponentForName/parameterComponentForName → null for the ref-only entry) + correct the comment. If swagger-php genuinely discards it (your earlier claim) → post the concrete repro proving it and reconcile why reviewer1 saw otherwise. No arguing without evidence.

Radiergummi:
🤖 **[coder]** addressed round 2; reviewer1 was right — settled #3 empirically and corrected my round-1 claim; pushed `5c0202d1`. Local gates green (test 2964 passed / pint passed / phpstan [OK] No errors).

**Empirical settlement of finding #3.** I rebuilt reviewer1's exact fixture and ran the real `OpenApi\Generator::generate(validate: false)` path (the same path `AuthoredAnnotationScanner` uses):

```
RESPONSES:
  response='Real' ref=(UNDEF)
  response='AliasNotFound' ref='#/components/responses/Real'
PARAMETERS:
  parameter='RealParam' ref=(UNDEF)
  parameter='AliasParam' ref='#/components/parameters/RealParam'
```

A class-level ref-only `@OA\Response`/`@OA\Parameter` (name + ref) **does** land in `components.responses`/`.parameters` with both name and ref — no "Skipping unknown". reviewer1 is correct; my round-1 conclusion was wrong.

**Why my round-1 probe differed (reconciled).** My round-1 probe added a fresh `.php` file to the fixtures dir and relied on composer autoload; the class was **not loadable at swagger-php scan time** in that process, so the scanner logged `Skipping unknown <Class>` and dropped the whole class. That was a test-harness artifact (class not reflectable at scan time), not swagger-php discarding ref-only entries. Once the class is genuinely loadable (explicit `require`, or the alias authored on an already-scanned fixture class), the entries index as shown above. So the `!is_defined(...->ref)` filter (`AuthoredAnnotationScanner.php:179` & `:196`) **is reachable via normal generation and is load-bearing** — it keeps a ref-only alias from being indexed as a definition.

**What I did:**
- Added `AliasNotFound` (ref-only `@OA\Response`) and `AliasParam` (ref-only `@OA\Parameter`) to the existing `ComponentDefinitions` fixture (already proven to scan), each a `$ref` to a real sibling component.
- New unit test in `AuthoredAnnotationScannerTest`: `does not index a ref-only response/parameter usage entry as a definition` — asserts `responseComponentForName('AliasNotFound')` → null and `parameterComponentForName('AliasParam')` → null. **Red-verified** by removing both `!is_defined(...->ref)` guards (test fails — the aliases get indexed as definitions).
- **Corrected** the misleading round-1 comment: it no longer claims the filter is "unreachable/defensive-only"; it now states the filter is reachable and load-bearing (keeps ref-only usage entries out of the definition index).

Test-only change (fixture + unit test); no production logic touched. Ready for re-review.

Radiergummi:
🤖 **[lead]** round 2 finding resolved (pushed 5c0202d1, test-only +23/-6, no production source touched, gates green: 2964 passed / pint / phpstan OK). Coder confirmed reviewer1 was RIGHT — the round-1 'unreachable' claim was a harness artifact (fixture class not loadable at swagger-php scan time → whole class skipped); with a loadable class the ref-only @OA aliases DO index, so the !is_defined filter is reachable + load-bearing. Coder added the filter test (red-verified by removing both guards) + corrected the comment. → re-review round 3. reviewer1: confirm your finding is closed. reviewer2: the delta since your round-2 approval is purely additive test + comment correction (no production change) — confirm your approval still stands.

Radiergummi:
🤖 **[reviewer2]** approval stands over the test-only delta in 5c0202d1.

Radiergummi:
🤖 **[reviewer1]** review round 3: approved, no concerns.

The round-2 finding is fully resolved (commit `5c0202d1`, test-only, no production source touched — so the production logic I cleared in round 2 stands).

**Verified the new filter test genuinely exercises the guard (red-verified by mutation, not just by reading):** removing the `!is_defined(...->ref)` guard from **both** `indexResponseComponents` and `indexParameterComponents` makes `it does not index a ref-only response/parameter usage entry as a definition` **fail** — `responseComponentForName("AliasNotFound")` / `parameterComponentForName("AliasParam")` stop returning null because the ref-only alias entries get wrongly indexed as definitions. With the guards in place the test passes. The `AliasNotFound` (ref-only `@OA\Response`) + `AliasParam` (ref-only `@OA\Parameter`) fixtures correctly reproduce what I observed empirically: swagger-php lands a ref-only entry with **both** a name and a `ref`, and the filter must keep it out of the definition index. Worktree restored clean after the mutation.

**The corrected comment is now accurate.** It states swagger-php lands a ref-only `@OA\Response`/`@OA\Parameter` (name + ref, an alias pointing at another component) in `components.responses`/`.parameters`, and the filter must keep those usage entries out of the definition index "or a ref would resolve to the alias." That matches reality and no longer invites a maintainer to delete a load-bearing filter. The coder also confirmed the root cause of its earlier "unreachable" claim was a test-harness artifact (the probe class wasn’t autoloadable at scan time → "Skipping unknown"), which squares with my round-2 result.

**Gates green in the worktree:** Pint passed, PHPStan `[OK] No errors`, changed-area suite 45 passed (95 assertions). Across all three rounds: parameter store faithfully mirrors the response path (idempotent first-wins, `#[Scoped]` per-run), empty-guard fix + parameters-only test, resolve-or-skip+log never dangles (response **and** parameter sides now both tested, red-verified), transitive pull-in cycle-safe (both sides tested, red-verified), ref-detection does not misclassify non-ref responses, the ref-only filter is tested + load-bearing, serialize+validate runs in the non-snapshot (5.8) path, Support/Contracts boundary intact, CHANGELOG + docs accurate.

No further code-review concerns. The generation-affecting Layer A survey gate (11 apps) is the remaining pre-merge step per the plan.

Radiergummi:
🤖 **[lead]** TWO independent approvals at final SHA 5c0202d1 (reviewer1 round-3 approved + mutation-verified the filter test; reviewer2 approval stands over the test-only delta), CI 11/11 green incl swagger-php 5.8 → survey gate (Layer A, 11 apps). Generation-affecting: the PR changes emitted output for apps authoring @OA component blocks; corpus apps mostly don't, so expected delta is additive/zero — confirming no regression elsewhere.

Radiergummi:

Baseline freshly regenerated on current `main` (`02a13147`); candidate generated from the PR branch (`5c0202d1`). Every deterministic metric is identical baseline↔candidate across all 11 corpus apps (paths / operations / responseSchemas / documentedResponses / requestBodies / componentSchemas), all generate-exits 0, no new crashes. As expected for this change: the corpus apps don't author `@OA\Response`/`@OA\Parameter` component blocks, so the harvester addition is a no-op for them — the delta is additive and localized to apps that *do* author such blocks. (Recorded by the lead; the surveyor completed the run but went idle before posting.)

| app | paths | operations | responseSchemas | componentSchemas |
|-----|-------|------------|-----------------|------------------|
| AdvisingApp | 523=523 | 532=532 | 24=24 | 61=61 |
| AureusERP | 111=111 | 160=160 | 66=66 | 36=36 |
| Bagisto | 406=406 | 491=491 | 9=9 | 117=117 |
| BookStack | 245=245 | 335=335 | 3=3 | 52=52 |
| Coolify | 304=304 | 375=375 | 16=16 | 3=3 |
| InvoiceNinja | 529=529 | 694=694 | 67=67 | 375=375 |
| Koel | 181=181 | 294=294 | 42=42 | 106=106 |
| Lychee | 213=213 | 273=273 | 82=82 | 280=280 |
| Pelican | 215=215 | 260=260 | 0=0 | 104=104 |
| SpeedtestTracker | 37=37 | 42=42 | 1=1 | 1=1 |
| Vito | 329=329 | 456=456 | 114=114 | 87=87 |

Radiergummi:
🤖 **[lead]** ALL gates green → escalating to Moritz for the merge call (per the LOC controversy rule), **not auto-merged**.

**Why human-gated:** the `src/` production diff is **+221/-10 = 233 changed lines**, over the ~150 auto-merge threshold. That is the *only* controversy trigger — it does **not** touch `Contracts/`, change stage order, add a config key, or alter/remove a public signature (the new `ComponentSchemaRegistry`/`ComponentsStage` methods are `@internal` Support infra, faithful mirrors of the shipped response path, both reviewers confirmed). The overage is the parameter-component plumbing flagged as the risk from the start (#399 precedent).

**Evidence it's ready to merge:**
- **Two independent reviewer approvals** at the final SHA `5c0202d1` (reviewer1 round-3 approved + mutation-verified every test; reviewer2 approval stands over the test-only delta). No open concerns; the one substantive round-2 finding (a wrong 'unreachable' rationale) was empirically settled — reviewer1 was right, fix landed + corrected comment.
- **CI 11/11 green** incl. the swagger-php 5.8 job (component shapes serialize + validate in the non-snapshot path).
- **Survey Layer A pass** — 0 regressions across 11 apps (baseline freshly regenerated on current main 02a13147).
- Diff did not deviate from the agreed plan.

PR left **ready + unmerged**, worktree in place at `.claude/worktrees/feat-201-swaggerphp-harvester-harvest-oa-response`. Moritz requested as reviewer. Recommend: merge as-is (the LOC is inherent to the parameter plumbing, no inlining was done to game the count, per your round-1 instruction).

Radiergummi:
🤖 **[lead]** merged by Moritz's decision. Fully green: two independent reviewer approvals (3-round adversarial loop, one real finding empirically settled), CI 11/11 incl. swagger-php 5.8, survey Layer A 0 regressions across 11 apps. The +233-line src diff (inherent parameter-component plumbing, no LOC-gaming) was the only thing holding auto-merge; Moritz cleared it. Closes #201.

PR #404, opened 2026-06-19
feat(#202): migration.oa-redundant-component-with-inference — remove redundant @OA\Response / @OA\Parameter component definitions

### LINKED ISSUE #202: Migration: remove redundant `@OA\Response` / `@OA\Parameter` component definitions
Parent: #122. Sibling of #196/#197/#198.

## Summary

Extend the `migration.*` family to remove redundant **reusable** `@OA\Response` / `@OA\Parameter` **component definitions** (`#/components/responses/*`, `#/components/parameters/*`) once inference reproduces an equivalent — the component-definition analog of the operation-level removal in #196.

## Blocked by harvesting

The redundancy oracle compares the authored contribution against the inference-only control document. Response/parameter component definitions are not harvested yet, so there is nothing to compare and no way to verify the same-API-surface invariant after removal. This must wait until the harvester learns to read and register them (the linked harvester issue).

## Approach (once unblocked)

Reuse the #122/#185 infrastructure: `InferenceControlDocument`, `SchemaEquivalence::subsumes()`, the dangling-`$ref` guard. Compare each authored response/parameter component against inference's equivalent; fire only on subsumption; `--fix` removes the `@OA\Response` / `@OA\Parameter` component block (docblock or attribute). Never remove one another surviving annotation still `$ref`s.

## Acceptance criteria

- Redundant authored response/parameter component definitions flagged + removed under `migration.*`.
- Golden fixtures + same-API-surface feature flavor; `docs/linting.md` row + `CHANGELOG.md` entry.

Radiergummi:
## Plan — #202: remove redundant `@OA\Response` / `@OA\Parameter` **component definitions**

Closes #202. Component-definition analog of the operation-level removal (#196,
`migration.oa-redundant-operation-with-inference`). Now unblocked: PR #402 (#201) merged to
`main` (HEAD `e34b9fc6`), so the harvester reads and registers reusable
`#/components/responses/*` / `#/components/parameters/*` definitions and resolves operation
`$ref`s to them.

### Tier — 1, and why

Reflection + a single bounded swagger-php scan (no method-body parsing, no dataflow). The
`AuthoredAnnotationScanner` already harvests the component pool via one `Generator` pass; this
rule reads from that index and from the inference-only `InferenceView`. **No Tier-2 dataflow is
needed** — the redundancy verdict is provenance-based (authored artifact vs. inference-only
control), identical in spirit to the shipped schema/operation rules.

### The non-obvious design decision (resolved here, not punted)

The issue's one-liner — *"compare each authored component against inference's equivalent"* —
hides a real subtlety I verified against `src/`:

**Vanilla Laravel inference never emits reusable `#/components/responses/*` or
`#/components/parameters/*` component definitions.** The Core plugin inlines responses and
parameters into operations. So there is **no per-name inference counterpart** to compare an
authored response/parameter *component* against — `InferenceView` exposes only `schemaForClass`,
`schemaForName`, `operationForRoute` (verified: `src/Lint/InferenceView.php`). A naive
"`inference->responseComponentForName($name)`" would always be null → the rule could never fire,
or (worse) fire blindly.

**The correct oracle is usage-site, threaded through the `$ref`** — exactly how the shipped
`OperationSubsumption` already reasons about responses (`src/Plugins/SwaggerPhp/Lint/Support/OperationSubsumption.php:99-112`):

> An authored response/parameter **component** is redundant iff, for **every** operation that
> `$ref`s it, the inference-only operation for that route already produces a per-status response
> (resp. a parameter) that **subsumes** what the component contributes — *and* the component is
> referenced by at least one such operation (an unreferenced/orphan component is not this rule's
> business; `component.orphaned` already covers dead components).

This keeps the same-API-surface invariant honest: removing the component is safe only because the
operations that used it collapse to an inference-equivalent shape. It reuses
`SchemaEquivalence::subsumes()` (per-content/per-schema, `$ref`-following — #198) with **no new
comparator semantics**; the new comparator is a thin component-level adapter over it.

This is resolvable within Tier-1 and consistent with the merged siblings, so **no escalation** —
but it is the load-bearing design choice the coder must implement, and reviewers should check it.

### Reuse points (named, verified in `src/`)

- `AuthoredAnnotationScanner` (`src/Plugins/SwaggerPhp/Support/AuthoredAnnotationScanner.php`) —
  already indexes `responseComponentsByName` / `parameterComponentsByName` with accessors
  `responseComponentForName()` / `parameterComponentForName()`. **Gap to close (small):** these
  are not indexed by declaring class, and `declaringClassOf()` accepts only `Schema|Operation`.
  The fixer needs the declaring class/member to locate the source. So extend the scanner to also
  record the declaring class/file per component name and widen `declaringClassOf()` to accept
  `Response|Parameter` (it reads `$annotation->_context`, which all `OA\AbstractAnnotation`s carry).
- `OaRedundancyEngine` (`…/Lint/Support/OaRedundancyEngine.php`) — the shared skeleton
  (`evaluate(authored, inferred, comparator, reflector, buildFinding, isEssential)`). Reuse
  verbatim. Pass the `isEssential` dangling-`$ref` guard.
- `SchemaEquivalence` (`…/Lint/Support/SchemaEquivalence.php`) — the subsumption primitive
  (`$ref`-by-name following, cycle-guarded; #198). Inject as today.
- `OperationSubsumption` (`…/Lint/Support/OperationSubsumption.php`) — reference for the
  per-status response subsumption + `anySubsumes` parameter matching; the new comparator borrows
  its `anySubsumesResponse` / `anySubsumes` shapes.
- Dangling-`$ref` guard — model on `AuthoredAnnotationScanner::isSchemaReferencedByOtherAuthored()`
  + `referencesRef()` (`AnnotationWalker::walk`). Add a sibling
  `isComponentReferencedByOtherAuthored($pointer, $excludingComponentName)` that scans operations
  and *other* component definitions for a `$ref` to this component's pointer
  (`ComponentReference::pointer(name, ComponentType::Responses|Parameters)`), so removing a
  component another **surviving** annotation still points at is refused.
- `RedundantOaAnnotationFixer` + `DocblockAnnotationRemover` / `OaAttributeRemover`
  (`…/Lint/Fix/`) — removes a docblock or `#[OA\*]` attribute given
  `CONTEXT_SOURCE_CLASS` (+ optional `CONTEXT_SOURCE_MEMBER`) + the `AuthoredAnnotationShape`.
  Component definitions live as a **class-level** docblock/attribute (member is null, as in the
  `SwaggerPhpComponents/ComponentDefinitions.php` fixture), so the existing class-targeting branch
  applies as-is. **Reuse the fixer unchanged** — the finding context (class + shape) is all it
  needs. (`removeBlocks` removes a single matching `@OA\*` block from a multi-annotation docblock;
  verify it removes only the targeted `@OA\Response`/`@OA\Parameter`, not sibling blocks — if it
  can only remove by class-node wholesale, that's a finding to raise, see Risks.)
- `AuthoredAnnotationShape` (`…/Lint/Support/AuthoredAnnotationShape.php`) — `detect(reflector)`
  + `FINDING_CONTEXT_KEY`. Reuse.
- Tree visitor seam — **there is no `ComponentResponseNode`/`ComponentParameterNode`** in the lint
  tree (verified: `src/Lint/Tree/` has only `ComponentSchemaNode`, `ResponseNode`, `ParameterNode`,
  `QueryParameterNode`; visitors mirror that). The rule therefore should **not** rely on a
  per-component node. Implement it as a `Rule` + `ApiRule` (`checkApi(ApiNode, LintContext)`) —
  the same document-level seam `DocumentAnnotationMigration` uses (verified:
  `implements Rule, ApiRule, NeedsInferenceDocument`; `checkApi(...)` at line 50). Walk the
  authored component pool from the scanner inside `checkApi`. Do **not** invent a new tree node.

### Files

**Create**
- `src/Plugins/SwaggerPhp/Lint/OaRedundantComponentWithInference.php` — the rule. Implements
  `Rule`, `ApiRule`, `FixableRule`, `NeedsInferenceDocument` (mirrors `DocumentAnnotationMigration`).
  `excludedStages()` returns
  `[HarvestAuthoredAnnotationsStage::class]`. One rule covering **both** responses and parameters
  (single `migration.*` ID, mirrors how the doc rule covers Info/Server/SecurityScheme/Tag
  together) — see Rule ID below. Severity `Severity::Improvable` (level-4), matching siblings.
- `src/Plugins/SwaggerPhp/Lint/Support/ComponentSubsumption.php` — thin comparator: given an
  authored `OA\Response`/`OA\Parameter` component and the set of inference-only operations that
  reference it (by route), returns true iff each such operation subsumes the component's
  contribution via `SchemaEquivalence`. (If reuse shows `OperationSubsumption` can be parameterised
  cheaply, fold into it instead — coder's call, but keep one comparator per concept.)

**Modify**
- `src/Plugins/SwaggerPhp/Support/AuthoredAnnotationScanner.php` — record declaring class per
  component name (new `responseComponentClasses` / `parameterComponentClasses` or a small VO);
  widen `declaringClassOf()` to `Schema|Operation|Response|Parameter`; add
  `isComponentReferencedByOtherAuthored()`. Also expose the authored component pools + the routes
  that reference each component (or let the rule walk `operationsByMethod` for `$ref`s — prefer
  reading what already exists).
- `src/Plugins/SwaggerPhp/SwaggerPhpPlugin.php` — `$registry->addRule(OaRedundantComponentWithInference::class);`
  in the migration-rule block.
- `docs/linting.md` — add the catalog row(s) between the `lint-rule-catalog` markers **and** a
  short paragraph in the `## Migration rules` section (mirror the existing
  `migration.oa-redundant-operation-with-inference` paragraph at ~line 294).
- `CHANGELOG.md` — `[Unreleased]` entry.

**Test fixtures** (under `tests/Fixtures/SwaggerPhp*`, reuse `SwaggerPhpComponents/`):
- A controller + component-definitions class where a `NotFound` response component / a `PageParam`
  parameter component are **reproduced by inference** (the route's inferred 404 / query param
  matches) → redundant.
- A component carrying something inference cannot reproduce (a human `description`, a constraint,
  an enum) → kept.
- A component still `$ref`d by another *surviving* authored annotation → kept (dangling guard).
- Both docblock and attribute declaration forms.

### Tests (TDD — failing first), with edge cases

Mirror `tests/Feature/SwaggerPhpOperationMigrationRuleTest.php` (the closest analog).

`tests/Feature/SwaggerPhpComponentMigrationRuleTest.php`:
1. **subsumed → flagged**: redundant `@OA\Response`/`@OA\Parameter` component appears in findings
   under `--only migration.*`.
2. **load-bearing → kept**: a component carrying a description/constraint inference cannot derive
   is **not** flagged.
3. **dangling guard → kept**: a component another surviving authored annotation still `$ref`s is
   **not** flagged.
4. **both forms**: parametrise (1)/(2) over docblock and attribute fixtures.
5. **off on a default-level run**; **disabled as a family by `--skip migration.*`** (copy the two
   existing guard tests verbatim against the new rule ID).
6. **same-API-surface (feature flavor)**: generate the harvested document and the inference-only
   control (`OpenApiGenerationOrchestrator::inferenceOnly('default', [HarvestAuthoredAnnotationsStage::class])`);
   assert the operation(s) that referenced the removed component resolve to the **same** response
   schema `$ref` / parameter shape on both sides (the operation rule's pattern at lines 123-136).
7. **fixer wiring**: rule exposes `RedundantOaAnnotationFixer` and a description mentioning
   "inference".

`tests/Unit/Plugins/SwaggerPhp/RedundantOaComponentFixerTest.php` (if the fixer path needs
component-specific coverage — likely yes for the "remove one `@OA\Response` block from a class
docblock carrying several" case): before/after golden assertion that **only** the targeted
component block is removed, siblings untouched. Edge cases: docblock form, attribute form, target
class not locatable → yields no fix (don't crash).

`tests/Unit/Plugins/SwaggerPhp/AuthoredAnnotationScannerTest.php` (extend): declaring-class is now
recorded for response/parameter components; `isComponentReferencedByOtherAuthored()` true/false
branches; ref-only *usage* entries (`AliasNotFound`/`AliasParam`) are not treated as definitions
(already a `#402` invariant — assert it still holds).

### Rule ID + docs catalog row

**One** new rule ID: `migration.oa-redundant-component-with-inference` (covers both response and
parameter components, as the doc-annotation rule covers four root kinds under one ID). Default
severity **Improvable (level 4)**, matching the migration family.

`docs/linting.md` catalog row (between the markers):

```
| `migration.oa-redundant-component-with-inference` | 4 | A hand-authored reusable @OA\Response / @OA\Parameter component definition the generator already reproduces inline via inference. (SwaggerPhp plugin; a `migration.*` rule — see Migration rules.) |
```

Plus a short paragraph in `## Migration rules` after the operation-level one.

`CHANGELOG.md [Unreleased]`:
> Added `migration.oa-redundant-component-with-inference`: flags (and `--fix` removes) reusable
> `@OA\Response` / `@OA\Parameter` component definitions inference already reproduces inline,
> with a dangling-`$ref` guard. Off by default; ships with the SwaggerPhp plugin.

### Controversy flags (for the lead's merge gate)

- **New lint rule → new default severity.** `Improvable` (level 4), consistent with the
  `migration.*` family; off on a default run, only fires under `--level 4` / `--only migration.*`.
- **Plugin-registered, off by default**, `class_exists`-guarded via the SwaggerPhp plugin (#122).
  **No `src/Support/` or `src/Contracts/` coupling** — the rule, comparator, scanner changes all
  live under `src/Plugins/SwaggerPhp/`. Low blast radius.
- Touches the shared `AuthoredAnnotationScanner` (widen `declaringClassOf`, add component
  declaring-class indexing + `isComponentReferencedByOtherAuthored`). In-plugin, but the scanner is
  used by the other migration rules + the harvest stage — regression surface is the existing
  SwaggerPhp test suite; keep changes additive.
- **No `Contracts` change, no stage-order change, no config key.** `InferenceView` is **not**
  extended (the oracle is usage-site, so no new inference accessor) — confirm during coding; if it
  turns out cleaner to add `InferenceView::operationForRoute`-based plumbing, that's still
  `src/Lint/` (internal, `@internal`-tagged), flag it if so.
- **Survey: N/A.** Lint rules don't change generation output; the same-API-surface test proves the
  generated document is unchanged by `--fix`. State explicitly in the merge note.

### Risk / open question for the coder (degrade gracefully)

`DocblockAnnotationRemover::removeBlocks` must remove **only** the targeted `@OA\Response`/
`@OA\Parameter` block when several component blocks share one class docblock (the
`ComponentDefinitions` fixture stacks many). Verify this in the unit test up front (TDD). If the
remover can only operate per class-node wholesale, that is a scoping finding to raise *before*
writing the fixer — it would mean the fixer needs a block-selector, which expands LOC. Plan
assumes per-block removal already works (the schema/operation fixers remove single blocks); confirm
first.

### LOC estimate (src/ only)

~180–230 LOC: rule ~120 (`OaRedundantComponentWithInference`), comparator ~60
(`ComponentSubsumption`), scanner additions ~40–60. Fixer reused unmodified. Tests + fixtures +
docs are additional. Squash-mergeable on green `tests` + `vendor/bin/pint --test` + PHPStan L8.

### Build sequence

1. Write failing Feature tests (subsumed/kept/dangling/both-forms/same-API-surface) + fixtures →
   verify red.
2. Extend scanner (declaring-class index, `declaringClassOf` widen,
   `isComponentReferencedByOtherAuthored`) → unit tests green.
3. `ComponentSubsumption` comparator over `SchemaEquivalence` → comparator unit test green.
4. `OaRedundantComponentWithInference` rule + plugin registration → Feature tests green.
5. Verify `--fix` removes only the targeted block (fixer reused) → fixer test green.
6. `docs/linting.md` row + paragraph; `CHANGELOG.md`.
7. `composer test` + `vendor/bin/pint --test` + `composer lint` all green.

---

## Plan-review amendments (reviewer1)

Verified every reuse claim against `main` (HEAD `e34b9fc6`, harvester #402 merged). The usage-site
oracle is **sound** and **Tier-1**. Three tightenings the coder MUST implement (each found by
reading the actual code, not the plan's prose):

1. **The comparator must resolve the operation's bare `$ref` before comparing — make this
   explicit.** In the *harvested* document an operation that uses a component keeps its response
   as a **bare `$ref`** (the harvester registers the component into `components->responses` but
   does **not** inline it — `HarvestAuthoredAnnotationsStage::mergeAuthoredOperation` line 128). The
   inference-only operation produces the same status response **inline**. So `ComponentSubsumption`
   cannot compare "authored operation vs inferred operation" the way `OperationSubsumption` does —
   in fact `OperationSubsumption::subsumes` line 108 *bails out* (`is_defined($response->ref) →
   return false`) precisely on a bare-`$ref` response, which is why the operation rule never fires
   on ref-using operations. The new comparator must, **per referencing operation**: resolve the
   bare `$ref` to the authored component body (via `scanner->responseComponentForName()` /
   `parameterComponentForName()`, already `$ref`-by-name, Tier-1) and call
   `SchemaEquivalence::subsumes(inferredResponseForThatStatus, resolvedComponentBody)`. State this
   resolution step in the comparator's contract; do not reuse `OperationSubsumption` verbatim for
   responses/parameters — its line-108 short-circuit makes it the *wrong* primitive here.

2. **"At least one referencing operation" is a HARD gate, not a footnote — and it is not built
   into the engine.** `OaRedundancyEngine::evaluate()` (line 50) only checks
   inferred-present + comparator + `isEssential`; it has **no** referencer-count check. If a
   component is `$ref`d by zero operations, "every referencing operation subsumes it" is
   **vacuously true** and the rule would fire on an orphan — wrong (that is `component.orphaned`'s
   job, and it risks removing a component still reachable transitively). The rule must explicitly
   skip a component with **zero** referencing operations before calling the engine (or fold the
   gate into the comparator returning false on an empty referencer set). Add a test:
   *orphan component (referenced by no operation) → NOT flagged*.

3. **The dangling-`$ref` guard is net-new code, not a copy of the schema guard.** The model the
   plan names, `isSchemaReferencedByOtherAuthored()` (scanner line 516), walks **only**
   `schemasByName` + `operationsByMethod` — it does **NOT** walk `responseComponentsByName` /
   `parameterComponentsByName`. The transitive component→component case the lead asked about is
   real and live in the fixture: `AliasNotFound` is a `ref`-only response component pointing at
   `NotFound`; `AliasParam` → `PageParam`. So the new
   `isComponentReferencedByOtherAuthored($pointer, $excludingComponentName)` MUST iterate the
   **response and parameter component pools** in addition to operations and schemas — `referencesRef`
   + `AnnotationWalker::walk` already traverse the full annotation tree, so the walk primitive is
   reusable, but the **set of nodes scanned must be widened**. Add a test:
   *removing `NotFound` while `AliasNotFound` survives → NOT flagged*.

Everything else verified accurate: the `ApiRule`/`checkApi` seam mirrors `DocumentAnnotationMigration`
(no new tree node — confirmed none exists); `InferenceView` has exactly three accessors and no
component accessor (claim holds); `RedundantOaAnnotationFixer` is reusable unchanged for class-level
(member=null) definitions; `DocblockAnnotationRemover::removeBlocks` already removes per-block
(located by `[start,end]` span), so the multi-block fixture is correctly fixable — **the plan's
TDD-first item on per-block removal is right to keep** (the `ComponentDefinitions` fixture stacks 5
blocks; assert siblings survive). Rule ID, Improvable/L4 severity, `docs/linting.md` catalog row +
`## Migration rules` paragraph (after line 294), and `CHANGELOG.md [Unreleased]` are all correctly
specified. Discriminator (definition vs ref-only usage) is `is_defined(name) && !is_defined(ref)` —
the `#402` invariant the plan's scanner test re-asserts.


Radiergummi:

Radiergummi:
🤖 **[lead]** plan posted (PR #404) → plan-review. Tier-1, ~180-230 src LOC, plugin-registered + off-by-default, no Support/Contracts/stage-order/config change, survey N/A (lint). **Central thing to verify:** the planner resolved that vanilla inference never emits reusable #/components/responses|parameters/* (it inlines), so the issue's literal 'component vs inference equivalent' oracle is impossible; the correct oracle is **usage-site, threaded through the $ref** — a component is redundant iff EVERY operation that $refs it collapses to an inference-equivalent response/param via SchemaEquivalence::subsumes(). Reviewer: confirm (a) that oracle is sound + truly Tier-1 (no dataflow) + consistent with the merged operation-level rule; (b) the dangling-$ref guard (isComponentReferencedByOtherAuthored) prevents removing a component another surviving annotation still references; (c) the AuthoredAnnotationScanner additions are minimal/additive; (d) the new rule ID + Improvable/L4 default severity + docs/linting.md row are specified; (e) the DocblockAnnotationRemover targeted-vs-wholesale risk the planner flagged is a real TDD item for the coder.

Radiergummi:
🤖 **[reviewer1]** plan-review: APPROVED (cleared to coding) with the plan amended in-body.

Verified every reuse claim against `main` (HEAD `e34b9fc6`, harvester #402 merged) — did not take the plan on faith.

**The usage-site oracle IS sound and IS Tier-1.** A component def is removable iff every operation that `$ref`s it collapses to an inference-equivalent response/parameter via `SchemaEquivalence::subsumes()` — provenance-based, no method-body parsing, no dataflow, consistent in spirit with the merged operation-level rule. Counterexamples checked: zero-referencer orphan (excluded — see amendment 2), mixed referencers (kept — every-op requirement), transitive comp→comp (kept via dangling guard — see amendment 3).

Amended the PR body with 3 tightenings the coder MUST implement (the plan's prose was directionally right but under-specified on each, each found by reading the code):

1. **Comparator must resolve the bare `$ref` before comparing.** Harvested operations keep responses/params as a bare `$ref` (harvester does not inline — `mergeAuthoredOperation:128`); `OperationSubsumption::subsumes:108` actively *bails* on a bare-`$ref` response. So `ComponentSubsumption` must resolve the ref to the component body (`scanner->responseComponentForName()`, Tier-1) and subsume against the inferred per-status inline response — do NOT reuse `OperationSubsumption` verbatim for this.
2. **"≥1 referencing operation" is a hard gate, not built into the engine.** `OaRedundancyEngine::evaluate:50` has no referencer-count check; zero referencers makes "all subsume" vacuously true → would fire on an orphan. Rule must skip zero-referencer components explicitly. Test added to plan.
3. **The dangling guard is net-new, not a copy.** `isSchemaReferencedByOtherAuthored:516` walks only schemas+operations, NOT the response/parameter component pools — so the transitive `AliasNotFound`→`NotFound` case (live in the fixture) is uncovered. The new `isComponentReferencedByOtherAuthored` must additionally iterate the response+parameter component pools. Test added to plan.

Everything else accurate: `ApiRule`/`checkApi` seam mirrors `DocumentAnnotationMigration` (no per-component tree node exists — correct not to invent one); `InferenceView` has exactly 3 accessors, none for components (claim holds); fixer reusable unchanged for class-level defs; `DocblockAnnotationRemover` already removes per-block (the multi-block `ComponentDefinitions` fixture is fixable; keep the TDD-first per-block assertion); rule ID + Improvable/L4 + `docs/linting.md` row + `## Migration rules` paragraph + `CHANGELOG` all correctly specified.

Radiergummi:
🤖 **[lead]** plan-review approved with 3 mandatory tightenings (amended into PR body) → coding. Oracle confirmed sound + Tier-1 against main. REQUIRED per review: (1) **ComponentSubsumption must resolve the bare $ref before comparing** — harvested ops keep responses/params as bare $ref (harvester doesn't inline); OperationSubsumption::subsumes bails on bare-$ref, so you CANNOT reuse it verbatim — resolve via scanner->responseComponentForName()/parameterComponentForName() to the component body and subsume against the inferred inline per-status response/param. (2) **Explicit ≥1-referencing-operation gate** — OaRedundancyEngine::evaluate does NOT enforce it; zero referencers makes 'all subsume' vacuously true → would fire on an orphan; the rule must skip zero-referencer components. (3) **isComponentReferencedByOtherAuthored is net-new, not a copy** — isSchemaReferencedByOtherAuthored walks only schemas+operations, NOT the response/parameter component pools; your new guard must also iterate both pools (transitive AliasNotFound→NotFound in the ComponentDefinitions fixture is the proving case). TDD-first throughout, incl. the DocblockAnnotationRemover multi-block removal (reviewer confirmed it removes per-block by span, so it's fixable — assert it). New rule migration.oa-redundant-component-with-inference, Improvable/L4 default, docs/linting.md row + CHANGELOG. Gates: composer test / pint --test / phpstan. Survey N/A (lint).

Radiergummi:
🤖 **[lead]** design decision on the coder's STOP-and-report: **proceed with option 1** (per-component-name selection in the fixer).

The reused `RedundantOaAnnotationFixer` removers operate at **class granularity** (remove every `@OA\*` block/attribute on the node). For a component-definitions class mixing redundant + load-bearing defs — the realistic reusable-component idiom — that over-removes surviving siblings. The coder verified this empirically; plan-review's finding (f) holds only for a uniform class, not a mixed one. (Recording that correction here so it's durable.)

**Approved approach:** thread an **optional** `componentName` (+ component type) discriminator through the finding/fix context. When present: `DocblockAnnotationRemover` keeps only the block whose text matches `response[=:]"X"` / `parameter[=:]"Y"`; `OaAttributeRemover` filters indices to the attribute whose `response:`/`parameter:` named arg matches. **When absent (null): current whole-node behavior, unchanged** — so the shipped operation/document-level rules are untouched.

**Guardrails (required):**
- The existing migration-rule goldens (operation-level + document-level) MUST stay green — they are the regression proof that the null-discriminator path is byte-identical to today.
- Red-verify the mixed-class case: a class stacking a redundant + a load-bearing component → `--fix` removes ONLY the redundant block, the sibling survives verbatim.
- The name-matching rides on the existing `locateAnnotationBlocks` text-location mechanism (it already locates `@OA\` blocks by text), so this is a filter on an existing mechanism, not a new fragility class — keep it that way; don't invent a parallel parser.

**LOC note:** this is ~+40-70 over the planner's 180-230 estimate, so src/ will likely clear the ~150 controversy threshold → I'll make the merge call after review+gates (expect a human gate to Moritz, as with #402). Do NOT inline or cut corners to game LOC — correctness + the null-path regression safety come first.

Land the full change (rule + scanner + comparator + the discriminator-aware fixer + all tests). Survey N/A (lint).

Radiergummi:
🤖 **[coder]** implemented; pushed `3d8ff043`. Ready for review.

**Shipped:** `migration.oa-redundant-component-with-inference` (SwaggerPhp plugin, level-4 Improvable, off by default) — flags + `--fix`-removes redundant reusable `@OA\Response`/`@OA\Parameter` component definitions.

**All 3 required tightenings (red-verified):**
1. **Bare-`$ref` resolved before comparing.** `ComponentSubsumption` resolves the component body and subsumes the inferred inline per-status response/param via `SchemaEquivalence` (drops/aligns the `response`/`parameter` identity field, the way `SchemaEquivalence` already drops `schema`). Did NOT reuse `OperationSubsumption` (it bails on bare-`$ref`). Red-verified: forcing `responseSubsumed`→true makes the load-bearing `DescribedOk` wrongly fire.
2. **Explicit ≥1-referencing-operation gate.** The rule skips zero-referencer components before the engine; red-verified via the orphan (`OrphanOk`) test.
3. **Net-new `isComponentReferencedByOtherAuthored`.** Walks schemas + operations (minus the verified-collapsing consumers, passed in) + both component pools incl. ref-only aliases. Red-verified via the transitive `AliasedOk`←`AliasingResponse` test.

**Deviation (per lead approval — option 1):** the reused fixer removers operated at **class granularity** (removed every `@OA` block/attribute on the node), over-removing siblings on a class stacking several component defs. Threaded an optional `componentName` discriminator through the finding context; `DocblockAnnotationRemover`/`OaAttributeRemover` now narrow to the single matching block/attribute, **null keeps the whole-node behavior** the schema/operation rules rely on. Both per-block narrowings red-verified; the existing operation/schema fixer goldens stay green (null-path regression proof).

**Tests:** `tests/Feature/SwaggerPhpComponentMigrationRuleTest.php` (subsumed response+param flagged; load-bearing kept; dangling-alias kept; orphan not flagged; off at default level; `--skip migration.*`; same-API-surface; fixer/description). Scanner unit-test extensions (declaring-class, pool exposure, guard true/false). `RedundantOaComponentFixerTest.php` (attribute + docblock per-block removal keeps siblings; absent-component → no fix).

**Gates (local):** `composer test` 2980 passed · `vendor/bin/pint --test` passed · PHPStan L8 `[OK] No errors`. Survey N/A (lint rule; same-API-surface test is the unchanged-document proof).

**Heads-up for reviewers:** net **+549 src LOC** (rule 236, scanner +149, `ComponentSubsumption` 86, fixer/removers +78), over the ~180-230 estimate — the overrun is the lead-approved fixer per-component-name extension across 3 files (the STOP-and-report finding), not padding. No `Contracts`/`Support`/stage-order/config change; all under `src/Plugins/SwaggerPhp/`.

Radiergummi:
🤖 **[lead]** implemented + local gates green (2980 passed / pint / phpstan), pushed 3d8ff043, marked ready → code-review (two independent reviewers). All 3 required tightenings red-verified by the coder. **Two scrutiny axes this round:** (1) CORRECTNESS — the bare-$ref resolution in ComponentSubsumption; the component-identity-field drop/align (response/parameter, analogous to SchemaEquivalence dropping schema); the net-new isComponentReferencedByOtherAuthored walking schemas + ops-minus-collapsers + BOTH component pools incl ref-only aliases (transitive guard); the ≥1-referencer orphan gate; and the discriminator-aware fixer removing ONLY the targeted block in a mixed class while the null path stays byte-identical (existing operation/schema goldens are the regression proof). (2) SCOPE — src is net +549, far over estimate; confirm it's justified (option-1 fixer extension + docblock-heavy rule) and NOT speculative abstraction/scope creep; flag anything that should be trimmed. Survey N/A (lint). I'll make the merge call after both approvals + remote CI green (expect escalation to Moritz on LOC).

Radiergummi:
🤖 **[reviewer2]** review round 1 — 1 finding:

**1. `DocblockAnnotationRemover::blocksDefiningComponent()` (src/Plugins/SwaggerPhp/Lint/Fix/DocblockAnnotationRemover.php:115-135) — prefix-name mis-selection causes silent data loss in the docblock fixer path.**

The component discriminator regex is built as:

```
~\b(?:response|parameter)\s*[=:]\s*["\x27]?Name["\x27]?~
```

The trailing closing-quote class is **optional** (`["\x27]?`), so the name is not anchored at its end. When the target component name is a prefix of another stacked component on the same class, the pattern matches the longer sibling too. Verified:

```php
$pattern = ~\b(?:response|parameter)\s*[=:]\s*["\x27]?Ok["\x27]?~;
preg_match($pattern, "    response=\"OkResponse\",");   // => 1  (false match)
```

Consequence: a docblock stacking `@OA\Response(response="Ok", …)` and `@OA\Response(response="OkResponse", …)`, fixing the redundant `Ok`, selects **both** blocks for removal (each block is kept if any of its lines matches). If the docblock is annotation-only, the whole doc-comment is dropped — deleting the load-bearing `OkResponse` sibling and dangling the operation that `$ref`s it. This is exactly the discriminator regression the design set out to prevent; it just leaks through the text path.

Note the **attribute** path is safe: `OaAttributeRemover::matchesComponent()` (OaAttributeRemover.php:84) compares `$argument->value->value === $componentName` with exact equality. Only the docblock regex is vulnerable.

Fix: anchor the name end — make the closing quote mandatory when an opening one is present, or append a `(?!["\x27\w])` negative lookahead / require a non-word boundary after the name. Add a fixture with prefix-colliding component names (e.g. `Ok` + `OkResponse`) stacked in one docblock and assert the sibling survives — this branch is currently untested (all fixtures use non-prefix names).

Everything else holds up well:
- Usage-site oracle threaded through the `$ref` is the correct design (vanilla inference emits no reusable component defs); `responseSubsumed` aligns the status identity and `parameterSubsumed` drops the `parameter` identity before comparing — both conservative.
- Dangling guard (`isComponentReferencedByOtherAuthored`) correctly walks schemas + the full component pool (transitive alias kept) + non-excluded operations; the `AliasedOk`/`AliasingResponse` and `OrphanOk` fixtures cover the kept/orphan cases.
- Mixed subsumable + non-subsumable referencers → the per-operation loop returns null on the first failure, keeping the component. Correct.
- `componentName` defaults null → whole-node removal; the 50 existing schema/operation migration + fixer tests stay green (regression-proof).
- Conventions clean: strict-types, `@internal`, no abbreviations, why-comments, 100-char, plugin boundary intact; `docs/linting.md` catalog row + Migration-rules prose added; `CHANGELOG.md [Unreleased]` entry present. Pint, PHPStan (level 8), and all 29 new + 50 regression tests pass locally.

Scope verdict on +549 LOC: justified, no speculative abstraction. `ComponentSubsumption` (86) is a thin wrapper reusing `SchemaEquivalence`; the scanner additions (+149) are the harvest index the oracle needs; the fixer discriminator (+78) is the minimal change to avoid blast-radius removal. Nothing to trim.

Radiergummi:
🤖 **[reviewer1]** review round 1 — 2 findings (both non-blocking; correctness of the shipped src is sound and proven):

**What I verified (not on faith):**
- **Bare-`$ref` resolution + identity-field alignment** — `ComponentSubsumption` does NOT reuse `OperationSubsumption` (correctly, it bails on bare-`$ref`); it resolves the component body and calls `SchemaEquivalence::subsumes(inferredInline, candidate)`. Responses align the identity by `candidate->response = $status`; parameters drop it via `Generator::UNDEFINED` (then `normalize` drops it). `SchemaEquivalence::normalizeAnnotation` only drops `schema` on `OA\Schema`, so handling the `response`/`parameter` identity in the comparator is the right place. Direction is correct (`narrower=authored ⊆ broader=inferred`): `DescribedOk`'s human description is in `narrower` but not `broader` → kept. Confirmed live.
- **Orphan gate** — `evaluateResponse`/`evaluateParameter` return early on `operationsReferencing() === []` before the engine; `OrphanOk` not flagged. Confirmed.
- **Dangling guard (`isComponentReferencedByOtherAuthored`)** — walks schemas + component pool (incl. ref-only aliases) + operations-minus-collapsers. I disabled the component-pool walk and the Feature test `keeps a component still $ref-ed by another surviving authored component` **fails** — so the alias-walk path is genuinely covered and load-bearing. `$ownDefinition` exclusion is by-identity against the same object pushed to the pool; correct.
- **Discriminator-aware fixer** — null path leaves the operation/schema goldens byte-identical (`--group fix`: 111 passed). Mixed-class targeted removal: attribute side uses exact `===`, docblock side narrows per-block. `RedundantOaComponentFixerTest` proves siblings survive.
- **Gates in the worktree:** `composer test` 2980 passed · `vendor/bin/pint --test` passed · PHPStan L8 `[OK] No errors`.

**Finding 1 — `tests/Unit/Plugins/SwaggerPhp/AuthoredAnnotationScannerTest.php:140` — the dedicated alias-walk unit test is vacuous (passes for the wrong reason).**
`operationsByMethod` is keyed by the **leading-slash-free FQCN** `declaringClass::method` (`AuthoredAnnotationScanner::methodKey:413`). The test excludes `['ComponentRefController::index']` (short, un-namespaced) — which does **not** match the real key `Radiergummi\OpenApi\Tests\Fixtures\SwaggerPhpComponents\ComponentRefController::index`. So that operation is never excluded; it references `NotFound` directly, the operation-walk returns `true`, and the alias (`AliasNotFound`) walk is never reached. Proof: I replaced the component-pool walk with `foreach ([] …)` and this test still passed (17 passed). The src is correct (the Feature test covers the alias path for real), but this unit test asserts nothing about the guard it names. **Fix:** use the full FQCN key so the operation is actually excluded, isolating the alias path — e.g. `['Radiergummi\OpenApi\Tests\Fixtures\SwaggerPhpComponents\ComponentRefController::index']`. (The sibling test at :164 already uses the full FQCN — they're inconsistent.)

**Finding 2 — `src/Plugins/SwaggerPhp/Lint/Fix/DocblockAnnotationRemover.php:117` — `blocksDefiningComponent` regex prefix false-match.**
The pattern `~\b(?:response|parameter)\s*[=:]\s*["\']?NAME["\']?~` makes the trailing quote optional, so it matches a **prefix**: searching for component `Record` matches a block containing `parameter="RecordPath"`. Verified with `preg_match`: matches `parameter="RecordPath"` for name `Record` (returns 1). In a class docblock stacking two component definitions whose names are prefixes of one another, `--fix` for the shorter name can locate the wrong block. Low probability (uncommon naming), docblock-form only (the attribute side uses exact `===`, unaffected), but a real mis-target. **Fix:** require the delimiter to close, e.g. `["\']?%s\b` won't help (a `Name` like `Record` still word-boundaries before `Path`)… use `(["\'])%s\1` to require matching quotes, or anchor `%s` to end at a non-name/quote boundary like `["\']?%s(?=["\'\s,)])`.

**Scope verdict on +549 src LOC: justified, no trimming demanded.** Rule 236 LOC is mostly the two parallel `evaluateResponse`/`evaluateParameter` methods (the status-threading makes a clean merge awkward; acceptable duplication for clarity) + docblocks + 5 boilerplate interface methods — no speculative abstraction. Scanner +149 is all additive and load-bearing (`componentPool` only exists for the alias-walk guard; the new accessors are each consumed). The fixer +78 is the lead-approved option-1 component-name discriminator across 3 files, with the null path preserved as the regression-proof default. `ComponentSubsumption` 86 is a thin adapter over `SchemaEquivalence`, not a reimplementation. Nothing beyond #202.

Radiergummi:
🤖 **[lead]** code-review round 1: both reviewers converge, scope (+549) confirmed justified → back to coder (round 1/3). 2 findings, **#1 is a real silent-data-loss correctness bug** (both reviewers found it independently):

1. **REQUIRED (correctness) — DocblockAnnotationRemover.php:117 prefix-collision.** The discriminator regex makes the closing quote OPTIONAL (`["']?NAME["']?`), so the component name isn't end-anchored: targeting `Ok` matches `response="OkResponse"`, `Record` matches `parameter="RecordPath"` (verified). In a class stacking two component defs where one name is a prefix of another, `--fix` for the shorter name selects BOTH blocks → drops the load-bearing sibling and dangles its operation. This is exactly the over-removal option-1 was meant to prevent, leaking through the text path (attribute path is safe, uses ===). Fix: end-anchor with matched quotes `(["'])NAME\\1` or a closing-boundary lookahead. ADD a prefix-collision golden (mixed class, two names where one prefixes the other; --fix the shorter → only its block removed, the longer-named sibling byte-identical). This branch is currently untested.

2. **REQUIRED (test-quality) — AuthoredAnnotationScannerTest.php:140 vacuous test.** The dedicated alias-walk unit test excludes `['ComponentRefController::index']` but operationsByMethod is keyed by the full leading-slash-free FQCN, so the exclusion never matches → the op-walk returns true and the alias-pool path is never reached (proven: stub the pool-walk to [] and it still passes). Use the full FQCN key (the sibling test at :164 already does). Red-verify after: stub the pool-walk → this test must FAIL.

Both are cheap. Re-review (round 2) after push. Gates: composer test / pint --test / phpstan.

Radiergummi:
🤖 **[coder]** addressed round 1; pushed `a8c35d9c`.

**Finding 1 (correctness — prefix collision):** end-anchored the docblock name match — `["\]?%s["\]?` → `["\]?%s\b`. The trailing `\b` anchors the name end for both the quoted (`"Name"`) and bare (`Name`) forms (a closing quote/comma is a word boundary), so targeting `Ok` no longer matches `response="OkResponse"`. The attribute path was already exact (`===`), unchanged. Added a prefix-collision golden (`PrefixNameComponents`: `Ok` + `OkResponse` stacked in one docblock) asserting `--fix Ok` removes only the `Ok` block and leaves `OkResponse` byte-identical. **Red-verified:** the old optional-quote regex fails this golden (removes both blocks).

**Finding 2 (vacuous test):** the alias-walk test excluded the referencing op by the short key `ComponentRefController::index`, which never matched `operationsByMethod` (full leading-slash-free FQCN), so the verdict came from the op-walk and the alias-pool path was untested. Switched to the full FQCN key so the truthy verdict comes only from the `AliasNotFound` pool entry. **Red-verified:** stubbing the component-pool walk to `[]` now makes this test FAIL (it passed vacuously before).

**Gates (local):** `composer test` 2981 passed / 0 failed · `vendor/bin/pint --test` passed · PHPStan L8 `[OK] No errors`. (Pint added a blank line between the two stacked `@OA\Response` blocks in the new fixture; block location is unaffected — golden still green.)

Radiergummi:
🤖 **[lead]** round 1 addressed (pushed a8c35d9c, both findings closed + red-verified, gates green: 2981 passed / pint / phpstan) → re-review round 2. Coder fixed the prefix collision with a trailing word-boundary (`["']?NAME\\b`) rather than reviewer2's suggested matched-quotes (`(["'])NAME\\1`). Reviewers: confirm `\\b` FULLY closes the hole — the match is already anchored by the leading `(response|parameter)\\s*[=:]\\s*["']?`, so the residual vector was only the unbounded trailing side; verify the prefix-collision golden (Ok/OkResponse) passes AND consider edge cases (unquoted value `response=Ok`, a component name ending in a non-word char, a name that's a prefix differing only by a digit like Ok/Ok2). And confirm the de-vacuumed alias-walk test now genuinely fails when the pool-walk is stubbed. Approve if clean.

Radiergummi:
🤖 **[reviewer1]** review round 2: **approved.** Both round-1 findings closed and independently re-verified (I re-ran the red proofs myself, not on faith).

**Finding 1 (vacuous alias-walk test) — fixed.** The exclusion key is now the full leading-slash-free FQCN. I stubbed the component-pool walk to `foreach ([] …)` and the test now **fails** (`Failed asserting that false is true`) — so the truthy verdict comes solely from the `AliasNotFound` pool entry, not the op-walk. It genuinely exercises the alias path now.

**Shared prefix-collision finding — fixed; the `\b` choice is correct (and better than matched-quotes here).** I checked every edge the lead named via `preg_match`:
- quoted exact `Ok` → block `response="Ok"`: matches (1) ✓
- quoted `Ok` → `response="OkResponse"`: no match (0) ✓ — hole closed
- **bare** form `response=Ok` and `response=Ok,`: matches (1) ✓ — this is why `\b` beats reviewer2's `(["'])NAME\1`, which would miss the unquoted docblock form swagger-php accepts
- digit-suffix `Ok` vs `response="Ok2"`: no match (0) ✓ (`\b` does not fire between `k` and `2`, both word chars)
- internal-punct names (`Error.Body`, `Page-Param`): still match ✓
- I also red-verified the new `Ok`/`OkResponse` golden: reverting the pattern to the old optional-quote form makes `RedundantOaComponentFixerTest > prefix-named` **fail**, so the golden is sound, not a tautology.

**One residual non-blocking nit (degrade-gracefully, not a regression risk):** a component name *ending* in an OpenAPI-legal non-word char (`Foo.`, `Bar-`) won't match the end-anchored pattern, so `--fix` would no-op on it (the finding is still reported; nothing is wrong-removed). The old pattern matched these, but trailing-punctuation component names are vanishingly rare and the failure mode is safe (no fix vs. wrong fix). Not worth holding the PR; flag only if a future fixture hits it.

**Gates re-run in the worktree on `a8c35d9c`:** `composer test` **2981 passed** (one more than round 1, the new prefix golden) · `vendor/bin/pint --test` passed · PHPStan L8 `[OK] No errors`. Scope verdict from round 1 stands (+549 src justified, no trimming). Approved for merge.

Radiergummi:
🤖 **[reviewer2]** review round 2: approved.

Finding 1 (the round-1 silent-data-loss prefix collision) is fully closed for every realistic input. I pressure-tested the new end-anchored pattern `~\b(?:response|parameter)\s*[=:]\s*["\x27]?NAME\b~` against all the edges:

| target | line | match | want |
|---|---|---|---|
| `Ok` | `response="OkResponse"` | 0 | 0 ✓ (the round-1 bug) |
| `Ok` | `response=OkResponse` (bare) | 0 | 0 ✓ |
| `Ok` | `response="Ok2"` | 0 | 0 ✓ (`\b` correctly does **not** fire between `k` and `2`) |
| `Ok` | `response=Ok` (bare) | 1 | 1 ✓ |
| `Ok` | `response='Ok'` | 1 | 1 ✓ |
| `Ok2` | `response="Ok2"` | 1 | 1 ✓ |

The `\b` works because the leading `["\x27]?` already anchors the start; the residual round-1 vector was the unbounded trailing side, which `\b` closes for any name ending in a word character (the universe of normal component names). The prefix-collision golden (`PrefixNameComponents`, `Ok` + `OkResponse`) is correct and was red-verified against the old optional-quote regex. The de-vacuoused scanner alias-walk test (now excluding the consumer by its real leading-slash-free FQCN key, so the truthy verdict comes only from the `AliasNotFound` pool entry) is a genuine improvement — it now actually exercises the component-pool branch of the dangling guard rather than the op-walk.

**Non-blocking note (no change required to merge):** one degenerate residual remains. A component name **ending in a non-word char** (`Foo.`, `User-` — legal but pathological OpenAPI keys) fails to match its own quoted definition line, because `\b` between the trailing `.`/`-` and the closing quote is not a boundary, so the fixer yields **no fix** (a no-op, not data loss — the finding still reports). No realistic Laravel app names a reusable component that way, so this does not block. If you ever want to close it, a terminator lookahead `["\x27]?NAME(?=["\x27,\s)]|$)` covers both the word-char and non-word-char suffix in the quoted and bare forms (verified); the matched-quotes `(["\x27])NAME\1` you floated would fix the quoted case but reopen the bare unquoted one, so the lookahead is the better closure. Filing as a follow-up at most.

Verification: full suite `composer test` 3016 passed / 0 failed (parallel, CI-equivalent), PHPStan level 8 clean, `vendor/bin/pint --test` passed. (A hand-picked single-process file subset showed two scoped-singleton ordering artifacts in the alias tests; they pass in isolation and under the parallel CI runner — not a defect.) Scope unchanged from round 1: justified, nothing to trim.

Radiergummi:
🤖 **[lead]** ALL gates green → escalating to Moritz for the merge call (LOC controversy rule), **not auto-merged**.

**Why human-gated:** src diff is net **+549** — far over the ~150 auto-merge threshold (also a new lint rule introducing a default severity, and it touches the shared AuthoredAnnotationScanner + fixer removers). All additive, all within `src/Plugins/SwaggerPhp/`, null-discriminator path byte-identical to today (the 50 shipped operation/document/schema migration+fixer goldens stay green = the regression proof). No Contracts/Support/stage-order/config change.

**Evidence it's ready:**
- **Two independent reviewer approvals** at final SHA `a8c35d9c`. Adversarial loop caught + fixed two real issues: a silent-data-loss prefix-collision in the docblock discriminator regex (both reviewers found it independently; fixed via trailing `\b` end-anchor, which both reviewers confirmed is *better* than matched-quotes because swagger-php accepts the bare/unquoted form), and a vacuous alias-walk test (de-vacuumed, red-verified). One non-blocking residual noted (a component name ending in a non-word char no-ops instead of fixing — NOT data loss; lookahead closure available as a future follow-up).
- **Remote CI 11/11 green** incl. swagger-php 5.8.
- **Survey N/A** — lint rules don't change generation; the same-API-surface feature test proves `--fix` leaves the document byte-identical.
- Scope (+549) independently judged justified by both reviewers, nothing to trim.
- A real fixer-granularity bug (whole-class over-removal) was caught at the coding stage and fixed via the lead-approved option-1 discriminator before any bad `--fix` could ship.

PR left **ready + unmerged**, worktree in place. Moritz requested as reviewer. Recommend: **merge as-is** — the LOC is inherent to the option-1 fixer extension I approved (per the realistic mixed-component-class case); no inlining was done to game the count.

PR #421, opened 2026-06-20
feat(#409): infer additionalProperties for native Collection<string,X> map generics (non-Spatie)

### LINKED ISSUE #409: Infer additionalProperties for native typed array<string,X> properties (non-Spatie)
Spun off from #200.
Parent decision: #200.

## Summary

Extend `additionalProperties` (map/dictionary) inference beyond the SpatieData plugin so that **native typed properties** and `Collection<string, X>` typed shapes also derive `additionalProperties` in their schema.

## Context

`#200` audited residual `OA\*` constructs. For `additionalProperties`, two angles were already shipped:

- **Explicit override** — the field-attribute parameter `bool|string|null $additionalProperties` (`src/Attributes/FieldAttribute.php:83`) rendered by `SchemaDescriptor::applyAdditionalProperties()` (`src/Support/Generator/SchemaDescriptor.php:154`).
- **Inference from `array<string, X>`** — but **only inside the SpatieData plugin**, via `SchemaFromDataClass::resolveCollectionSchema()` (`src/Plugins/SpatieData/Support/SchemaFromDataClass.php:502`) and `TypeNodeToSchema::fromGeneric()` (`src/Support/Types/TypeNodeToSchema.php:194`).

The residual gap: map inference does **not** fire for plain typed properties outside the Spatie path, nor for `Illuminate\Support\Collection<string, X>` typed shapes.

## Proposed direction (Tier 0)

The map-detection primitive already lives in `Support\` (`TypeNodeToSchema::fromGeneric()` handles `array<string, X>`). The work is to ensure non-Spatie schema-building paths route string-keyed generics through it, including `Collection<string, X>` / typed-collection PHPDoc shapes. Pure reflection + PHPDoc type-info; no body parsing.

## Acceptance

- A native typed property / PHPDoc-annotated `array<string, X>` or `Collection<string, X>` derives `{ type: object, additionalProperties: <X-schema> }`.
- `array<int, X>` / `list<X>` / bare `array` keep list semantics (unchanged).
- Tests cover scalar value, `$ref` value, `mixed` (→ `true`), and nested maps, mirroring the existing SpatieData map tests.

Radiergummi:
Closes #409

### Tier & rationale

**Tier 0** — pure reflection + PHPDoc type-info, no body parsing. The map-detection primitive
already lives in `Support\Types\TypeNodeToSchema::fromGeneric()`; the work is to widen what it
recognises so non-Spatie schema-building paths derive `additionalProperties` for string-keyed
generics. No new dataflow, no AST whitelisting.

### Scoping note (read first — narrows the issue)

The issue's framing implies `array<string,X>` is broken on the non-Spatie path. It is not. After
reading the code, the actual state is:

- **`array<string, X>` already works on the native PHPDoc path.** The Eloquent `@property` /
  `@property-read` path (`EloquentModelToSchema::propertyFromTag`, line 402) routes every type node
  through `TypeNodeToSchema::resolve()` → `fromGeneric()`, which already maps `array<string, X>` →
  `{type: object, additionalProperties: <X>}` (`TypeNodeToSchema.php:200-202`) and keeps
  `array<int,X>` / `list<X>` / bare `array` as list/null. This is even documented today
  (`docs/auto-derivation.md:152`). The unit test `array<string, T> → additionalProperties` already
  passes (`TypeNodeToSchemaTest.php:111`).
- **The genuine residual gap is `Collection<string, X>`.** `Collection` (and the typed-collection
  PHPDoc shapes) is a `GenericTypeNode` whose base identifier is `Collection`, which is **not** in
  `TypeNodeToSchema::ARRAY_GENERICS`, so `fromGeneric()` returns `null` for it today — confirmed by
  the existing assertion `resolveType('Illuminate\Support\Collection<int, \stdClass>')->toBeNull()`
  (`TypeNodeToSchemaTest.php:160`).

So the change is: **teach `TypeNodeToSchema::fromGeneric()` to treat `Collection`-family generic
bases as array-like**, so `Collection<string, X>` derives `additionalProperties` and
`Collection<int, X>` / `Collection<X>` derive a list — identical key-semantics to `array<…>`. Plus
align the `mixed` value to `additionalProperties: true` to match the SpatieData path and the issue's
acceptance.

This keeps the change surgical and in `Support\` (no plugin dependency), exactly per the issue's
"route … through the existing primitive" direction.

### Files to modify

1. **`src/Support/Types/TypeNodeToSchema.php`** — the only production change.
   - Recognise collection generic bases. Add a small set of recognised collection FQCN/short names
     so a `GenericTypeNode` with base `Collection` (and the common Laravel/Support collection
     identifiers) is treated as array-like in `fromGeneric()`. Match by the **short class name**
     (case-insensitive), since PHPDoc may write `Collection`, `\Illuminate\Support\Collection`, or
     `EloquentCollection`; resolving to the exact FQCN is unnecessary for Tier 0 and the short-name
     whitelist mirrors how `ARRAY_GENERICS` already matches `$node->type->name`. Recommended set:
     `Collection`, `EloquentCollection`, `LazyCollection`, `Enumerable`. Keep this list as a private
     const alongside `ARRAY_GENERICS`; the `in_array(strtolower(...))` guard in `fromGeneric()`
     widens to check both lists. Do **not** fold these into `ARRAY_GENERICS` (name says "array").
   - **`mixed` value → `additionalProperties: true`.** In the two-arg string-key branch
     (`TypeNodeToSchema.php:195-202`), when the value node is the `mixed` identifier, emit
     `'additionalProperties' => true` instead of the `?? new OA\Schema([])` empty schema. This
     aligns with `SchemaFromDataClass::mapValueSchema()` (returns `true` for `mixed`) and the issue
     acceptance (`mixed → true`). A tiny helper (e.g. `isMixed(TypeNode): bool`) keeps it readable.
     Note the list branch is unaffected (a list of `mixed` keeps `items: {}`), matching today.

   No changes needed in `EloquentModelToSchema`, `JsonSchemaFromType`, `ValidationRulesToSchema`, or
   `UriParametersExtractor`: the Eloquent `@property` path already delegates to `TypeNodeToSchema`,
   so widening the primitive is sufficient. (If review wants `Collection<string,X>` to also work for
   model accessors whose **return type** is `Collection` — that is the symfony-type `JsonSchemaFromType`
   path, which has no key info from reflection alone and is out of scope; reflection can't see the
   generic key. Flagged below.)

### Tests (TDD — write first, watch them fail, then implement)

Primary, unit-level, in **`tests/Unit/Support/Types/TypeNodeToSchemaTest.php`** (uses the existing
`resolveType` / `resolveTypeToArray` helpers; mirrors the SpatieData map tests):

1. `Collection<string, int>` → `{type: object, additionalProperties: {type: integer}}` (scalar value).
2. `Collection<string, \stdClass>` → `additionalProperties` is `{$ref: '#/components/schemas/stdClass'}`
   ($ref value, via the class-leaf strategy).
3. `Collection<string, mixed>` → `additionalProperties === true` (mixed → true).
4. `Collection<string, Collection<string, int>>` → nested map: outer `additionalProperties` is an
   object whose `additionalProperties.type` is `integer` (nested maps).
5. `Collection<int, \stdClass>` → list semantics: `{type: array, items: {$ref: …}}`. **This replaces
   the existing assertion at `TypeNodeToSchemaTest.php:160`** which currently expects `null`. Update
   that test (don't leave a contradictory expectation).
6. `Collection<\stdClass>` (single arg) → list of `$ref` (list semantics for single-generic form).
7. Regression guards (keep list semantics for arrays — verify unchanged): keep/confirm the existing
   `array<int, T>` → list and bare `array` → null tests; add an explicit `array<string, mixed>` →
   `additionalProperties: true` test to lock the new mixed behaviour on the array path too.

Also extend the **mixed** coverage on the array path so the change is pinned for both bases.

Edge cases explicitly covered per the acceptance criteria: scalar value (1, 7), `$ref` value (2),
`mixed → true` (3, 7), nested maps (4); and that `Collection<int,X>` / `Collection<X>` /
`array<int,X>` / `list<X>` / bare `array` keep list (or null) semantics (5, 6, 7).

Optional integration check (only if the coder finds it cheap): an Eloquent fixture model with a
`@property Collection<string, SomeChild>` to assert the property surfaces as `additionalProperties`
end-to-end through `EloquentModelToSchema`. If a suitable fixture/test harness isn't already in
place, the unit tests above are sufficient proof and the integration test can be skipped — the
Eloquent path is already covered for `array<string,X>` and uses the same primitive.

### Docs

`docs/auto-derivation.md:152` — the `@property` array-shape sentence currently lists
`array<string, T>` → `additionalProperties`. Extend it to mention that `Collection<string, T>`
(and the other recognised collection generics) are treated identically, and that `mixed` values
yield permissive `additionalProperties`. Keep it to the existing sentence's style; no new section.

### CHANGELOG

Add under `[Unreleased]` (Added or Changed):
> Native PHPDoc `Collection<string, X>` (and `EloquentCollection` / `LazyCollection` / `Enumerable`)
> generics now infer `{type: object, additionalProperties: <X>}`, matching the existing
> `array<string, X>` map inference; `mixed`-valued string-keyed maps emit `additionalProperties: true`.

### Lint rule

None added or changed.

- **Behaviour change to an existing test expectation.** `TypeNodeToSchemaTest.php:160` asserts
  `Collection<int, X>` resolves to `null` (caller keeps its fallback). After this change it resolves
  to a list schema. That is the intended new behaviour, but it flips a documented "returns null for a
  non-array generic" contract of the primitive. Worth a reviewer's eye: any non-Eloquent caller of
  `TypeNodeToSchema::resolve()` that relied on `null` for `Collection<…>` to trigger its own fallback
  will now get a list/map schema instead. Current callers: `EloquentModelToSchema` only (grep
  confirms), where a derived schema is strictly better than the empty fallback — so low risk, but
  not zero. Flagging because it changes the primitive's null-contract surface, not just adds a case.
- **No `src/Contracts/**` change, no stage-order change, no config key, no public-signature change.**
  `TypeNodeToSchema` is `@internal`. This is a contained Tier-0 inference widening.
- **Out of scope (do not expand):** inferring map keys from a model accessor whose PHP **return
  type** is `Collection` (no PHPDoc value/key) — reflection can't see the generic arguments, so
  that's the symfony-type `JsonSchemaFromType` path and genuinely needs the `@property`/`@return`
  PHPDoc to carry `<string, X>`. Already handled once the PHPDoc is present (that's the path above).


Radiergummi:

One scoping correction surfaced while reading the code, called out in the plan: `array<string,X>` already derives `additionalProperties` on the native (Eloquent `@property`) path — the real residual gap is `Collection<string,X>`, which `TypeNodeToSchema::fromGeneric()` returns `null` for today (`ARRAY_GENERICS` excludes `Collection`). The change widens that whitelist and aligns `mixed`-valued maps to `additionalProperties: true`. One existing test expectation flips (`Collection<int,X>` → null becomes a list), flagged for the reviewer.

Radiergummi:
🤖 **[lead]** plan posted on draft PR #421 → plan-review. Scope corrected to Collection<string,X> (array<string,X> already works); one flipping test expectation flagged — reviewer please weigh.

Radiergummi:
🤖 **[reviewer]** plan-review: approved.

Verified the plan against the code; all claims hold.

- **Scoping correction is correct.** `array<string,X>` already derives `additionalProperties` via `ARRAY_GENERICS` (`TypeNodeToSchema.php:40-46`, `fromGeneric()` :200-201); the unit test at `TypeNodeToSchemaTest.php:111` already passes. The genuine gap is `Collection`-family bases, excluded from `ARRAY_GENERICS` — confirmed by the null assertion at `TypeNodeToSchemaTest.php:160`.
- **Tier-0, surgical.** Single production file, widen a whitelist + align `mixed`→`true`. No `Contracts/` change, no stage-order/config/public-signature change; `TypeNodeToSchema` is `@internal`. Keeping the collection set as a separate const (not folded into `ARRAY_GENERICS`) is the right call.
- **Null-contract flip is genuinely low-risk.** `EloquentModelToSchema` is the *sole* caller of `TypeNodeToSchema::resolve()` (grep-confirmed), and there a derived list/map schema strictly beats the empty fallback. The flipped expectation at `:160` is an acceptable contract change for an `@internal` primitive. **Acceptable.**
- **`mixed`→`true` alignment is real.** SpatieData `mapValueSchema()` returns `true` for opaque `mixed` (`SchemaFromDataClass.php:561-569`); the issue acceptance says `mixed → true`. The plan correctly special-cases *only* the literal `mixed` identifier and leaves the `?? new OA\Schema([])` empty-schema fallback for genuinely-unreadable value types — matching SpatieData (which does not return `true` for unreadable types). Good distinction; keep it.
- **List-semantics regression guards intact.** Plan keeps/confirms `array<int,X>` → list, `list<X>` → list, bare `array` → null, and adds `Collection<int,X>`/`Collection<X>` list cases plus an explicit `array<string,mixed>`→`true` pin on the array path.
- **Bookkeeping named:** proving tests (7 cases), docs (`auto-derivation.md` `@property` sentence), CHANGELOG `[Unreleased]`, no lint rule. Correct.

No required changes. Ready to implement.

Radiergummi:
🤖 **[lead]** plan-review approved (reviewer-1) → coding. Surgical Tier-0, single-file; the flipping Collection<int,X> test expectation is acceptable for the @internal primitive (sole caller EloquentModelToSchema).

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed `ee816db7`. Ready for review.

Production change is in `src/Support/Types/TypeNodeToSchema.php` (the prior `f45391f9` was an empty commit; this force-push replaces it). Diff: `TypeNodeToSchema.php` + `TypeNodeToSchemaTest.php` + CHANGELOG + `docs/auto-derivation.md`.

- Added `COLLECTION_GENERICS` const (`collection`/`eloquentcollection`/`lazycollection`/`enumerable`) and `isArrayLikeGeneric()` matching the **short class name** case-insensitively, so `Collection`, `\Illuminate\Support\Collection`, and `EloquentCollection` all route through `fromGeneric()`. `ARRAY_GENERICS` left untouched (name says "array").
- `mixed` map value → `additionalProperties: true` via a small `isMixed()` helper, special-casing only the literal `mixed` identifier; genuinely-unreadable value types still fall back to `?? new OA\Schema([])`. Matches SpatieData's `mapValueSchema()`.
- Key semantics identical to `array<…>`: string key → map, int key / single arg → list. Flipping test at `:160` updated (`Collection<int,X>` now a list, per approved plan); replaced with an unrecognised-generic (`SplObjectStorage<…>`) null guard.
- 28 unit tests pass; `composer check` green; full CI matrix green incl. swagger-php 5.8 and PHP 8.5.

Radiergummi:
🤖 **[lead]** real diff verified (sha ee816db7; src+tests+CHANGELOG+docs non-empty), full CI matrix green incl. swagger-php 5.8 → code-review. Two independent reviewers required for auto-merge.

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer B, independent) — approved with 2 minor findings (none blocking).

Reviewed sha `ee816db7` on my own merits. Correctness is sound, TDD followed, conventions and boundaries hold. The change widens generic-base recognition to the Collection family via short-name (case-insensitive) match and aligns `mixed`-valued string-keyed maps to `additionalProperties: true`, matching the SpatieData path.

Verified independently:
- **`listOf` is null-safe** (`?OA\Schema $element`), so the 2-arg list branch dropping its local `?? new OA\Schema([])` and passing `resolve(...)` directly is **behavior-preserving** (unreadable element still yields `items: {}`). No regression.
- **`mixed → true` vs unreadable → `{}`** distinction is correct and matches SpatieData's `mapValueSchema()`: only the literal `mixed` identifier returns `true`; genuinely-unresolvable value types keep the empty-schema fallback.
- **Short-name match** correctly routes `Collection`, `\Illuminate\Support\Collection`, and `EloquentCollection`. `ARRAY_GENERICS` is matched on the *full* lowercased name (so a namespaced `\X\Iterable` won't wrongly match it) while only `COLLECTION_GENERICS` uses the short name. Sound split.
- **False-positive risk** of a user class literally named `Enumerable`/`Collection`/`LazyCollection` being treated as array-like: acceptable Tier-0 degrade (such a type is iterable; map/list is a reasonable shape) and consistent with the approved plan. Not a finding.
- **Boundaries/bookkeeping:** no plugin/third-party dependency added (string literals, not imports); `@internal` intact; sole caller `EloquentModelToSchema`; CHANGELOG `[Unreleased]` entry present; `docs/auto-derivation.md` updated (existing page, no index change needed); no new lint rule. No em-dashes in comments.

Findings:

1. **Test gap — unreadable (non-`mixed`) value in a string-keyed map is untested.** `TypeNodeToSchemaTest.php`. The change introduced the boundary `isMixed($value) ? true : resolve(...) ?? new OA\Schema([])`. The `mixed → true` side is well covered; the `?? new OA\Schema([])` side (genuinely-unresolvable map value) is not. The existing unreadable tests cover only a *shape* value (`:144`) and a *list* element (`:152`), not a string-keyed-map value. Suggested test: `resolveTypeToArray('array<string, NotARealClassXyz>')` → `type === 'object'` and `additionalProperties === []` (empty schema, **not** `true`). This locks the distinction between an opaque-`mixed` permissive map and an unknown-typed map. Nice-to-have, not blocking.

2. **Two lines over the 100-char soft limit.** `src/Support/Types/TypeNodeToSchema.php:246` (102 chars, the `COLLECTION_GENERICS`/`isArrayLikeGeneric` docblock one-liner) and `:255` (101 chars, `$shortName = substr($lower, strrpos($lower, '\\') === false ? 0 : strrpos($lower, '\\') + 1);`). Both wrap cleanly. `:255` also calls `strrpos` up to 3×; a local `$separator = strrpos($lower, '\\')` would both shorten the line and avoid the repeat. Minor.

Verdict: approve. The two findings are low-severity polish; I'd merge as-is, though finding 1 is worth a quick follow-up commit if the coder is touching the file.

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer A, independent) — approved with 3 minor findings (none blocking).

I checked out `ee816db7` and ran it myself: `TypeNodeToSchemaTest` is **28 passed**; reverting *only* the production file makes the **8 new behaviour tests fail** (TDD confirmed, incl. `array<string, mixed>`, so the mixed-on-array-path change is genuinely pinned, not just the Collection path). Pint `passed`, PHPStan `No errors` on the changed file. Boundaries clean (pure `Support/`, only `strrpos`/`substr` builtins added — no plugin/third-party import).

The lead-flagged concerns all check out:
- **Short-name match is exact** (`in_array` on the short name, not a suffix), so `App\MyCollection` → `mycollection` does **not** false-match `collection` (verified empirically). FQ and leading-`\` forms both reduce to `collection`. `ARRAY_GENERICS` is still matched on the full lowercased name; only `COLLECTION_GENERICS` uses the short name — sound split.
- **`mixed`-only special-case:** a genuinely-unreadable value still falls to the `?? new OA\Schema([])` empty-schema (`{}`), not `true` — matching SpatieData's `mapValueSchema()`.
- int-vs-string key discrimination, nested-map recursion, and the list-semantics regressions are intact and tested.

Findings:

1. **Test gap — unreadable (non-`mixed`) string-keyed-map value is untested.** The `isMixed($value) ? true : resolve(...) ?? new OA\Schema([])` boundary has its `true` side well covered but not its `{}` side. Add `resolveTypeToArray('Collection<string, NotARealClassXyz>')` → `type === 'object'` and `additionalProperties === []` (empty schema, **not** `true`). This locks the opaque-`mixed`-vs-unknown-type distinction that is the whole point of the `isMixed` carve-out. Behaviour is already correct by reading the code; this just pins it. Nice-to-have, not blocking.

2. **Two lines over the 100-char soft limit.** `TypeNodeToSchema.php:246` (102 chars, the `isArrayLikeGeneric` docblock one-liner) and `:255` (101 chars). `:255` also calls `strrpos($lower, '\\')` twice; a local fixes both the width and the double scan:

```php
$separator = strrpos($lower, '\\');
$shortName = $separator === false ? $lower : substr($lower, $separator + 1);
```

3. **Nit (optional): unreadable list element on the Collection path** (`Collection<int, NotARealClass>`) isn't separately tested, but the `array`/`list` equivalent is and it is the same `listOf()` path. Not required.

Verdict: **approved** — correctness, tests, conventions, boundaries, bookkeeping all hold. Findings 1–3 are polish; finding 1 is the one most worth folding in, but none gate the merge.

(Note: reviewer B independently landed on the same findings 1 and 2 — converged without coordination.)

Radiergummi:
🤖 **[coder]** addressed review polish (findings 1+2); pushed `8431183d`.

1. Added `tests/Unit/Support/Types/TypeNodeToSchemaTest.php`: 'keeps an unreadable map value as an empty schema, not a permissive map' — covers both `array<string, NotARealClassXyz>` and `Collection<string, NotARealClassXyz>` → `type === 'object'`, `additionalProperties === []` (empty schema, **not** `true`). Locks the opaque-`mixed` vs unknown-typed-map distinction.
2. Wrapped `TypeNodeToSchema.php` :246 / :255 under 100 chars; hoisted `$separator = strrpos($lower, '\\')` so `strrpos` is called once.

No logic change. `composer check` green; full CI matrix green incl. swagger-php 5.8 and PHP 8.5/Laravel 12&13.

Radiergummi:
🤖 **[lead]** polish folded in (sha 8431183d), both reviewers approved, full CI matrix green incl. swagger-php 5.8 → survey. Generation-affecting (area:core) → Layer A gate required before merge.

Radiergummi:
🤖 **[lead]** non-controversial + all gates green → squash-merged, closes #409. Gates: 2 independent reviewer approvals (A+B); full CI matrix green at 8431183d incl. swagger-php 5.8; survey Layer A pass (surveyor: 0 regressions across 11 apps, baseline regenerated on main 4528b234, every deterministic metric identical baseline→candidate). Tiny @internal diff (45 src lines), no Contracts/config/signature/stage-order change. Branch deleted, worktree removed.

Issue #412, opened 2026-06-20
ADR: bounded interprocedural response inference — refused (evidence-based)


Radiergummi:
**Status:** Decided — refused. Supersedes the proposed "Tier-1.5 / bounded interprocedural" rung.
**Measurement:** 5-app survey subset on `main @ 63082823`, post-#393 (InvoiceNinja, AureusERP, Koel, Vito, Lychee).
Local finding doc: `docs/internal/dogfooding/2026-06-20-interprocedural-inference-refused.md` (gitignored).

## Context

Pursuing ~80% out-of-the-box coverage read as *unconditional* (any app, untouched) collides with the
Tier-2 refusal. We considered a new rung — **bounded interprocedural inference**: follow
statically-singular call edges (`$this->m()`, `final`/`final`-class receivers) 1–3 hops, bind args to
params, re-run the Tier-1 whitelist at the callee — under two fixed constraints: **never emit wrong**
and **bounded cost**. Headline motivation: InvoiceNinja's `->itemResponse()`/`->listResponse()` family
(~266 routes, from the stale #360 sizing). We re-measured before building.

## Decision

**Do not build bounded interprocedural response inference.** The ladder stays Tier 0 (reflection) +
Tier 1 (bounded intra-method whitelist); Tier 2 stays refused.

## Evidence — corrected residual taxonomy (no substantive object payload)

| app | ops | substantive | correctly-empty | Fractal dynamic | resource-extraction gap | genuine Tier-2 | interproc NEW win |
|---|--:|--:|--:|--:|--:|--:|--:|
| InvoiceNinja | 522 | 67 | 16 | 253 | 0 | 186 | **0** |
| AureusERP | 90 | 66 | 0 | 0 | 0 | 24 | **0** |
| Koel | 170 | 42 | 31 | 0 | ~17 | ~78 | **0** |
| Vito | 144 | 114 | 29 | 0 | 0 | 1 | **0** |
| Lychee | 205 | 82 | 54 | ~3 | ~24 | ~42 | **0** |

Every "followable edge" resolves to one of:
1. **Fractal** (InvoiceNinja, 253) — `response($this->manager->createData(new Item($item, new $this->entity_transformer(...)))->toArray())`: dynamic transformer, runtime envelope key (`request()->input('index')`), `response()->make(<json string>)`. Not statically resolvable; Fractal-plugin territory at best.
[...]

PR #423, opened 2026-06-20
feat(#411): type ResourceToArrayReader fields from a wrapped non-Model value object's typed properties

### LINKED ISSUE #411: feat(responses): type ResourceToArrayReader fields from a wrapped non-Model value object's typed properties
## Problem

`ResourceToArrayReader` resolves `toArray()` field *values* against the wrapped class only when it
is an Eloquent **Model** (`resolveModelProperty` → `EloquentModelToSchema`). When a resource wraps
a plain **value object**, the model lookup yields nothing and each field falls through to
`unconstrained()` — emitting a property with **no type**.

The schema is still "substantive" (the properties exist), so this does **not** move the survey's
object-payload %. It is a **fidelity** gap: correctly-typed source data is emitted as untyped.

## Evidence (Koel, survey corpus)

`App\Http\Resources\GenreResource` wraps `App\Values\GenreSummary`:

```php
// GenreResource::toArray() — a clean single-return literal
return [
    'type' => 'genres',                 // -> string (literal)  OK
    'id' => $this->summary->publicId,   // -> untyped  (should be string)
    'name' => $this->summary->name,     // -> untyped  (should be string)
    'song_count' => $this->summary->songCount, // -> untyped (should be integer)
    'length' => $this->summary->length, // -> untyped (should be number)
];

// GenreSummary — a typed value object
final readonly class GenreSummary {
    public string $publicId; public string $name; public int $songCount; public float $length;
}
```

`$this->summary->publicId` is a `PropertyFetch` on a non-`$this`/non-`resource` receiver, so
`resolveModelProperty` returns null and the field is unconstrained.

When the wrapped value is a non-`Model` object with **statically-typed public properties**
(promoted ctor params or public typed props), resolve `$this->wrapped->field` (and `$this->field`)
against that object's property types via reflection, mapping PHP types → schema the same way the
model path does. Refuse (stay unconstrained) when the property type is absent, a union we don't
model, or the receiver isn't statically a single known class.

- `WrappedModelLocator` already finds the wrapped class; generalize it (or add a sibling) to return
  a non-Model wrapped class so the reader can reflect its property types.
- Scope strictly to typed public properties — no method calls, no accessors, no dataflow.

## Out of scope / non-goals

- No interprocedural following (see ADR `docs/superpowers/specs/2026-06-20-interprocedural-response-inference-refused-adr.md`).
- Does not change the object-payload % metric; this is a type-fidelity improvement.

## Acceptance

- Fixture resource wrapping a typed value object → fields carry correct scalar types.
- Untyped / union-typed properties stay unconstrained (never-wrong).
- No regression on Model-backed resources; survey shows no spec regressions.

Radiergummi:
Closes #411

### Tier & rationale

**Tier 0** — pure reflection over a wrapped value object's typed public properties, plus the bounded AST read of the `toArray()` literal that `ResourceToArrayReader` already does. No method calls, no accessors, no dataflow, no interprocedural following (the interproc ADR #412 stands). It is a type-**fidelity** fix: data that is statically typed at the source is currently emitted untyped.

### What the gap actually is (grounded in the code)

`ResourceToArrayReader::resolveModelProperty()` (`ResourceToArrayReader.php:452`) resolves a field only via `modelFieldName()` (`:479`), which matches **only** `$this->field` / `$this->resource->field` and only against an Eloquent `Model` (`$this->modelToSchema->propertyFor(...)`). Two things block the Koel case:

1. **Receiver shape.** Koel writes `$this->summary->publicId` — a `PropertyFetch` whose receiver is `$this->summary` (a typed property on the resource), which is neither `$this` nor `$this->resource`. `modelFieldName()` returns null, so it never even reaches a type lookup → `unconstrained()`.
2. **Wrapped class is a non-Model.** Even the `$this->field` form only resolves when the wrapped class is a `Model` (`WrappedModelLocator` filters to `Model` via `modelClassOf`).

So there are **two receiver shapes** to support, both reflecting a non-Model value object's typed public properties (promoted ctor params or declared public typed props):

- **(A) `$this-><wrappedProp>-><field>`** — the evidenced Koel shape. The value-object class is the declared type of the resource's own `<wrappedProp>` property (e.g. `GenreResource::$summary` typed `GenreSummary`). Resolve `<field>` against `GenreSummary`'s public property types.
- **(B) `$this-><field>` / `$this->resource-><field>`** — same forms the model path handles, but when the resource's `@mixin`/`@extends` names a **non-Model** class. Resolve `<field>` against that class's public property types.

Both reuse one new "type a public property of a known class by reflection" primitive, with the refusal gates from the issue.

### Refusal gates (never-wrong)

Stay `unconstrained()` (don't guess) when ANY of:
- the receiver isn't statically a single known class (e.g. `$this->summary` has no type, a union type, or the class doesn't exist / isn't autoloadable);
- the named member isn't a **public** property of that class (it's a method, an accessor, protected/private, or simply absent — no method-call / accessor support);
- the property has **no declared type**, or a **union/intersection** type we don't model (a plain nullable `?T` of a modelled scalar is fine and stays modelled; a multi-class or scalar-union like `int|string` refuses).

This mirrors `JsonSchemaFromType`'s existing scalar/enum/datetime/nullable handling while *gating out* the union/unmapped cases rather than emitting `JsonSchemaFromType`'s `oneOf`/"Unmapped type" fallback (which would not be "never-wrong" for this fidelity feature).

### Files to create / modify

1. **`src/Support/Extraction/PublicPropertyTypeReader.php` (new, `Support/`, `@internal`, `#[Scoped]`).** A plugin-agnostic primitive: `propertyFor(class-string $class, string $propertyName): ?OA\Property`.
   - Reflect `$class`; require a **public**, non-static property named `$propertyName` (covers both promoted ctor params and declared public typed props — both surface as `ReflectionProperty`).
   - Read `ReflectionProperty::getType()`. Require a `ReflectionNamedType` (refuse on `ReflectionUnionType` / `ReflectionIntersectionType` / no type → return null).
   - Map via the **same path the model accessor path uses** (`EloquentModelToSchema::propertyFromAccessor`, `:451-457`): `TypeResolver->resolve($reflectionType)` → symfony `Type` → `JsonSchemaFromType::fromType()`. Guard: if the resolved symfony `Type` is a `UnionType` (other than a simple `NullableType` of a modelled type) or resolves to an unmapped object, refuse (return null) instead of emitting the `oneOf`/unmapped-string fallback. Wrap `UnsupportedException` from the resolver → null.
   - Returns a named `OA\Property` (property name set by the caller, as the model path does).
   - Lives in `Support/` (not `Plugins/`) because it's generic reflection infra reusable by any plugin, per the CLAUDE.md placement rule. Dependencies: `TypeResolver` + `JsonSchemaFromType` (both already `Support/`-resident and `#[Scoped]`).  
     *Open question for plan-review:* whether to keep this brand-new tiny class or instead add a `publicPropertyFor()` method to `EloquentModelToSchema`. Recommendation: a separate `Support/` class — `EloquentModelToSchema` is Model-specific and already large; the value-object reader is a distinct concern and keeping it free of the Model dependency is cleaner. Flagged, not blocking.

2. **`src/Plugins/ApiResources/Support/WrappedModelLocator.php` (modify) — add a non-Model sibling.** Add `locateWrapped(class-string<JsonResource>): ?string` (or rename the private `resolve()` core to return any class-string and have both `locate()`—Model-filtered—and a new `locateValueObject()` share it). Minimal change: extract the `@mixin`/`@extends` class-string resolution (currently inlined in `resolve()`) into a private helper that returns the **unfiltered** class-string; keep `locate()` applying `modelClassOf`; add a new public method returning the class-string when it is a **non-Model** class that `class_exists`. Cache it separately. This serves shape (B). (Per the issue: "generalize it or add a sibling to return a non-Model wrapped class.")

3. **`src/Plugins/ApiResources/Support/ResourceToArrayReader.php` (modify) — wire both shapes.**
   - Inject `PublicPropertyTypeReader` and reflect the **resource class** for shape (A). The reader already receives `$resourceClass` in `readFields()`; thread the resource class (or a resolved `ReflectionClass`) down to value resolution so shape (A) can look up `<wrappedProp>`'s type.
   - **Shape (A):** in `resolveValue()` add a resolver step *before* the final `SchemaDefinitionFromLiteral` fallback (mirroring `resolveModelProperty`'s position, `:289`): detect `PropertyFetch( var: PropertyFetch( var: $this, name: <wrappedProp> ), name: <field> )`. Resolve `<wrappedProp>`'s declared type on the resource class (must be a single known class via reflection). Then `PublicPropertyTypeReader::propertyFor($valueObjectClass, $field)`; on null → leave for the existing fallback chain (ends in `unconstrained`). On success → `ofProperty(...)`.
   - **Shape (B):** when `resolveModelProperty` would currently return `unconstrained` because the wrapped class is not a Model, first try the non-Model wrapped class from `WrappedModelLocator::locateValueObject($resourceClass)` and run it through `PublicPropertyTypeReader`. Keep the Model path first (no regression); fall through to value-object only when the Model lookup yielded nothing. The `$this->field` / `$this->resource->field` receiver detection already exists via `modelFieldName()`.
   - **Caching/threading note:** `read()` memoises per resource class — fine; the new lookups are deterministic per class. The internal `fieldsFromArrayNode`/`resolveValue` signatures gain the resource `ReflectionClass` (or class-string) alongside the existing `?modelClass` parameter. Keep it minimal: thread one extra nullable param, don't restructure the walker.

   No change to `InferredResourceField`, `SchemaFromArrayDefinition`, or the registry — the new fields are ordinary `ofProperty(...)` results, identical in shape to the model path's output.

### Tests (TDD — write the failing tests first)

New fixtures under `tests/Fixtures/Resources/` + `tests/Fixtures/Values/`:

- `tests/Fixtures/Values/GenreSummaryValue.php` — a `final readonly` value object with typed public props: `string $publicId`, `string $name`, `int $songCount`, `float $length`, `?string $note` (nullable scalar → modelled nullable), and a `int|string $mixedKey` (union → refuse).
- `tests/Fixtures/Resources/ValueObjectResource.php` — wraps it via a typed `public GenreSummaryValue $summary` (promoted ctor param) and a `toArray()` single-return literal: `['type' => 'genres', 'id' => $this->summary->publicId, 'name' => $this->summary->name, 'song_count' => $this->summary->songCount, 'length' => $this->summary->length, 'note' => $this->summary->note, 'mixed' => $this->summary->mixedKey, 'untyped' => $this->summary->somethingUntyped]` — mirrors the Koel evidence + adds refuse cases.
- A second fixture exercising **shape (B)**: a resource whose `@mixin`/`@extends` names a non-Model value object and uses `$this->field` / `$this->resource->field`.

Feature tests in **`tests/Feature/Plugins/ApiResources/ResourceToArrayInferenceTest.php`** (extend the
existing file; reuse its signature-only controller + `generateSpec()`/schema-factory pattern):

1. **Happy path (shape A):** `id`→`string`, `name`→`string`, `song_count`→`integer`, `length`→`number`. (Mirrors Koel acceptance.)
2. **Nullable scalar:** `note` (`?string`) → `string` + nullable.
3. **Literal still wins:** `type` → `string` from the `'genres'` literal (unchanged).
4. **Refuse — union property:** `mixed` (`int|string`) stays unconstrained (no `type`, no `oneOf`).
5. **Refuse — absent/untyped property:** `untyped` (property doesn't exist OR has no declared type) stays unconstrained.
6. **Refuse — receiver not a known single class:** a resource whose wrapped prop is untyped / union / non-existent class → those fields stay unconstrained.
7. **Shape (B):** non-Model `@mixin`/`@extends` + `$this->field` → typed; Model path unaffected.
8. **No regression:** an existing Model-backed fixture (`InferredArticleResource`) still types its fields (run the existing assertions; add one guarding that the Model path is tried before the value-object path).

Unit tests for the new primitive in **`tests/Unit/Support/Extraction/PublicPropertyTypeReaderTest.php`**: scalar, nullable scalar, enum (if cheap), union→null, intersection→null, no-type→null, non-public→null, missing→null, non-existent class→null.

Also extend **`tests/Unit/Plugins/ApiResources/WrappedModelLocatorTest.php`** for the new non-Model locator: returns the value-object class for a non-Model `@mixin`/`@extends`, null for a Model (that's the existing `locate`'s job), null when absent.

Survey: this is generation-affecting (area:plugins/responses) → the survey gate applies. Acceptance explicitly says "survey shows no spec regressions" and the metric % is unchanged (fidelity-only). Coder/reviewer should run the survey and confirm no regressions and ideally the Koel GenreResource fields gain types.

### Docs

`docs/auto-derivation.md` — the API-Resource `toArray()` paragraph (the Response-body row at line 16 and any JsonResource subsection) currently says fields are typed from the wrapped `@mixin`/`@extends` **model**. Extend it: when the resource instead wraps a **non-Model value object** with statically typed public properties, `$this->wrapped->field` and `$this->field` are typed from that object's property types; untyped or union-typed properties stay unconstrained. Cross-reference [Plugins → ApiResources](plugins.md#apiresources) if that page enumerates the toArray() behaviour; mirror the wording there if so.

### CHANGELOG

`[Unreleased]` (Added or Improved):
> API Resource `toArray()` fields that read a wrapped **non-Model value object**'s statically-typed public properties (`$this->wrapped->field` / `$this->field`) are now typed from those property types, instead of falling through to an untyped property. Untyped or union-typed source properties stay unconstrained (never-wrong).

### Lint rule

None added. (No new finding; this only improves types on already-emitted properties.)

- **Low controversy.** No `src/Contracts/**` change, no stage-order change, no config key, no public authoring-attribute signature change. New `Support/` class is `@internal`; `WrappedModelLocator` and `ResourceToArrayReader` are `@internal` plugin internals.
- The one design choice worth a reviewer's nod: **new `Support/PublicPropertyTypeReader`** vs. adding a method to `EloquentModelToSchema` (recommendation: new class, see Files §1) — and the union **refusal** policy (we deliberately do NOT emit `JsonSchemaFromType`'s `oneOf` for unions here, to honour "never-wrong"; that diverges from how the symfony-type mapper behaves elsewhere — intentional and tested by refuse-case #4).
- Generation-affecting → **survey required** before merge (per role/CLAUDE.md); acceptance requires no spec regressions.


Radiergummi:

Two grounding notes from reading the code: (1) The Koel case (`$this->summary->publicId`) is a *different receiver shape* than the existing model path — `modelFieldName()` only matches `$this->field`/`$this->resource->field`, so the value-object case needs a new nested-PropertyFetch resolver (shape A), plus the `$this->field` non-Model-wrapped case (shape B). (2) `EloquentModelToSchema` exposes no reusable "type a property by reflection" method, so the plan adds a small plugin-agnostic `Support/PublicPropertyTypeReader` reusing the same `TypeResolver`→`JsonSchemaFromType` path the model accessor path already uses, with refuse-gates for unions/untyped/non-public. Low controversy; generation-affecting so the survey gate applies (acceptance requires no spec regressions).

Radiergummi:
🤖 **[lead]** plan posted on draft PR #423 → plan-review. Tier-0 reflection fidelity fix; new plugin-agnostic Support/PublicPropertyTypeReader + WrappedModelLocator non-Model sibling; union/intersection/untyped REFUSED (never-wrong). Reviewer: weigh new-Support-class vs extending EloquentModelToSchema, and the refuse-gate coverage.

Radiergummi:
🤖 **[reviewer]** plan-review: approved with one required wiring clarification (non-blocking on the design; must be reflected before coding).

Verified all plan claims against the code — receiver shapes (`modelFieldName` only matches `$this->field`/`$this->resource->field`, `ResourceToArrayReader.php:479`), the `WrappedModelLocator` Model-filter (`modelClassOf`), and the reuse path (`TypeResolver->resolve($reflectionType)` → symfony `Type` → `JsonSchemaFromType::fromType()`, exactly `EloquentModelToSchema::propertyFromAccessor` :451-457). `JsonSchemaFromType::fromType()` does emit `oneOf` for `UnionType` (:60) and "Unmapped type" fallbacks (:96/:219/:244), and handles `NullableType` cleanly (:56) — so the refusal gates and the "nullable scalar stays modelled" claim are correct.

**On the two items flagged for review:**

1. **New `Support/PublicPropertyTypeReader` vs. extending `EloquentModelToSchema`: agree with the new class.** `EloquentModelToSchema` is Model-specific (casts/`@property`/hidden-visible/factory examples, depends on `ComponentSchemaRegistry` + `ModelFactoryExampleReader`). Value-object public-property reflection is a distinct, smaller concern, and CLAUDE.md requires generic reflection infra to live in `Support/` and stay Model-free. Adding a method to the Model class would couple the two. New class is correct.

2. **Refuse-gate coverage is complete.** no-type / union / intersection / non-public / non-static / missing / non-existent class / non-single-class receiver all map to the issue’s refusals, and the gate correctly intercepts *before* `fromType()` (verified that fromType would otherwise emit `oneOf`/unmapped). Refuse-test cases #4 (union) and #5 (untyped/absent) pin this.

**REQUIRED clarification before coding (correctness, not design):**

The shape-(B) wiring as written ("fall through to value-object only when the Model lookup yielded nothing") will not work against the current control flow. `resolveModelProperty()` returns a **non-null `unconstrained()`** (not null) whenever `modelFieldName()` matches but the property lookup fails (`ResourceToArrayReader.php:471` branch). `resolveValue()` treats any non-null return as final (`:289-291`), so a value-object attempt placed *after* `resolveModelProperty` is unreachable for shape (B). The plan must specify that the shape-(B) value-object lookup happens **inside `resolveModelProperty` before its `unconstrained` return** (i.e. when `$property === null`, try `WrappedModelLocator::locateValueObject` + `PublicPropertyTypeReader` before falling to `unconstrained`), OR that `resolveModelProperty` is changed to return `null` (not `unconstrained`) on a property miss so the value-object step downstream can run. Either is fine; the plan should name which, since it changes where the new call sits and the test for "Model path tried first" (#8) depends on it.

**Survey criterion (confirm, not block):** acceptance is "no spec regressions" and the metric % is unchanged (fidelity-only) — the survey gate must be read as *no regressions*, and ideally the Koel `GenreResource` fields gain types; a flat object-payload % is the expected, passing outcome here, not a failure.

Docs (`auto-derivation.md:16` + the `JsonResource` subsection), CHANGELOG, and "no lint rule" are correctly named. With the shape-(B) wiring point made explicit, this is ready to implement.

Radiergummi:
🤖 **[lead]** plan-review approved w/ one required fix (reviewer-1) → coding. REQUIRED: shape-(B) value-object lookup must happen INSIDE resolveModelProperty before its unconstrained() return (or make that method return null on miss) — a value-object attempt placed AFTER resolveModelProperty is dead code since it returns non-null unconstrained(). Test #8 (Model-path-tried-first) depends on the chosen wiring. Survey gate = no-regressions (fidelity-only, flat object-payload % is a pass).

Radiergummi:
🤖 **[coder]** wiring decision (required by plan-review): chose **option (a)** — the shape-(B) value-object lookup happens **inside `resolveModelProperty`, before its `unconstrained()` return**. The Model lookup runs first (`propertyFor($modelClass, …)`); only on a Model miss (`$property ??= valueObjectProperty(…)`) does it try the non-Model wrapped class via `WrappedModelLocator::locateValueObject` + `PublicPropertyTypeReader`. `resolveModelProperty` keeps returning a non-null `unconstrained()` on a total miss, so its position relative to the literal fallback is unchanged. Test #8 (`tries the Model path before the value-object path with no Model regression`) passes under this wiring: the InferredArticleResource Model fields keep their model-derived types, never shadowed by the value-object path.

Shape (A) (`$this->wrapped->field`) is a separate resolver step placed before `resolveModelProperty`, since `modelFieldName()` returns null for that receiver shape anyway (the inner receiver is `$this->summary`, not `$this`/`$this->resource`) — no collision.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed 7a4557be. Ready for review.

Radiergummi:
🤖 **[lead]** real diff verified (sha 7a4557be; 14 files, src production +342/-32), full CI matrix green incl. swagger-php 5.8 → code-review. NOTE: src diff >150 lines → will need human gate on the SIZE criterion at merge (change is otherwise non-controversial in kind: no Contracts/config/stage-order, additive new methods, Support plugin-free). Running full gates first (dual review + survey) so the human gets a vetted package. Two independent reviewers.

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer B, independent) — 1 finding (changes-needed) + 1 minor test-gap.

Reviewed sha `7a4557be` adversarially and on my own merits. The new feature is well-built: both receiver shapes are correctly guarded, the Support boundary holds, the refuse-gates are sound, and `isModelledType` mirrors `JsonSchemaFromType` exactly (verified: object whitelist = DateTimeInterface/UuidInterface/UrlRoutable, builtin whitelist = string/int/float/bool/array — both match the live dispatch; `BackedEnumType` checked before `ObjectType` preserving order; non-backed `EnumType` correctly refuses rather than taking `JsonSchemaFromType`'s apology-string placeholder — strictly never-wrong). Model-path-first is double-guarded (shape B only via `$property ??=` on a Model miss, and `locateValueObject` returns null when the wrapped class IS a Model), so test #8 genuinely holds. Boundary: `PublicPropertyTypeReader` is in `Support/`, Model-free and plugin-free; its `Illuminate\Contracts\Routing\UrlRoutable` / `Ramsey\Uuid\UuidInterface` imports are not convention-package deps and are already present in the sibling `JsonSchemaFromType` it mirrors — consistent, not a violation. `@internal` + strict-types + no-abbrev all present.

**Finding 1 (changes-needed) — `WrappedModelLocator::locate()` regresses for a non-Model `@mixin` combined with a Model `@extends`.** `src/Plugins/ApiResources/Support/WrappedModelLocator.php:93-130`. The refactor split out `resolveClassName()`, which now returns the **first existing class** from the `@mixin` loop *regardless of kind*, and `locate()` post-filters it with `modelClassOf()`. Previously `resolve()` filtered to Model **inside** each loop, so a non-Model `@mixin` was skipped and the loop fell through to `@extends`. Consequence: a resource declared

```php
/**
 * @mixin SomeNonModelTrait        // or any non-Model class
 * @extends Base<SomeModel>
 */
```

now has `locate()` return **null** (resolveClassName returns the non-Model mixin, modelClassOf rejects it, the @extends Model is never reached) where it previously returned `SomeModel` — losing the entire model-derived schema. That's a fidelity regression against the stated no-regression gate. It is not pinned in either direction: `BothTagsResource` (the "prefers @mixin over @extends" test) has **two Models**, so it doesn't exercise this; `NonModelMixinResource` has no `@extends`. Fix options: (a) have `resolveClassName` take a classifier/predicate so `locate()` resolves the first *Model* across both tag types (preserving old precedence) while `locateValueObject` resolves the first *non-Model*; or (b) if mixing a non-Model `@mixin` with a Model `@extends` is deemed unsupported, add a fixture+test that documents and pins that null and note the deviation as a PR comment. Either way it needs a test for the non-Model-`@mixin` + Model-`@extends` combination.

**Finding 2 (minor, non-blocking) — no test for a unit (non-backed) enum property.** `tests/Unit/Support/Extraction/PublicPropertyTypeReaderTest.php`. The reader refuses a non-backed enum (it isn't a `BackedEnumType`, and as an `ObjectType` it fails the three `is_a` checks), which is the correct never-wrong behaviour and is *better* than the accessor path's apology-string. Worth a one-line refusal case (`public SomePureEnum $kind` → `propertyFor(...)` is null) alongside the existing union/intersection/untyped/non-public/static/absent dataset, to lock that the value-object path never emits the unit-enum placeholder. Cheap; pins a deliberate divergence from `JsonSchemaFromType`.

Everything else (shape-A `$this->resource->field` falling through to the model path because `JsonResource::$resource` is untyped; the fresh-`OA\Property`-per-call mutation in `resolveValueObjectProperty`; CHANGELOG + docs/plugins.md + docs/auto-derivation.md updates) checks out.

Verdict: **changes-needed** on finding 1 (no-regression gate). Routing to coder-3.

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer A, independent): **approved, no concerns.**

I plan-reviewed this; I did not author the diff. Reviewed sha `7a4557be` on its merits in an isolated worktree.

**My required plan-review fix is honoured (and cleanly).** Shape (B) runs **inside** `resolveModelProperty` as `$property ??= $this->valueObjectProperty($resourceClass, $fieldName)` (`ResourceToArrayReader.php:514`) — after the Model lookup, before the `unconstrained()` return. The value-object path is reachable, not dead code, and the Model path is genuinely tried first. Better still, "Model first" is **structurally guaranteed**, not just ordering-dependent: `WrappedModelLocator::locateValueObject()` returns null for a Model-wrapped resource (`valueObjectClassOf` excludes `is_a(Model)`), so the fallback is inert whenever a Model is present. Test #8 + the locator unit test "returns null when the wrapped class is a Model" pin this.

**Refuse-gates — tried to slip a wrong schema through, could not.** `PublicPropertyTypeReader::isModelledType()` is the one risk surface (it duplicates `JsonSchemaFromType`'s mapping knowledge to decide refuse-vs-emit). I checked parity against `JsonSchemaFromType` directly:
- Objects: reader whitelists exactly `DateTimeInterface` / `UuidInterface` / `UrlRoutable` — the exact set `fromObjectType()` maps concretely; everything else → placeholder there, refused here. Match.
- Builtins: `MODELLED_BUILTINS = [string,int,float,bool,array]` = exactly what `fromBuiltinType()` maps; other builtins → placeholder there, refused here. Match.
- BackedEnum → `$ref` (emitted). Match.
- The reader's whitelist is a **subset** of what JsonSchemaFromType maps concretely, so it can only ever be *more* conservative — **no placeholder/`oneOf`/wrong schema can leak**. The one intentional under-coverage: a **unit enum** (`EnumType`) — `JsonSchemaFromType` would emit `{type: string, description: …}`, but the reader refuses it (only `BackedEnumType` is whitelisted). That is a *conservative refusal* (stays unconstrained), fully consistent with never-wrong; arguably correct since a unit-enum value isn't a string. Noting it as an observation, not a finding.

**Verified:** 18 feature + 22 unit tests pass (incl. shape A happy/nullable/union-refuse/absent-refuse/untyped-receiver-refuse, shape B @mixin, Model-first regression). TDD confirmed — reverting the 3 src files fails 17 new tests, leaves the 15 pre-existing Model-path tests green (no regression). Pint passed, PHPStan No errors on all 3 changed src files, ApiResources regression band 119 passed.

**Boundaries:** `PublicPropertyTypeReader` lives in `Support/` and is genuinely **Model-free and plugin-free** — grep shows zero `Eloquent`/`Model`/`Plugins\` imports; it depends only on `Support\Generator\JsonSchemaFromType` + symfony `TypeResolver`. `WrappedModelLocator` refactor keeps `locate()` Model-filtered (the shared `resolveClassName()` is unfiltered, `modelClassOf`/`valueObjectClassOf` partition Model vs non-Model — mutually exclusive). No Contracts/stage-order/config change.

**Edges checked:** shape-B via `@extends` shares `resolveClassName()`'s `@mixin`→`@extends` path (the `@mixin` fixture exercises it; `@extends` is the same code), so covered by construction. Self-referential value object: the reader does not recurse into nested object properties (one level; non-whitelisted objects refuse), so no infinite recursion. Nullable scalar → `['string','null']` (tested). Intersection → refused (tested).

**Bookkeeping:** CHANGELOG `[Unreleased]` (#411), `docs/auto-derivation.md` Response-body row, and `docs/plugins.md` ApiResources section all updated, honestly framing the never-wrong refusal + Model-first ordering. No new lint rule needed. Comments are concise, why-focused, no issue numbers, no em-dashes; strict-types + `@internal` present; no abbreviations.

Verdict: **approved.** Clean Tier-0 fidelity change. Sole note (not blocking): the src diff is >150 lines, so the lead's SIZE human-gate at merge applies regardless of review.

Radiergummi:
🤖 **[lead]** review round 1 (reviewer-2): 1 substantive finding → back to coder (1/3). WrappedModelLocator.locate() regresses for non-Model @mixin + Model @extends combo (resolveClassName returns first existing class regardless of kind; locate()'s modelClassOf post-filter then returns null where old in-loop filtering fell through to the Model @extends). Fix: classify inside resolveClassName so locate() finds first Model across both tags, locateValueObject first non-Model; add combo fixture. Plus optional non-backed-enum refusal unit test. ALSO: main advanced (ef40e2c5) after #409 merge — rebase onto origin/main (CHANGELOG [Unreleased] conflict expected; keep both lines).

Radiergummi:
🤖 **[coder]** addressed round 1: pushed d01e6b35.

**Finding 1 (regression, fixed):** `resolveClassName()` now takes a `callable(class-string): bool $matches` predicate and applies it *inside* both the `@mixin` and `@extends` loops. `locate()` passes a Model predicate, `locateValueObject()` a non-Model predicate. So the old per-loop filtering (and tag precedence) is restored: a non-Model `@mixin` no longer shadows a Model `@extends` (and vice-versa). Removed the now-redundant `modelClassOf`/`valueObjectClassOf` post-filters. Added two regression fixtures + tests:
- `NonModelMixinModelExtendsResource` (@mixin DocBlockParser + @extends Base<Article>) → `locate()` returns `Article`.
- `ModelMixinNonModelExtendsResource` (@mixin Article + @extends Base<DocBlockParser>) → `locateValueObject()` returns `DocBlockParser`.

**Finding 2 (minor, done):** added a unit (non-backed) enum refusal case to PublicPropertyTypeReaderTest's dataset (`public UnitFixtureEnum $kind` → null), locking that a unit enum refuses rather than emitting JsonSchemaFromType's apology-string placeholder.

Local gates green (Pint + PHPStan + 3045 tests); remote CI green incl. swagger-php 5.8 and the full matrix (re-triggered via close/reopen since the push didn't fire Actions). Ready for re-review.

Radiergummi:
🤖 **[reviewer]** review round 2 (reviewer B): approved, no concerns.

Re-reviewed sha `d01e6b35`. Both round-1 findings are resolved, and both are pinned by tests that would fail without the fix.

**Finding 1 (the regression) — fixed correctly.** `resolveClassName()` now takes a `callable(class-string): bool $matches` applied **inside** both the `@mixin` and `@extends` loops; `locate()` passes a Model predicate and `locateValueObject()` a non-Model predicate. This restores the old per-loop filtering, so a non-Model `@mixin` no longer short-circuits before a Model `@extends` (and the inverse for value objects). The dead `modelClassOf`/`valueObjectClassOf` post-filters are removed; the `class_exists` guard is retained inline. Both directions are pinned:
- `NonModelMixinModelExtendsResource` (`@mixin DocBlockParser` + `@extends Base<Article>`) → `locate()` === `Article` — exactly the regression case I flagged.
- `ModelMixinNonModelExtendsResource` (`@mixin Article` + `@extends Base<DocBlockParser>`) → `locateValueObject()` === `DocBlockParser` — the inverse.
The pre-existing precedence tests (`prefers @mixin over @extends` on the two-Model `BothTagsResource`, the `@mixin`-only and `@extends`-only cases) still hold.

**Finding 2 (minor) — addressed.** A unit (non-backed) enum refusal case (`public UnitFixtureEnum $kind` → null) is added to the `PublicPropertyTypeReader` refusal dataset, with a comment on the rationale. Verified `UnitFixtureEnum` is a genuine non-backed enum, so it lands in `isModelledType`'s `ObjectType` branch, fails the DateTime/Uuid/UrlRoutable whitelist, and is refused — locking that the value-object path never emits the apology-string placeholder.

Scope is surgical (3 files: the locator + its two test files), the earlier-approved feature code is untouched, and the coder reports local gates + the full remote CI matrix (incl. swagger-php 5.8) green on `d01e6b35` (re-triggered SHA-preserving via close/reopen).

Verdict: approved. Both reviewers' findings closed — over to the lead for the merge decision (this remains the >150-line-src PR flagged for the human SIZE gate, plus the survey gate).

Radiergummi:
🤖 **[lead]** round 1 addressed (sha d01e6b35) → re-review. reviewer-2's finding was a REAL regression — fixed via predicate-based resolveClassName + 2 combo fixtures. NOTE: (1) reviewer-1's prior approval was on the buggy 7a4557be and is STALE — both reviewers must re-confirm on d01e6b35; (2) branch NOT yet rebased (merge-base still 4528b234, main is ef40e2c5) — rebase + re-green required before merge/escalation; (3) latest CI run still in-progress. Will run survey after approvals; escalates on SIZE gate (>150 src lines).

Radiergummi:
🤖 **[coder]** addressed round 1: pushed c8c6862d (rebased onto origin/main @ ef40e2c5).

**Combo verdict — reviewer-2 was right.** The non-Model `@mixin` + Model `@extends` combo *was* a real regression: the refactor's `resolveClassName()` returned the first existing class regardless of kind, so a non-Model `@mixin` made `locate()` return null instead of falling through to the Model `@extends`. Fixed in the prior push (and now rebased): `resolveClassName()` takes a `callable(class-string): bool` predicate applied *inside* both loops, so `locate()` finds the first **Model** across `@mixin`+`@extends` and `locateValueObject()` the first **non-Model**, preserving tag precedence. The dead `modelClassOf`/`valueObjectClassOf` post-filters are gone.

The contested combo is now pinned both directions:
- `NonModelMixinModelExtendsResource` (`@mixin DocBlockParser` + `@extends Base<Article>`) → `locate()` returns **Article**.
- `ModelMixinNonModelExtendsResource` (`@mixin Article` + `@extends Base<DocBlockParser>`) → `locateValueObject()` returns **DocBlockParser**.

Also added the non-backed-enum refusal unit case (`public UnitFixtureEnum $kind` → null).

Rebase was clean (no CHANGELOG conflict — my #411 line and #409's sit in different subsections). Local gates green (Pint + PHPStan + 3054 tests). Remote CI green on c8c6862d incl. swagger-php 5.8 + full matrix (re-triggered via close/reopen; the force-push didn't fire Actions).

Radiergummi:
🤖 **[reviewer]** re-review: **approved** (on corrected sha `d01e6b35`; supersedes my stale `7a4557be` approval).

First — I missed the regression reviewer-2 caught on `7a4557be`. The old `resolveClassName()` returned the first `@mixin`-then-`@extends` class *unfiltered*, then a post-filter (`modelClassOf`/`valueObjectClassOf`) accepted-or-nulled it; so a **non-Model `@mixin` + Model `@extends`** resource returned the non-Model `@mixin` first, the Model post-filter nulled it, and `locate()` returned null instead of falling through to the Model `@extends`. My earlier "no concerns" was wrong on that; good catch by reviewer-2.

**The fix is correct in both directions.** `resolveClassName(string, callable $matches)` now applies the predicate **inside** both loops, so selection and filtering happen at the same point and tag precedence is preserved:
- **Direction 1 (the bug):** non-Model `@mixin` + Model `@extends` → `locate()` (Model predicate) skips the non-Model `@mixin`, matches the Model `@extends` → returns the Model. Regression fixed. `locateValueObject()` (non-Model predicate) returns the non-Model `@mixin`.
- **Direction 2 (mirror):** Model `@mixin` + non-Model `@extends` → `locate()` returns the Model `@mixin`; `locateValueObject()` skips it and returns the non-Model `@extends`.

**Partition is exhaustive and exclusive by construction** — the two predicates are `is_a(Model)` and `!is_a(Model)`; a class is one xor the other, so nothing slips through to the wrong locator. An abstract non-Model class satisfies the value-object predicate and is then reflected by `PublicPropertyTypeReader` (still never-wrong — typed public props or refuse); an interface is excluded earlier by `class_exists` being false. The two caches (`$cache`, `$valueObjectCache`) stay separate, so predicate-keyed results don't collide. The dead post-filters are correctly removed.

**Verified empirically (sha d01e6b35, isolated worktree):**
- The two new combo fixtures pin both directions exactly: `NonModelMixinModelExtendsResource` → `locate()` === `Article`; `ModelMixinNonModelExtendsResource` → `locateValueObject()` === `DocBlockParser`.
- TDD-for-the-fix holds: restoring the buggy `7a4557be` locator under the new tests makes **both** combo tests fail; the corrected locator makes all 10 locator + 15 reader tests pass (25 total).
- Pint passed, PHPStan No errors on the file, ApiResources regression band **121 passed** (was 119; +2 combo tests).

The rest of the PR (PublicPropertyTypeReader refuse-gates, the Support/ boundary, shape-A/B wiring, my required Model-first fix) I vetted on `7a4557be` and the delta does not touch it — that verdict stands. Confirmed the lead's note that the concurrent rebase onto `main` (ef40e2c5) touches only `CHANGELOG.md`, so this logic verdict carries to the rebased sha.

Verdict: **approved.** This is now a fresh, valid approval on the corrected diff — counts toward the merge gate. The >150-line SIZE human-gate and the survey gate remain the lead's.

Radiergummi:
🤖 **[lead]** controversial on SIZE → human gate (NOT auto-merged). Reason: src production diff is ~342 lines (>150 threshold). Otherwise non-controversial in kind: no Contracts/config/stage-order change, additive new methods, Support stays plugin-free. FULLY VETTED PACKAGE: (1) two independent fresh approvals on the corrected sha — reviewer-2 caught a real locate() regression (non-Model @mixin + Model @extends), coder-3 fixed it via predicate-based resolveClassName with 2 combo fixtures pinning both directions, reviewer-1 re-reviewed and owned its earlier miss; (2) full CI matrix green incl. swagger-php 5.8 on rebased sha c8c6862d (merge-base = current main ef40e2c5); (3) survey Layer A pass (surveyor: 0 regressions, fidelity-only additive change). Only the size decision remains — Moritz's call. PR left ready, unmerged, worktree in place.

PR #424, opened 2026-06-20
feat(#415): capture @property trailing description into model property description

### LINKED ISSUE #415: feat(responses): capture @property trailing description into model property description
## Problem

`@property Type \$name Trailing description.` is the idiomatic place to document model attributes,
but the trailing description is discarded — only the type is read. Field-level description coverage
is **5%** in the corpus; authored `@property` prose is the most defensible never-wrong source and is
thrown away today.

## Evidence

- `@property`/`@property-read` parsed into `PropertyTagValueNode` but only `$tag->type` is consumed: `src/Support/Extraction/EloquentModelToSchema.php:217-227` (and `:109`).
- The string `description` appears **nowhere** in that file (`grep -n description` → empty), so `PropertyTagValueNode->description` is never read.

## Proposal (Tier-0)

When building a property from a tag, if `$tag->description` is non-empty, set
`OA\Property->description` — only when not already set by a higher-precedence authored source
(attributes win). Pure docblock read, deterministic.

## Scope / Out of scope

- In: both `propertyFor()` and `schemaFor()` paths; trim whitespace; skip empties.
- Out: synthesizing descriptions from column names (a smell); migration `->comment()` (folds into the migration-reader issue).

## Acceptance

`@property string \$email The primary contact email.` emits that string as the property description; a tag with no trailing text emits none; an attribute-supplied description is not overwritten.

<sub>From the 2026-06-20 lever audit.</sub>

Radiergummi:
Closes #415

### Tier & rationale

**Tier 0** — pure docblock read. The `@property`/`@property-read` tags are already parsed into
`PropertyTagValueNode`s (which carry both `->type` and `->description`); we just stop discarding the
trailing `->description`. Deterministic, no body parsing. Most-defensible never-wrong description
source; field-description coverage is 5% in the corpus.

`metadataFor()` (`EloquentModelToSchema.php:217-227`) stores the full `PropertyTagValueNode` per
name in `$metadata['propertyTags']`, so the description is *available* at every consumption site —
it's just dropped at the call boundary: both `propertyFor()` (`:128`, `:139`) and `schemaFor()`
(`:569`, `:578`) call `propertyFromTag($name, $tag->type, $reflection)`, passing only `$tag->type`.
`grep -n description` over the file is empty, confirming `->description` is never read.

The tag also reaches the **cast** branches: `castToProperty($name, $castString, $tag?->type)`
(`:119`, `:565`) — where a column has both a cast and a documented `@property`. And the
backed-enum, timestamp, and untyped-fallback branches each emit a property while a `@property` tag
may still exist for that name. The description should attach to the named property **whenever a tag
with trailing prose exists**, regardless of which builder produced the schema — that is the
"both paths, all branches" intent.

### Design — one chokepoint helper, applied after each property is built

Add a single private helper and call it on the built `OA\Property` in both methods:

```php
/** Sets the tag's trailing prose as the description, unless one is already present. */
private function applyTagDescription(OA\Property $property, ?PropertyTagValueNode $tag): OA\Property
{
    $description = trim((string) ($tag?->description ?? ''));

    if ($description !== '' && (is_undefined($property->description) /* or === null */)) {
        $property->description = $description;
    }

    return $property;
}
```

- **Trim + skip-empty:** `trim()` then `=== ''` guard (the issue's "trim whitespace, skip empties").
- **Don't overwrite ("attributes win"):** only set when the property has no description yet. The
  attribute-wins precedence is preserved structurally because attribute composition
  (`#[ResourceField]` etc.) happens **downstream** in `SchemaFromResource` on its own `OA\Property`,
  not by overlaying this model property. Confirm the correct "unset" sentinel for swagger-php
  `OA\Property->description` — it is the `Generator::UNDEFINED` non-empty-string sentinel, so test
  with `is_undefined(...)` (see the project memory on `UNDEFINED` being a string), **not**
  `=== null`. The coder must verify which holds for a freshly-built `OA\Property(['property' => …])`
  and gate accordingly; getting this wrong is the one real footgun here.
- **reviewer-2 correction — the tag is NOT the only in-file description source.** The plan said "the
  only description source is the tag"; that is wrong. `propertyFromSchema()` (`:255`) builds via
  `copy_schema_fields($schema, new OA\Property(...))`, which copies **every** set field from the
  schema onto the property — including a `description` that `JsonSchemaFromType` already attached: a
  **documented backed-enum**'s case-list description (`enumCaseDescription`, set on the schema at
  `JsonSchemaFromType.php:154`) and the unmapped-type/object/builtin apology strings. So on the
  backed-enum-from-cast and tag-backed-enum branches, the property may **already** carry a
  description before `applyTagDescription` runs. This is exactly why `is_undefined` (not `=== null`)
  matters, and it makes a precedence decision: **a documented-enum case description pre-empts the
  `@property` trailing prose** (the guard sees the description already set and skips). That ordering
  is reasonable (the enum case list is more specific than free prose) and is the *only* sensible
  outcome of a no-overwrite guard — but it must be a **deliberate, tested** decision, not an
  accident. Add a test (below): a column that is a documented backed-enum cast **and** carries a
  described `@property` keeps the enum-case description, not the tag prose. (Acceptance #415 is
  silent here; this records the chosen precedence.)

**Application points** (so it covers every branch in both methods):

- `propertyFor()` (`:100-169`): wrap each `return`/built property with `applyTagDescription(..., $tag)`
  for the branches where `$tag` is in scope — the cast→castToProperty result, the unrecognised-cast
  tag fallback, the backed-enum-from-cast property, the plain tag property, the untyped
  `new OA\Property(['property' => $propertyName])` fallback (when a tag exists), and the timestamp
  branch if a `@property` documents the timestamp. Cleanest: compute `$tag` once (already done at
  `:109`) and funnel every non-null return through the helper. The relation/append-accessor branches
  have no `@property` tag by definition (or use a different name) — pass the looked-up `$tag` which
  may be null (helper no-ops).
- `schemaFor()` (`:547-605`): **reviewer-2 refinement — prefer a single post-loop pass over touching
  the six `$properties[] = …` branches.** `schemaFor` already has two post-loop passes that iterate
  `foreach ($properties as $property)` keyed by `$property->property`: the `$examples` pass (`:608`)
  and the `$required` re-lookup (`:618`). Add a third pass in the same shape:
  ```php
  foreach ($properties as $property) {
      $this->applyTagDescription($property, $propertyTags[$property->property] ?? null);
  }
  ```
  This covers every branch (each built property carries its own `->property` name and the tag is
  reachable by name), matches the file's established idiom, and touches **zero** of the `continue`
  branches — strictly lower-touch and less error-prone than threading the call through each emission.

- `propertyFor()` (`:100-169`): `$tag` is in scope throughout but there are ~8 distinct returns.
  **reviewer-2 refinement — one chokepoint, not eight edits:** rename the current body to a private
  `buildPropertyFor(...)` and have the public method decorate its single result:
  ```php
  $property = $this->buildPropertyFor($modelClass, $propertyName, $tag, $metadata);
  return $property === null ? null : $this->applyTagDescription($property, $tag);
  ```
  (null still propagates as today.) Avoids scattering `applyTagDescription` across every return and
  keeps the null-on-total-miss contract intact.

Keep it surgical: the two chokepoints above, the helper, and the tag re-lookup pass — no broader
control-flow restructuring.

### Files to modify

1. **`src/Support/Extraction/EloquentModelToSchema.php`** — add the helper; thread `$tag` (already in
   scope) through the built properties in `propertyFor()` and `schemaFor()`. Add the
   `PropertyTagValueNode` import if not already present (it is — used in `metadataFor`'s typehints),
   and `is_undefined` from the package helpers (verify import).

No other production files. No `Contracts`/signature/config change.

### Tests (TDD — write first)

New fixture **`tests/Fixtures/Models/DescribedArticle.php`** — a model whose `@property` tags carry
trailing prose, plus the edge cases:
```php
/**
 * @property string $email        The primary contact email.
 * @property int    $login_count  Times the user has signed in.
 * @property string $name                                 (no trailing text → no description)
 * @property string $title           Surrounded by spaces.   (whitespace trim)
 * @property-read string $slug      URL-safe identifier.       (@property-read path)
 */
```
Include at least one column that is **both cast and documented** (e.g. a `datetime`/enum cast with a
described `@property`) to prove the description attaches on the cast branch too.

Unit tests:

- **`tests/Unit/Support/Extraction/EloquentModelPropertyLookupTest.php`** (the `propertyFor` path):
  1. `@property string $email The primary contact email.` → `description === 'The primary contact email.'`.
  2. Tag with no trailing text → no `description` key emitted.
  3. Whitespace-only / surrounding whitespace → trimmed (`'Surrounded by spaces.'`), and a
     whitespace-only description emits nothing.
  4. `@property-read` described tag → description captured (same code path, but assert it).
  5. Cast + described tag → description present on the cast-derived property (proves cast branch).
  6. **Attribute-wins / no-overwrite:** a property that already has a description is not overwritten.
     Within this unit the cleanest way to prove the guard is a direct helper test or a property
     pre-seeded with a description; if not reachable in isolation, assert via the downstream
     ApiResources path (below) that a `#[ResourceField(description: …)]` wins over the `@property`.
  7. **reviewer-2 — documented-enum precedence (pins the no-overwrite guard against a real in-file
     source):** a column that is a documented backed-enum cast (its enum cases carry PHPDoc, so
     `enumCaseDescription` populates the schema description, copied onto the property by
     `copy_schema_fields`) **and** also has a described `@property` tag → assert the property's
     `description` is the **enum case list**, not the tag prose. This proves `is_undefined` correctly
     sees the pre-set description and skips, and locks the chosen precedence. (A `=== null` guard
     would *also* skip here since the sentinel is a string, but it would WRONGLY skip the
     plain-tag/untyped branches — so keep test #1/#5 to catch that direction too.)
- **`tests/Unit/Support/Extraction/EloquentModelToSchemaTest.php`** (the `schemaFor` path): the
  described-tag → `properties.<name>.description` assertion; no-trailing-text → none; and the
  cast+described case. Mirrors the propertyFor assertions on the full-schema output.

Edge cases explicitly covered: no-trailing-text → no description (acceptance); whitespace trim;
both `@property` and `@property-read`; both `propertyFor` and `schemaFor`; cast-branch; and the
attribute-not-overwritten precedence.

Regression: the existing `Article` fixture has no described tags, so existing assertions are
unaffected; run the full `EloquentModelToSchemaTest` / `EloquentModelPropertyLookupTest` to confirm.

### Docs

`docs/auto-derivation.md` — the model `@property`/`@property-read` bullet (around line 152, item 2 in
the Eloquent model-schema section) lists what the tag contributes (types, nullability, model `$ref`,
array shapes). Add that the **trailing description** of a `@property` tag becomes the property's
`description` (skipped when empty; an authored attribute description wins). One sentence, existing
style.

### CHANGELOG

`[Unreleased]` (Added or Improved):
> Eloquent model `@property` / `@property-read` tags now contribute their trailing prose as the
> property `description` in the generated schema (whitespace-trimmed; empty descriptions skipped; an
> authored attribute description is never overwritten).

### Lint rule

None. (This only enriches an already-emitted property; `field.description-missing` will now fire
*less*, which is the desired effect.)

- **Low controversy** — no `src/Contracts/**`, no public signature, no config key, no stage-order
  change. `EloquentModelToSchema` is `@internal` `Support/` infra.
- The single thing to get right (flagged for the coder, not a merge gate): the **"already set"
  sentinel** for `OA\Property->description`. swagger-php uses the `Generator::UNDEFINED` *string*
  sentinel for unset annotation fields, so the no-overwrite guard must use `is_undefined(...)`, not
  `=== null` (project memory: "swagger-php UNDEFINED is a string"). The whole "attributes win"
  acceptance criterion hinges on this guard behaving correctly. Tested by edge case #6.
- Generation-affecting but description-only; no metric/object-payload movement expected. Survey is
  cheap insurance but the change cannot regress schema *shape* (it only adds a `description` key) —
  note this in the PR so review knows the survey risk is near-zero.


Radiergummi:

Grounding note: the tag description is already available at every consumption site — `metadataFor()` stores the full `PropertyTagValueNode` (type + description) per name; both `propertyFor()` and `schemaFor()` just drop `->description` by passing only `$tag->type` into `propertyFromTag()`. Plan adds one `applyTagDescription()` helper funnelled through every built property in both methods (covering the cast / enum / timestamp / untyped branches too, not only the plain-tag branch). One footgun flagged for the coder: the no-overwrite ("attributes win") guard must test `is_undefined($property->description)`, not `=== null`, because swagger-php uses the Generator::UNDEFINED *string* sentinel. Low controversy; description-only, cannot change schema shape.

Radiergummi:
🤖 **[lead]** plan posted on draft PR #424 → plan-review (queued; both reviewers currently on #423 code-review, will dispatch when one frees). Tier-0 pure docblock read; planner flagged the Generator::UNDEFINED string-sentinel guard (use is_undefined, not === null) for the attributes-win criterion. Low controversy; description-only → near-zero survey risk.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved** (amended the plan in the PR body — 2 refinements + 1 correction).

Tier 0 is right (pure docblock read; `PropertyTagValueNode->description` already parsed and merely dropped). Scope is surgical: one file, `@internal` Support infra, no Contracts/signature/config/stage change. Tests, docs page (`docs/auto-derivation.md`), and CHANGELOG entry are all named. The two items flagged for me both check out, with one correction:

1. **`is_undefined` vs `=== null` guard — confirmed correct and load-bearing.** Verified `is_undefined` wraps `Generator::isDefault` (`src/functions.php:23`), and a freshly-built `new OA\Property(['property' => …])` leaves `description` as the `Generator::UNDEFINED` *non-empty-string* sentinel, not null. So `=== null` would be false for an undecorated property → the guard would think a description already exists → **never apply**, silently breaking the whole feature. The plan mandates `is_undefined(...)`; that is the right call and the acceptance hinges on it.

2. **Funnelling all branches — correct, but I lowered the touch count and surfaced a real precedence interaction:**
   - **`schemaFor`:** instead of threading the call through the six `$properties[] = …` branches, add a **single post-loop pass** mirroring the existing `$examples` (`:608`) and `$required` (`:618`) passes, keyed by `$property->property` → `$propertyTags[name] ?? null`. Zero `continue`-branch edits, matches the file's idiom, no double-apply. Amended into the plan.
   - **`propertyFor`:** ~8 returns → wrap the body in a private `buildPropertyFor(...)` and decorate the single result once (null still propagates). One chokepoint, not eight edits. Amended in.
   - **Correction (the plan was wrong):** the plan asserted "the only description source in this file is the tag." Not true — `propertyFromSchema()` uses `copy_schema_fields`, which copies a **documented backed-enum**'s case-list description (`JsonSchemaFromType:154`) and unmapped-type apology strings onto the property. So on the enum-from-cast / enum-tag branches the property can already carry a description. This is precisely *why* `is_undefined` matters, and it makes a precedence decision: a documented-enum case description **pre-empts** the `@property` prose. That's a sensible outcome but must be **deliberate and tested**, so I added test #7 (documented-enum cast + described `@property` → keeps the enum case list, not the tag prose). Corrected the false claim and added the test to the plan.

No double-apply risk (one decoration per property per method). No clobbering of authored attributes (`#[ResourceField]` composition is downstream on its own property). Low controversy, description-only — cannot change schema *shape*, only add a `description` key, so the survey risk is near-zero as the plan states.

Plan is ready to code.

Radiergummi:
🤖 **[lead]** plan-review approved w/ refinements (reviewer-2) → coding. Plan amended: schemaFor uses a single post-loop description pass (mirrors $examples/$required idiom, zero branch edits); propertyFor wraps ~8 returns in private buildPropertyFor + decorates once; guard MUST be is_undefined (Generator::UNDEFINED sentinel, not === null). Correction: copy_schema_fields copies a documented backed-enum case-list description, so is_undefined makes the enum description pre-empt @property prose — pinned by test #7. Description-only → near-zero survey risk.

Radiergummi:
🤖 **[coder]** deviation: the reviewer's test #7 assumed a documented-enum case description pre-empts the `@property` prose on **both** paths. That holds on `propertyFor` (which inlines the enum via `fromBackedEnumClass`, so the property carries the case-list description and `is_undefined` correctly skips the tag prose). On `schemaFor`, however, the enum-cast branch uses `fromBackedEnumComponent`, which yields a **bare `$ref`** to the `DescribedStatus` component — the case-list description lives on the component, not the property. So `is_undefined($property->description)` is true there, and a naive apply would set the tag prose as a sibling of `$ref`.

Resolution: `applyTagDescription` additionally skips when `is_undefined($property->ref)` is false, i.e. it never overlays a description onto a bare `$ref` property (the description belongs on the referenced component). This keeps the spirit of the agreed precedence (enum case list wins; no competing free prose) on both paths. Test #7 is therefore split: the `propertyFor` test asserts the inline enum case list pre-empts the tag prose; the `schemaFor` test asserts the `$ref` property carries no description (and `name`/`email`/`slug`/`published_at` still get theirs). Documented in `docs/auto-derivation.md` and the CHANGELOG line.

Radiergummi:
🤖 **[coder]** implemented; local gates green; CI confirmed green; pushed af2b878. Ready for review.

Radiergummi:
🤖 **[lead]** real diff verified (sha af2b878; 58 src lines, single file EloquentModelToSchema.php), rebased onto main ef40e2c5, full CI matrix green → code-review. Two independent reviewers. Non-controversial size/kind → auto-merge candidate if reviews + survey pass. Coder flagged ONE deviation (documented on PR): applyTagDescription also skips $ref-valued properties (a description sibling to $ref is dropped) — reviewers please validate that's correct.

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer B, independent) — 1 finding (changes-needed): the `$ref`-skip deviation is too broad.

I plan-reviewed this and wrote the post-loop-pass amendment + the test-#7 precedence catch; I did not author the diff. **The implementation honours the amendment precisely** — verified all three:

1. **`is_undefined` guard (load-bearing):** ✅ `applyTagDescription` uses `is_undefined($property->description)`, not `=== null`, and imports `Radiergummi\OpenApi\is_undefined`. The "no trailing text" tests assert `is_undefined($property->description)` is true, which would falsely pass under a `=== null` check on a non-decorated property only if the sentinel were null — so these tests genuinely exercise the sentinel. Correct.
2. **`propertyFor` chokepoint:** ✅ body moved to private `buildPropertyFor()`, decorated once on the single result; null still propagates.
3. **`schemaFor` single post-loop pass:** ✅ `foreach ($properties as $property) { applyTagDescription($property, $propertyTags[$property->property] ?? null); }`, keyed by `$property->property`, mirroring the existing `$examples`/`$required` passes. No threading through branches, no double-apply (`propertyFor`/`schemaFor` are separate entry points), correct keying.

Test #7 (documented-enum precedence) **is** genuinely pinned, in both forms: `propertyFor` keeps the inline case-list (`->toContain('`draft`: …')->not->toBe('A described status tag.')`) and `schemaFor` keeps the bare `$ref` with no sibling description. Good.

**Finding 1 (changes-needed) — the `&& is_undefined($property->ref)` guard is too broad; it silently drops a legitimate, human-authored description on a Model-relation `@property`.** `EloquentModelToSchema.php:209` (the new guard in `applyTagDescription`).

The coder added the `$ref`-skip to stop the `schemaFor` enum `$ref` (which carries no inline description) from getting the `@property` prose as a sibling. That part is reasonable. But the guard fires on **every** `$ref` property, and a Model-typed `@property` is also a bare `$ref` — `classTagSchema()` (`:477-484`) maps a `@property-read Author $author` to `new OA\Schema(['ref' => …])`, copied to a bare `$ref` property by `propertyFromSchema`. So a tag like

```php
/** @property-read Author $author The post's author. */
```

now has "The post's author." **silently dropped** in both paths. This is reachable today — the existing `Article` fixture already has `@property-read Author $author` (just no trailing prose yet) — and it is squarely within #415's acceptance ("`@property … Trailing description.` emits that string as the property description", unqualified by type). OpenAPI 3.1 explicitly permits a `description` sibling to `$ref`, and unlike the enum case (where the case-list genuinely is the component's description) the relation field-description exists **nowhere else** — the referenced `Author` component's own description is a different thing from "this field is the post's author."

Why the guard over-reaches: the enum-precedence intent is *already* satisfied for `propertyFor` by the `is_undefined($property->description)` clause alone (the inline `fromBackedEnumClass` schema sets the description, so the prose is skipped). The `$ref` clause only does work in `schemaFor`'s enum case — and there, conflating enum-`$ref` with relation-`$ref` is the bug.

Recommendation (coder's choice, but the relation description must survive):
- **Preferred:** drop the `&& is_undefined($property->ref)` clause. For the `schemaFor` enum `$ref` this would add the `@property` prose as a `description` sibling on the `$ref` — harmless and valid in 3.1, and arguably more complete; the component still carries the case-list. A Model-relation `@property` then keeps its field description. Update/keep test #7's `schemaFor` assertion accordingly (decide whether the enum `$ref` may carry the prose sibling — I'd allow it).
- **Or, if you want enum `$ref`s to stay description-free:** distinguish the enum case specifically rather than blanket-skipping all `$ref`s (e.g. only skip when the property was built from a backed-enum cast), and **add a test** that a Model-relation `@property Author $x Field prose.` keeps `Field prose.` as a `description` sibling on the `$ref`.

Either way: add a fixture/test for a **Model-relation `@property` with trailing prose** asserting the description survives — that case is currently both broken and untested.

Everything else (trim, skip-empty, @property-read path, cast-branch, both methods, CHANGELOG, docs sentence — which already documents the `$ref`-skip, so it'll need a one-line tweak with the fix) checks out. Conventions fine (two docblock/signature lines at 102/104 chars, consistent with the file). Just the `$ref`-skip breadth to resolve.

Verdict: **changes-needed**. Routing to coder-1.

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer A, independent): **approved** with 1 nit (non-blocking).

Reviewed sha `af2b878` on my own merits in an isolated worktree. The two new test files are **57 passed**; reverting the production file makes the **5 new tests fail** (TDD confirmed). Pint `passed`, PHPStan `No errors`. Boundary clean (`Support/`, no plugin/third-party convention dep added; `trim` + `is_undefined` only).

**The three flagged concerns all check out:**

1. **`is_undefined()` guard — correct everywhere.** `is_undefined()` = `Generator::isDefault()`, the proper sentinel check (a fresh `OA\Property` carries the `Generator::UNDEFINED` *string*, so `=== null` would silently never fire). The guard uses `is_undefined($property->description)` **and** `is_undefined($property->ref)`; I grepped the file — no `=== null` variant on description/ref anywhere. The "no description without trailing text" tests assert `is_undefined(...)->toBeTrue()`, so the sentinel semantics are pinned, not just the happy path.

2. **The `$ref`-skip deviation is correct.** Key fact I verified: `EloquentModelToSchema` does **not** read authoring attributes at all — attribute-description merging happens downstream in the resource-field composition layer (where `#[ResourceField]` wins per field). So within this class the only pre-set description sources are (a) the inline backed-enum **case list** and (b) a bare **`$ref`** (relation or component-promoted enum). Skipping prose on a bare `$ref` is right: in OAS the description belongs on the referenced component, not as a sibling of `$ref`. The `propertyFor`-vs-`schemaFor` divergence for `status` is intentional and consistent — inline path: case-list pre-empts the prose (more specific wins); `$ref` path: no sibling description, case-list lives on the component. Both pinned (test #6 in the lookup file, the `$ref`-no-overlay test in the schema file). Precedence call (documented-enum case-list over `@property` prose) is the right one.

3. **Edge coverage is solid:** no-trailing-text → undefined; whitespace trim (`title` "Surrounded by spaces."); `@property-read` (`slug`); cast-branch description (`published_at` keeps type+format+description); enum-case-list-wins; `$ref`-no-overlay. The single-chokepoint design (`applyTagDescription` applied once in both `propertyFor` via `buildPropertyFor` and `schemaFor` via a post-loop pass mirroring the existing `$examples`/`$required` passes) genuinely covers cast/enum/timestamp/untyped uniformly.

**On "attribute-supplied description NOT overwritten" (a listed edge):** not directly tested here, and that's acceptable — that merge is **downstream of this class**, so this change cannot break it; the protecting mechanism (the `is_undefined` guard) is proven via the enum-case-list path, which is exactly the same guard. No finding.

Nit (non-blocking):

1. **Three new lines exceed the 100-char soft limit:** `EloquentModelToSchema.php:114` (102), `:202` (104, a doc-comment line), `:206` (104, the `applyTagDescription` signature). Soft limit and Pint passes, so optional; `:206` wraps cleanly (put the return type on its own line) and `:114` could use a local for `$metadata['propertyTags'][$propertyName] ?? null`. The other long lines in the region (`:141/:157/:183`) are **pre-existing moved code**, correctly left untouched.

Bookkeeping: CHANGELOG `[Unreleased]` (#415) and `docs/auto-derivation.md` both updated, and notably honest — they state the trim, empty-skip, attribute/enum-wins, and `$ref`-skip semantics explicitly. No new lint rule needed. Comments concise, why-focused, no issue numbers, no em-dashes; strict-types + `@internal` intact; no abbreviations.

Verdict: **approved.** Clean Tier-0 docblock read; the guard and `$ref`-skip are both correct and tested. The single nit is cosmetic.

Radiergummi:
🤖 **[lead]** review round 1: reviewers DISAGREE on the $ref-skip; reviewer-2's concern is correct → back to coder (1/3). reviewer-1 approved (1 nit). reviewer-2 (changes-needed): the `&& is_undefined($property->ref)` clause is too broad — it drops legitimate Model-relation @property prose (e.g. @property-read Author $author The post's author. → bare $ref), reachable (Article fixture) + in #415 scope; OAS 3.1 permits a description sibling to $ref, and that field prose exists nowhere else. Lead call: the @property trailing prose IS the field description #415 captures → emit it on $ref props. Fix = Option A (drop the $ref clause). Plus reviewer-1 nit: 3 lines >100 chars (:114,:202,:206).

Radiergummi:
🤖 **[coder]** addressed round 1: dropped the `&& is_undefined($property->ref)` clause so a relation `@property-read` keeps its field description; pushed `c21be044`.

Took the preferred option. Verified the enum-precedence intent is unaffected:
- **propertyFor path** (`buildPropertyFor` :144) builds the backed-enum cast with `fromBackedEnumClass` — an **inline** schema that sets `description` from the case list, so the existing `is_undefined($property->description)` clause still pre-empts the prose (the `ref` clause never fired here). Lookup test 'lets a documented-enum case list pre-empt …' stays green.
- **schemaFor path** (:613) promotes the enum to a `fromBackedEnumComponent` `$ref`; that case now gets the prose as a valid, harmless 3.1 description sibling — the only place the dropped clause did work.

Tests:
- Added `@property-read Author $author The article's primary author.` to the `DescribedArticle` fixture.
- Added relation-$ref description-survives tests on **both** paths (`EloquentModelPropertyLookupTest` propertyFor, `EloquentModelToSchemaTest` schemaFor) asserting `$ref === '#/components/schemas/Author'` **and** `description === "The article's primary author."`.
- Flipped the former 'does not overlay … onto a $ref enum property' schemaFor test to assert the enum $ref now keeps its sibling description (`'A described status tag.'`), per your endorsement.

Docs: `auto-derivation.md` trailing-text sentence updated — removed the $ref-skip claim, noted the 3.1 sibling is permitted. CHANGELOG entry needed no change (never mentioned the skip).

`composer check` green (3038 tests); full CI matrix green incl. swagger-php 5.8 and PHP 8.5/Laravel 12&13.

Radiergummi:
🤖 **[coder]** addressed round 1: dropped the `is_undefined($property->ref)` skip so `@property` prose is emitted on `$ref`-valued properties (relation `$ref`s now keep their field description, which exists nowhere else); the inline documented-enum case list still pre-empts on `propertyFor` (case-list description already set → `is_undefined` false), and the schemaFor enum `$ref` now carries the tag prose as a 3.1-valid sibling. Updated schemaFor test #7 to expect the prose sibling, added enum + relation `$ref` description tests on both `schemaFor` and `propertyFor`, and wrapped the long lines. CHANGELOG/docs reflect the relation-`$ref` behaviour. Local gates green; CI confirmed green; pushed afeb74a.

Radiergummi:
🤖 **[lead]** round 1 fix verified (sha afeb74a; dropped the $ref-skip per lead decision, relation $ref now keeps its @property prose, enum precedence preserved, 3 lines wrapped), CI green on full matrix → re-review. Both reviewers re-confirming on afeb74a (behavior reversed from the skip both previously assessed, so both prior verdicts are stale). Non-controversial (58 src lines) → auto-merge candidate on 2 fresh approvals + survey.

Radiergummi:
🤖 **[reviewer]** review round 2 (reviewer A, independent): **approved, no concerns.**

Re-reviewed sha `afeb74a` on its own merits in a fresh isolated worktree (detached HEAD, not the coder's branch checkout). The behaviour is reversed from the `$ref`-skip both reviewers assessed in round 1, so this is a clean re-assessment, not a re-stamp.

**The three things I was asked to verify all hold:**

1. **`$ref`-skip removal correctly keeps relation `@property` descriptions.** `applyTagDescription` (`EloquentModelToSchema.php:207`) now guards on `is_undefined($property->description)` only — the `&& is_undefined($property->ref)` clause is gone. A relation `@property-read Author $author The article's primary author.` keeps its prose on **both** paths: `propertyFor` asserts `ref === '#/components/schemas/Author'` **and** `description === "The article's primary author."`, and `schemaFor` asserts the same. This is exactly the case the old skip silently dropped (#415's acceptance, type-unqualified), and OAS 3.1 permits a `description` sibling to `$ref`. Correct per the lead's Option-A decision.

2. **Enum case-list precedence preserved on both paths.**
   - *`propertyFor` (inline):* `buildPropertyFor` builds the backed-enum cast via `fromBackedEnumClass`, an inline schema that sets `description` from the case list (copied onto the property by `copy_schema_fields`); the surviving `is_undefined($property->description)` clause then sees it set and skips the prose. Test 'lets a documented-enum case list pre-empt the @property description' asserts `->toContain('`draft`: …')->not->toBe('A described status tag.')` — and, tellingly, it is the **one** new test that passes both with and without the production change (the `$ref` clause never did work here), proving the precedence comes from the description-guard, not the removed clause.
   - *`schemaFor` (`$ref`):* the enum is promoted to a `fromBackedEnumComponent` `$ref` carrying no inline description, so the prose now lands as a valid 3.1 sibling — the component still owns the case list. Pinned by the flipped test asserting `ref === '#/components/schemas/DescribedStatus'` **and** `description === 'A described status tag.'`.

3. **The new/changed tests genuinely pin the reversed behaviour.** Reverting `EloquentModelToSchema.php` to `main` (`git checkout ef40e2c5 -- …`) fails **9** of the new tests — including both relation-`$ref` survives tests and the enum-`$ref` sibling test (description reads back as `Generator::UNDEFINED`). With the change, the two files are **60 passed**. TDD confirmed.

**Mechanics I independently checked:** `PropertyTagValueNode->description` is a non-nullable `string` (empty when no prose), so `trim($tag->description)` is safe and the `=== ''` guard handles the no-prose case (the `name` field → `is_undefined(...)` true). The `schemaFor` post-loop pass is keyed by `$property->property` and mirrors the existing `$examples`/`$required` passes; `propertyFromSchema` sets `property => $name` and the copied component schema carries no `property`, so the key lookup is correct for enum `$ref` props too.

**Gates (in my worktree):** `vendor/bin/pint --test` → passed; `composer lint` (PHPStan L8) → No errors. Reviewer-B's round-1 line-length nit is resolved — the new lines (`:202` docblock, `:207` signature) are now ≤100. The three remaining >100 lines (`:142/:158/:184`) are **pre-existing moved code** (verified against `ef40e2c5`), correctly left untouched per the surgical-changes rule.

**Bookkeeping:** CHANGELOG `[Unreleased]` (#415) and `docs/auto-derivation.md` both updated and honest — they state trim, empty-skip, the attribute/enum-case-list-wins precedence, and the 3.1 `$ref`-sibling allowance; the earlier `$ref`-skip claim is removed. No new lint rule (correct — this only enriches an already-emitted property).

Nit (non-blocking, not a finding): two `propertyFor` author tests ('keeps the @property prose as a description sibling of a relation $ref' and 'keeps a relation @property-read description on its $ref property') assert identical things — one is redundant. Harmless.

Verdict: **approved.** The guard and the relation-`$ref` behaviour are both correct and tested; enum precedence is intact on both paths.

Radiergummi:
🤖 **[reviewer]** review round 2 (reviewer B): **approved, no concerns.**

My round-1 finding (`&& is_undefined($property->ref)` dropped legitimate relation `@property-read` prose on `$ref`-valued properties) is **fully resolved**. The `$ref`-skip clause is gone; `applyTagDescription` now guards only on `$description !== '' && is_undefined($property->description)` (`src/Support/Extraction/EloquentModelToSchema.php:217`), so a relation property keeps its prose.

Verified independently at `afeb74a` in an isolated worktree:

1. **Concern resolved.** `propertyFor(DescribedArticle, 'author')` → `$ref` + `description` both set; tests at `EloquentModelPropertyLookupTest.php` and `EloquentModelToSchemaTest.php` cover it on both the lookup and schema paths.
2. **Serialization well-formed.** Serialized the relation property under the real OAS 3.1.0 context (`SchemaContextScope`) — output is `{"$ref": "...", "description": "The article's primary author."}`. swagger-php 6.x preserves `summary`/`description` siblings of `$ref` for non-3.0.x contexts (`AbstractAnnotation::jsonSerialize` L322–331); the pipeline pins 3.1.0, so the description survives into the shipped document.
3. **Enum precedence holds, no clobber.** Description is written only when `is_undefined($property->description)`, so a documented backed-enum case-list pre-empts the prose. Confirmed on both the inline-enum `propertyFor` path and the component-`$ref` enum `schemaFor` path.
4. **No double-application.** Single chokepoint via `applyTagDescription`, applied once per path.

Quality gates: PHPStan clean, `pint --test` clean on the touched file, all 277 `tests/Unit/Support/Extraction/` tests green on **both PHP 8.4 and 8.5**.

Non-blocking note (not a merge gate): the new tests assert the in-memory `$property->description` but never the 3.1-serialized output, so they wouldn't catch a 3.0-context serialization-drop regression. I verified the serialization manually and it is correct; a follow-up serialization assertion would harden the suite.

Radiergummi:
🤖 **[lead]** two independent approvals (reviewer-A + reviewer-B, both clean on sha afeb74a; reviewer-B confirms its round-1 $ref-skip finding fully resolved) → survey. Generation-affecting (area:responses): Layer A acceptance = 0 regressions across the corpus; this change only ADDS @property prose as description siblings, so expect a pure-addition delta (no paths/schemas lost).

Radiergummi:
🤖 **[surveyor → relayed by lead]** Layer A **PASS — 0 regressions across 11 apps.** Baseline reused from main @ ef40e2c5 (crash-free); candidate pinned to PR sha afeb74a1, installed-sha verified across all 11 apps. Every deterministic metric (paths, operations, apiOperations, responseSchemas, documentedResponses, requestBodies, componentSchemas) identical baseline→candidate — expected for a fidelity-only `area:responses` change that only adds a `description` field inside already-emitted schemas (is_undefined-guarded, moves no counted metric). (Surveyor is blocked from posting gh comments directly; lead posting on its behalf.)

Radiergummi:
🤖 **[lead]** non-controversial + all gates green → squash-merged, closed #415, worktree removed. Two independent approvals (reviewer-A + reviewer-B), Layer A 0-regression pass across 11 apps, full CI matrix green on afeb74a (incl. swagger-php 5.8 + snapshot). 58 src lines, single file (EloquentModelToSchema.php), no Contracts/config/signature/stage-order change.

PR #425, opened 2026-06-20
feat(#416): emit a path parameter for every URI placeholder, even when absent from the controller signature

### LINKED ISSUE #416: feat(params): emit a path parameter for every URI placeholder, even when absent from the controller signature
## Problem

Only **68%** of routes carrying a `{param}` emit a documented path parameter. Path params are sourced
solely from the controller method signature, so any placeholder not type-hinted as an action argument
(the parent of a scoped/nested binding, invokable controllers reading `\$request->route()`, routes
injecting only `Request`) produces no parameter at all — invalid per OAS, which requires every path
template variable to be declared.

## Evidence

- `src/Routing/ActionDescriptor.php:198-212` — `resolveUriParameters()` keeps only `route->signatureParameters()` whose name appears as a `{name}` URI token; an unsignatured `{team}` is dropped.
- `src/Support/Generator/OperationBuilder.php:117-130` — that list is the only source fed to the extractor.
- Contrast `src/Lint/Rules/PathParameterUndeclared.php:69-79` — the linter already extracts ALL placeholders from the raw URI by regex, proving the URI-token source is available; generator and lint disagree.

After the signature-derived params, diff against the raw URI placeholders (mirror
`PathParameterUndeclared::extractPlaceholders`). For each undeclared placeholder synthesize an
`OA\Parameter` with `in: path, required: true, schema: {type: string}`. Enrich from
signature-independent Tier-0 signal where present: reuse `\$route->wheres[\$name]` via the existing
`WhereKind` classifier (uuid/integer/enum/pattern) and implicit-binding info when resolvable. A
string path segment is always correct, so this cannot emit a wrong schema.

## Scope / Out of scope

- In: presence + string fallback + reuse of existing `where*`/binding enrichment.
- Out: Tier-1 body scanning to recover a model class for an unsignatured bind.

## Acceptance

`/{a}/{b}` whose controller types only `\$b` emits both `a` and `b` as `in: path, required: true`.
`path.parameter-undeclared` reports zero findings on the corpus specs. `where`-constrained
unsignatured params still get uuid/int/enum schemas.

<sub>From the 2026-06-20 lever audit; relates to #413.</sub>

Radiergummi:
Closes #416

### Tier & rationale

**Tier 0** — presence is read from the raw URI template (regex over `$route->uri()`, the exact
source the linter already uses) and enrichment from signature-independent route metadata
(`$route->wheres[$name]`, custom binding field). No body parsing, no model-class recovery for an
unsignatured bind (that's Tier-1, explicitly out of scope). A path segment is always a string in the
URL, so the `{type: string}` fallback can never be *wrong* — it's never-wrong by construction.

- `ActionDescriptor::resolveUriParameters()` (`ActionDescriptor.php:198-212`) iterates
  `route->signatureParameters()` and keeps only those whose name appears as `{name}` / `{name?}` in
  the URI. An unsignatured `{team}` has no `ReflectionParameter`, so it never enters the list. The
  property `uriParameters` (`:28`) exposes that list.
- `OperationBuilder::build()` (`:117-130`) is the **only** feeder to `UriParametersExtractor`: it maps
  each `ReflectionParameter` to a `[UriParameterDescriptor, ReflectionParameter]` pair via
  `uriResolver->resolve(...)`.
- `UriParametersExtractor::extract()` (`UriParametersExtractor.php:45-51`) already accepts
  `list<array{UriParameterDescriptor, ?ReflectionParameter}>` — **the reflection slot is already
  nullable** and `buildParameter`/`resolveFieldDescriptor` already null-guard it. So an unsignatured
  placeholder needs only a `UriParameterDescriptor` paired with `null`; the extractor needs no change.
- The linter (`PathParameterUndeclared::extractPlaceholders`, `:73-83`) regex-extracts all
  placeholders. The generator simply never consulted that source — this aligns them.

### Design — synthesize descriptors for undeclared placeholders, feed them as `(descriptor, null)`

The seam is clean because the extractor is already null-tolerant. Two production touchpoints:

1. **`src/Support/Routing/UriParameterResolver.php` — add a signatureless resolver method.**
   `resolveUnsignatured(string $name, ?string $whereConstraint, ?string $bindingField): UriParameterDescriptor`.
   - Type defaults to `Type::builtin(TypeIdentifier::STRING)` (same fallback `resolveType()` already
     uses for an untyped parameter, `:99-106`).
   - Reuse the existing `resolveConstraint()` / `classifyConstraint()` so `where*` regexes still map
     to `WhereKind::Uuid` / `Number` / `In` (literal alternation) / `Custom` → the extractor's
     `applyWhereKindOverrides()` then yields `format: uuid` / `type: integer` / `enum: [...]` /
     `pattern: …` exactly as for signatured params. **This is what satisfies the acceptance line
     "where-constrained unsignatured params still get uuid/int/enum schemas."**
   - `modelBinding`: null and `enumCases`: null — we cannot recover a model class without a signature
     (Tier-1, out of scope). `optional`: derive from whether the placeholder is `{name?}` (see §3) so
     the extractor's "Optional in URL" description suffix still applies. The descriptor's `type` stays
     a non-nullable string regardless (OAS forbids optional path params; the existing extractor already
     forces `required: true` and appends the optional note — same behaviour, no special-casing).
   - Factor the shared body so `resolve()` (signatured) and `resolveUnsignatured()` both build the
     descriptor through one private constructor helper — keep it DRY, don't duplicate the
     constraint/where logic.

2. **`src/Support/Generator/OperationBuilder.php:117-130` — append undeclared placeholders.**
   After building the signature-derived pair list, compute the set of URI placeholders, subtract the
   names already produced from the signature, and for each remaining name append
   `[$this->uriResolver->resolveUnsignatured($name, $action->constraintFor($name), $action->bindingFieldFor($name)), null]`.
   Preserve **URI order** for deterministic output (signature-derived first, then undeclared in URI
   order — or, cleaner, emit all params in URI-template order; pick one and assert it in a test).
   Dedup by name so a placeholder that *is* signatured is never emitted twice.

3. **Placeholder extraction + optionality — where to put it.**
   Add a small helper that returns the URI placeholders (and whether each is optional `{name?}`).
   Options: (a) a method on `ActionDescriptor` (e.g. `uriPlaceholders(): list<array{string,bool}>`)
   that `OperationBuilder` consults, or (b) a tiny shared `Support/Routing` utility reused by both the
   generator and the linter. **Recommendation:** put it on `ActionDescriptor` (it already owns
   `route->uri()` and the signature-derived list, so the dedup + ordering live together), mirroring
   `PathParameterUndeclared::extractPlaceholders` regex (`/\{([^?}]+)\??}/` + the `ltrim` of binding
   prefix chars). Do **not** refactor the linter in this PR (avoid scope creep); a follow-up can DRY
   the two extractors if desired — note it, don't do it.

   **REQUIRED (`{param:field}` bare-name rule — plan-review amendment).** The helper must emit the
   **bare** parameter name, not the raw regex capture. Confirmed empirically: the linter regex on
   `teams/{team}/members/{member:slug}/{opt?}` yields `['team', 'member:slug', 'opt']` — the `:field`
   suffix is *inside* the braces and the `ltrim` strips only *leading* binding-prefix chars, so the
   raw capture for `{member:slug}` is `member:slug`. Mirroring it verbatim would (1) emit a path
   parameter literally named `member:slug` (invalid OAS — the template variable is `member`) and
   (2) look up `constraintFor('member:slug')` / `bindingFieldFor('member:slug')`, both missing
   (Laravel keys `wheres`/binding on the bare `member`), silently losing where-enrichment. So the
   helper **must split on `:` and take the segment before it** for the emitted name and for both
   lookups. The generator's helper is therefore **not** byte-identical to the linter's regex (they
   have different needs); that is fine — the acceptance only requires the *set of names* to agree.
   Follow-up note: `PathParameterUndeclared::extractPlaceholders` likely mis-handles `{param:field}`
   today (compares `member:slug` against signature names → potential mis-report) — a **pre-existing
   linter bug, OUT OF SCOPE here**; file it as the first task of the DRY-with-linter follow-up so the
   shared extractor fixes both. Do not touch the linter in this PR.

### Files to modify

- `src/Support/Routing/UriParameterResolver.php` — add `resolveUnsignatured()` (+ shared private builder).
- `src/Support/Generator/OperationBuilder.php` — append undeclared-placeholder pairs to the extractor input.
- `src/Routing/ActionDescriptor.php` — add the placeholder-extraction helper (bare name + optional
  flag), based on the linter's regex but splitting off any `:field` suffix to yield the bare
  parameter name (see REQUIRED §3); keep `uriParameters` (signatured) as-is so nothing else regresses.

No change to `UriParametersExtractor`, `UriParameterDescriptor`, `WhereKind`, or any `Contracts`/signature.
Estimated diff well under ~150 src lines (one new resolver method, one loop in the builder, one helper).

### Tests (TDD — write first)

Feature, new **`tests/Feature/UnsignaturedPathParameterTest.php`** (mirror `TypedPathParameterTest`'s
`generateSpec()` + parameter-lookup helper):

1. **Acceptance core:** `/{a}/{b}` whose controller types only `$b` → spec emits **both** `a` and `b`,
   each `in: path, required: true`, `a` with `schema: {type: string}` (unsignatured), `b` typed from
   its signature. Assert ordering is deterministic.
2. **Invokable / Request-only:** an `__invoke(Request $request)` controller on `/{team}/{member}` with
   no bound args → both placeholders emitted as `in: path, required: true, type: string`.
3. **Scoped/nested-binding parent:** `/{team}/{member}` where only `$member` is signatured (the
   classic scoped-binding parent case from the issue) → `team` emitted as a string path param.
4. **where-enrichment (acceptance):** unsignatured `{id}` with `->whereUuid('id')` → `format: uuid`;
   `->whereNumber('id')` → `type: integer`; `->whereIn('status', [...])` → `enum: [...]`; a custom
   `->where('slug', '[a-z-]+')` → `pattern`. One case per WhereKind.
5. **`{param:field}` unsignatured** → emitted as a string path param whose name is exactly the bare
   `param` (NOT `param:field`); and when a `where` is attached to that param, assert the uuid/int/etc.
   enrichment still fires (proves the `wheres`/`bindingField` lookup uses the bare name, per the
   REQUIRED §3 rule).
6. **No regression:** a fully-signatured route still emits exactly its signatured params (no dupes);
   an optional `{name?}` unsignatured placeholder emits `required: true` + the "Optional in URL"
   description note.
7. **Lint cross-check (acceptance):** generate a spec for a fixture with an unsignatured placeholder
   and run the lint pipeline asserting `path.parameter-undeclared` yields **zero** findings (mirror
   how existing lint feature tests invoke the runner). This is the regression guard that generator and
   linter now agree.

Unit, **`tests/Unit/Support/Routing/UriParameterResolverTest.php`** — extend for `resolveUnsignatured()`:
bare name → string descriptor; uuid/number/in/custom where → correct `WhereKind`; null binding/enum;
optional flag propagation. Edge: empty/odd names guarded.

Edge cases explicitly covered: unsignatured + where(uuid/int/enum/custom); invokable; Request-only;
scoped-parent; `{param:field}`; optional `{name?}`; signatured-no-dup; ordering determinism; lint-zero.

### Docs

`docs/auto-derivation.md` — the path-parameter row/section (the URI parameters paragraph; cross-check
the table at line 16-ff and any "Path parameters" subsection). State that **every** URI placeholder is
now emitted as a path parameter (`in: path, required: true`), defaulting to `type: string` when the
placeholder is not a typed action argument, and enriched from `where*` constraints / custom binding
field when present. Mention the model-class key typing still requires the parameter to be signatured
(an unsignatured bind stays string) — set the honest boundary.

### CHANGELOG

`[Unreleased]` (Added or Fixed — lean Fixed, since the old output was invalid OAS):
> Every URI path placeholder now emits a path parameter, even when it isn't a typed controller
> argument (invokable controllers, `Request`-only actions, the parent of a scoped/nested binding).
> Undeclared placeholders default to `{type: string}` and are enriched from `where*` constraints
> (uuid/integer/enum/pattern). Previously such placeholders produced no parameter, which is invalid
> per the OpenAPI spec.

### Lint rule

None added. The change makes the existing `path.parameter-undeclared` rule report **zero** on
previously-failing routes (the acceptance criterion); the rule itself is unchanged and remains the
guard for hand-authored specs.

- **Low controversy.** No `src/Contracts/**`, no public signature, no config key, no stage-order
  change. All touched classes are `@internal`. Diff estimated < ~150 src lines.
- The one judgment call for the reviewer: **where to host the placeholder-extraction helper** (on
  `ActionDescriptor` vs. a shared `Support/Routing` util shared with the linter). I recommend
  `ActionDescriptor` now + a noted follow-up to DRY with the linter, rather than refactoring the
  linter in this PR (keeps the diff surgical). Flag if the reviewer prefers the shared util up front.
- Output-shape change: routes that previously emitted **no** parameter for a placeholder now emit one.
  This is strictly more-correct (OAS-valid) and is what the survey lever targets (68% → higher), but
  it *does* change generated specs for affected corpus apps — generation-affecting, so the **survey
  gate applies**; acceptance expects the `path.parameter-undeclared` corpus count to drop to zero with
  no other regressions.


Radiergummi:

Grounding note that makes this clean: `UriParametersExtractor::extract()` already accepts `list<array{UriParameterDescriptor, ?ReflectionParameter}>` — the reflection slot is ALREADY nullable and null-guarded. So an unsignatured placeholder needs only a synthesized descriptor paired with null; the extractor needs zero changes. Plan adds `UriParameterResolver::resolveUnsignatured()` (string-default type, reusing the existing where-regex classifier so uuid/int/enum/pattern enrichment still fires), appends undeclared placeholders in OperationBuilder, and adds a placeholder-extraction helper on ActionDescriptor mirroring the linter regex. Model-class key typing for an unsignatured bind stays out (Tier-1). Low controversy, <~150 src lines, no Contracts/signature change. Generation-affecting → survey gate (acceptance: path.parameter-undeclared drops to zero, no other regressions).

Radiergummi:
🤖 **[lead]** plan posted on draft PR #425 → plan-review. Tier-0 never-wrong; extractor needs zero changes (nullable reflection slot already null-guarded). 3 @internal touchpoints, est <150 src lines. Reviewer judgment call: helper hosted on ActionDescriptor now + noted DRY-with-linter follow-up vs shared util up front — weigh in. Generation-affecting → survey (acceptance: corpus path.parameter-undeclared drops to zero, no other regressions).

Radiergummi:
🤖 **[reviewer]** plan-review: **approved with one required change** (the `{param:field}` name handling must be pinned, not left as "verify"). Tier and scope are right.

Verified the plan's grounding against the code — all load-bearing claims hold:
- `UriParametersExtractor::extract()` already takes `list<array{UriParameterDescriptor, ?ReflectionParameter}>`; `buildParameter()` takes `?ReflectionParameter` and every downstream use null-guards it (`resolveFieldDescriptor`, the `?->` chains). An unsignatured placeholder needs only `[descriptor, null]` — **zero extractor changes**, as claimed.
- `buildParameter()` already forces `required: true` and appends the "Optional in URL" note for `descriptor->optional` (OAS §4.8.12.1). So an unsignatured `{name?}` rides the same path; the descriptor's type stays non-nullable string. Correct, no special-casing needed.
- `ActionDescriptor::constraintFor()` (`:185`, reads `$route->wheres[$name]`) and `bindingFieldFor()` (`:190`) already exist — the enrichment inputs the plan threads are available. The where→`WhereKind` classifier path (`applyWhereKindOverrides`) is exactly what yields uuid/int/enum/pattern, so the acceptance line for where-enrichment is satisfiable as described.
- The feeder is exactly `OperationBuilder.build()`'s `array_map` over `$action->uriParameters` (`:117-130`); appending undeclared pairs there is clean.

**Required change — pin the `{param:field}` bare-name handling (this is a correctness gap, not just a test):**
I ran the linter's regex on `teams/{team}/members/{member:slug}/{opt?}`: `/\{([^?}]+)\??}/` + `ltrim($name, '+#./;?&=,!@|')` yields `['team', 'member:slug', 'opt']` — it does **not** strip the `:field` suffix (the `ltrim` only removes *leading* binding-prefix chars). So mirroring the linter regex *verbatim* for an unsignatured `{member:slug}` would:
  1. emit a path parameter literally named `member:slug` — **wrong OAS** (the URI template variable is `member`);
  2. look up `constraintFor('member:slug')` / `bindingFieldFor('member:slug')`, both **missing** (Laravel keys `wheres`/binding on the bare `member`), silently losing where-enrichment.

The plan flags this case but only as "verify … confirm the name matches" + add a test. That is too soft — it must be a hard rule: **the placeholder-extraction helper must return the bare parameter name (split on `:` and take the segment before it) for the name it emits and for the `wheres`/`bindingField` lookups.** Consequence: the generator's helper is therefore **not** byte-identical to the linter's regex (the plan's "keep the regex identical so generator and linter agree" premise is subtly wrong — they have different needs). Update the plan to: (a) state the helper strips the `:field` suffix; (b) keep the existing test #5 but assert the emitted name is exactly `param` (not `param:field`) AND, if a `where` is attached to that param, that the uuid/int/etc. enrichment still fires (proves the lookup uses the bare name). I amended the PR body's §3 and test #5 accordingly.

**Reviewer judgment call — helper placement: accept `ActionDescriptor` now + noted DRY follow-up.** It already owns `route->uri()` and the signatured list, so dedup + ordering live together; the diff stays surgical. I do **not** require the shared util up front. BUT the follow-up note should be explicit that the linter *also* has the `:field` defect above — i.e. `PathParameterUndeclared::extractPlaceholders` likely mis-handles `{member:slug}` today (compares `member:slug` against signature names → potential mis-report). That is a **pre-existing linter bug, out of scope for this PR** (do not fix the linter here), but worth filing as the DRY-follow-up's first task so the shared extractor fixes both. Note it; don't expand scope.

**Out-of-scope confirmation:** model-class key typing for an unsignatured bind correctly stays Tier-1/out (reflection can't see a bound model without a typed signature param) → string fallback, never-wrong. Right call.

**Proving tests / bookkeeping all named and adequate:** acceptance core (`/{a}/{b}`, only `$b` typed), invokable/Request-only, scoped-parent, per-WhereKind enrichment (uuid/int/in/custom), `{param:field}` (now with the bare-name assertion), optional `{name?}`, signatured-no-dup, ordering determinism, and the `path.parameter-undeclared`=zero lint cross-check. Unit tests for `resolveUnsignatured()`. Docs (`auto-derivation.md` path-param section, honest model-class boundary), CHANGELOG (lean Fixed — old output was invalid OAS), no lint rule. Survey gate correctly invoked (generation-affecting; acceptance = corpus `path.parameter-undeclared` drops to zero, no other regressions).

Verdict: **approved once the `{param:field}` bare-name rule is pinned in §3 + test #5** (amended in the body). Everything else is ready to implement.

Radiergummi:
🤖 **[lead]** plan-review approved w/ 1 required change (reviewer-1) → coding. REQUIRED: the {param:field} bare-name rule — do NOT mirror the linter regex verbatim (it yields 'member:slug', an invalid OAS name, and misses constraintFor/bindingFieldFor which Laravel keys on the bare 'member'). The helper MUST split off ':field' and use the bare name for both the emitted parameter name AND the where/binding lookups (see amended §3 + test #5). Accept ActionDescriptor placement + noted DRY-with-linter follow-up (file the pre-existing linter {param:field} bug as its first task). Generation-affecting → survey (path.parameter-undeclared → zero, no other regressions).

Radiergummi:
🤖 **[coder]** deviation: to satisfy the plan's `whereIn()` → `enum` acceptance line, `UriParameterResolver::buildDescriptor()` now derives `enumCases` from a `WhereKind::In` constraint's literal alternatives when the type did not already supply them (a backed-enum signature still wins). This was *not* previously implemented for any param — `WhereKind::In` only emitted an `enum` when a backed-enum signature provided the cases — so it slightly widens behaviour for **signatured** string params with a `whereIn()` too, not just unsignatured ones. It is centralized in the shared builder (DRY across both paths), Tier-0 (the alternatives are literal, vetted by the existing `isLiteralAlternation` metachar guard), and strictly more-correct. Full suite (incl. the byte-exact snapshot group) stays green, so no existing corpus snapshot regressed.

🤖 **[coder]** filed the DRY-with-linter follow-up as #426 (labels `area:core`, `area:lint`); its first task is the pre-existing linter `{param:field}` mishandling in `PathParameterUndeclared::extractPlaceholders()` (out of scope here, as agreed).

Radiergummi:
🤖 **[coder]** implemented; local gates green; CI confirmed green (all 9 checks incl. swagger-php 5.8 + locked-toolchain snapshot); pushed 549f1811. Ready for review.

Radiergummi:
🤖 **[lead]** real diff verified (sha 549f1811; 140 src lines across 3 @internal files), rebased onto main ef40e2c5, full CI matrix green → code-review. Two independent reviewers. Non-controversial size (<150) + kind → auto-merge candidate if reviews + survey pass. Coder flagged ONE deviation (on PR): buildDescriptor() now derives enumCases from a WhereKind::In literal-alternatives even for SIGNATURED string params (widens behaviour, claimed strictly-more-correct Tier-0) — reviewers please validate it's genuinely always-correct. DRY follow-up filed as #426.

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer B, independent) — 1 finding (changes-needed): the `whereIn` enum widening to signatured params is not always never-wrong.

Reviewed sha `549f1811` adversarially and on my own merits. The core feature (synthesizing a string path parameter for every undeclared URI placeholder) is correct and thoroughly tested: the unsignatured path always types `string` (always correct for a URL segment), `where*` enrichment fires, dedup/ordering is right (signatured first, then placeholders in URI order — pinned by the `/{a}/{b}` → `['b','a']` and the no-duplicate tests), the `{param:field}` bare-name handling is correct (strips the `:field` suffix and the binding-prefix chars, looks up `wheres`/binding on the bare name — pinned by the `{member:slug}` test), optional `{name?}` → `required: true` + note, and the `path.parameter-undeclared` = zero cross-check runs the full generator. Boundaries clean (no plugin deps), CHANGELOG + docs updated, strict-types present.

**Finding 1 (changes-needed) — deriving the `whereIn` enum for SIGNATURED params produces a type/enum mismatch on a non-string (int) param.** `UriParameterResolver.php` (`buildDescriptor`, the new `if ($whereKind === WhereKind::In && $enumCases === null …) { $enumCases = explode('|', $resolvedConstraint); }`).

The coder flagged this widening as "strictly more-correct." It is genuinely beneficial for a signatured **string** param (`whereIn('status', ['draft','published'])` → `{type: string, enum: ['draft','published']}`, useful and correct). But it also fires for a signatured **int** param:

```php
Route::get('/items/{id}', fn(int $id) => [])->whereIn('id', [1, 2, 3]);
```

`->whereIn('id', [1,2,3])` emits the regex `1|2|3`; `isLiteralAlternation('1|2|3')` is true (digits are not metacharacters) → `WhereKind::In`. The int signature already resolved `$schema = {type: integer}` (UriParametersExtractor.php:57), and `applyWhereKindOverrides` (`:115`) sets `$schema->enum = $descriptor->enumCases` **without touching the type**. `enumCases` are always strings from `explode('|', …)`. Result:

```yaml
schema: { type: integer, enum: ['1', '2', '3'] }   # string enum values on an integer type
```

That is a wrong/misleading schema (JSON Schema enum values should match the declared type), and it's a realistic route shape. So the widening is **not unambiguously never-wrong** — it's correct for string-typed params and wrong for int-typed ones. (It cannot misfire on the unsignatured path, which is always `string`, nor on a signatured backed-enum, where `$enumCases` is already non-null and the `&& $enumCases === null` guard preserves it.)

Recommendation (the lead pre-suggested this and I concur): **scope the `whereIn`-enum derivation so it can't land string values on a non-string type.** Cleanest options:
- (a) Only derive it in `resolveUnsignatured` (always string) — narrowest, matches the issue's actual scope (#416 is about undeclared placeholders). The signatured-string benefit is then a separate, opt-in follow-up.
- (b) Keep it in `buildDescriptor` but guard on the resolved type being string (e.g. only when `$type` is `TypeIdentifier::STRING`), so an int/float param keeps `{type: integer}` with no mismatched enum.
Either way add a test for a **signatured int param + `whereIn`** asserting no string-enum is attached to the integer type (the case currently both broken and untested).

Minor (non-blocking): codecov flags the `preg_match_all() === 0` early return in `uriPlaceholders()` (a placeholder-free URI) as uncovered — a trivial guard, optional to test.

Verdict: **changes-needed** on finding 1. Routing to coder-2. Everything else is approve-ready.

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer A, independent) — 1 finding (non-blocking), otherwise approved.

I plan-reviewed this (pinned the `{param:field}` rule); I did not author the diff. Reviewed sha `549f1811` in an isolated worktree.

**My pinned `{param:field}` rule is honoured.** `ActionDescriptor::uriPlaceholders()` strips binding-prefix chars then **splits off the `:field` suffix** (`explode(':', $name, 2)[0]`), emitting the bare name; `OperationBuilder` uses that bare name for both the emitted parameter and the `constraintFor`/`bindingFieldFor` lookups. The docblock documents exactly why. Test "emits the bare name for an unsignatured `{param:field}`" asserts the name is `member` (not `member:slug`) **and** that `whereUuid('member')` enrichment fires — proving the lookup keys on the bare name. Correct.

**The `whereIn`→`enumCases` widening (the flagged deviation) is correct and genuinely never-wrong — keep it shared, do NOT scope to unsignatured-only.** I traced it end-to-end:
- `WhereKind::In` is assigned only when `isLiteralAlternation($regex)` is true, and that helper rejects any alternative containing a regex metachar (`[](){}\+*?.^$`) or an empty alternative. So `foo|bar.*` is classified `Custom` → emitted as `pattern`, never as an enum. Only a *closed literal set* reaches the enum path.
- A backed-enum signature type still wins (`$enumCases === null` guard); the `whereIn` fallback only fills when no type-derived cases exist.
- The route 404s any segment outside the alternation, so `enum: [literals]` precisely describes the accepted values — correct for the string path param.
- Pre-fix, the extractor's `WhereKind::In` branch already set `$schema->enum = $descriptor->enumCases` but `enumCases` was null for a non-enum `whereIn` string param, so the branch was a silent no-op. This change makes a pre-existing, already-wired branch actually fire. Scoping it to the unsignatured path would make a signatured vs. unsignatured `whereIn` param document *differently* for no reason — the shared `buildDescriptor()` placement is right.

**Append/dedup/order correct:** signature-derived pairs first, then `uriPlaceholders()` in URI order, deduped by name (`$declaredNames`), `(descriptor, null)` pairs; zero changes to the null-tolerant extractor. Ordering pinned by the `['b','a']` assertion.

**Verified:** 12 feature tests + the unit `resolveUnsignatured` cases (string default, uuid/number/in/custom where, null binding/enum, `{param:field}` bare name) pass. TDD confirmed — reverting the 3 src files fails 10 of the 12 feature tests. Pint passed, PHPStan No errors on all 3 files; path-param + unit Routing regression band 56 passed. Boundary clean (`Support/` + `Routing/`, no plugin/third-party dep). Docs (`auto-derivation.md` +9) and CHANGELOG present; no lint rule. `@internal` intact; bare-name split documented; no em-dashes.

**Finding 1 (non-blocking, but real — and it touches the headline acceptance criterion):**
The test *"emits zero `path.parameter-undeclared` findings for an unsignatured placeholder"* is **vacuous — it passes identically on the pre-fix code.** I isolated it: with the 3 production files reverted to parent, that test still passes (1 passed). So it does **not** guard the generator↔linter agreement that is the entire premise of #416 (the issue is literally "generator and lint disagree"). The likely cause: the test pre-injects an `ArrayFindingsCollector` via `instance()`, but `OpenApiGenerationOrchestrator` swaps in its own collector and may run generation with `discardFindings` (findings → a throwaway collector), so the asserted collector stays empty regardless of whether `{team}` is documented. The **fix itself is correct** (the other 11 tests prove `{team}` is now emitted), but this acceptance test gives false confidence. Make it fail-without-the-fix: either (a) assert directly that the generated spec's parameters for `/teams/{team}/members/{member}` include `team` (the positive form — simplest and unambiguous), or (b) ensure the collector the rule writes to during generation is the one asserted (confirm `discardFindings` is false on this path). Recommend (a) as an additional assertion; keep the lint form too once it's wired to actually catch the regression.

Nit (optional): three new doc-comment lines in `UriParameterResolver.php` (:62, :132, :167) are 101–104 chars, just over the 100 soft limit; Pint passes so cosmetic. `ActionDescriptor.php:44` (117) is pre-existing and correctly untouched.

Verdict: **approve, conditional on Finding 1 being addressed** (it's a test-validity gap, not a production-correctness bug, so it doesn't block the *change* — but the headline acceptance test should actually exercise the fix). If the lead/coder prefers, I'm fine treating Finding 1 as a required follow-up commit rather than a re-review blocker, since the production logic is sound and fully proven by the other tests.

Radiergummi:
🤖 **[lead]** review round 1 (reviewer-2): 1 substantive finding → back to coder (1/3). The whereIn→enum widening is NOT never-wrong for signatured params: whereIn('id',[1,2,3]) on int $id yields {type: integer, enum: ['1','2','3']} (string enum values on integer type) — realistic route, wrong schema. Fix: scope enumCases-from-whereIn to the unsignatured path only (or guard on resolved type === string); add a signatured-int + whereIn regression test. Core unsignatured feature is otherwise approve-ready (reviewer-2). reviewer-1's round-1 verdict still pending.

Radiergummi:
🤖 **[coder]** addressed round 1 (finding 1): scoped the `whereIn`→`enum` derivation so string enum values can no longer land on a non-string typed param.

**Chosen option: (b)** — kept the derivation in `buildDescriptor()` but guarded it on the resolved type being `TypeIdentifier::STRING`. Picked (b) over (a) because the intended behavior includes a **signatured `string` param** with `whereIn` getting the enum (e.g. `string $status` + `whereIn([draft,published])`), which option (a) — restricting derivation to the unsignatured path — would have dropped. The STRING guard preserves that while excluding `int`/other types; a `?string` still qualifies (symfony `isIdentifiedBy` recurses through the nullable wrapper).

Regression tests (both in `UriParameterResolverTest`):
- signatured `int $id` + `whereIn` `1|2|3` → `whereKind = In`, `enumCases` null (no string enum on the integer schema)
- signatured `string $status` + `whereIn` `draft|published` → enum `[draft, published]` (intended behavior preserved)

Also took the non-blocking codecov nit: added an `ActionDescriptorTest` case covering the `uriPlaceholders()` no-placeholder early return (static route → `[]`).

Local gates green: Pint passed, PHPStan no errors, 3049 tests passed. Pushed `61273ea2`.

Radiergummi:
🤖 **[coder]** addressed round 1: scoped the whereIn->enum derivation to string-typed params + fixed the vacuous acceptance test; pushed `61273ea2`.

**Reviewer B finding 1 (must-fix) — took option (b), the resolved-type guard.** `buildDescriptor()` now only fills `enumCases` from a `whereIn` alternation when `$type->isIdentifiedBy(TypeIdentifier::STRING)`. A signatured `int $id` + `whereIn('id', [1,2,3])` (regex `1|2|3`, classified `WhereKind::In`) now stays a bare `{type: integer}` instead of getting `enum: ['1','2','3']`. Kept it in the shared `buildDescriptor` (per reviewer A's end-to-end trace that the shared placement is right) rather than scoping to unsignatured-only, so a signatured **string** param keeps the useful enum.
  - Tests (`UriParameterResolverTest`): 'surfaces a whereIn alternation as enum cases for a signatured string param' (positive) and 'does not attach a string whereIn enum to a signatured int param' (the case that was broken+untested). Verified the int test **fails without the guard** and passes with it.

**Reviewer A finding 1 (test-validity) — fixed the vacuous test.** The old 'emits zero path.parameter-undeclared findings' test asserted on an `ArrayFindingsCollector` the generator never wrote to (it passed on pre-fix code). Replaced it with the positive form: assert the generated spec's parameters for `/teams/{team}/members/{member}` include `team` (`in: path`, `required: true`) — which **fails on the pre-fix code** where `{team}` was dropped. Removed the now-orphaned imports.

**Optional codecov nit — covered.** Added 'exposes no placeholders for a static route' in `ActionDescriptorTest` for the `preg_match_all() === 0` early return in `uriPlaceholders()`.

Docs (`auto-derivation.md`) + CHANGELOG updated to qualify the whereIn->enum behaviour as string-typed-only. `composer check` green (3049 tests); full CI matrix green incl. swagger-php 5.8 and PHP 8.5/Laravel 12&13.

Radiergummi:
🤖 **[lead]** round-1 fix verified (sha 61273ea2): option-b STRING guard — whereIn→enum now only derives on TypeIdentifier::STRING, so signatured int+whereIn stays {type:integer} (no string enum); signatured-string keeps the useful enum. Coder also folded reviewer-A's finding-1 (vacuous test → positive form asserting generated /teams/{team}/members/{member} includes 'team', fails pre-fix) + the codecov static-route nit. Full CI matrix green incl. swagger-php 5.8 + snapshot + PHP 8.5×Laravel12/13. → re-review: reviewer-B re-checks its finding, reviewer-A re-confirms the changed diff.

Radiergummi:
🤖 **[reviewer]** review round 2 (reviewer A, independent): **approved, no concerns.**

Re-reviewed sha `61273ea2` on its own merits in a fresh isolated worktree. Both things I was asked to re-confirm hold; my round-1 finding-1 is properly addressed.

**1. The `whereIn`→`enum` STRING guard is correct.** I traced the enum source end-to-end: `UriParametersExtractor::applyWhereKindOverrides()` (`:115-117`) only copies `$descriptor->enumCases` onto the schema — it does **not** independently derive an enum from the constraint. So `buildDescriptor()` in `UriParameterResolver` is the **sole** source of `enumCases` for a `WhereKind::In` constraint, and it now gates the derivation on `$whereKind === WhereKind::In && $enumCases === null && $resolvedConstraint !== null && $type->isIdentifiedBy(TypeIdentifier::STRING)` (`UriParameterResolver.php:138-146`). There is no second, unguarded path. Consequences I verified:
- Signatured `int $id` + `whereIn('id', [1,2,3])` (regex `1|2|3`, classified `In`): `enumCases` stays `null`, so the extractor hits the `: null` no-op branch and the schema stays a bare `{type: integer}`. No string enum on an integer type. ✔
- Signatured `string $status` + `whereIn(['draft','published'])`: STRING-identified → `enum: ['draft','published']` preserved (the useful case option (a) would have dropped). ✔
- `?string` still qualifies (`isIdentifiedBy` recurses the nullable wrapper) — the round-1 reviewer note about nullable strings holds; the unit suite's optional-flag test confirms STRING identity. ✔
- Backed-enum signature still wins via the `$enumCases === null` clause; unsignatured path is always STRING so it can only ever benefit. ✔

**TDD on the guard (proved, not assumed):** removing only the `&& $type->isIdentifiedBy(TypeIdentifier::STRING)` clause makes *'does not attach a string whereIn enum to a signatured int param'* **fail** (1 failed); restoring it passes, and *'surfaces a whereIn alternation as enum cases for a signatured string param'* still passes (guard doesn't over-restrict). Both regression tests are in `UriParameterResolverTest` and pin the exact boundary.

**2. My round-1 finding-1 (the vacuous acceptance test) is properly addressed.** The old test asserted on an `ArrayFindingsCollector` the generator never wrote to (passed on pre-fix code). It's **gone** — `grep` finds **zero** `ArrayFindingsCollector`/`FindingsCollector` references in the test file, and the now-orphaned imports were removed (the file imports only `Request` + `Route`). It's replaced with the positive form *'it documents the unsignatured parent of a scoped binding so the linter has nothing to flag'* (`UnsignaturedPathParameterTest.php:154-160`), which reads the **generated spec's** parameters for `/teams/{team}/members/{member}` and asserts `team` is present (`in: path`, `required: true`). I confirmed it **fails on pre-fix code**: reverting all three production files (`ActionDescriptor`, `OperationBuilder`, `UriParameterResolver`) to `main` fails **11 of the 12** feature tests including this one (round 1: the vacuous version passed pre-fix; now it doesn't). The inline comment documents exactly why the positive assertion is the regression guard. Correct fix.

**Other checks:** the codecov nit is covered — *'it exposes no placeholders for a static route'* (`ActionDescriptorTest`) exercises the `preg_match_all() === 0` early return, passes. Full touched-test run: **37 passed** (12 feature + 4 unit ActionDescriptor + 21 unit resolver). `vendor/bin/pint --test` on the 3 prod files → passed; `composer lint` (PHPStan L8) → No errors. Boundary clean (`Support/Routing` + `Routing/`, no plugin/third-party dep). Docs (`auto-derivation.md`) + CHANGELOG updated and now correctly qualify the `whereIn`→enum behaviour as string-typed-only. No new lint rule (correct — this drives `path.parameter-undeclared` to zero on the affected routes).

Nit (non-blocking, same class as round 1): one new comment line `UriParameterResolver.php:132` is ~104 chars, just over the 100 soft limit; Pint passes so cosmetic. The other >100 lines in the touched files are pre-existing and correctly untouched (zero added lines in `OperationBuilder` exceed 100).

Verdict: **approved.** The STRING guard is the right, minimal fix and is TDD-pinned in both directions; my finding-1 test-validity gap is closed with a genuinely fail-without-the-fix positive assertion.

Radiergummi:
🤖 **[reviewer]** review round 2 (reviewer B): **approved, no concerns.**

My round-1 finding (the `whereIn`→`enum` derivation in `buildDescriptor()` landed string enum values on non-string typed params, e.g. `int $id` + `whereIn` → `{type: integer, enum: ['1','2','3']}`) is **fully resolved**. The derivation is now guarded on `$type->isIdentifiedBy(TypeIdentifier::STRING)` (`src/Support/Routing/UriParameterResolver.php:141`), kept in shared `buildDescriptor` so both signatured and unsignatured string params still get the useful enum.

Verified independently at `61273ea2` in an isolated worktree:

1. **Concern resolved.** `int $id` + `whereIn('1|2|3')` → `enumCases` null → bare `{type: integer}`; `string $status` + `whereIn('draft|published')` → enum preserved. Both covered by the new unit tests in `UriParameterResolverTest.php`.
2. **TDD confirmed.** Removing the STRING clause makes the int-param regression test fail (enum wrongly attached) while the string-param test still passes — the guard is load-bearing.
3. **Edge cases (the flagged risk) all correct:**
   - `?string $status` → enum **kept** — `Type::isIdentifiedBy` recurses through the `NullableType` wrapper to the inner `string`, so nullable strings are not missed.
   - `?int $id` → enum **dropped**.
   - `int|string $x` → enum kept (union has a string member). Not a defect: string values validate against the string branch and it's an unusual path-param type; unlike pure-`int` it is never a guaranteed mismatch. Consistent with the documented "string type" intent.
   - untyped param → resolves to `string` → enum kept.
4. **Full emit chain checked.** `UriParametersExtractor::applyWhereKindOverrides` sets `$schema->enum` only when `enumCases !== null`, so the int case emits no enum in the real spec.

Bookkeeping + gates: CHANGELOG and `docs/auto-derivation.md` both updated accurately (string-only enum, int caveat); PHPStan clean, `pint --test` clean on the touched file; all 104 `--group=routing` tests green on **PHP 8.4 and 8.5**.

Radiergummi:
🤖 **[lead]** two independent approvals (reviewer-A + reviewer-B, both clean on sha 61273ea2; reviewer-B confirms its round-1 whereIn-guard finding resolved, edge cases ?string/​?int/​int|string all correct; reviewer-A confirms finding-1 fix is a genuine fail-without-fix guard) → survey. Generation-affecting (area:params/core): synthesizes a path parameter for every URI placeholder absent from the signature. Layer A acceptance = 0 regressions (no paths/operations/schemas lost); expect MORE path params emitted + path.parameter-undeclared findings → 0. Queued behind #424 survey (surveyor ×1). Non-blocking nit carried: comment line UriParameterResolver.php:132 ~104 chars — fold into a sync, not a gate.

Radiergummi:
🤖 **[lead]** rebased onto origin/main after #424 merged (sha 61273ea2 → fa48e600). Only conflict was the expected CHANGELOG `[Unreleased]` append-overlap — resolved by keeping both #415 and #416 entries (the #416 line carries the string-typed-only whereIn qualification from the fix commit). Code unchanged; docs auto-merged. CI re-running on fa48e600; survey will run on the rebased sha once green.

Radiergummi:
🤖 **[surveyor → relayed by lead]** Layer A **PASS — 0 regressions across 11 apps. Target sha fa48e600.** Baseline regenerated on current main `181ba65f` (post-#424); candidate run against `fa48e600`, both checkouts verified clean. All deterministic count metrics flat baseline→candidate (paths/operations/apiOperations/responseSchemas/documentedResponses/requestBodies/componentSchemas), 0 crash regressions. Direct path-parameter delta (computed from specs; corpus has no `parameters` metric): **+823 path params added corpus-wide, 0 dropped** — exactly the intended effect (Coolify +346, AdvisingApp +169 from invokable controllers; SpeedtestTracker flat, no undeclared placeholders). (Surveyor blocked from gh comments; lead relaying.)

Radiergummi:
🤖 **[lead]** all gates green → squash-merged, closed #416, worktree removed. Two independent adversarial approvals (reviewer-A + reviewer-B, both TDD-verified; reviewer-B confirmed the approval transfers byte-identically to the rebased fa48e600). Layer A survey PASS on fa48e600: 0 count/crash regressions across 11 apps, path.parameter-undeclared findings 320→0 corpus-wide (+823 path params added, 0 dropped). Full CI matrix green incl. swagger-php 5.8 + snapshot. Touches 3 @internal files (UriParameterResolver/OperationBuilder/ActionDescriptor), 148/-16 src lines — right at the ~150 soft threshold; merged on the strength of adversarial 2-reviewer + survey + full-CI vetting and zero sensitive-surface flags (no Contracts/config/public-signature/stage-order change). Borderline size noted for the record.

Issue #427, opened 2026-06-20
feat(fractal): recognize the spatie/laravel-fractal API with a literal transformer ($this->fractal root + direct transformWith)


Radiergummi:
## Scope (narrowed — see decision comment below)

Extend the existing #263 Fractal call-shape reader to recognize **more of the
`spatie/laravel-fractal` package's own public API** when the transformer is a **literal**. This is
library-level convention support (same category as reading FormRequest / API Resources / Spatie
Data) — **not** app-specific idiom reading.

Concretely, broaden `FractalHelperResponseResolver::matchChainedShape` / `isFractalEntrypoint` to:

1. Accept an injected `Spatie\Fractal\Fractal` (Fractalistic) instance as a chain root — i.e. a
   `\$this->fractal->item(...)` / `\$this->fractal->collection(...)` property root — not only the
   `fractal()` helper and `Fractal::create()` facade already handled by #263.
2. Read the transformer from a **direct** `->transformWith(<literal>)` link in the chain, where
   `<literal>` is `new SomeTransformer()` or `SomeTransformer::class` — the forms the spatie package's
   own documentation shows.
3. Keep item-vs-collection from the literal `->item()`/`->collection()` method already in the chain;
   reuse `SchemaFromTransformer` (the transformer→schema engine) unchanged.

Never-wrong boundary unchanged: matches only on a **literal** transformer + literal envelope method;
degrades gracefully (logged note) otherwise. `#[FractalResponse]` remains authoritative.

## Explicitly out of scope (rejected — see comment)

- **No InvoiceNinja base-controller detector** (`\$entity_transformer` property + `itemResponse`/
  `listResponse`). That is one app's private base controller, not a shared convention.
- **No indirection unwrapping** (e.g. Pelican's `\$this->getTransformer(X::class)`). `getTransformer()`
  is the app's own method, not spatie's API. Reading it would be app-specific.

Bespoke indirection between "the transformer exists" and "the controller returns it" is, by the
library's inference philosophy, authoring-attribute territory — that is what `#[FractalResponse]` is
for.

## Acceptance

- An action that returns `\$this->fractal->collection(\$d)->transformWith(new XTransformer())` (or
  `XTransformer::class`) emits the collection envelope + transformer-derived payload `\$ref`; the
  `->item(...)` form emits the item envelope. Both work with the `fractal()` helper and facade roots
  too (no regression to #263).
[...]

PR #428, opened 2026-06-20
feat(#413): three-way response-coverage metric (substantive · correctly-empty · genuinely-missing)

### LINKED ISSUE #413: feat(survey): three-way response-coverage metric (substantive · correctly-empty · genuinely-missing)
## Problem

The survey's two headline numbers are each wrong, in opposite directions:

- **object-payload %** (`responseSchemas / apiOperations`) is *pessimistic* — it counts a correct
  empty 2xx (a `void` / `noContent` action) as a miss.
- **`completenessPercent`** is *optimistic* — it credits a *give-up* empty 2xx (the action returns a
  body we couldn't resolve) as "complete."

Example (InvoiceNinja, `main @ 63082823`): `completenessPercent` = **88.3%**, but only
**67 substantive + 16 correctly-empty ≈ 16%** are honestly good; the other ~440 are give-up empties
(Fractal + dynamic) that completeness counts as complete. The true coverage is between the two, and
neither metric finds it.

## Proposal

Classify each in-prefix operation into **three** buckets instead of two, by reflecting +
AST-classifying the route's action (not just reading the spec):

1. **substantive** — a 2xx with a real object/array/scalar/map payload (today's `responseSchemas`).
2. **correctly-empty** — the action is genuinely no-content (`void`/`never` return, `return;`,
   `return null;`, `response()->noContent()`, single `return $this->noContent()`); an empty 2xx is
   the *right* answer.
3. **genuinely-missing** — the action *returns a body* but the generator emitted empty/thin
   (dynamic `json()`, Fractal, multi/conditional returns, unresolved service builders).

`genuinely-missing` is the real denominator future inference levers should be sized against.

A working prototype of the classifier exists (the spike used for ADR #412): join
`artisan route:list --json` to the generated spec, reflect each residual action, bucket by return
shape. Fold a lean version into `tools/survey/metrics.php` (and surface in `completeness.php`).

## Scope

- Harness-only; the library is not modified.
- Add the three counts (+ a `genuinelyMissingByShape` rollup) to the per-app metrics JSON and the
  synthesis candidates.
- Keep `completenessPercent` for continuity, but add the honest three-way breakdown alongside it.

## Follow-up

Once landed, re-baseline the full 11-app corpus on current main and use `genuinely-missing` to
choose the next inference lever (foreign-annotation reading vs Fractal maturation). Relates to #411,
#412.

Radiergummi:
Closes #413.

## Plan

The survey's two headline numbers each lie in opposite directions: `responseSchemas/apiOperations`
is pessimistic (counts a correct empty 2xx as a miss); `completenessPercent` is optimistic (credits a
give-up empty 2xx as complete). Split each in-prefix operation into three honest buckets.

The "correctly-empty vs returns-a-body" signal is **not in the spec** — it needs the action's return
shape. So:

1. **`tools/survey/classify.php` (new, app-context)** — a lean productionisation of the ADR-#412 spike
   (`analyze-routes.php`): boots the consumer app, reflects each in-prefix action, AST-classifies its
   return expression into a source-shape string, emits `classify.json` (one record per route).
2. **`surveyMetrics()` (pure) gains an optional `$classification` arg** — joins the source shapes with
   the spec's substantive-ness it already computes, and emits a `responseCoverage` block:
   `{substantive, correctlyEmpty, genuinelyMissing, genuinelyMissingByShape}`. The three buckets sum to
   `apiOperations`; `genuinelyMissing` is the real denominator to size future levers against. Emitted
   only when a classification is supplied; `completenessPercent` is kept untouched for continuity.
3. **`metrics.php` CLI** loads `classify.json` from the app dir when present.
4. **`completeness.php`** surfaces the three-way summary line when given `--classify=<classify.json>`.

Bucketing (per in-prefix op): substantive (spec) → substantive; else if the action's shape/return-type
is no-content (`void`/`never`, `return;`, `response()->noContent()`, `return null;`) → correctly-empty;
else → genuinely-missing (rolled up by shape; unmatched shapes count conservatively as genuinely-missing).

Harness-only — the library is not modified.

## Verify
- New tests in `tests/Tools/SurveyMetricsTest.php` for the three-way split + the no-classification path.
- `composer test`, `vendor/bin/pint --test`, `composer lint` green.

## Follow-up (separate)
Re-baseline the 11-app corpus on current main and use `genuinely-missing` to size levers (#411, #427).

### DIFF
--- a/tests/Tools/SurveyMetricsTest.php
+++ b/tests/Tools/SurveyMetricsTest.php
@@ -66,6 +66,64 @@ it('counts only substantive 2xx into responseSchemas, contentless ops into docum
         ->and($m['completenessPercent'])->toBe(66.7);
 });
 
+it('omits responseCoverage when no classification is supplied', function (): void {
+    $spec = json_decode((string) file_get_contents(__DIR__ . '/../Fixtures/survey/spec.json'), true);
+    $run = ['generateExit' => 0, 'lintExit' => 0, 'generateStderr' => false, 'bootOutcome' => 'booted'];
+
+    $m = surveyMetrics($spec, ['findings' => []], $run, '/api');
+
+    expect($m)->not->toHaveKey('responseCoverage');
+});
+
+it('splits in-prefix ops into substantive / correctly-empty / genuinely-missing with a classification', function (): void {
+    // Three in-prefix ops: one substantive, one genuinely no-content (void), one that returns a
+    // body the generator could not resolve (empty-schema 200). The contentless vs give-up split
+    // can only come from the action source shape, not the spec.
+    $spec = ['paths' => [
+        '/api/show/{id}' => ['get' => ['responses' => ['200' => ['content' => [
+            'application/json' => ['schema' => ['type' => 'object', 'properties' => ['id' => ['type' => 'integer']]]],
+        ]]]]],
+        '/api/destroy/{id}' => ['delete' => ['responses' => ['204' => []]]],
+        '/api/dynamic' => ['get' => ['responses' => ['200' => ['content' => [
+            'application/json' => ['schema' => ['type' => 'object']],
+        ]]]]],
+    ]];
+
+    $run = ['generateExit' => 0, 'lintExit' => 0, 'generateStderr' => false, 'bootOutcome' => 'booted'];
+
+    $classification = [
+        ['uri' => '/api/show/{id}', 'verb' => 'get', 'shape' => 'resource::make', 'returnType' => 'JsonResource'],
+        ['uri' => '/api/destroy/{id}', 'verb' => 'delete', 'shape' => 'no return (void-like)', 'returnType' => 'void'],
+        ['uri' => '/api/dynamic', 'verb' => 'get', 'shape' => 'response()->json(<non-literal>)', 'returnType' => 'JsonResponse'],
+    ];
+
+    $m = surveyMetrics($spec, ['findings' => []], $run, '/api', null, $classification);
+
+    expect($m['responseCoverage']['substantive'])->toBe(1)
+        ->and($m['responseCoverage']['correctlyEmpty'])->toBe(1)
+        ->and($m['responseCoverage']['genuinelyMissing'])->toBe(1)
+        ->and($m['responseCoverage']['genuinelyMissingByShape'])->toBe(['response()->json(<non-literal>)' => 1])
+        // The three buckets partition apiOperations exactly.
+        ->and($m['responseCoverage']['substantive'] + $m['responseCoverage']['correctlyEmpty'] + $m['responseCoverage']['genuinelyMissing'])
+        ->toBe($m['apiOperations']);
+});
+
+it('counts a non-substantive op with no classification record as genuinely-missing (conservative)', function (): void {
+    $spec = ['paths' => [
+        '/api/mystery' => ['get' => ['responses' => ['200' => ['content' => [
+            'application/json' => ['schema' => ['type' => 'object']],
+        ]]]]],
+    ]];
+    $run = ['generateExit' => 0, 'lintExit' => 0, 'generateStderr' => false, 'bootOutcome' => 'booted'];
+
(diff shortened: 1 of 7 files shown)


Radiergummi:
🤖 **[lead]** admitted for independent code-review (review-only run; will escalate to needs-human with findings, not auto-merge).

Radiergummi:
🤖 **[reviewer]** review round 1: approved — 4 non-blocking findings.

Reviewed adversarially against #413. The three-way bucketing is correct and deterministic: substantive comes from the spec's existing `$hasSubstantiveResponse`; non-substantive ops join the classification index by normalised `VERB /path` (param-collapsed, verb-uppercased on both sides — consistent), then split correctly-empty (`survey_isNoContentShape`: `void`/`never` return type, or one of the no-content shapes / `return null;`) vs genuinely-missing (everything else, bucketed by shape; missing/`null` record → `unclassified`, conservatively genuinely-missing — matches the spec). The three buckets partition `apiOperations` exactly. Local gates green on the branch: `pest` (23 passed, the new metrics + synthesize tests included), `pint --test`, `phpstan` (no errors). Harness-only, no `src/` change → no CHANGELOG/docs/linting.md obligation; `methodology.md` updated and matches the implementation.

Findings (all nice-to-have, none blocking):

1. **`tools/survey/classify.php:409` — invokable controllers are mislabeled `closure/invokable-or-unknown` and thus land in `genuinelyMissing` even when `__invoke` is void/`noContent`.** This is **already filed and scoped out as #450** ("found while auditing the #413 re-baseline"), so it is a knowingly-deferred limitation, not a defect introduced here — but it means the `genuinelyMissing` numbers this PR emits are by-design inflated until #450 lands. Calling it out so the re-baseline follow-up reads them with that caveat.

2. **`tools/survey/synthesize.php:756` — `arsort($coverageByShape)` relies on implementation-defined tie ordering.** Ties (equal counts) resolve by insertion order in current PHP, which is deterministic *given* the fixed results order + alphabetical `ksort` upstream in metrics.php, so the snapshot test passes — but this is inconsistent with the recurring-lint rollup, which uses an explicit `count DESC then key ASC` tie-break (and has a test proving it). Recommend the same explicit tie-break here for robustness + consistency; the new test (counts 3 vs 2) does not exercise a tie.

3. **`tools/survey/classify.php:252` — dead capture.** The `$methodReturns` closure does `use (&$methodReturns)` but never recurses into itself (it uses the inner `$walk`); the by-ref capture is unused. Cosmetic.

4. **`tools/survey/metrics.php:612` — same-key path collision.** Two distinct in-prefix routes normalising to the same `VERB {}` key (e.g. `/users/{id}` vs `/users/{slug}`) collapse to the last record in the index; both spec ops then join it. Rare, best-effort for a measurement metric — noting only.

The crux the lead flagged (thin empty-`$ref` envelopes counting as substantive; 204 handling) is governed by the pre-existing `survey_substantive` / `$hasSubstantiveResponse` path, unchanged by this PR; 204/no-content is handled correctly here via the classification join. Nothing blocks merge.

Radiergummi:
🤖 **[lead]** Independent code-review complete → **approved**, 4 non-blocking findings. Bucketing is correct + deterministic; local gates green. Note F1 = the already-filed **#450** invokable-mislabel, which by-design inflates genuinelyMissing until #450 lands — read the re-baseline numbers with that caveat; F2 (explicit count-DESC/key-ASC tie-break in synthesize.php:756 for consistency with the lint rollup) is worth doing. Harness-only. Ready for your sign-off.

Radiergummi:
Looks good to me

Radiergummi:
🤖 **[lead]** Triage rebase onto current `main` (was 41 commits behind — green CI was stale). Clean rebase; force-pushed `48c1c44b`; CI re-running against current main. **Still needs from @Radiergummi:** sign-off — approved, harness-only (`area:survey`). Non-blocking F2 (explicit count-DESC/key-ASC tie-break at `synthesize.php:756`) optional; read genuinelyMissing with the #450 caveat noted earlier.

Issue #418, opened 2026-06-20
feat(params): use @param docblock descriptions as path/query parameter descriptions


Radiergummi:
## Problem

Parameter descriptions are only ever sourced from `#[PathParam]`/`#[QueryParam]` attributes. A plain
`@param string \$slug The post slug.` on the controller method is discarded, so well-documented
vanilla controllers emit parameters with no description.

## Evidence

- `src/Support/Routing/DocCommentParser.php:42-46` — the route-method docblock is parsed by the string-based `DocCommentParser`, which strips all `@`-lines; the richer `DocBlockParser` (phpstan/phpdoc-parser, which exposes `ParamTagValueNode.description`) is not applied to action methods.
- `src/Support/Extraction/UriParametersExtractor.php:82-84,151-163` — parameter description comes only from a `FieldAttribute`.

## Proposal (Tier-0, never-wrong)

Parse the action method docblock `@param` tags (via the existing `DocBlockParser`) into a
`name → description` map on `ActionDescriptor`. In `UriParametersExtractor` and the query resolvers,
fall back to that description when no attribute/validation description is present. Attributes and
inline-validate descriptions keep precedence, so this only fills empties.

## Scope / Out of scope

- In: path + query parameter description text; lowest precedence.
- Out: using `@param` types (signature/`where` types already win); request-body field descriptions.

## Acceptance

`@param string \$slug The post slug.` on an action with `/{slug}` yields that text as the path
parameter description; an explicit `#[PathParam(description: …)]` still wins; absence is a no-op.

<sub>From the 2026-06-20 lever audit.</sub>

Radiergummi:
🤖 **[lead]** admitted (pipeline had room; #424 merged) → planning. Tier-0 area:core/params: use @param docblock descriptions as path/query parameter descriptions.

PR #430, opened 2026-06-21
feat(#417): surface FormRequest rules as query params on GET/HEAD and stop emitting a GET body

### LINKED ISSUE #417: feat(params): surface FormRequest rules as query parameters on GET/HEAD (and stop emitting a GET request body)
## Problem

A `FormRequest` type-hinted on a GET/HEAD action is rendered as a **request body**, never as query
parameters — and GET bodies are discouraged by OAS. Inline `validate()` on GET is already correctly
flattened to query params, so the two idioms behave inconsistently and FormRequest-validated
index/search endpoints document no query params.

## Evidence

- `src/Plugins/Core/Resolvers/FormRequestRequestSchemaResolver.php:33` — gates only on `\$descriptor->method === null` (the reflection method), **no HTTP-verb check**.
- `src/Support/Extraction/RequestBodyExtractor.php:40-54,71` — resolvers run for every verb; only POST/PUT/PATCH gate the empty-finding.
- Contrast `src/Plugins/Core/Resolvers/CoreQueryParameterResolver.php:169-262` — inline `validate()` on bodyless verbs → query params (with dotted→wire flattening).

On GET/HEAD, `FormRequestRequestSchemaResolver` returns null (suppress the body); a query path reads
the same FormRequest `rules()` and flattens them to query params via the existing
`ValidationRulesToSchema` + the `CoreQueryParameterResolver::flattenField` wire-notation flattener.
Same rules read, redirected by verb.

## Scope / Out of scope

- In: GET/HEAD only; reuse the existing mapper + flattener; drop the GET body.
- Out: DELETE (ambiguous body-vs-query — keep current behavior); FormRequests whose `rules()` can't be read statically (already degrades with a notice on the inline path).

## Acceptance

`index(SearchRequest \$r)` on a GET route emits its rules as query params (wire notation) and no
request body; the same FormRequest on POST still emits a body. No GET operation in the corpus carries
a requestBody.

<sub>From the 2026-06-20 lever audit; also fixes a latent GET-request-body correctness bug.</sub>

Radiergummi:
Closes #417

> **Reviewer amendments (plan-review, fold in before coding):**
> 1. **`FormRequestRulesReader::read()` returns the RAW `rules()` array + degrade reason, not a mapped result.** True symmetry with `InlineValidatorRulesReader` (which returns raw rules; `process()` lives in the consumer). The reader does only: instantiate + `SpecTimeRequest::resolveConstructorDeps`/`configure` + `rules()` + try/catch → `{rules: ?array, degradeReason: ?string}` (a small `FormRequestRulesResult` VO). Both `SchemaFromFormRequest::buildSchema()` and the new query source call `rulesMapper->process()` themselves. This keeps the body path's existing `process()` + downstream (`setHasFileFields`, constant overrides, example synthesis) untouched — the extract-method stays a pure refactor and the snapshot byte-identical.
> 2. **No `descriptions` in the FormRequest query source.** `ValidationRulesToSchema::process()` returns only `{fields, additionalPropertiesField}` — no descriptions; FormRequest fields have no AST trailing-comment source. Pass `descriptions: []` to `flattenField`. Drop `descriptions` from the reader contract.
> 3. **Precedence: after the inline-validate loop, before the attribute loop** (accessor → inline-validate → FormRequest → `#[QueryParam]`). `#[QueryParam]` still wins; FormRequest sits at the inline-validate tier. Update the `CoreQueryParameterResolver` class docblock source list.
> 4. CorePlugin wiring needs **no edit** (resolvers are class-string registered + container-resolved; all `#[Scoped]`). Full rationale in the plan-review comment.

### Tier & rationale

**Tier 0** — same `FormRequest::rules()` read the body path already does, redirected by HTTP verb;
reuses the existing `ValidationRulesToSchema` mapper and the `CoreQueryParameterResolver::flattenField`
wire-notation flattener. No new inference, no body parsing beyond what already happens. Query params
for a GET search endpoint are *more* correct than a discouraged GET request body, and the rules are
the authoritative source — never-wrong.

- `FormRequestRequestSchemaResolver::resolveRequestSchema()` (`:33`) gates only on
  `$descriptor->method === null` — **no HTTP-verb check** — so a `FormRequest` on a GET action becomes
  a request body via `RequestBodyExtractor` (`RequestBodyExtractor.php:40-54`, which runs resolvers for
  every verb).
- The inline `validate()` idiom already does the right thing: `CoreQueryParameterResolver`
  (`:169-262`) reads inline rules on bodyless verbs (GET/HEAD), maps them via
  `ValidationRulesToSchema::process()` → `FieldDescriptor`s, and flattens each through `flattenField`
  (`:275-340`, wire notation `filter[name]` / `ids[]`, arrays-of-objects dropped). DELETE is skipped
  (body-vs-query ambiguous). So the two idioms are inconsistent: inline→query, FormRequest→body.
- The FormRequest rules read (instantiate + `SpecTimeRequest::configure()` + `rules()` + try/catch +
  degrade finding) lives in `SchemaFromFormRequest::buildSchema()` (`:107-135`), producing the same
  `rulesMapper->process()` result (`['fields' => FieldDescriptor[], 'additionalPropertiesField' => …]`)
  that `flattenField` consumes.

### Design — two halves, both verb-gated on GET/HEAD

**Half 1 — suppress the GET/HEAD body.** In `FormRequestRequestSchemaResolver::resolveRequestSchema()`,
return `null` when `$descriptor->httpMethod` is GET or HEAD (mirror
`CoreQueryParameterResolver::BODYLESS_METHODS = [HttpMethod::Get, HttpMethod::Head]`). DELETE stays
unchanged (out of scope; same reasoning the inline path already uses). This alone stops the GET
request body. `RequestBodyExtractor`'s empty-finding only fires on POST/PUT/PATCH (`:71`), so
returning null on GET emits no spurious `request.empty` finding — verify, no change needed there.

**Half 2 — surface the FormRequest rules as query params on GET/HEAD.** Add a third source to
`CoreQueryParameterResolver` alongside the inline-validate and accessor sources: when the action is
GET/HEAD and a `FormRequest` is type-hinted on the method, read its `rules()` → `FieldDescriptor`s and
run each through the existing `flattenField`. Precedence: place it so `#[QueryParam]` attributes still
win (attributes are the authoring escape hatch); relative to inline `validate()` and accessor reads,
order by the same "later wins on name collision" rule already documented in the resolver's class
docblock — slot the FormRequest source between accessor reads and inline-validate, or alongside
inline-validate (a route rarely has both a FormRequest *and* inline `validate()`); pick one and assert
it. **Recommendation:** treat the FormRequest source at the same precedence tier as inline `validate()`
(both are "validated input"), applied right after it, since a method won't typically have both.

**Reuse without duplication (the key structural choice).** The FormRequest rules-read currently only
exists inside `SchemaFromFormRequest::buildSchema()` coupled to schema-building. Extract the
rules→descriptors step into a small reusable reader so both the body path and the new query path share
it (symmetric with how `InlineValidatorRulesReader` already serves the inline path):

- New **`src/Plugins/Core/Support/FormRequestRulesReader.php`** (`@internal`, `#[Scoped]`):
  `read(class-string<FormRequest>): ?FormRequestRulesResult` returning
  `{fields: array<string,FieldDescriptor>, additionalPropertiesField: ?FieldDescriptor,
  descriptions: array<string,string>, degradeReason: ?string}` — encapsulating the instantiate +
  `SpecTimeRequest::resolveConstructorDeps`/`configure` + `rules()` + try/catch +
  `rulesMapper->process()`. Returns null (with a degrade reason) when `rules()` isn't statically
  readable — exactly how the inline path degrades with a notice.
  - Refactor `SchemaFromFormRequest::buildSchema()` to consume this reader for the rules→descriptors
    step (keeping its schema assembly, `#[RawSchema]`, file-field detection, constant-override, and
    degraded-finding behaviour intact). This is a **surgical extract-method refactor**, not a rewrite;
    the body-path output must stay byte-identical (guarded by the existing FormRequest body tests +
    the snapshot group).
  - `CoreQueryParameterResolver` gains `FormRequestRulesReader` + `PayloadParameterScanner` (to detect
    the FormRequest on the method, the same way `FormRequestRequestSchemaResolver` does via
    `scanner->candidateOfType($method, FormRequest::class)`) as constructor deps, and a new private
    `formRequestParameters(ActionDescriptor): array<string,OA\Parameter>` that mirrors
    `inlineValidationParameters()`: GET/HEAD gate, read rules, log a notice + degrade when unreadable,
    `flattenField` each descriptor (reusing the existing private method untouched), and emit the
    dropped-key notice for keys that can't be query-expressed.

  *Whether `descriptions` exist for FormRequest fields:* the inline path pulls field descriptions from
  the validator scan; FormRequest fields may have none — pass an empty `descriptions` map (or whatever
  `rulesMapper->process()` already surfaces) so `flattenField` simply omits descriptions. Confirm the
  mapper result shape during coding; do not invent a description source.

### Files

- **Modify** `src/Plugins/Core/Resolvers/FormRequestRequestSchemaResolver.php` — GET/HEAD → null.
- **New** `src/Plugins/Core/Support/FormRequestRulesReader.php` (+ a small result VO, e.g.
  `FormRequestRulesResult`) — shared rules→descriptors reader.
- **Modify** `src/Plugins/Core/Support/SchemaFromFormRequest.php` — consume the new reader (extract-method).
- **Modify** `src/Plugins/Core/Resolvers/CoreQueryParameterResolver.php` — add the FormRequest query
  source; new constructor deps; new private method; update the class docblock's source list.
- **Plugin wiring:** check `src/Plugins/Core/CorePlugin.php` for how these resolvers are constructed —
  the new constructor deps on `CoreQueryParameterResolver` and the new reader must be registered
  (likely already container-resolved since everything is `#[Scoped]`; confirm and adjust the plugin's
  resolver registration if it hand-news them).

No `Contracts`/public-signature/config change. The `RequestSchemaResolver` /
`QueryParameterResolver` interfaces are untouched (only the concrete resolvers' internals change).

### Tests (TDD — write first)

Mirror `tests/Feature/InlineValidationSchemaTest.php:183-187` and `MultiVerbGatingTest.php` (the inline
GET→query, no-requestBody pattern). New **`tests/Feature/FormRequestQueryParameterTest.php`**:

1. **Acceptance core:** `index(SearchRequest $r)` on `Route::get` → operation has query parameters in
   wire notation matching the FormRequest's rules, and **no `requestBody`**. Assert specific params
   (a scalar, a `nullable`/`sometimes` → not-required, a nested `filter.name` → `filter[name]`, a
   scalar array → `ids[]`).
2. **Same FormRequest on POST still emits a body** (no regression): `store(SearchRequest $r)` on
   `Route::post` → `requestBody` present, no flattened query params from the FormRequest.
3. **HEAD** behaves like GET (body suppressed, query params emitted).
4. **DELETE unchanged** (out of scope): a FormRequest on a DELETE route keeps today's behaviour
   (still a body) — lock it so the scope boundary is explicit.
5. **Unreadable rules() degrades** on GET: a FormRequest whose `rules()` can't be read statically →
   no query params + a logged notice (assert via the test logger, as the inline path tests do), no
   crash, no body.
6. **Arrays-of-objects / unrepresentable keys dropped** with a notice (reuse a fixture with a
   `filter.*.id`-shaped rule) — proves the flattener's drop path fires for the FormRequest source too.
7. **`#[QueryParam]` still wins** over a flattened FormRequest field of the same name (precedence guard).
8. **Corpus-shape guard / acceptance:** add a test (or extend an existing multi-verb test) asserting a
   GET route with a FormRequest carries no `requestBody` key at all.

Unit: **`tests/Unit/Core/.../FormRequestRulesReaderTest.php`** — reads a fixture FormRequest's rules to
descriptors; returns null + degrade reason when `rules()` throws / is unreadable; reuse existing
fixtures (`SimpleFormRequest`, `WildcardFormRequest`, a throwing one). And a focused test that
`SchemaFromFormRequest` body output is unchanged after the refactor (the existing
`FormRequestSchemaTest` + snapshot group already cover this — run them, add one if a gap shows).

Edge cases explicitly covered: GET vs POST vs HEAD vs DELETE; nested/dotted→wire; scalar array→`[]`;
arrays-of-objects dropped; unreadable rules degrade; attribute precedence; no-requestBody on GET.

### Docs

`docs/auto-derivation.md` — the request-body / query-parameter sections. State that a `FormRequest`
type-hinted on a **GET/HEAD** action is surfaced as **query parameters** (wire notation), not a
request body (GET bodies are discouraged by OAS); on body verbs it remains a request body. Mention
DELETE keeps its current body behaviour (ambiguous) and that unreadable `rules()` degrades with a
notice, consistent with the inline `validate()` path. Cross-check `docs/request-bodies.md` for a
FormRequest section and add the verb nuance there too if present.

### CHANGELOG

`[Unreleased]` (Changed + Fixed):
> A `FormRequest` type-hinted on a GET/HEAD action is now surfaced as query parameters (wire
> notation) instead of a request body, matching the inline `validate()` behaviour and avoiding a
> discouraged GET request body. POST/PUT/PATCH continue to emit a request body; DELETE is unchanged.

### Lint rule

None added. Existing findings unaffected; this removes invalid GET request bodies that no rule
currently flags (could note a future `request-body.on-get` rule, but **out of scope**).

- **Output-shape change for affected routes** (GET FormRequest: body → query params). This is the
  intended lever + a latent correctness fix, but it changes generated specs → **generation-affecting,
  survey gate applies.** Acceptance: no GET operation in the corpus carries a `requestBody`; the same
  FormRequest on POST still emits a body.
- **Refactor of `SchemaFromFormRequest`** (extract the rules reader) touches a hot, snapshot-covered
  path. Low risk if done as a pure extract-method, but the reviewer should confirm the body-path
  output is byte-identical (snapshot group is the guard; the swagger-php 5.x job excludes snapshots,
  so also run the normal suite).
- **No `Contracts`/public-signature/config/stage-order change.** Concrete resolvers gain constructor
  deps (all `#[Scoped]`, container-resolved) — confirm `CorePlugin` wiring doesn't hand-construct them
  with a fixed arg list; if it does, that's the one wiring edit to make.
- Precedence decision (FormRequest query source vs inline-validate vs accessor) is a small judgment
  call — recommended above (same tier as inline-validate, after it); flag for the reviewer to confirm.


Radiergummi:

Two halves, both GET/HEAD-gated: (1) FormRequestRequestSchemaResolver returns null on GET/HEAD (suppress the body; the empty-finding only fires on POST/PUT/PATCH so no spurious request.empty); (2) CoreQueryParameterResolver gains a FormRequest source that reads the same rules() and runs the descriptors through the existing flattenField. The key structural choice: the FormRequest rules-read currently lives inside SchemaFromFormRequest::buildSchema() — plan extracts it into a shared FormRequestRulesReader (symmetric with InlineValidatorRulesReader) consumed by both the body path and the new query path, as a surgical extract-method so the body-path output stays byte-identical (snapshot-guarded). DELETE stays out of scope.

Flags: generation-affecting (GET FormRequest: body→query) → survey gate, acceptance = no GET op carries a requestBody. The SchemaFromFormRequest refactor touches a snapshot-covered hot path — reviewer should confirm byte-identical body output. No Contracts/signature/config change; one wiring check in CorePlugin for the new constructor deps. Precedence call (FormRequest source same tier as inline-validate) flagged for confirmation.

Radiergummi:

Radiergummi:
🤖 **[lead]** plan posted on draft PR #430 → plan-review. Tier-0 verb redirect (GET/HEAD): FormRequestRequestSchemaResolver returns null on GET/HEAD (suppress GET body), CoreQueryParameterResolver gains a FormRequest query source via the EXISTING flattenField. Structural choice to scrutinize: extract the FormRequest rules-read out of SchemaFromFormRequest::buildSchema into a shared FormRequestRulesReader (symmetric w/ InlineValidatorRulesReader) — reviewer MUST confirm body-path output stays byte-identical (snapshot group is the guard; swagger-php 5.x CI excludes snapshots, so run normal suite too). Also confirm: new ctor deps on CoreQueryParameterResolver wiring in CorePlugin; precedence (FormRequest source == inline-validate tier); DELETE out of scope. Generation-affecting → survey later (acceptance: no GET op carries a requestBody).

Radiergummi:
🤖 **[reviewer]** plan-review (reviewer, independent): **approve-with-changes.** Tier-0 is right, scope is surgical, the two-half design is correct, and all four flagged judgment calls check out against the code. Two refinements to fold into the plan before coding (one is a real symmetry mismatch in the proposed reader contract), plus confirmations.

**Confirmations (grounded against the code):**

1. **GET/HEAD body suppression + no spurious `request.empty`** — confirmed. `RequestBodyExtractor::emitEmptyFindingIfWriteMethod()` (`RequestBodyExtractor.php:69-71`) gates strictly on `[Post, Put, Patch]`, so `FormRequestRequestSchemaResolver` returning null on GET/HEAD emits no `request.empty`. No change needed there, as the plan says. The current gate at `FormRequestRequestSchemaResolver.php:33` is indeed verb-blind (only `method === null`).

2. **CorePlugin wiring (flag #4) — no edit needed.** `CorePlugin` registers resolvers by **class-string** (`addQueryParameterResolver(CoreQueryParameterResolver::class)`, `CorePlugin.php:62`), and `OpenApiRegistry::addQueryParameterResolver(string $class)` stores the class-string for later container resolution. Every collaborator is `#[Scoped]`. So adding ctor deps to `CoreQueryParameterResolver` and a new `#[Scoped] FormRequestRulesReader` is container-resolved automatically — the plan's "one wiring edit if it hand-news them" hedge resolves to **no edit**. Coder should still grep for any direct `new CoreQueryParameterResolver(` (there is none today) but not spend effort on plugin registration.

3. **DELETE out of scope** — acceptable and consistent. The inline path already refuses DELETE (`CoreQueryParameterResolver.php:174,186-198`, "validated fields may be body or query string"). Matching that for FormRequest keeps the two idioms symmetric; the body stays today's behaviour. Lock it with test #4 as planned.

4. **Body-path byte-identical guard** — the guard exists: `FormRequestSchemaTest` + the `snapshot` group. The plan correctly notes the swagger-php 5.x CI job excludes the snapshot group, so the **normal suite must also assert the body shape** for the refactored path — `FormRequestSchemaTest` is non-snapshot and does that. Good.

**Change 1 (required) — the proposed `FormRequestRulesReader` contract breaks the symmetry it claims, and over-reaches.** The plan models the new reader after `InlineValidatorRulesReader` and has it return `{fields: FieldDescriptor[], additionalPropertiesField, descriptions, degradeReason}` — i.e. **already mapped through `rulesMapper->process()`**. But `InlineValidatorRulesReader::read()` returns **raw rules** (`InlineValidationScanResult` with `->rules`, `->descriptions`, `->skippedFields`); the `process()` call lives in the *consumer* (`CoreQueryParameterResolver.php:223` and `SchemaFromFormRequest.php:135`). For a truly symmetric, single-responsibility seam, `FormRequestRulesReader::read()` should return the **raw `rules()` array + degrade reason** (the instantiate + `SpecTimeRequest::resolveConstructorDeps`/`configure` + `rules()` + try/catch), and leave `process()` to each consumer — exactly as the inline reader does. That also keeps `SchemaFromFormRequest`'s existing `process()` call (and everything downstream of it: `setHasFileFields`, constant overrides, example synthesis) **on the body path untouched**, which is what makes the refactor a clean extract-method and the snapshot byte-identical. Mapping inside the reader would instead force the body path to consume a pre-mapped result and risk shifting that hot path. Recommend: reader returns `{rules: ?array, degradeReason: ?string}` (a small `FormRequestRulesResult` VO); both consumers call `rulesMapper->process()` themselves.

**Change 2 (required) — drop `descriptions` from the FormRequest source; don't pass a map the source can't populate.** Verified `ValidationRulesToSchema::process()` returns only `{fields, additionalPropertiesField}` (`:131-135`, `:173-176`) — **no descriptions**. Inline descriptions come from the AST trailing-comment scan (`InlineValidatorRulesReader.php:264-270`), which has no FormRequest analogue. So `formRequestParameters()` must pass `descriptions: []` to `flattenField` (the plan hints at this but still lists `descriptions` in the reader's return type — remove it from the reader contract per Change 1, and just pass `[]` at the call site). `flattenField` already tolerates an empty map (`$descriptions[$dottedPath] ?? null`). No invented description source — correct instinct, just tighten the contract.

**Precedence (flag #3) — endorse "after inline-validate, before attributes".** Current order in `resolveQueryParameters()` (`:57-71`) is accessor → inline-validate → attributes (later wins). Slotting the FormRequest source **after** the inline-validate loop and **before** the attribute loop keeps `#[QueryParam]` winning (the authoring escape hatch) and treats FormRequest as the same "validated input" tier as inline `validate()`. A route having both inline `validate()` and a FormRequest is pathological; on collision, last-write (FormRequest) wins, which is defensible. Update the class docblock source list (`:29-38`) to document the new source #2.5/#3 and the new ordering — the plan calls for this; make sure it lands.

**One mutation-safety note (no change needed, just confirm in coding):** `flattenField` mutates the `FieldDescriptor` it's handed (`->description`, `applyTo`). The body and query sources each call `process()` independently and are verb-disjoint (GET→query, POST→body), so there is no shared-descriptor aliasing between paths. Fine as-is; just don't cache/share a single `process()` result across both.

**Tests / docs / CHANGELOG / lint** — proving tests are well chosen (acceptance GET→query no-body; POST still body; HEAD like GET; DELETE unchanged; unreadable-rules degrade-with-notice; arrays-of-objects dropped via `WildcardFormRequest`-shape; `#[QueryParam]` precedence; no-`requestBody`-key guard). Add the `FormRequestRulesReaderTest` unit (raw-rules read + degrade on throwing `rules()`; reuse `SimpleFormRequest`/`WildcardFormRequest` + a throwing fixture — note no throwing FormRequest fixture exists yet, so the coder will add one). Docs target (`auto-derivation.md`, cross-check `request-bodies.md`) and the CHANGELOG `[Unreleased]` Changed+Fixed entry are named. No new lint rule (correct; a future `request-body.on-get` is rightly out of scope). Generation-affecting → survey gate applies (acceptance: no GET op carries a `requestBody`).

Net: the design is sound and lands at the lowest viable tier with no Contracts/signature/config/stage-order change. Fold in Change 1 (reader returns raw rules, mapping stays in consumers — preserves the byte-identical body path and the real `InlineValidatorRulesReader` symmetry) and Change 2 (no `descriptions` in the FormRequest source), confirm the precedence/docblock update, and this is ready to code.

Verdict: **approve-with-changes.**

Radiergummi:
🤖 **[lead]** plan-review approved-with-changes (reviewer-A) → coding. Two REQUIRED reader-contract refinements folded into the plan body: (1) FormRequestRulesReader returns {rules, degradeReason} RAW (NOT pre-mapped) so both consumers call rulesMapper->process() themselves — keeps the SchemaFromFormRequest extract-method pure + body output byte-identical (symmetric w/ InlineValidatorRulesReader); (2) no descriptions — ValidationRulesToSchema returns only {fields, additionalPropertiesField}, FormRequest has no AST-comment source, so pass descriptions: [] to flattenField + drop it from the reader contract. CorePlugin wiring = zero edits (class-string registered + container-resolved #[Scoped]). Precedence: accessor→inline-validate→FormRequest→attributes (attributes still win).

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed `07700416`. Ready for review.

Followed the plan + the 4 reviewer amendments:

- **`FormRequestRulesReader`** returns the RAW `{rules, degradeReason}` (no mapping, no descriptions) — symmetric with `InlineValidatorRulesReader`. Both `SchemaFromFormRequest::buildSchema()` and the new query source call `rulesMapper->process()` themselves. The reader stays pure (no logging/findings); each consumer handles degradation in its own idiom (body path: warning + `request-body.schema-degraded` finding; query path: generation-log notice). Body-path output is **byte-identical** — the `snapshot` group is green.
- **Half 1:** `FormRequestRequestSchemaResolver` returns null on GET/HEAD.
- **Half 2:** new `CoreQueryParameterResolver::formRequestParameters()`, slotted right after the inline-validate loop (same precedence tier; `#[QueryParam]` still wins), reusing `flattenField` untouched with `descriptions: []`.
- **CorePlugin:** zero wiring edits — resolvers are class-string registered + container-resolved, new `#[Scoped]` ctor deps resolve automatically. Confirmed.

Tests: new `FormRequestQueryParameterTest` (GET→query+no-body, POST-still-body, HEAD-like-GET, DELETE unchanged, unreadable-rules degrade, arrays-of-objects dropped, `#[QueryParam]` precedence, no-`requestBody` guard) + `FormRequestRulesReaderTest` (raw rules, nested/array rules, throws→degrade, no-`rules()`→degrade). Updated the 4 test sites that hand-construct `SchemaFromFormRequest`/`CoreQueryParameterResolver` for the new ctor deps.

Docs: `auto-derivation.md` (summary table + the GET/HEAD query-source paragraph) and `request-bodies.md` (verb nuance in the FormRequest intro). CHANGELOG `[Unreleased]` entry appended.

Radiergummi:
🤖 **[coder]** deviation from the plan (one factual correction on DELETE):

The plan and the request-bodies docs both state a DELETE FormRequest "keeps its current body behaviour (still a body)". That premise is **empirically wrong**: `OperationDescriptor::shouldAttachBody()` only attaches a `requestBody` for **POST/PUT/PATCH** — a DELETE (and GET/HEAD/OPTIONS) request body is stripped at serialization regardless of what the resolver builds. So a DELETE FormRequest never produced a body on `main`.

Consequence for scope: this change leaves DELETE genuinely **unchanged** — no body (as before) and no query params (the query source gates on GET/HEAD only, mirroring the inline-validate path which also excludes DELETE). I updated the DELETE feature test to assert the true behaviour (no `requestBody`, no query params) and added a comment citing `shouldAttachBody`, and worded the DELETE line in `request-bodies.md` / `auto-derivation.md` as "unchanged (no body, no query params)" rather than "still a body". No behaviour was changed for DELETE; only the scope-boundary assertion is now accurate.

Note: because `shouldAttachBody` already strips GET/HEAD bodies at serialization, the Half-1 resolver gate is partly belt-and-suspenders — but it is still worth keeping: returning null on GET/HEAD avoids registering an orphan component schema (and running the FormRequest body build side effects) for a body that would be stripped anyway.

Radiergummi:
🤖 **[lead]** real diff verified (sha 07700416; 222 src lines across 5 files in src/Plugins/Core, much of it the designed extract-method move into FormRequestRulesReader/FormRequestRulesResult), full CI matrix green (11/11 incl. swagger-php 5.8 + snapshot). → code-review (reviewer-A who plan-reviewed + reviewer-B, independent). NOTE for the merge gate: 222 src lines EXCEEDS the ~150 controversy threshold → this will ESCALATE to human at merge even when reviews + survey pass (not an auto-merge); we drive it to green+approved+surveyed so Moritz gets a merge-ready PR. Coder flagged a sound plan correction (deviation comment): DELETE never had a body (OperationDescriptor::shouldAttachBody is POST/PUT/PATCH only), so DELETE is genuinely unchanged; docs/test corrected. Branch is behind current main (post-#425) — final rebase + re-CI required before merge; defer until after review so the sha doesn't move under reviewers.

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer A, independent): **approved, no concerns.**

I plan-reviewed this (wrote the 4 amendments); I did not author the diff, so code-review is in bounds. Reviewed in a fresh isolated worktree.

**Sha note (important for the merge gate).** The task pointed me at `07700416`, but the branch head is now **`97739ce6`** — the branch was **rebased onto post-#425 main and force-pushed** (`07700416` is not an ancestor of `97739ce6`). I verified the rebase only *replayed* the #417 work: diffing the net #417 change at the old base (`181ba65..07700416`) against the new base (`63d7e7e7..97739ce6`), the **only** difference is one CHANGELOG context line (#425's `[Unreleased]` entry now sits above #417's) — every source/test/doc/fixture line of the #417 work is byte-identical across the rebase. So this review of `97739ce6` is equivalent to reviewing `07700416`, on top of correctly-merged current main. (The lead's "defer rebase until after review" note is already overtaken by events; no action needed — `97739ce6` is the right thing to merge, and CI is green on it.)

**All four plan amendments + both halves landed correctly:**

1. **Reader returns RAW `{rules, degradeReason}`; consumers map.** `FormRequestRulesReader::read()` does only instantiate + `SpecTimeRequest::configure()` + `rules()` + try/catch → `FormRequestRulesResult` (no `process()`, no logging, no findings) — genuinely symmetric with `InlineValidatorRulesReader`. Both `SchemaFromFormRequest::buildSchema()` and `CoreQueryParameterResolver::formRequestParameters()` call `rulesMapper->process()` themselves.

2. **Body path is byte-identical (extract-method verified, not assumed).** The `SchemaFromFormRequest` diff is a pure extract: the moved try/catch → reader; the degrade branch now keys on `$rulesResult->rules === null` and uses `$rulesResult->degradeReason` (which the reader sets to `$exception->getMessage()`), so the warning text and the `request-body.schema-degraded` finding message are unchanged; `emitDegradedFinding(Throwable)` → `(?string $reason)` only because its sole use was `->getMessage()`. The distinct **`!method_exists` "does not declare a rules() method"** branch is **retained before** the reader call (`SchemaFromFormRequest.php:94-104`), so that case still returns its own schema with no finding — the reader's own `method_exists` degrade is defensive for the query consumer and never reached on the body path. Proof: `tests/Feature/FormRequestSchemaTest` + `SchemaFromFormRequestTest` (27 passed) and the **`snapshot` group (9 passed)** are green; the swagger-php 5.8 job (which excludes snapshots) is covered by the non-snapshot body tests.

3. **GET/HEAD body suppression + no spurious `request.empty`.** `FormRequestRequestSchemaResolver` returns null on `[Get, Head]` *before* scanning, so no component schema is registered for a GET/HEAD FormRequest. `RequestBodyExtractor::emitEmptyFindingIfWriteMethod` fires only on POST/PUT/PATCH, so returning null on GET emits no `request.empty`. The coder's note that this gate is partly belt-and-suspenders (GET bodies are stripped at serialization by `shouldAttachBody` anyway) is correct, and keeping it is the right call — it avoids registering an orphan component + running the body-build side-effects for a body that would be stripped.

4. **Query source: precedence + `descriptions: []`.** `formRequestParameters()` is slotted in `resolveQueryParameters` **after** the inline-validate loop and **before** `attributeParameters`, keyed by name with last-write — so `#[QueryParam]` still wins (proven by the attribute-wins test). GET/HEAD-gated (`!in_array(httpMethod, BODYLESS_METHODS)`), DELETE excluded mirroring the inline path. Reuses `flattenField` untouched with `descriptions: []` (no FormRequest description source — correct). Degrade + dropped-key notices mirror the inline idiom.

**DELETE deviation — verified correct.** `OperationDescriptor::shouldAttachBody()` (`:115-124`) attaches a body only for POST/PUT/PATCH, so a DELETE FormRequest never produced a body on `main`. The plan/docs claim that DELETE "keeps its body" was factually wrong; the coder corrected the DELETE test to assert reality (no body, no query params) with a comment citing `shouldAttachBody`, and reworded the docs. The deviation is sound and documented as a PR comment per the workflow. DELETE is genuinely unchanged.

**Tests pin the behavior (TDD confirmed).** `FormRequestQueryParameterTest` (8 scenarios — GET→query in wire notation with `filter[name]`/`ids[]` + required/optional/types, POST-still-body, HEAD-like-GET, DELETE-unchanged, unreadable-rules degrade+notice, array-of-objects dropped+notice, `#[QueryParam]`-wins, no-requestBody guard) + `FormRequestRulesReaderTest` (4: raw rules, nested/array rules, throws→degrade, no-`rules()`→degrade) — **12 passed**. Reverting the two resolvers to base fails the 4 positive GET→query assertions (the absence-asserting tests pass trivially when the feature is off, as expected). All scenarios read the real generated spec, not a mocked collector.

**Gates:** `vendor/bin/pint --test` on the 5 changed src files → passed; `composer lint` (PHPStan L8) → No errors. The 4 hand-construction test sites for the new `#[Scoped]` ctor deps (`rulesReader:` / `formRequestRulesReader:` + `scanner:`) are updated; CorePlugin needs no wiring edit (class-string registered + container-resolved), as planned. Boundary clean (all in `src/Plugins/Core`; reader/result are `@internal`, the body extractor untouched). Docs (`auto-derivation.md`, `request-bodies.md`) + CHANGELOG present and now honest about DELETE.

Nit (non-blocking, same class as my prior reviews): ~10 new **doc-comment** lines are 101–105 chars, just over the 100 soft limit (class docblocks on `FormRequestRulesReader`/`FormRequestRulesResult` and the `CoreQueryParameterResolver` source-list; several are unavoidable inline `{@see FQCN}` references). Pint passes, and the pre-existing region already runs to 112 chars, so cosmetic. Codecov's 1 uncovered line is a degrade/notice branch — coverage is 98.5% on the patch and the degrade path is tested at the feature level; not a finding.

Verdict: **approved.** The extract-method is behavior-preserving (snapshot byte-identical), the verb gating and precedence are correct, the DELETE correction is right and documented, and the tests pin the behavior. (Per the lead: 222 src lines exceeds the ~150 controversy threshold, so this escalates to human at merge regardless — reviewed on its merits as merge-ready.) A final rebase is already effectively done; just re-confirm CI on the post-review sha before merge.

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer B, independent): **approve** — 2 minor non-blocking findings, no changes required to merge.

Reviewed sha `07700416` via the current PR head `97739ce6` (= `07700416` cleanly rebased over #425; `git diff 07700416 97739ce6` is exactly the #425 path-param content, so the #430 change is preserved). All checks below run against `97739ce6` in an isolated worktree.

**1. Body-path byte-identical refactor — verified clean (no semantic drift).** The extract into `FormRequestRulesReader`/`FormRequestRulesResult` preserves every observable on the body path:
- The early `method_exists('rules')` short-circuit in `SchemaFromFormRequest::buildSchema()` (`:94`) is kept verbatim (same message, trailing period), so the reader's own `method_exists` branch is dead on the body path and the "no rules()" output is unchanged.
- The try/catch covers the identical unit (`resolveConstructorDeps` + `new` + `configure` + `rules()`); the degraded path keeps the same order (warning → `emitDegradedFinding` → `setHasFileFields(false)` → same degraded `OA\Schema`).
- `emitDegradedFinding` derives file/line from `ReflectionClass($formRequestClass)` (`:204-207`), **not** the exception, so narrowing its param from `Throwable` to `?string $reason` loses nothing — `degradeReason` carries `$exception->getMessage()`, the exact string the warning + finding used before.
- Guard holds empirically: `--group=snapshot` (9) + all FormRequest tests (57) green.

**2. GET/HEAD gating — correct.** Half-1 returns null on `BODYLESS_METHODS=[Get,Head]` via strict-enum `in_array` (`httpMethod` is `?HttpMethod`, so identity comparison is right; null falls through safely). Half-2's `formRequestParameters()` faithfully mirrors `inlineValidationParameters()` — same gate, same `additionalPropertiesField → '*'` dropped-key handling, same `flattenField` loop and notice wording — with the two intended deltas: `descriptions: []` (no AST source) and no DELETE branch (gates strictly on GET/HEAD). `#[QueryParam]` precedence preserved: the FormRequest source is slotted before the attribute loop (`resolveQueryParameters` `:72-78`), so attributes still win on name collision (precedence test confirms the authored description wins).

**3. DELETE deviation — sound and accurately reflected.** `OperationDescriptor::shouldAttachBody()` is POST/PUT/PATCH-only, so DELETE never had a body; the change leaves it genuinely unchanged (no body, no query params). The DELETE test asserts this with a `shouldAttachBody` citation, and both `auto-derivation.md` and `request-bodies.md` say "unchanged (no body, no query params)" rather than the original "still a body". Correct call.

**4. TDD confirmed.** Reverting both prod halves fails 4 query-emission feature tests (GET→query, HEAD-like-GET, degrade-notice, array-of-objects-drop). The reader unit tests (raw rules / nested+array / throws→degrade / no-`rules()`→degrade) are well-targeted.

Gates: PHPStan clean, `pint --test` clean on `src/Plugins/Core/`, full suite 3072 passed on PHP 8.4; new tests + snapshot green on 8.5 too.

---

**Finding A (minor, non-blocking) — unused local.** `CoreQueryParameterResolver.php:302` assigns `$actionName = $this->actionName($method);` in `formRequestParameters()` but never uses it (both notices reference `$formRequestClass`, not `$actionName`). It's a copy-paste remnant from `inlineValidationParameters()` (which does use `$actionName`). PHPStan level 8 doesn't flag unused locals, so CI is green — but it's dead code worth deleting.

**Finding B (nit, non-blocking) — slightly inaccurate docblock.** `FormRequestRequestSchemaResolver` class docblock says "POST/PUT/PATCH/**DELETE** keep the request body." Per your own deviation finding, DELETE never gets a body. The body-suppression logic is correct; only the docblock prose overstates DELETE. Suggest "POST/PUT/PATCH keep the request body; DELETE is unchanged (already bodyless)."

Neither finding blocks merge. If you fold in A (and optionally B), even cleaner. Approving as-is.

Radiergummi:
🤖 **[lead]** TWO independent approvals (reviewer-A + reviewer-B, both on 97739ce6, both TDD-verified: body-path byte-identical extract, GET/HEAD gating, #[QueryParam] precedence, DELETE deviation correct). → survey (generation-affecting, area:requests/params). In parallel, coder folds in reviewer-B's 2 non-blocking nits (A: dead unused $actionName CoreQueryParameterResolver.php:302; B: FormRequestRequestSchemaResolver docblock wrongly says DELETE keeps a body) — neither is generation-affecting, so the survey verdict transfers. MERGE PATH: 222 src lines > ~150 → ESCALATE to Moritz once green+approved+surveyed (NOT auto-merge).

Radiergummi:
🤖 **[lead]** approval-transfer note: head moved 97739ce6 → 79e9e707 (reviewer-B's nits A+B folded in). Verified the diff is ONLY those two trivial, non-functional changes — `git diff 97739ce6 79e9e707`: remove the dead `$actionName` local in CoreQueryParameterResolver (−1), correct the FormRequestRequestSchemaResolver docblock re DELETE (+2/−1). No generation or behavioral change. Both independent approvals (reviewer-A + reviewer-B, given on 97739ce6) carry forward to 79e9e707; no re-review needed. CI green on 79e9e707. Pending only the Layer A survey, then escalation to Moritz.

Radiergummi:

Baseline: `main` @ `63d7e7e7` (fresh, same-main). Candidate: PR branch @ `79e9e707` (verified via manifest `libraryCommit`). PHP 8.4.

### Re: the `componentSchemas` drop (resolved — expected shift, **not** a regression)

The drop is real and reproducible on a clean same-main baseline:

| app | componentSchemas (base → cand) |
|-----|-------------------------------|
| AdvisingApp | 61 → 59 (−2) |
| AureusERP | 36 → 33 (−3) |
| InvoiceNinja | 375 → 238 (−137) |
| Koel | 106 → 93 (−13) |
| Lychee | 280 → 229 (−51) |
| Pelican | 104 → 77 (−27) |

All other metrics (`paths`, `operations`, `apiOperations`, `responseSchemas`, `documentedResponses`, **`requestBodies`**) are unchanged across all 11 apps.

**Root cause (localized on Koel, −13):** every one of the 13 dropped component schemas is a `*Request` FormRequest class:
`AlbumListRequest, ArtistListRequest, CheckDownloadableCountRequest, DownloadSongsRequest, FetchSongsForQueueRequest, FetchSongsToQueueByGenreRequest, GetUserInvitationRequest, LastfmCallbackRequest, PaginateSongsByGenreRequest, SearchRequest, SongListRequest, SongPlayRequest, ViewSongOnITunesRequest`.

In the **baseline** spec, all 13 had **zero `$ref` references** — they were orphaned/dangling component schemas. They were registered as components but never wired into any operation, because these are GET routes and a GET requestBody is stripped at serialization. That is precisely why `requestBodies` does **not** move: the baseline never counted them as request bodies either.

PR #430 stops registering a GET/HEAD FormRequest as a component schema and instead surfaces its `rules()` as **query parameters**. The information is preserved and improved — it is now actually attached to the operation. On Koel the candidate **gains +64 query parameters across 37 GET operations**, including exactly the routes backing the removed schemas (`/api/search`→`SearchRequest`, `/api/invitations`→`GetUserInvitationRequest`, `/lastfm/callback`→`LastfmCallbackRequest`, `/download/songs`→`DownloadSongsRequest`, `/itunes/song/{album}`→`ViewSongOnITunesRequest`, `/api/download/check`→`CheckDownloadableCountRequest`, `/api/queue/fetch`→`FetchSongsForQueueRequest`, …).

**Verdict:** the `componentSchemas` delta is the intended GET-FormRequest-body → query-param shift. It removes dead, unreferenced components and moves the validation shape onto the operations as query params — the spec is strictly more correct. No information lost, no regression.

(The prior session's FAIL was a measurement artifact: its `candidate-430` was generated against `libraryCommit 97739ce6` — neither `main` nor this PR tip. Discarded and re-run clean.)

Radiergummi:
🤖 **[lead]** ESCALATE to Moritz:

Merge-ready, NOT auto-merged. All gates pass:
 - 2 independent approvals
 - Layer A survey PASS (0 regressions; the componentSchemas delta is the intended GET-body→query-param shift — removed schemas were orphaned/dangling GET FormRequest components with zero $ref, info preserved + improved as +query params, root-caused by surveyor not rubber-stamped)
 - full CI matrix GREEN on the post-#431/#432 merge head 5288d528 (11/11 incl. swagger-php 5.8 + snapshot).

Rebased onto current main via merge (CHANGELOG + docs/auto-derivation.md conflicts resolved keeping both #417 + #418/#419 entries; force-push declined by policy so a non-force merge commit was used — collapses on squash).  
Held for human review ONLY because the src diff is 222 lines > the ~150 auto-merge controversy threshold; no Contracts/config/signature/stage-order touched.  
Squash-merge when ready.

Issue #426, opened 2026-06-20
Share URI-placeholder extraction between the generator and the linter (and fix the linter's {param:field} mishandling)


Radiergummi:
Follow-up from #416 / PR #425.

The generator now extracts URI placeholders in `ActionDescriptor::uriPlaceholders()` (bare name + optional flag). The linter independently extracts placeholders in `PathParameterUndeclared::extractPlaceholders()`. These two should share a single utility so they cannot drift.

**First task (a real, pre-existing bug):** `PathParameterUndeclared::extractPlaceholders()` does *not* strip a `{param:field}` custom-key suffix. Its regex (`/\{([^?}]+)\??}/`) plus `ltrim($name, '+#./;?&=,!@|')` yields `member:slug` for `{member:slug}` (the `ltrim` only removes *leading* binding-prefix chars). It then compares `member:slug` against the operation's declared parameter names — which are the bare `member` — so a route with a `{param:field}` segment can be mis-reported by `path.parameter-undeclared`. The generator's `uriPlaceholders()` already handles this correctly (splits on `:` and uses the bare name); the shared extractor should fix the linter to match.

**Scope:**
- Extract one shared `Support\Routing` utility returning the bare placeholder name (and optional flag).
- Have both `OperationBuilder` (via `ActionDescriptor`) and `PathParameterUndeclared` consume it.
- Add a linter test for a `{param:field}` segment to lock the fix.

Out of scope for #416/#425 (kept surgical there); filed for later.

Radiergummi:
🤖 **[lead]** admitted → planning. Shared URI-placeholder extractor + linter {param:field} bugfix. area:core+area:lint; survey applies (generation touches uriPlaceholders), keep diff surgical.

PR #436, opened 2026-06-22
feat(responses): read migration column types to enrich model field schemas

### LINKED ISSUE #414: feat(responses): read migration column types to enrich model field schemas (formats, constraints, enum, comments)
## Problem

`EloquentModelToSchema` derives field metadata from `$casts` + `@property` tags only. `$casts`
carries date/datetime/decimal/enum — so uuid PKs, `email`/`url`/`ip` string columns, `json`
columns, and bounded `varchar`/`decimal` all reach the schema as bare `type: string`. The 11-app
corpus measured **field-level format coverage at 7%**; the model side contributes almost none of it.
Migrations are the canonical, authored, never-wrong description of column shape — and are never read.

## Evidence

- No column/migration reader exists anywhere: grep `getColumnType|listTableColumns|getColumnListing|Schema::getColumns|information_schema|doctrine/dbal` over `src/` → zero (the only `doctrine` hits are `doctrine/annotations` for the SwaggerPhp harvester).
- Deferred-by-design: `docs/superpowers/plans/2026-06-03-eloquent-model-response-schema.md:1512` ("DB/migration column reading" under Deferred).
- Cast side emits format only for date/datetime: `src/Support/Extraction/EloquentModelToSchema.php:292-308`; `decimal:N` drops scale → `type: string` (`:294`).
- Reusable format vocabulary already exists on the validation side: `src/Support/Extraction/ValidationRulesToSchema.php:428-509`.

Bounded AST reader over `database/migrations/*.php` matching only the `Blueprint` fluent-call
whitelist inside `Schema::create()/table()` closures, keyed by table (`$model->getTable()`):
`uuid`→format uuid, `ipAddress`→format ip, `date`→date, `dateTime`/`timestamp`→date-time,
`json`→object, `decimal($p,$s)`→`number`+`multipleOf`, `string($n)`/`char($n)`→`maxLength`,
`unsigned*`→`minimum:0`, `enum([...])`→enum, `->nullable()`→nullable, `->comment('…')`→description.
Fill only fields not already set by a cast/`@property`/attribute (those win). Reuse
`ValidationRulesToSchema`'s format mapping so request and response sides agree. Degrade silently +
log when a shape isn't matched (Tier-1 discipline).

## Scope / Out of scope

- In: the standard `Blueprint` whitelist; precedence below casts/tags/attributes; per-table memoization; opt-out flag.
- Out: `->change()`/alter-tracking (Tier-2), raw `DB::statement` DDL, any live-DB introspection (stay static-file, never touch a connection).

## Acceptance

A model whose table declares `$table->uuid('id'); $table->ipAddress('last_ip')->nullable(); $table->decimal('amount',8,2)` and no casts for them emits format uuid / format ip+nullable / number+multipleOf 0.01. A column matched by both a cast and a migration keeps the cast output. Unmatched shapes emit nothing + a debug log.

<sub>From the 2026-06-20 lever audit; relates to #413 (coverage metric).</sub>

Radiergummi:
Closes #414

### Tier

**Tier-1** (bounded AST whitelist), the lowest tier that captures the idiom. Migrations are
authored, deterministic, never-wrong column descriptions. We parse `database/migrations/*.php`
statically and match **only** a whitelist of `Blueprint` fluent-call shapes inside
`Schema::create()/table()` closures — no variable tracking, no live connection, no `->change()`
alter-flow (that would be Tier-2). When a shape isn't matched we emit nothing and log at debug.

Not controversial on its own: no `src/Contracts/**` change, no stage-order change, no public
signature change. It **does add one config key** (the opt-out flag) and reuses the
`ValidationRulesToSchema` format vocabulary — both called out below for the human gate.

### Approach

The join key is the **table name** (`$model->getTable()`), not a file path: migrations live in one
directory and are not co-located with models. So:

1. A new **`MigrationColumnReader`** (scoped singleton) parses every `database/migrations/*.php`
   file **once per run**, building a per-table column index `table → column → ColumnMetadata`.
   It reuses the existing `MethodBodyScanner` parser machinery pattern (php-parser
   `createForNewestSupportedVersion` + `NameResolver`, try/catch → log + skip on parse failure),
   matching the proven `Schema::create/table` → `Blueprint` traversal validated over 1759 corpus
   migration files (issue sizing comment).
2. `EloquentModelToSchema` consults the reader for each property name, filling **only** fields not
   already set by a cast / `@property` / attribute (cast precedence is the acceptance criterion).

### Files

**Create**
- `src/Support/Extraction/MigrationColumnReader.php` — the reader. `#[Scoped]`. Public surface:
  `columnsForTable(string $table): array<string, ColumnMetadata>`. Internally:
  - Discover files via `glob(database_path('migrations') . '/*.php')` (guard the directory absent
    → empty index; consistent with `ExampleFileLoader`/`ThrowsErrorContributor` using `base_path`).
    Add a `--path`/subdir sweep is **out of scope**; module-organized apps that register extra
    migration paths are a follow-up (note in PR, not built — keeps this surgical).
  - Parse each file, find `Schema::create('t', fn(Blueprint $b) => …)` and `Schema::table(...)`
    static calls (facade match like `InlineValidatorRulesReader::facadeMatches`: FQCN
    `Illuminate\Support\Facades\Schema` **or** bare `Schema`). First arg must be a string literal
    (`AstLiteralEvaluator`); dynamic table name → skip that block + debug log.
  - Walk the closure body's top-level statements; for each method-call **chain** rooted at the
    `Blueprint` parameter variable, read the head column-defining call and the trailing modifier
    calls. `->change()` anywhere in the chain → skip the column (Tier-2 alter-flow, out of scope).
- `src/Support/Extraction/ColumnMetadata.php` — a small readonly value object carrying the resolved
  partial schema fields: `format`, `maxLength`, `minimum`, `multipleOf`, `enum`, `nullable`,
  `default`, `description`, plus a coarse `type` hint (only where the migration implies one the cast
  wouldn't, e.g. `json`→object, `decimal`→number). All nullable/undefined-by-default.

**Modify**
- `src/Support/Extraction/EloquentModelToSchema.php`:
  - Inject `MigrationColumnReader` (constructor) — plus a `bool $readMigrationColumns` flag bound
    from config (mirrors how `ModelFactoryExampleReader` takes `?int $seed` from config).
  - Add a single chokepoint `applyMigrationColumns(OA\Property $property, ?ColumnMetadata $column)`
    that fills each schema field **only when `is_undefined()`** (same guard the readOnly/example
    passes already use), so cast/tag/attribute always win. Call it:
    - in `buildPropertyFor()` (single-property lookup path), and
    - in the `schemaFor()` properties loop (full-schema path),
    keyed by `$metadata` table (add `'table' => $model->getTable()` to the metadata memo).
  - `nullable` from the migration only **relaxes** to nullable; it never marks a property required
    and never overrides a `@property` non-null that already drives `required` (don't touch the
    `required` derivation — migration nullability adds `null` to the type, nothing more).
- `config/openapi.php`: add the opt-out flag. **Resolved at plan-review:** the config has no
  `responses` or `models` block (root keys are flat: `examples`, `lint`, `routes`, `filters`, …),
  so a flat `'read_migration_columns' => true` root key with a short comment is the correct shape —
  matches the flat-config convention. Default = enabled.
- `OpenApiServiceProvider` — bind the flag into the `EloquentModelToSchema` resolution (the only
  glue; reader itself needs no args).

### The Blueprint whitelist (column-defining heads → schema fields)

Reuse `ValidationRulesToSchema`'s format spellings so request and response sides agree
(`uuid`→`format: uuid`, `ipAddress`→`format: ip`, `email` n/a here — migrations have no email type):

| Blueprint call | schema |
|---|---|
| `uuid(col)` / `foreignUuid(col)` | `format: uuid` |
| `ipAddress(col)` | `format: ip` |
| `macAddress(col)` | `pattern` (reuse `applyMacAddress` spelling) |
| `date(col)` | `format: date` |
| `dateTime`/`dateTimeTz`/`timestamp`/`timestampTz`(col) | `format: date-time` |
| `time`/`timeTz`(col) | `format: time` is non-standard → **omit** (no format; just string) |
| `year(col)` | `type: integer` |
| `json`/`jsonb`(col) | `type: object` |
| `decimal(col, p, s)` / `unsignedDecimal` | `type: number`, `multipleOf: 10^-s` (e.g. s=2 → 0.01) |
| `string(col, n)` / `char(col, n)` | `maxLength: n` (only when `n` is a literal int) |
| `unsignedInteger`/`unsignedBigInteger`/`unsignedTinyInteger`/… / `increments`/`*Increments` | `minimum: 0` |
| `enum(col, [..])` / `set(col, [..])` | `enum: [..]` (literal string list only) |
| modifier `->nullable()` (no arg, or arg truthy literal) | `nullable` |
| modifier `->default(<literal>)` | `default` (literal scalar only; `DB::raw`/expr → skip) |
| modifier `->comment('…')` | `description` (literal string only) |

Anything not in this list (custom macros, `$table->foo()`) → no contribution + debug log once.
`->default()` and `$attributes`-array defaults: the **default keyword** half (issue comment 1) is
**in scope** for the migration side (`->default()` → `default`). The sibling Tier-0 read of the
model's static `$attributes` property is a **small add** also in scope (same never-wrong precedence,
fills `default` only when unset); coder folds it into the metadata memo.

> **[reviewer] scope note:** `$attributes`-default reading is *additive* beyond #414's stated
> whitelist (the issue's acceptance criteria don't mention it). Keep it genuinely small — a single
> Tier-0 reflection read of the static `$attributes` array, filling `default` only when undefined,
> with its own test case. If it grows tendrils (casts interaction, nested arrays, accessor-derived
> defaults), **defer it to a follow-up issue** rather than expand this PR. It must not delay the
> migration-column core.

New `tests/Unit/Support/Extraction/MigrationColumnReaderTest.php` with fixture migration files
(testbench `database_path` points at the package's test app; add fixtures under the test app's
`database/migrations`, or feed the reader an explicit directory in unit tests — coder picks the
seam that keeps the reader injectable). Cover, per the acceptance criteria and edge/degrade paths:

- **Happy path (acceptance):** table with `uuid('id')`, `ipAddress('last_ip')->nullable()`,
  `decimal('amount',8,2)` and no casts → `format: uuid` / `format: ip` + nullable /
  `type: number` + `multipleOf: 0.01`.
- **Cast precedence:** a column matched by **both** a cast and a migration keeps the cast output
  (e.g. `decimal:2` cast vs `string` migration → cast wins). Covered in
  `EloquentModelToSchemaTest`.
- **`@property` / attribute precedence:** tag-typed property unchanged by migration.
- **maxLength / minimum / enum / comment / default** each get a dedicated case.
- **Degrade paths (emit nothing + debug log, never throw):** unparseable migration file;
  dynamic table name (`Schema::create($name, …)`); dynamic column name; `enum` with a
  non-literal member; `->default(DB::raw(...))`; `->change()` chain skipped; a `Blueprint` macro
  not on the whitelist; absent `database/migrations` directory.
- **Union/nullable interaction:** migration `nullable` adds `null` to an existing `type: string`
  (→ `['string','null']`) without disturbing `required`.
- **Opt-out flag off:** reader not consulted; schema identical to today (regression guard).
- **Memoization:** parsing happens once per file across many `columnsForTable` calls.

Existing suites that must stay green: `EloquentModelToSchemaTest`,
`EloquentModelPropertyLookupTest`, `EloquentModelResponseTest`, and the byte-exact `snapshot`
group (new fixtures will shift snapshots → regenerate intentionally and review the diff).

### Docs & CHANGELOG

- `docs/auto-derivation.md` — extend the Eloquent-model-response section: "migration columns now
  enrich field formats/constraints (uuid/ip/date formats, maxLength, unsigned minimum, decimal
  multipleOf, enum, nullable, default, comment→description), below casts/tags/attributes in
  precedence; opt out with the config flag." (Material → may warrant the lead's `[docs]` sub-phase.)
- `docs/config.md` — document the new `read_migration_columns` flag.
- `CHANGELOG.md [Unreleased] → Added` — extend the existing "Response schemas inferred from Eloquent
  model metadata (…, migration columns, …)" bullet to spell out the new column-derived signals, and
  reference `#414`.

### Out of scope (restate, to keep it surgical)

`->change()`/alter-tracking, raw `DB::statement` DDL, any live-DB introspection, extra/module
migration paths beyond `database/migrations`, table disambiguation across renamed tables. No new
authoring attribute (the flag is the only escape hatch needed).

### Human-gate flags for the lead

- Adds **one config key** (`read_migration_columns`) — config-surface change.
- Reuses `ValidationRulesToSchema` format vocabulary across request/response sides (intentional, to
  keep formats consistent) — a small coupling worth a glance.
- New fixtures shift the `snapshot` group; the regenerated snapshots need a human eyeball.
- Generation-affecting (`area:core` + `area:responses`) → **survey gate applies**.


Radiergummi:

Tier-1 bounded-AST migration-column reader (`MigrationColumnReader` + `ColumnMetadata`), consulted by `EloquentModelToSchema` below cast/`@property`/attribute precedence. Reuses the `MethodBodyScanner` parser pattern + `ValidationRulesToSchema` format vocabulary. Human-gate flags: adds one config key (`read_migration_columns`), shifts `snapshot` fixtures, survey gate applies (area:core + area:responses).

Radiergummi:
🤖 **[lead]** planning done → plan-review. Reviewer: validate Tier-1 discipline (bounded whitelist, precedence below casts/tags/attrs, degrade+log), and the config-key + snapshot-fixture human-gate flags the planner raised.

Radiergummi:
🤖 **[reviewer]** plan-review: approved (tightened config-shape + scoped the `$attributes` add).

Validated against the codebase, not just the prose:

- **Tier-1 discipline — holds.** Bounded `Blueprint` whitelist inside `Schema::create/table` closures only; reuses the proven `MethodBodyScanner` (`createForNewestSupportedVersion` + `NameResolver`, `#[Scoped]`, try/catch → log+skip) and `AstLiteralEvaluator`/`facadeMatches` patterns. `->change()` → skip (correctly called Tier-2 alter-flow), dynamic table/column → skip+debug, unparseable file → skip+debug, absent `database/migrations` → empty index. No variable tracking. Genuinely never-wrong authored source.
- **Precedence — verified at the real sites.** Fills only `is_undefined()` fields (same guard `EloquentModelToSchema` uses for readOnly/example), so cast/`@property`/attribute always win. `nullable` only relaxes the type and does **not** touch the `required` derivation (which keys off non-null `@property` tags). Cast-vs-migration precedence is an explicit test case.
- **Boundaries — clean.** Reader lives in `src/Support/Extraction/`, no plugin/third-party dependency; reuses `ValidationRulesToSchema` format spellings (uuid→`format: uuid`, ipAddress→`format: ip`, mac→`pattern`) — all confirmed present. No Contracts change, no stage-order change.
- **Scope — surgical.** Out-of-scope list (alter-tracking, DB introspection, extra migration paths, `DB::raw` defaults) is explicit and right. No new authoring attribute; the config flag is the only escape hatch.
- **Tests — TDD-first, edges enumerated:** uuid/ip/decimal happy path, cast precedence, `@property`/attribute precedence, maxLength/minimum/enum/comment/default, and the full degrade set (unparseable, dynamic table/column, non-literal enum member, `DB::raw` default, `->change()`, off-whitelist macro, absent dir, union/nullable, opt-out regression, memoization). Good.

**Changes I made to the plan:**
1. **Config shape resolved** (was deferred to coder): config has no `responses`/`models` block — root keys are flat (`examples`, `lint`, `routes`, `filters`). So it's a **flat `read_migration_columns` root key**, not nested. Dead branch removed.
2. **Scoped the `$attributes`-default add:** it's *additive* beyond #414's stated whitelist, so I added a reviewer note — keep it to a single Tier-0 read filling `default` when undefined (with its own test), and **defer to a follow-up** if it grows. It must not gate the migration-column core.

**Human-gate flags confirmed real** (for the lead's merge-time controversy gate):
- Adds one config key `read_migration_columns` (config-surface change).
- New fixtures shift the byte-exact `snapshot` group → regenerated snapshots need a human eyeball.
- Generation-affecting (`area:core` + `area:responses`) → survey gate applies.

Verdict: **approved-for-coding.**

Radiergummi:
🤖 **[lead]** plan approved by reviewer-1 → coding. Coder: implement per the amended plan — flat root config key `read_migration_columns` (default enabled); keep the $attributes-default add to a single Tier-0 read filling `default` when undefined (own test), defer if it grows, must NOT gate the core whitelist. TDD: failing tests first. Regenerate snapshot fixtures and call them out for human eyeball.

Radiergummi:
🤖 **[coder]** implemented; CI green locally (pint/phpstan/test, 3115 passing). Pushed d45a5a0d. Ready for review.

**What landed (per the amended plan):**
- `MigrationColumnReader` (`#[Scoped]`, Tier-1 bounded AST over `database/migrations/*.php`, keyed by `getTable()`) + `ColumnMetadata` VO, in `src/Support/Extraction/`. Reuses the `AstLiteralEvaluator` + `NameResolver` parser pattern and `ValidationRulesToSchema` format spellings (uuid→`format: uuid`, ipAddress→`ip`, mac→`pattern`).
- `EloquentModelToSchema` consults it in both the single-property (`propertyFor`) and full-schema (`schemaFor`) paths via one `applyMigrationColumn` chokepoint that writes **only `is_undefined()` fields**, so cast / `@property` / attribute always win. `nullable` only relaxes a scalar type to `[..., null]` and never touches the `required` derivation. A `$ref` property (relation/enum cast) takes only a `description` sibling — no type-shaping grafted onto the `$ref`.
- Config: flat root key `read_migration_columns` (default `true`). Wired in `OpenApiServiceProvider` (reader gets `database_path(migrations)`; flag from config).

**Degrade set (all silent + debug log, never throws), each tested:** dynamic table name, dynamic/absent column name, `->change()` alter chain, off-whitelist macro, non-literal `enum` member, `DB::raw(...)` default, unparseable file, absent migrations dir. Plus: cast-vs-migration precedence, `@property`/attribute precedence, maxLength/minimum/enum/comment/default, union/nullable, opt-out regression, and per-file memoization.

**Scope notes / deviations:**
- The `$attributes`-default Tier-0 add was **deferred to a follow-up** (not built here) to keep the PR surgical — only the migration `->default()` half is in. The migration-column core does not depend on it.
- Added a `PLAIN_HEADS` whitelist (plain `integer`/`text`/`time`/etc.) so a recognised-but-signal-less column still attaches its `->default()`/`->comment()`/`->nullable()` modifiers. Without it, `integer(weight)->default(0)` would have been dropped as off-whitelist, losing the default. Format heads now also carry `type: string` (every whitelisted format is a string format), so nullable relaxation produces `[string, null]`.
- **Snapshot fixtures did NOT shift** — the byte-exact `snapshot` group passes unchanged. The test/testbench harness has no `database/migrations` directory, so the reader returns an empty index there and generation output is identical. (Flagging anyway since the plan anticipated a shift: please confirm you are comfortable that no snapshot eyeball is needed because the harness exercises the reader only via the dedicated unit tests, not the snapshot apps.)

🚩 Human-gate flags: one new config key (`read_migration_columns`); reuses `ValidationRulesToSchema` format vocabulary; generation-affecting (area:core + area:responses) → survey gate applies.

Radiergummi:
🤖 **[lead]** coder-2 pushed d45a5a0d, full remote CI green (incl. swagger-php 5.8 + Snapshot — snapshots did NOT shift, testbench has no migrations dir). → code review (2 independent approvals). NOTE for merge gate: this adds config key `read_migration_columns` → controversial → will escalate to human at merge even if all gates pass; reviewers still do a full pass. Scope deviation: $attributes-default add deferred to a follow-up (acceptable, keeps PR surgical).

Radiergummi:
🤖 **[reviewer-1]** review round 1 — 1 finding (+2 minor notes).

Reviewed d45a5a0d adversarially in an isolated worktree (removed after), independently of reviewer-2. Ran the gates myself: Pint clean, PHPStan L8 clean, reader+model tests 73 passed, full suite **3114 passed / 0 failures**. The architecture is sound — bounded AST, blueprint-rooted chain unwind, literal-only via `AstLiteralEvaluator`, the `applyMigrationColumn` chokepoint guards **every** write with `is_undefined` (verified by neutralizing it: 7 of 10 new model tests fail, incl. the load-bearing `price` cast-wins-but-fills-multipleOf precedence case — non-vacuous). `$ref` branch takes only a description; `relaxToNullable` only widens scalar/union types and never touches `required`. Degrade set is complete and each path is tested. Config key flat-root + provider wiring + opt-out regression all correct. The `phpstan.neon` ignore is correctly scoped to `tests/Fixtures/Migrations/*` with justification.

**Finding 1 (correctness — requires change): `ulid`/`foreignulid` → `format: uuid` is wrong and contradicts the package's own established convention.** `src/Support/Extraction/MigrationColumnReader.php:72-73` maps ULID columns to `format: uuid`. A ULID is a 26-char Crockford-base32 identifier, *not* an RFC-4122 UUID (36-char hex), so `format: uuid` mis-describes the column and a consumer validating against it would reject valid ULIDs. This directly contradicts the deliberate, tested, documented convention elsewhere in the package:
- `src/Support/Routing/UriParameterResolver.php:285-286`: "ULID keys are strings, but there is no standard OpenAPI format for them" — it intentionally emits **no** format for `HasUlids`.
- `tests/Fixtures/Models/UlidArticle.php:11-12` and `docs/auto-derivation.md:85`: "a `HasUlids` model → `string` (ULID has no standard OpenAPI format)".

It also makes **`docs/auto-derivation.md` self-contradictory**: the new line `:173` says "`uuid`/`foreignUuid`/`ulid` → `uuid`" while `:85` (unchanged) says ULID has no format. The plan's whitelist listed only `uuid`/`foreignUuid` — `ulid`/`foreignulid` were added beyond plan and beyond the issue. **Fix:** remove `ulid`/`foreignulid` from `FORMAT_HEADS` (they then fall through to plain-string handling, consistent with the rest of the package), add a reader test asserting a `ulid('x')` column yields no format (a real gap — ULID is currently untested), and correct the `docs/auto-derivation.md:173` list. Small, self-contained.

**Note A (minor, non-blocking): fixture comment inaccuracy.** `tests/Fixtures/Models/Widget.php:11-12` says "A `name` cast deliberately collides with the migration's `string('name', 120)`", but `name` is **not** in `$casts` (only `$fillable`). The `maxLength: 120` test in fact relies on `name` being *uncast* so the migration fills it. The comment describes the opposite of what the fixture does — reword to "`name` is intentionally **uncast** so the migration's `string('name', 120)` maxLength is observable."

**Note B (minor, non-blocking): a few lines exceed the 100-char soft limit** in `MigrationColumnReader.php` (e.g. `:398` at 115, `:347` at 109). Pint passes so these aren't hard violations and the codebase tolerates the occasional long line, but `:398` (the `macaddress` match arm) wraps cleanly if you're touching the file for Finding 1.

The `$attributes`-default deferral is correctly absent and the core does not depend on it (verified) — that deviation matches the plan-review scope note. Reader memoization, format-vocabulary reuse (`uuid`/`ip`/mac-`pattern` agree with `ValidationRulesToSchema`), and the snapshot-unchanged claim (harness has no `database/migrations`) all check out.

Sending back to the coder for Finding 1. Notes A/B are optional but cheap while you're in there.

Radiergummi:
🤖 **[reviewer-2]** review round 1: approved — 1 minor, non-blocking finding (a test gap, not a correctness bug).

Reviewed adversarially at d45a5a0d in an isolated worktree (removed after). I drove the reader directly against 11 hand-built adversarial column shapes rather than trusting the unit tests:

**Blast radius — tight, no mis-matches found:**
- `decimal(8,2)`→`multipleOf 0.01`, `decimal(5,4)`→`0.0001` (scale math exact, no float drift in output); scaleless `decimal`→`number`, no `multipleOf`.
- `string("code", 1)`→`maxLength 1` (no off-by-one); `string("nolen")`→no `maxLength`.
- Modifier reorder `->default("x")->nullable()->comment("c")` captured all three regardless of order (the unwind-then-fold is order-independent).
- `enum([...])->nullable()` composes; non-literal enum member drops the enum entirely; `->nullable(false)` correctly re-asserts NOT NULL.
- `unsignedInteger`→`minimum 0`; `integer("weight")->default(0)` keeps the default via `PLAIN_HEADS` (confirmed it would otherwise be dropped). Off-whitelist `geometry()` / dynamic names contribute nothing + debug log.

**Multi-migration / table reuse:** create-then-`Schema::table` alter keyed by `getTable()` merges additively (`[...existing, ...new]`); `->change()` is skipped at both head and modifier position, so an alter never clobbers the original column. glob() timestamp order means create precedes alter. Correct.

**Precedence chokepoint:** every write in `applyMigrationColumn` is `is_undefined`-guarded — cast wins (proven: `price` keeps `type: string` from the `decimal:2` cast while the migration fills only the undefined `multipleOf`). `relaxToNullable` only widens `->type` to `[.., "null"]`; it never touches the `required` derivation. A `$ref` property takes only a `description` sibling — no type/enum graft (proven on the enum-cast `status`).

**Config / lifecycle:** `read_migration_columns` defaults true, opt-out path tested; `->scoped()` binding (+ `#[Scoped]`) gives a fresh lazy `$index` per run, no cross-run leak.

**Gates:** PHPStan clean package-wide; Pint clean on the diff; remote Snapshot (locked toolchain), swagger-php 5.8, PHP 8.5, Coverage, quality all pass. Snapshot byte-identical confirms zero generator drift (harness has no `database/migrations`, reader returns empty). Bookkeeping accurate: config block, `docs/config.md` row, `docs/auto-derivation.md` precedence ladder (slotted step 5, below cast/tag/attribute), CHANGELOG (#414).

**Finding 1 (minor, non-blocking):** `->default(null)` is a real shape that takes a distinct branch (`$arguments[0]->value instanceof Expr\ConstFetch` in `applyModifiers`, MigrationColumnReader.php:515) yielding `hasDefault=true, default=null`, and reaches the property as a valid `"default": null` (I verified end-to-end — distinct from the UNDEFINED sentinel). There is no regression test pinning it. Suggest a reader case: a `->default(null)` column asserting `hasDefault === true` and `default === null` (and optionally an integration case asserting the emitted `default: null`). Not a merge blocker; the behavior is correct today.

Approving on my pass.

Radiergummi:
🤖 **[lead]** review round 1: reviewer-2 approved (1 minor note); reviewer-1 changes-requested (1 correctness finding). → back to coder (round 1/3). Fix the union of findings, re-push, then back to in-review for re-check.

Radiergummi:
🤖 **[coder]** addressed round 1: ulid≠uuid + minor fixes; pushed 5c371cde. All CI green.

**Blocking (reviewer-1, Finding 1 — ulid):**
- Dropped `ulid`/`foreignUlid` from `FORMAT_HEADS`. They now fall through to `PLAIN_HEADS` (recognised so their modifiers still attach, but emit a bare `string` with no format), matching the `HasUlids` convention at `UriParameterResolver.php:285-286` / the `UlidArticle` fixture / `docs/auto-derivation.md:85`.
- Added reader test `emits no format for a ulid column` (asserts both `format` and `type` are null on a `ulid(reference)` column).
- Corrected `docs/auto-derivation.md` step 5: ulid columns stay a bare string (no format), consistent with :85.

**Minor:**
- reviewer-1 (A): reworded the `Widget` fixture docblock — `price` (cast `decimal:2` vs migration `decimal(8,2)`) is the cast-collision example; the rest are uncast.
- reviewer-1 (B): extracted the mac-address regex to a `MAC_ADDRESS_PATTERN` constant (was the 115-char line) and wrapped the remaining over-width logic lines (`isSchemaBlueprintCall` via `in_array`, the `->change()` check, the nullable assignment). The only lines still slightly over 100 are `logger->debug(...)` message strings, which do not wrap cleanly and match how the rest of the codebase formats log calls.
- reviewer-2: added reader test `records an explicit null default as a set default` — `->default(null)` yields `hasDefault === true && default === null` (distinct from no default).

`composer check` green (3117 tests); snapshot group still byte-identical.

Radiergummi:
🤖 **[coder]** rebased onto `origin/main` (picks up #437); pushed a1284b37. CHANGELOG `[Unreleased]` conflict resolved by keeping both entries (Fractal #427 + migration #414). `composer check` green on the rebased tree (3127 tests) and all remote CI green. No code conflict — #437 is entirely in `src/Plugins/Fractal/`.

Radiergummi:
🤖 **[lead]** round-1 fixes + rebase onto main (a1284b37), CI green → re-review (round 2/3). reviewer-1: confirm the ulid→uuid finding is resolved (no format on ulid, doc fixed). reviewer-2: confirm the changed diff (dropped FORMAT_HEADS entries, added regression tests) still holds your approval.

Radiergummi:
🤖 **[coder]** rebased onto main e2c69b49 (CHANGELOG: both kept); CI re-gated; pushed a1284b37.

(Already done in the prior turn — flagging again here for clarity since two rebase requests crossed. Branch tip `a1284b37` sits on top of `e2c69b49` (#437); CHANGELOG `[Unreleased]` keeps both the Fractal #427 and migration #414 entries; `composer check` green on the rebased tree, 3127 tests; all remote CI green on `a1284b37`.)

Radiergummi:
🤖 **[reviewer-2]** re-review round 2: approved, no concerns. My round-1 finding is closed and the delta improves correctness.

Re-confirmed the delta d45a5a0d → a1284b37 (diffed src + tests directly, ran the affected suites at the new head):

- **ulid fix (reviewer-1) is a genuine correctness improvement I missed in round 1.** `ulid`/`foreignUlid` moved out of `FORMAT_HEADS` into `PLAIN_HEADS` — a ULID is 26-char base32, not a UUID, and has no standard OpenAPI format, so emitting `format: uuid` was wrong. Verified `ulid` now appears in exactly one head table (no duplication) and that this aligns the reader with the **pre-existing** route-binding behavior (`tests/Fixtures/Models/UlidArticle.php` already documents ULID → bare string). New test `emits no format for a ulid column` pins `format === null && type === null`. Good catch.
- **My `->default(null)` finding is closed.** New fixture column `nickname` declared `->default(null)` + test `records an explicit null default as a set default` asserts exactly `hasDefault === true && default === null`. Precisely the pin I asked for.
- **Cosmetic / behavior-preserving rest:** `MAC_ADDRESS_PATTERN` const (identical string), `in_array([create,table], strict: true)` for the two `===`, `||`/`&&` line wraps, Widget docblock reword. The Fractal hunk is rebase noise (PR #437, separately approved).
- **No regression to the round-1-validated behavior:** precedence chokepoint, degrade set, nullable-doesnt-touch-required, `$ref` immunity all unchanged.

Gates at a1284b37: 75 tests pass (was 73; +2 new), PHPStan clean, Pint clean; remote Snapshot (byte-identical), swagger-php 5.8, PHP 8.5 ×2, Coverage, quality all green.

Final approval on the merged state stands.

Radiergummi:
🤖 **[reviewer-1]** re-review round 2: approved, Finding 1 resolved.

Verified the fix at a1284b37 in an isolated worktree (removed after). All gates green: Pint clean, PHPStan L8 clean, reader tests 26 passed, full suite **3126 passed / 0 failures**.

**Finding 1 (ulid → uuid) — fully resolved, all three parts:**
1. `ulid`/`foreignulid` removed from `FORMAT_HEADS` and moved to `PLAIN_HEADS` with an explanatory comment, so a ulid column now falls through to a bare string (no format) — matching the `HasUlids` convention. Verified the path is **genuinely correct, not vacuously passing**: I revert-probed by re-adding `ulid => 'uuid'` to `FORMAT_HEADS` and the new test fails with "Failed asserting that 'uuid' is null"; removing it passes. The test locks the fix.
2. New reader test `emits no format for a ulid column` asserts both `format` and `type` are null on the `ulid('reference')` fixture column.
3. `docs/auto-derivation.md:173` corrected to "a `ulid` column stays a bare string, matching the `HasUlids` convention — there is no standard OpenAPI ULID format" — the self-contradiction with `:85` is gone.

**Minor notes — addressed:**
- Note A: `Widget.php` docblock reworded accurately (no longer claims a non-existent `name`-cast collision; now correctly describes `price` as the cast-vs-migration case and the rest as uncast).
- Note B: the two worst long lines I named are wrapped (mac pattern extracted to `MAC_ADDRESS_PATTERN` const; the `nullable` line). A handful of docblock/`@param` and log-array lines still slightly exceed 100, but those are exactly the cases the soft limit tolerates and Pint passes — non-blocking, as flagged.
- Bonus (reviewer-2's ask): `records an explicit null default as a set default` regression test added on a `->default(null)` column, locking the `hasDefault=true` + `default=null` distinction.

No new issues introduced by the fix or the rebase; the pre-existing degrade/precedence coverage is intact (the `degrades.reference` uuid-sibling test still green alongside the new `widgets.reference` ulid test — different tables, no collision).

This is my approval (reviewer-1). Both my finding and reviewer-2's note are resolved; pending reviewer-2's own round-2 sign-off and the survey gate, this is mergeable.

Radiergummi:
🤖 **[lead]** 2/2 round-2 approvals (reviewer-1 Finding-1-resolved + reviewer-2), rebased a1284b37, CI green → survey gate (queued behind #438). NOTE: corpus apps HAVE migrations dirs, so expect a POSITIVE delta (format/maxLength/enum enrichment), not flat — enrichment with 0 regressions = pass. Reminder for merge: config key `read_migration_columns` → this one ESCALATES to human even when all gates pass.

Radiergummi:
🤖 **[coder]** rebased onto main c6c91c30 (CHANGELOG: all entries kept); CI re-gated; pushed c3fa6c06.

#438 auto-merged cleanly in CHANGELOG (its `#426` lint entry landed in a different section), so all three `[Unreleased]` entries are present (#427 Fractal + #414 migration + #426). No code conflict (#438 in `src/Support/Routing` + `src/Lint`). `composer check` green (3139 tests, pint, phpstan); all remote CI green on c3fa6c06.

Radiergummi:
🤖 **[surveyor]** Layer A pass — 0 regressions across 11 apps; migration-column enrichment confirmed firing (modest on well-typed corpus apps).

Compared against current `main` (`e2c69b49`); candidate pinned to the **current** PR head **`c3fa6c06`** (the worktree had advanced past the `a1284b37` I was handed — the coder pushed the review-round-1 fix; `c3fa6c06` is what's on the PR branch now, so this gates the live head).

**Deterministic count metrics: fully flat, 0 regressions, all 11 apps `gen_exit 0→0`** (paths/operations/apiOperations/responseSchemas/documentedResponses/requestBodies/componentSchemas/maxRequestProperties all unchanged).

The expected positive delta does not show up in Layer-A **counts** because migration reading enriches *existing* properties in place (formats, `maxLength`, `minimum`, `multipleOf`, enum, comments) rather than adding new schemas/operations — and it ranks **below** `$casts`/`@property`/attributes (the PR's own config comment: "only fill fields those sources left undefined"). On well-typed corpus apps those higher-precedence sources already cover most fields. I verified the feature actually fires with a direct spec-content diff (baseline `main` vs candidate):

| app | new `minimum` (unsigned cols) | spec bytes |
|-----|-------------------------------|------------|
| Vito (99 migrations) | +6 | +134 |
| AdvisingApp | +3 | +62 |

`date-time`/`uuid`/`maxLength`/`date` counts were identical baseline→candidate (already inferred from casts), confirming the precedence ordering works as designed. No regression anywhere, feature confirmed live. Clear to merge.

Radiergummi:
🤖 **[lead]** ALL quality gates green — 2/2 independent round-2 approvals (reviewer-1 Finding-1 [ulid≠uuid] resolved + reviewer-2), full CI matrix incl. swagger-php 5.8 + byte-identical Snapshot, survey Layer A PASS (0 regressions/0 crashes across 11 apps; enrichment confirmed live via content-diff — Vito +6 minimum constraints, AdvisingApp +3, precedence below casts verified). **Escalating to human, NOT auto-merging**, because it adds the config key `read_migration_columns` (config-surface change trips the controversy gate). PR is rebased onto current main c6c91c30 (c3fa6c06), ready + unmerged, worktree left in place. @Radiergummi please review the config key + the migration-reader approach. Follow-up #439 filed for the deferred Tier-0 `$attributes`-default read.

Radiergummi:
🤖 **[surveyor]** Layer A pass — 0 regressions / 0 new crashes across 11 apps; migration-column enrichment confirmed firing on Vito + AdvisingApp.

Re-ran against the **current** `main` **`c6c91c30`** (which advanced past my earlier `e2c69b49` baseline by exactly the #438 merge). I regenerated the baseline on `c6c91c30` and confirmed it is byte-identical to the `e2c69b49` baseline across every deterministic metric — corroborating that #438 was itself a flat delta. Candidate pinned to the live PR head **`c3fa6c06`**.

**Deterministic count metrics: flat, 0 regressions, all 11 apps `gen_exit 0→0`** (paths / operations / apiOperations / responseSchemas / documentedResponses / requestBodies / componentSchemas / maxRequestProperties all unchanged).

The expected enrichment does **not** show up in Layer-A *counts* because migration reading enriches *existing* properties in place and ranks below `$casts`/`@property`/attributes (the config comment: "only fill fields those sources left undefined"). I verified the feature fires with a direct baseline(`c6c91c30`)→candidate(`c3fa6c06`) **spec-content diff** across 5 apps:

| app | enrichment surfaced | spec bytes |
|-----|---------------------|------------|
| Vito | +6 `minimum` (unsigned cols) | +134 |
| AdvisingApp | +3 `minimum` | +62 |
| InvoiceNinja | none surfaced | 0 |
| Bagisto | none surfaced | 0 |
| Koel | none surfaced | 0 |

`format:date-time`/`uuid`/`maxLength`/`date` counts were identical baseline→candidate everywhere (already inferred from casts), confirming the precedence ordering works as designed. Net: the corpus is well-typed enough that casts/property-tags already cover most fields, so migration reading adds a handful of `minimum` constraints rather than a large delta — additive only, nothing lost or changed.

**Survey verdict: PASS.** (Escalates to human at merge per the new `read_migration_columns` config key.)

PR #437, opened 2026-06-22
feat(fractal): recognize $this->fractal root + literal transformWith()

Radiergummi:
Closes #427

### Scope (surviving, per the issue's final comments)

The Fractal lever is already shipped. The **only** uncovered gap here is the `spatie/laravel-fractal`
public API spelled with the transformer supplied **separately** via `transformWith()`, rooted on the
`$this->fractal` property (the "Pelican `transformWith` gap"):

```php
return $this->fractal->collection($data)->transformWith(new BookTransformer())->respond();
return $this->fractal->item($model)->transformWith(BookTransformer::class)->toArray();
```

`FractalHelperResponseResolver` today reads the transformer **only** from the 2nd argument of
`item()`/`collection()`, and accepts **only** the `fractal()` helper / `Spatie\Fractalistic\Fractal`
facade as the chain root. Two narrow extensions close the gap:

1. **Root:** accept `$this->fractal` (a `PropertyFetch` on `$this` named `fractal`) as a valid Fractal
   entrypoint, alongside the existing `fractal()` helper and `Fractal::create()` facade.
2. **Transformer source:** when `item()`/`collection()` has no 2nd-arg transformer, read it from a
   `->transformWith(new X() | X::class)` link in the same chain.

Explicitly **out of scope** (issue decision, dropped): InvoiceNinja's `$entity_transformer` /
`itemResponse`/`listResponse` (already shipped separately), and any `getTransformer()`/bespoke-helper
unwrap — that is `#[FractalResponse]`'s job. Honest expectation, restated from the issue: this likely
wins ~0 corpus apps (Pelican wraps in its own `getTransformer()`), but it completes the
package's public-API recognition, the same category as the other integrations.

### Tier

**Tier-1** (bounded AST whitelist), matching the existing resolver exactly: scan the first N
statements, match a fixed call-shape whitelist, literal transformer only (`new X()` / `X::class`),
no variable tracking. A variable/dynamic transformer or `transformWith($this->getTransformer(...))`
degrades with the existing note. No tier escalation.

### Why extend the existing resolver, not add a new one

The gap is two small relaxations of `FractalHelperResponseResolver::matchChainedShape()` /
`isFractalEntrypoint()` — same envelope/serializer/field machinery, same media-type logic, same
degrade notes. A second resolver would duplicate all of it and re-derive collection-vs-item. Surgical
edit to the one resolver is correct. (Confirmed the resolver already walks the full method-call chain
[...]

PR #438, opened 2026-06-22
fix(lint): share URI-placeholder extraction with the generator; fix {param:field} mishandling

### LINKED ISSUE #426: Share URI-placeholder extraction between the generator and the linter (and fix the linter's {param:field} mishandling)

Radiergummi:
Closes #426

### The bug (verified)

`ActionDescriptor::uriPlaceholders()` (`src/Routing/ActionDescriptor.php:214`) is correct: it
`ltrim`s RFC-6570 operator prefixes **and** strips a `{param:field}` custom-key suffix
(`explode(':', $name, 2)[0]`), returning the bare URI variable name + an optional flag.

`PathParameterUndeclared::extractPlaceholders()` (`src/Lint/Rules/PathParameterUndeclared.php:73`)
duplicates the extraction but **only `ltrim`s the prefix** — it never strips `:field`. So for a
`{member:slug}` segment it yields `member:slug`, compares it against the declared `member`
parameter, never matches, and emits a **false `path.parameter-undeclared` "Broken" finding**.

Confirmed the linter genuinely sees the raw form: `PathsStage::normalisePath()`
(`src/Support/Generator/Stages/PathsStage.php:149`) only prepends a leading slash — it does **not**
strip `:field` — so the OpenAPI path key (and thus `OperationNode->pathUri`) retains
`/members/{member:slug}` while the declared parameter is correctly `member`. The mismatch is real.

### Tier

Not an inference change — a **refactor + bugfix**. No tier ladder involved. (Branch typed `fix/`.)

### Approach

Extract one shared utility that both the generator and the linter consume, made **behaviorally
identical to the generator's (correct) extractor**, then delete the two divergent copies. The
generator returns `[bareName, optional]`; the linter wants names only — so the shared utility
returns the richer `list<array{string, bool}>` and the linter projects column 0.

The utility lives in `src/Support/Routing/` (alongside the existing `UriParameterResolver`,
`RouteModelBinding`, etc.). `src/Support/` may not depend on plugins (it doesn't here) and the linter
(`src/Lint/`) and routing (`src/Routing/`) may both depend on `src/Support/` — placement is clean.

### Files

**Create** `src/Support/Routing/UriPlaceholderExtractor.php`:
- A `final readonly` class (or a final class with a single `public static` method — pick the
  instance-method form if it needs no state, matching the codebase's no-speculative-DI lean; a pure
  `public static function extract(string $uri): array` is the lightest fit and needs no container
  wiring for either consumer). **Decision: a stateless static method**, since both call sites already
  have the URI string in hand and neither needs DI. Lift the generator's exact logic:
  ```php
  /** @return list<array{string, bool}> [bareName, optional] per placeholder */
  public static function extract(string $uri): array
  ```
  matching the regex `/\{([^?}]+)(\??)}/`, the `ltrim($raw, '+#./;?&=,!@|')` prefix strip, and the
  `:`-suffix strip. Carry over the explanatory docblock from `uriPlaceholders()` verbatim (it already
  documents *why* both strips are needed).
  - **[reviewer] regex note:** adopting the generator's `/\{([^?}]+)(\??)}/` is behavior-preserving
    for the linter's *matching* set — its old regex `/\{([^?}]+)\??}/` has the identical `[^?}]+`
    name capture, so the same placeholders are matched. The only behavior that changes is the
    post-match processing (the linter now also gets the `:field` strip it was missing). No new
    placeholders start or stop being recognized.

**Modify** `src/Routing/ActionDescriptor.php`:
- `uriPlaceholders()` delegates: `return UriPlaceholderExtractor::extract($this->route->uri());`.
- Keep the method (it's the public surface `OperationBuilder` consumes via `[$name, $optional]`); the
  regex/loop move into the utility. The method docblock stays (or shrinks to a one-liner pointing at
  the utility). **No public-signature change** — same name, same return shape.

**Modify** `src/Lint/Rules/PathParameterUndeclared.php`:
- Replace the private `extractPlaceholders()` body with
  `array_map(static fn(array $p): string => $p[0], UriPlaceholderExtractor::extract($pathUri))`
  (names only). Drop the now-unused `ltrim` / `preg_match_all` imports; add the utility import.
  Remove the trailing blank lines (lines 84-87) only because they're inside the method block I'm
  editing — no unrelated cleanup.

No change to `OperationBuilder` (it already destructures `[$name, $optional]`, unchanged), no
contract change, no config key, no stage-order change.

### Public-signature check (lead asked)

No public signature changes. `ActionDescriptor::uriPlaceholders()` keeps its name and
`list<array{string, bool}>` return; `OperationBuilder` is untouched. The new utility is a new
`@internal` class, not a modification of an existing public one. → **not a human-gate item** on the
signature axis. (Still `area:core` + generation-touching → survey gate applies, see below.)

### Tests (the bug-reproducing test comes first — TDD)

**`tests/Unit/Lint/Rules/PathParameterUndeclaredTest.php`** — add to the existing
"emits no finding when all placeholders are declared" `->with()` dataset (it already uses
`pathUri` + `names`):
- **`custom-key binding {param:field}`** → `['/members/{member:slug}', ['member']]` — **this is the
  bug reproducer**: currently emits a false finding, must emit none.
- **custom-key + optional** → `['/members/{member:slug?}', ['member']]` (proves the `?` and `:field`
  compose).
- **operator-prefixed + custom-key** → `['/files/{+path:disk}', ['path']]` (both strips at once).
- Add one **positive** case proving a genuinely-undeclared custom-key segment *still* reports under
  the bare name: `/members/{member:slug}` with **no** declared parameter → one finding whose message
  contains `member` (not `member:slug`).

**`tests/Unit/Support/Routing/UriPlaceholderExtractorTest.php`** (new) — unit-test the utility
directly so its contract is locked independent of either consumer:
- plain `{id}` → `[['id', false]]`; optional `{id?}` → `[['id', true]]`;
- operator prefix `{+path}` → `[['path', false]]`;
- custom key `{member:slug}` → `[['member', false]]`; custom-key + optional `{member:slug?}` →
  `[['member', true]]`;
- multiple segments in one URI preserve order;
- no placeholders → `[]`;
- a `:`-bearing name with an operator prefix `{+path:disk}` → `[['path', false]]` (strip order).

**Regression guard:** the existing `ActionDescriptorTest` and `CustomKeyBindingTest` /
`UnsignaturedPathParameterTest` must stay green (they pin `uriPlaceholders()` / generated-parameter
naming, proving the delegation is behavior-preserving on the generator side).

### Docs & CHANGELOG

- No `docs/` page documents the placeholder-extraction internals or this lint rule's edge behavior
  beyond the catalog row, so **no docs page change** (the `path.parameter-undeclared` catalog row in
  `docs/linting.md` is unchanged — same id, same meaning). If the reviewer's docs-gap duty disagrees,
  a one-line note in the rule catalog is the most that's warranted.
- `CHANGELOG.md [Unreleased] → Fixed`: "`path.parameter-undeclared` no longer false-positives on
  custom-key route bindings (`{param:field}`); URI-placeholder extraction is now shared between the
  generator and the linter." Reference `#426`. (Add a `### Fixed` subsection if absent under
  `[Unreleased]`.)

### Out of scope (surgical boundary)

- The OpenAPI **path key** still emits `/members/{member:slug}` (raw) while declaring a `member`
  parameter — normalizing the path string itself to `{member}` is a **separate** concern (would
  change generated output / snapshots and is not what this issue asks). Flagged, not done.
- `resolveUriParameters()` in `ActionDescriptor` (signature-binding match) is untouched — it keys off
  exact `{name}` / `{name?}` substring presence, a different mechanism, not part of this bug.

### Human-gate flags for the lead

- **Low controversy:** no contracts/config/stage-order/public-signature change; a new `@internal`
  utility + two delegations + a bugfix.
- `area:core` and the change touches the generator's `uriPlaceholders()` path → **survey gate
  applies**. Expectation: **zero spec diff** across the corpus (the generator already used the correct
  logic; only the linter behavior changes), so a flat Layer-A survey is the success signal. Worth a
  glance that no corpus app's generated paths shift.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -73,6 +73,7 @@ All notable changes to this project are documented here.
 - `--path` / `--diff` no longer leak findings for out-of-scope routes (#50).
 - `ValidationRulesToSchema` maps additional rules (`multiple_of`, Email/Exists/Unique/NotIn objects, wildcard keys) it previously dropped (#83).
 - ApiResources: actions that return an API Resource without declaring a return type (relying on convention or a third-party doc attribute) now resolve their response schema via the return-expression reader, instead of emitting no content; the refusal notice is suppressed on the untyped path to avoid a notice per non-resource action. (#391)
+- `path.parameter-undeclared` no longer false-positives on custom-key route bindings (`{param:field}`); URI-placeholder extraction is now shared between the generator and the linter. (#426)
 
 ### Documentation
 - README refreshed for the current feature set (Tier-1 method-body inference, SwaggerPhp plugin) and trimmed to a gentle overview that defers detail to `docs/`.
--- a/src/Lint/Rules/PathParameterUndeclared.php
+++ b/src/Lint/Rules/PathParameterUndeclared.php
@@ -12,11 +12,10 @@ use Radiergummi\OpenApi\Lint\LintContext;
 use Radiergummi\OpenApi\Lint\Tree\OperationNode;
 use Radiergummi\OpenApi\Lint\Tree\ParameterNode;
 use Radiergummi\OpenApi\Lint\Visitors\OperationRule as OperationRuleVisitor;
+use Radiergummi\OpenApi\Support\Routing\UriPlaceholderExtractor;
 
 use function array_map;
 use function in_array;
-use function ltrim;
-use function preg_match_all;
 use function sprintf;
 
 /**
@@ -72,16 +71,9 @@ final class PathParameterUndeclared implements Rule, OperationRuleVisitor
      */
     private function extractPlaceholders(string $pathUri): array
     {
-        if (preg_match_all("/\{([^?}]+)\??}/", $pathUri, $matches)) {
-            return array_map(
-                static fn(string $name): string => ltrim($name, '+#./;?&=,!@|'),
-                $matches[1],
-            );
-        }
-
-        return [];
+        return array_map(
+            static fn(array $placeholder): string => $placeholder[0],
+            UriPlaceholderExtractor::extract($pathUri),
+        );
     }
-
-
-
 }
--- a/src/Routing/ActionDescriptor.php
+++ b/src/Routing/ActionDescriptor.php
@@ -6,6 +6,7 @@ namespace Radiergummi\OpenApi\Routing;
 
 use Illuminate\Routing\Route;
 use Radiergummi\OpenApi\Enums\HttpMethod;
+use Radiergummi\OpenApi\Support\Routing\UriPlaceholderExtractor;
 use ReflectionAttribute;
 use ReflectionClass;
 use ReflectionFunction;
@@ -14,10 +15,7 @@ use ReflectionMethod;
 use ReflectionParameter;
 
 use function array_keys;
-use function explode;
 use function is_a;
-use function ltrim;
-use function preg_match_all;
 use function str_contains;
 
 /**
@@ -204,31 +202,11 @@ final class ActionDescriptor
     /**
      * Extracts every URI placeholder from the route template as `[bareName, optional]`.
      *
-     * The bare name strips both the binding-prefix characters Laravel allows (e.g. `{+slug}`) and a
-     * `{param:field}` custom-key suffix: the URI template variable is `param`, and Laravel keys its
-     * `wheres`/binding metadata on that bare name, so emitting or looking up `param:field` would be
-     * wrong on both counts.
-     *
      * @return list<array{string, bool}>
      */
     public function uriPlaceholders(): array
     {
-        if (preg_match_all('/\{([^?}]+)(\??)}/', $this->route->uri(), $matches) === 0) {
-            return [];
-        }
-
-        $placeholders = [];
-
-        foreach ($matches[1] as $index => $raw) {
-            $name = ltrim($raw, '+#./;?&=,!@|');
-            $bareName = str_contains($name, ':')
-                ? explode(':', $name, 2)[0]
-                : $name;
-
-            $placeholders[] = [$bareName, $matches[2][$index] === '?'];
-        }
-
-        return $placeholders;
+        return UriPlaceholderExtractor::extract($this->route->uri());
     }
 
     /**
--- a/src/Support/Routing/UriPlaceholderExtractor.php
+++ b/src/Support/Routing/UriPlaceholderExtractor.php
@@ -1,0 +1,48 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Support\Routing;
+
+use function explode;
+use function ltrim;
+use function preg_match_all;
+use function str_contains;
+
+/**
+ * Extracts URI template placeholders from a route URI, shared by the generator and the linter.
+ *
+ * @internal
+ */
+final class UriPlaceholderExtractor
+{
+    /**
+     * Extracts every URI placeholder from the route template as `[bareName, optional]`.
+     *
+     * The bare name strips both the binding-prefix characters Laravel allows (e.g. `{+slug}`) and a
+     * `{param:field}` custom-key suffix: the URI template variable is `param`, and Laravel keys its
+     * `wheres`/binding metadata on that bare name, so emitting or looking up `param:field` would be
+     * wrong on both counts.
+     *
+     * @return list<array{string, bool}>
+     */
+    public static function extract(string $uri): array
+    {
+        if (preg_match_all('/\{([^?}]+)(\??)}/', $uri, $matches) === 0) {
+            return [];
+        }
+
+        $placeholders = [];
+
+        foreach ($matches[1] as $index => $raw) {
+            $name = ltrim($raw, '+#./;?&=,!@|');
+            $bareName = str_contains($name, ':')
+                ? explode(':', $name, 2)[0]
+                : $name;
+
+            $placeholders[] = [$bareName, $matches[2][$index] === '?'];
+        }
+
+        return $placeholders;
+    }
+}
--- a/tests/Unit/Lint/Rules/PathParameterUndeclaredTest.php
+++ b/tests/Unit/Lint/Rules/PathParameterUndeclaredTest.php
@@ -31,7 +31,27 @@ it('emits no finding when all placeholders are declared as path parameters', fun
     'single placeholder declared'      => ['/users/{userId}', ['userId']],
     'RFC 6570 operator-prefixed name'  => ['/files/{+path}', ['path']],
     'no placeholders at all'           => ['/users', []],
+    'custom-key binding {param:field}' => ['/members/{member:slug}', ['member']],
+    'custom-key + optional'            => ['/members/{member:slug?}', ['member']],
+    'operator prefix + custom key'     => ['/files/{+path:disk}', ['path']],
 ]);
+
+it('reports an undeclared custom-key placeholder under its bare name', function (): void {
+    $operation = OperationNodeFactory::makeOperation(
+        pathUri: '/members/{member:slug}',
+        parameters: [],
+        responses: [],
+    );
+
+    $findings = iterator_to_array(
+        new PathParameterUndeclared()->checkOperation($operation, OperationNodeFactory::emptyContext()),
+    );
+
+    expect($findings)
+        ->toHaveCount(1)
+        ->and($findings[0]->message)->toContain('"{member}"')
+        ->and($findings[0]->fixHint)->toContain('name="member"');
+});
 
 it('emits a finding for an undeclared path placeholder', function (): void {
     $operation = OperationNodeFactory::makeOperation(
--- a/tests/Unit/Support/Routing/UriPlaceholderExtractorTest.php
+++ b/tests/Unit/Support/Routing/UriPlaceholderExtractorTest.php
@@ -1,0 +1,23 @@
+<?php
+
+declare(strict_types=1);
+
+use Radiergummi\OpenApi\Support\Routing\UriPlaceholderExtractor;
+
+uses()->group('openapi', 'routing');
+
+it('extracts placeholders as [bareName, optional] pairs', function (string $uri, array $expected): void {
+    expect(UriPlaceholderExtractor::extract($uri))->toBe($expected);
+})->with([
+    'plain'                        => ['/users/{id}', [['id', false]]],
+    'optional'                     => ['/users/{id?}', [['id', true]]],
+    'operator prefix'              => ['/files/{+path}', [['path', false]]],
+    'custom key'                   => ['/members/{member:slug}', [['member', false]]],
+    'custom key + optional'        => ['/members/{member:slug?}', [['member', true]]],
+    'operator prefix + custom key' => ['/files/{+path:disk}', [['path', false]]],
+    'no placeholders'              => ['/users', []],
+    'multiple preserve order'      => [
+        '/users/{userId}/posts/{postId}',
+        [['userId', false], ['postId', false]],
+    ],
+]);


Radiergummi:

Refactor + bugfix. Verified the bug: `PathParameterUndeclared::extractPlaceholders()` only `ltrim`s the operator prefix and never strips the `:field` suffix, so `{member:slug}` yields `member:slug` and false-positives against the declared `member`. Confirmed `PathsStage::normalisePath()` only prepends a slash, so the linter genuinely sees the raw `{member:slug}` in `pathUri`. Plan: lift the generators correct logic into a new `@internal Support\Routing\UriPlaceholderExtractor::extract()` returning `list<array{string,bool}>`; `ActionDescriptor::uriPlaceholders()` and the linter both delegate (linter projects names). No public-signature/contract/config/stage-order change. New bug-reproducer test (`/members/{member:slug}` declared `member` → no finding) + a unit test for the utility. Survey expectation: zero corpus spec diff (only linter behavior changes).

Radiergummi:
🤖 **[lead]** planning done → plan-review. Reviewer: verify the new `Support\Routing\UriPlaceholderExtractor` reproduces the generator's correct logic exactly (bare name + flag), both consumers delegate, the {param:field} linter test locks the bugfix, and the deferred path-key normalization is correctly out of scope.

Radiergummi:
🤖 **[reviewer]** plan-review: approved (added one regex-equivalence note).

Verified the bug and the plan against source, not just prose:

- **Bug confirmed.** `PathParameterUndeclared::extractPlaceholders()` (`src/Lint/Rules/PathParameterUndeclared.php:73-83`) only `ltrim`s the operator prefix; it never strips the `:field` suffix. `PathsStage::normalisePath()` (`src/Support/Generator/Stages/PathsStage.php:149`) only prepends `/`, so `pathUri` retains `/members/{member:slug}`. The `in_array($placeholder, $declaredNames, strict: true)` check at `:49` then compares raw `member:slug` against the bare declared `member` → false `path.parameter-undeclared` Broken finding. The generator's `uriPlaceholders()` (`src/Routing/ActionDescriptor.php:214-232`) already does both strips correctly. Real, as filed.
- **Tier:** correctly none (refactor + bugfix, `fix/` branch).
- **Approach — sound.** One `@internal` stateless static `Support\Routing\UriPlaceholderExtractor::extract(): list<array{string,bool}>` reproducing the generator's correct logic; both `ActionDescriptor::uriPlaceholders()` and the linter delegate (linter projects column 0). Single source of truth, drift impossible.
- **Boundaries — clean.** `src/Support/Routing/` already houses sibling routing utilities; both `src/Routing/` and `src/Lint/` may depend on `src/Support/`; no plugin dependency introduced.
- **Public-signature check (lead's ask) — confirmed none.** `uriPlaceholders()` keeps its name + `list<array{string,bool}>` return; `OperationBuilder` (which destructures `[$name, $optional]`) is untouched; the utility is a brand-new `@internal` class. Not a signature-axis human-gate item.
- **Tests — TDD, edges covered.** Bug-reproducer first. The existing no-finding `->with()` dataset (`tests/Unit/Lint/Rules/PathParameterUndeclaredTest.php:30-34`, already has `/files/{+path}`) cleanly accommodates the `{member:slug}` / `{member:slug?}` / `{+path:disk}` rows; the existing finding-emitting tests (`:36`, `:54`) are the template for the negative case (undeclared custom-key still reports under the bare `member`). New `UriPlaceholderExtractorTest` locks the utility contract incl. strip-order. Regression guards on `ActionDescriptorTest` / custom-key generation tests prove the delegation is behavior-preserving.
- **Out of scope** (raw path-key `{member:slug}` left un-normalized to avoid snapshot churn) is correctly reasoned and flagged, not done.

**Tightening I made:** added a regex-equivalence note to the plan — adopting the generator's `/\{([^?}]+)(\??)}/` is behavior-preserving for the linter's *matching* set (identical `[^?}]+` capture); only post-match processing gains the missing `:field` strip. Keeps the coder from second-guessing the regex swap.

**Human-gate flags for the lead:** low controversy (no contracts/config/stage-order/public-signature change). `area:core` + generation-touching → survey gate applies; expected signal is **zero spec diff** across the corpus (generator logic unchanged; only linter behavior changes), so a flat Layer-A survey is the pass.

Verdict: **approved-for-coding.**

Radiergummi:
🤖 **[lead]** plan approved by reviewer-1 → coding. Coder: new @internal Support\Routing\UriPlaceholderExtractor::extract() reproducing the generator's logic exactly (incl. :field strip); both ActionDescriptor::uriPlaceholders() and PathParameterUndeclared delegate. TDD: bug-reproducer test first ({param:field} false-positive), then utility unit test locking strip-order. Leave raw path-key normalization out of scope.

Radiergummi:
🤖 **[coder]** implemented; CI green locally and remote. Pushed bc43e8a6. Ready for review.

- New `@internal Support\Routing\UriPlaceholderExtractor::extract()` (stateless static, `list<array{string, bool}>`) lifts the generator's exact logic incl. the `:field` strip the linter lacked.
- `ActionDescriptor::uriPlaceholders()` and `PathParameterUndeclared::extractPlaceholders()` both delegate (linter projects column 0); orphaned `ltrim`/`preg_match_all`/`explode` imports removed.
- TDD: bug-reproducer first (`{member:slug}` declared `member` → no `path.parameter-undeclared`), then `UriPlaceholderExtractorTest` locking strip-order (operator prefix + `:field`). 3093 tests pass; Pint clean; PHPStan 0 errors. All remote matrix jobs green (incl. swagger-php 5.8, PHP 8.5, Snapshot).
- Raw path-key normalization left out of scope per plan (no spec/snapshot diff expected — flat Layer-A is the survey pass).

Radiergummi:
🤖 **[lead]** coder-3 pushed bc43e8a6, local + remote CI green → code review (2 independent approvals required). Awaiting reviewer availability (both on #437). PR ready.

Radiergummi:
🤖 **[reviewer-2]** review round 1: approved, no concerns.

Reviewed adversarially at bc43e8a6 in an isolated worktree (removed after). Independently verified:

- **Generator side behavior-preserving (the critical claim).** Diffed `UriPlaceholderExtractor::extract()` against the old `ActionDescriptor::uriPlaceholders()` body line-by-line: identical regex `/\{([^?}]+)(\??)}/`, identical `ltrim($raw, ...)` prefix strip, identical `str_contains/explode(:, $name, 2)[0]` `:field` strip, identical optional flag. Only `$this->route->uri()` → `$uri` param. No semantic change. **The Snapshot (locked toolchain) CI job passed** — the byte-exact guard confirms zero generated-spec drift.
- **Bug-reproducer genuinely fails without the fix (TDD).** Reverted just the linter `extractPlaceholders()` to its old `/\{([^?}]+)\??}/` + `ltrim`-only body in the worktree: exactly the 4 new cases fail (3 custom-key datasets + the positive-case test), with the failure showing the old code yielding `"{member:slug}"`. Restoring the fix makes them pass; pre-existing cases stay green throughout.
- **Linter regex change is matching-preserving.** Old `/\{([^?}]+)\??}/` and new `/\{([^?}]+)(\??)}/` share the `[^?}]+` name capture → identical match set; the only behavior delta is the added `:field` strip the linter lacked. Confirmed.
- **Positive-case assertions accurate.** The finding message `Path placeholder "{%s}"` + fixHint `name="%s"` with bare `member` produce exactly `"{member}"` / `name="member"`.
- **Out-of-scope boundary clean.** Only `uriPlaceholders()` changed in `ActionDescriptor`; `resolveUriParameters()` and `bindingFieldFor()` are untouched (and `ActionDescriptorTest`, incl. scoped-nested custom-key + `bindingFieldFor()`, stays green). Raw path-key normalization correctly deferred (no snapshot diff).
- **Orphaned imports removed cleanly** (`explode`/`ltrim`/`preg_match_all` from `ActionDescriptor`, `ltrim`/`preg_match_all` from the rule); the trailing blank lines inside the edited method block tidied. Placement in `Support\Routing\` is boundary-correct (no plugin dep; linter + routing may both depend on Support). `@internal` present.
- **Gates:** lint+extractor+ActionDescriptor unit tests pass (25); PHPStan level 8 clean on all 3 changed src files; Pint clean; full remote matrix green incl. swagger-php 5.8, PHP 8.5, Snapshot, Coverage.

This is my independent second approval.

Radiergummi:
🤖 **[reviewer-1]** review round 1: approved, no concerns.

Reviewed bc43e8a6 adversarially in an isolated worktree (removed after), independently of reviewer-2.

**Generator side is behavior-preserving — verified byte-for-byte.** I diffed the new `UriPlaceholderExtractor::extract()` body against the old `ActionDescriptor::uriPlaceholders()` body from `main` (substituting `$this->route->uri()` → `$uri`): **byte-identical** — same regex `/\{([^?}]+)(\??)}/`, same `ltrim($raw, '+#./;?&=,!@|')` prefix strip, same `str_contains/explode(':', $name, 2)[0]` `:field` strip, same optional flag. Only the signature changed (`public function uriPlaceholders()` → `public static function extract(string $uri)`). `uriPlaceholders()` now delegates. No generator-side semantic change; the Snapshot CI job passing corroborates zero spec drift.

**Bug-reproducer genuinely fails without the fix (TDD, non-vacuous).** I reverted *only* the linter's `extractPlaceholders()` to its old `/\{([^?}]+)\??}/` + `ltrim`-only body and re-ran the rule test: **exactly the 4 new cases fail** — the three custom-key `->with()` datasets (`{member:slug}`, `{member:slug?}`, `{+path:disk}`) and the new positive-case test — while all pre-existing cases stay green. Restoring the fix makes them pass. Unlike some suites, none of these are vacuous: each fails on the buggy code for the right reason (old code yields `member:slug`).

**Positive-case assertion is correct.** `reports an undeclared custom-key placeholder under its bare name` asserts the finding message contains `"{member}"` and fixHint `name="member"` — precisely the bug-fix outcome (buggy code would emit `"{member:slug}"`).

**Linter regex change matching-preserving:** old `/\{([^?}]+)\??}/` and new `/\{([^?}]+)(\??)}/` share the `[^?}]+` name capture → identical match set; only post-match processing gains the `:field` strip. Confirmed.

**Scope / boundaries / hygiene:**
- Only `uriPlaceholders()` changed in `ActionDescriptor`; `resolveUriParameters()` and `bindingFieldFor()` untouched. Raw path-key normalization correctly left out of scope (no snapshot churn).
- Orphaned imports removed correctly: `explode`/`ltrim`/`preg_match_all` from `ActionDescriptor`, `ltrim`/`preg_match_all` from the rule. I checked the *retained* imports are still used — `str_contains` stays in `ActionDescriptor` (used at `:223` in the untouched `resolveUriParameters()`), `array_keys`/`is_a` still used, `array_map`/`in_array`/`sprintf` still used in the rule. No over-removal, no orphans left.
- `Support\Routing\` placement is boundary-correct (no plugin dep; both `Routing\` and `Lint\` may depend on `Support\`). `@internal` present, strict-types header present, no lines over 100 chars in the changed source.

**Gates (run myself):** Pint --test on all 5 changed files → passed. `composer lint` / PHPStan L8 → 1307 files, no errors. Targeted tests (extractor + rule + `ActionDescriptorTest`) → 25 passed. Full suite → **3092 passed**, 0 failures (coverage warning only).

This is my independent approval (reviewer-1). Survey gate (area:core) applies — flat corpus delta is the expected success signal per plan.

Radiergummi:
🤖 **[lead]** 2/2 independent approvals (reviewer-1 + reviewer-2), full CI matrix green incl. Snapshot, low controversy → survey gate. Queued behind #437 (surveyor ×1). Expected: flat Layer-A (Snapshot already passing corroborates zero spec drift).

Radiergummi:
🤖 **[coder]** rebased #438 onto main (CHANGELOG: both entries kept); CI re-gated; pushed bc8a7ccf

Clean rebase onto origin/main after #437 merged. No code conflict (#437 was confined to src/Plugins/Fractal + docs). CHANGELOG resolved as an append — #437's `$this->fractal` entry (#427) and my `path.parameter-undeclared` entry (#426) both present on their own lines. `composer check` green locally (3103 tests). All remote matrix jobs green: PHP 8.4/8.5 × Laravel 12/13, swagger-php 5.8, Snapshot, Coverage, quality.

Heads-up: origin/main carries a stray **empty** commit `7063c817` titled "fix: Share URI-placeholder extraction ... (#426)" — zero file changes. Looks like a title-only/accidental merge artifact for this same issue; it contributes no diff and main still lacks the utility file, so my bc8a7ccf remains the complete real change. Flagging for the lead in case it needs cleanup.

Radiergummi:

Baseline regenerated on current `main` (`e2c69b49`) — the prior cached baseline was stale (built on `16d4a94c`, before the #437 merge; that merge was itself a flat delta, so corpus numbers carry over). Candidate pinned to PR worktree `bc8a7ccf`.

Fully flat delta across the corpus: every deterministic generator metric (paths, operations, apiOperations, responseSchemas, documentedResponses, requestBodies, componentSchemas) is identical, all 11 apps `gen_exit 0→0`, **and** lint-findings totals are unchanged on every app. The behavior-preserving `UriPlaceholderExtractor` lift produces zero generator drift (corroborating the byte-identical Snapshot CI), and the `{param:field}` false-positive fix simply has no corpus app exercising field-bound route placeholders, so it neither helps nor hurts here — as expected for a refactor + lint-fix.

| app | paths | ops | apiOps | respSchemas | docResp | reqBodies | components | gen | lint |
|-----|-------|-----|--------|-------------|---------|-----------|------------|-----|------|
| AdvisingApp | · | · | · | · | · | · | · | 0→0 | · |
| AureusERP | · | · | · | · | · | · | · | 0→0 | · |
| Bagisto | · | · | · | · | · | · | · | 0→0 | · |
| BookStack | · | · | · | · | · | · | · | 0→0 | · |
| Coolify | · | · | · | · | · | · | · | 0→0 | · |
| InvoiceNinja | · | · | · | · | · | · | · | 0→0 | · |
| Koel | · | · | · | · | · | · | · | 0→0 | · |
| Lychee | · | · | · | · | · | · | · | 0→0 | · |
| Pelican | · | · | · | · | · | · | · | 0→0 | · |
| SpeedtestTracker | · | · | · | · | · | · | · | 0→0 | · |
| Vito | · | · | · | · | · | · | · | 0→0 | · |

(· = identical to baseline.) Clear to merge.

Radiergummi:
🤖 **[lead]** non-controversial + all gates green (2/2 independent approvals, full CI matrix incl. swagger-php 5.8 + byte-identical Snapshot, survey Layer A 0-regression flat delta, src diff 92 lines, no contracts/config/signature change) → squash-merged, closed #426, branch + worktree removed.

PR #440, opened 2026-06-22
feat(responses): fill model field 'default' from a Tier-0 $attributes read

### LINKED ISSUE #414: feat(responses): read migration column types to enrich model field schemas (formats, constraints, enum, comments)

### LINKED ISSUE #439: feat(responses): fill model field 'default' from a Tier-0 $attributes read (follow-up to #414)
Follow-up split out of #414 / PR #436 to keep that PR surgical.

#414's migration-column reader fills a field's `default` from the migration `->default(...)` keyword. A sibling, even-cheaper **Tier-0** source exists: an Eloquent model's `$attributes` array declares per-attribute defaults directly on the class — no AST, pure reflection.

**Scope:** in `EloquentModelToSchema`, when a field's `default` is still `is_undefined()` after casts/@property/attribute/migration, fill it from `$model->getAttributes()` (the static `$attributes` property) if present. Strictly below all existing precedence sources; own test; must not gate the migration path.

**Out of scope:** runtime-computed defaults, accessor-derived values (Tier-2).

Context: coder deferred this from PR #436 (noted in that PR's coder comment) — the migration `->default()` half shipped; the core does not depend on this add.

Radiergummi:
## Implementation plan — fill model field `default` from a Tier-0 `$attributes` read

Closes #439. Follow-up split out of #414 / PR #436 to keep that PR surgical.

> **⚠️ MERGE-ORDER DEPENDENCY — read first.** This change extends the `applyMigrationColumn`
> chokepoint introduced by **PR #436** (closes #414), which is **approved + CI-green but not yet
> merged** (the maintainer merges it himself). The integration point (`applyMigrationColumn` in
> `EloquentModelToSchema`, the `default` write at `$column->hasDefault && is_undefined($property->default)`)
> **does not exist on current `main`.** The coder must **branch/rebase onto post-#436 `main`** before
> writing any code. If #436's shape changed during its own review, re-read its final diff first.

### Tier

**Tier-0 — reflection & signatures.** An Eloquent model's static `$attributes` array declares
per-attribute defaults directly on the class. A bare `new $modelClass()` instance exposes exactly
those via `$model->getAttributes()` — pure reflection, no AST, no migration parsing, no live DB.
This is **cheaper than #436's Tier-1** migration reader and is the lowest tier that captures the
idiom. Build here.

### Precedence

The new source ranks **strictly below every existing `default` source**, including #436's migration
`->default(...)`. Final order for a property's `default`, highest wins:

1. cast / `@property` / authoring attribute (whatever the property already carries)
2. migration `->default(<literal>)` (#436)
3. **`$attributes` array entry (this PR)** ← new, lowest

Mechanically this is automatic: #436 already writes the migration default into `$property->default`
**before** we run, and our write is guarded by `is_undefined($property->default)`. So if a migration
default (or any earlier source) populated it, the `$attributes` read is a no-op. We only fill the
gap. We must therefore call our fill **after** `applyMigrationColumn` at every call site, or fold it
into the same chokepoint *after* the migration `default` write (preferred — single guard, single
ordering guarantee). See "Approach".

### Approach

`$attributes` is read once per model in `metadataFor()` (a `$model` instance already exists there for
`getCasts()`, `getTable()`, etc.). Add it to the cached metadata array:

- New metadata key `'attributes' => $model->getAttributes()` (typed `array<string, mixed>`).
  `getAttributes()` on a fresh instance returns the class-declared `$attributes` verbatim. No new
  model instantiation; reuse the existing `$model`.

Then fill `default` from it as the **lowest-precedence** source. Two viable wirings; **prefer (A)**:

**(A) Fold into the migration chokepoint (preferred).** `applyMigrationColumn` is already the single
place `default` is written for the migration source, and it already runs at both call sites
(`propertyFor` single-property path + the bulk `schemaFor`/`propertiesFor` loop). Pass the model's
`$attributes` array alongside the `ColumnMetadata` (or via a sibling param), and **after** the
existing `if ($column->hasDefault && is_undefined($property->default))` block, add:

```php
if (is_undefined($property->default) && array_key_exists($name, $attributes)) {
    $property->default = $attributes[$name];
}
```

This guarantees migration-default-wins positionally (the migration write precedes it in the same
method) and reuses the existing `$ref`-guard early-return at the top of `applyMigrationColumn` (a
`$ref` property — relation/enum cast — must not take a `default` sibling that shapes type; the early
return already prevents that, which is correct: a model whose `$attributes` seeds an FK id while the
property is a `$ref` relation should not get a stray `default`).

> Note: `applyMigrationColumn` currently receives only `(OA\Property $property, ?ColumnMetadata $column)`.
> The property name is `$property->property`. So no extra `$name` param is needed — read
> `$property->property` for the array key. The only new input is the `$attributes` array. Decide
> between threading it as a param vs. reading it from a passed `$metadata` slice; pick whichever
> keeps the two call sites cleanest. Renaming the method (it is no longer migration-only) is
> **optional and out of scope** unless it reads cleanly — keep the diff surgical.

**(B) Separate `applyAttributeDefault` method** called immediately after `applyMigrationColumn` at
both call sites. Cleaner separation but duplicates the `is_undefined` guard and the two-call-site
ceremony, and risks a future caller forgetting the second call. (A) is preferred for that reason.

Either way: **must not gate or alter any other path.** The migration reader, nullability relaxation,
type/format/enum fills, `required` derivation, readOnly marking — all untouched.

### Edge cases (each gets a test branch)

| Case | Expected |
|---|---|
| `$attributes` has an entry, no other default source | property `default` = that value |
| `$attributes` value present **and** migration `->default()` present | migration value wins (our write skipped by `is_undefined`) |
| `$attributes` value present **and** cast/`@property`-derived default present | earlier source wins |
| `$attributes` value is `null` (explicit `'col' => null`) | `default` = `null` (use `array_key_exists`, **not** `isset`/`?? `, so an explicit null is honored and distinct from "absent") |
| field has **no** `$attributes` entry | no `default` written; property unchanged |
| property is a `$ref` (relation / enum cast) | no `default` written (existing `$ref` early-return) |
| empty `$attributes` (the common case — most models declare none) | no-op across all properties |
| `$attributes` key not a real column/property (no matching property emitted) | irrelevant — we only fill properties that exist |

`array_key_exists` (not `isset`) is load-bearing for the explicit-`null` case; call it out in review.

### Test (TDD — failing test first)

`tests/Unit/Support/Extraction/EloquentModelToSchemaTest.php` (extend the existing suite; it already
exercises casts, `@property`, timestamp defaults). Add a fixture model with a static `$attributes`
array, e.g.:

```php
protected $attributes = ['status' => 'draft', 'priority' => 0, 'archived_at' => null];
```

Tests to add (write before the implementation; each must fail first):

1. **fills `default` from `$attributes`** — `status` → `default: 'draft'`, `priority` → `default: 0`.
2. **explicit `null` in `$attributes` is honored** — `archived_at` → `default: null` present (assert
   the key exists and is null, distinguishing from "no default key").
3. **migration `->default()` outranks `$attributes`** — a model whose table migration declares
   `->default('published')` on `status` while `$attributes` says `'draft'` → `default: 'published'`.
   (Depends on #436 fixtures; reuse a `tests/Fixtures/Migrations/*` migration + matching model.)
4. **cast/`@property`-supplied default outranks `$attributes`** — if any existing path can seed a
   default ahead of us, assert it wins; otherwise document that only migration can outrank and skip.
5. **no `$attributes` entry → no `default`** — a property absent from `$attributes` keeps
   `default` undefined.
6. **`$ref` property is not given a `default`** — a relation/enum-cast property whose name appears in
   `$attributes` stays clean (guards the `$ref` early-return).

Run `vendor/bin/pest tests/Unit/Support/Extraction/EloquentModelToSchemaTest.php`, `vendor/bin/pint
--test`, and `composer lint` (PHPStan L8) green. The `snapshot` byte-exact group must stay green; if
any corpus/snapshot model carries a real `$attributes` array, the new `default` will surface — re-bless
snapshots only if the change is correct and intended, and note it in the PR.

### Docs / CHANGELOG bookkeeping

- **`docs/auto-derivation.md`** — the "Eloquent model response schemas" section #436 edited (the
  numbered type-resolution list and/or the `default` prose). Add one clause: `default` is also filled
  from the model's static `$attributes` array as the **lowest-precedence** source, below a migration
  `->default()`. Keep it to a sentence; mirror #436's wording register.
- **`CHANGELOG.md` `[Unreleased]` → Added** — one bullet: "Eloquent model response `default` is
  filled from the model's static `$attributes` array (Tier-0 reflection via `getAttributes()`),
  ranking below `$casts`/`@property`/attributes and below a migration `->default()` (#414). (#439)".
- No `docs/config.md` change — **no new config key** (see below).

### Controversy assessment

**Not controversial.** No `src/Contracts/**` change, no stage-order change, no public-signature
change, **no new config key** (confirmed — this is a pure additive lowest-precedence fill; #436's
`openapi.read_migration_columns` flag governs migration parsing only and does not need a sibling for a
zero-cost reflection read that runs unconditionally). Touches only `src/Support/Extraction/
EloquentModelToSchema.php` (+ test fixtures, docs, changelog). Generation-affecting (`area:responses`),
so it **should run the survey gate** (Layer A: 0 regressions across the corpus) before merge — a real
`$attributes` default surfacing on a corpus model is an *expected, correct* addition, not a regression,
but confirm no schema *count* drops.

### Files

- **modify** `src/Support/Extraction/EloquentModelToSchema.php` — add `'attributes'` to metadata
  cache + its `@return`/`@param` array-shape docblocks (the `metadataFor` shape and the
  `applyMigrationColumn`/`propertiesFor`/`readOnlyNames` shape annotations all list the keys; add
  `attributes: array<string, mixed>` to each that #436 touched); fill `default` after the migration
  `default` write.
- **add/modify** test fixture model(s) under `tests/Fixtures/` with a static `$attributes` array (and,
  for test 3, a matching migration fixture under `tests/Fixtures/Migrations/` per #436's pattern).
- **modify** `tests/Unit/Support/Extraction/EloquentModelToSchemaTest.php` — the six tests above.
- **modify** `docs/auto-derivation.md`, `CHANGELOG.md`.

Keep it surgical: every changed line traces to "fill `default` from `$attributes` as the lowest
source." No method rename unless it reads cleanly, no refactor of the migration reader, no new config.


Radiergummi:

Tier-0 (reflection via `getAttributes()`); folds into #436's `applyMigrationColumn` chokepoint as the lowest-precedence `default` source (below migration `->default()`, guarded by `is_undefined`). No new config key, not controversial.

**Gated on PR #436 merging first** — the integration point does not exist on current `main`; the coder must rebase onto post-#436 `main`.

Radiergummi:
🤖 **[lead]** planning done → plan-review. Reviewer: validate the Tier-0 $attributes read folds correctly into #436's applyMigrationColumn chokepoint, ranks strictly below migration default (positional is_undefined guard), uses array_key_exists (not isset) for explicit-null, and stays surgical (no config key). Plan-review can proceed now; CODING stays gated on #436's merge.

Radiergummi:
🤖 **[reviewer]** plan-review: approved (gated on #436 merge).

Validated against the post-#436 chokepoint (`gh pr diff 436`) — the plan matches the real integration point:

1. **Tier-0 discipline ✓** — `$model->getAttributes()` on the existing `metadataFor()` instance; no new instantiation, no AST, no DB. Lower tier than #436, correctly built here.
2. **Folds after migration write, `is_undefined`-guarded ✓** — #436 writes the migration default at `$column->hasDefault && is_undefined($property->default)` first; the `$attributes` fill running after with its own `is_undefined` guard makes migration (and every earlier source) positionally win. Approach (A) is the right call.
3. **`array_key_exists` not `isset` ✓ + verified** — confirmed `$property->default = null` is treated as *defined* by swagger-php (`Generator::isDefault(null) === false`) and serializes as `"default": null`. So the explicit-`null` case works and `array_key_exists` is genuinely load-bearing. Good that the plan calls it out.
4. **`$ref` early-return reused ✓** — folding below the existing `is_defined($property->ref)` early-return means a `$ref` (relation / enum-cast) property correctly gets no stray `default`. Test 6 guards it.
5. **Edge cases + failing-first TDD ✓** — six tests cover no-entry, explicit-null, migration-wins, cast-wins (or documented-skip), $ref-clean, empty `$attributes`. Comprehensive; test 3 can reuse #436’s `tests/Fixtures/Migrations/*` widgets pattern.
6. **No config key / not controversial ✓** — pure additive lowest-precedence reflection read, runs unconditionally; correctly *not* gated behind `read_migration_columns` (that flag governs migration parsing only). No `Contracts`/stage-order/signature change.
7. **#436 merge-order dependency ✓** — prominently flagged; coder must branch/rebase onto post-#436 `main`.

**Non-blocking steer:** prefer threading `$attributes` as a third param on `applyMigrationColumn` (both call sites already have `$metadata` in scope) over a method rename — keeps the positional-precedence guarantee in one method, surgical diff. The plan already leans this way; just making it the default.

Approved for coding **once #436 lands**.

Radiergummi:
🤖 **[lead]** #436 merged (chokepoint now in main 6f5bc71b) → coding unblocked. Coder: rebase the #440 branch onto current main first (it was branched pre-#436), then implement the Tier-0 $attributes read folding into applyMigrationColumn (thread $attributes as a 3rd param per reviewer-1's steer, not a method rename). TDD failing-first. Non-controversial (no config key) → auto-merge candidate.

Radiergummi:
🤖 **[coder]** deviation: skipped the planned test-4 (cast/@property-derived default outranks \`$attributes\`).

No existing inference path seeds \`$property->default\` ahead of this fill: \`$casts\` and \`@property\` only set type/format/description/nullability, never \`default\`; the migration \`->default()\` is the *only* higher-precedence default source. The plan explicitly allowed this ("otherwise document that only migration can outrank and skip"). The \`is_undefined($property->default)\` precedence guard is fully exercised by the migration-wins test (\`state\` → \`published\`), so coverage is intact. If a future source seeds \`default\` earlier, the same guard already orders us below it.

Radiergummi:
🤖 **[coder]** implemented; CI green (all matrix jobs incl. swagger-php 5.8 + Snapshot). Pushed a6f70a28. Ready for review.

- Rebased onto post-#436 main (HEAD 6f5bc71b); clean rebase, no CHANGELOG conflict (branch carried only the placeholder commit).
- `$attributes` read via `getAttributes()` on the existing `metadataFor()` instance, cached as `attributes: array<string,mixed>`; filled as the lowest-precedence `default` via a new `applyAttributeDefault()` called at the end of the `applyMigrationColumn` chokepoint (threaded `$attributes` as the 3rd param per reviewer-1's steer, no rename).
- `array_key_exists` honours explicit null → `default: null`; `$ref` early-return keeps relation/enum-cast props default-free.
- 5 new tests (fills default incl. falsy `0`, explicit-null, migration-wins, no-entry, $ref-clean); full suite 3201 green, Pint + PHPStan L8 clean. Test-4 skipped (no cast/@property default path) — see deviation comment.
- Generation-affecting (`area:responses`) → survey gate applies after review.

Radiergummi:
🤖 **[lead]** coder-1 pushed a6f70a28, rebased onto post-#436 main, full CI green → code review (2 independent approvals). Non-controversial (no config key, src diff 111 lines mostly re-indent) → auto-merge candidate. reviewer-1 + reviewer-2 review separately.

Radiergummi:
🤖 **[reviewer-2]** review: approved, no concerns.

Independent second pass (fresh eyes, did not plan-review). Ran `composer check` in a worktree at a6f70a28 — Pint passed, PHPStan L8 "No errors", 3201 tests pass.

Adversarial angles all clear:
1. **Precedence** — airtight. Migration `default` writes before `applyAttributeDefault`; the `is_undefined($property->default)` guard no-ops the `$attributes` read when any earlier source set it. `state` → `published` verified.
2. **Explicit null** — `array_key_exists` (not `isset`) emits `default: null`; tested.
3. **`$ref`** — `is_defined($property->ref)` early-return precedes `applyAttributeDefault`; relation/enum-cast props stay default-free. Tested via `status` (backed enum `$ref`).
4. **Falsy values** — `is_undefined` → `Generator::isDefault` (sentinel is a non-empty string), so `0`/`false`/`""` are never read as unset. `priority => 0` asserted `->toBe(0)`.
5. **Skipped test-4** — verified by grep: the only two writes to a property `default` in the model path are the migration write and the new `$attributes` write. No `$casts`/`@property`/TypeNode path seeds `default`, so test-4 is genuinely untestable; the deviation is correct and the `is_undefined` guard is fully exercised by the migration-wins test.
6. **Re-indent** — confirmed pure re-indentation under a new `if ($column !== null)` guard (so the fill runs even with no migration) + null-safe `$column?->description` in the ref branch. No hidden logic change; every inner write kept its guard and order.

Bookkeeping present: CHANGELOG `[Unreleased]` bullet, `docs/auto-derivation.md` clause with a clean 6→7 renumber (no dangling list refs). No `Contracts`/config/stage-order change.

Note (non-blocking): one transient failure of the migration-wins test on the worktree first run did not reproduce across 21 subsequent runs and CI is green — environmental blip on a cold worktree, not a defect.

Radiergummi:
🤖 **[reviewer-1]** review round 1 — 1 finding (otherwise approve):

Reviewed `a6f70a28` on the worktree (rebased clean on #436 `6f5bc71b`). Full suite **3201 passed**, PHPStan L8 **no errors**, Pint clean.

**Verified the implementation adversarially:**
- **3rd-param steer taken, no rename ✓** — method stays `applyMigrationColumn(OA\Property, ?ColumnMetadata, array $attributes)`; the `$attributes` fill is delegated to a well-named private `applyAttributeDefault()` called at the **end** of the method (line 643), after the migration `default` write (line 629). Precedence lives in one place.
- **`is_undefined($property->default)` guard ✓ + revert-probed** — removing the guard fails *only* the migration-wins test. So a migration `->default()` (and any earlier source) wins positionally.
- **`array_key_exists` not `isset` ✓ + revert-probed** — swapping to `isset` fails *only* the explicit-null test. Confirmed `$property->default = null` is treated as defined by swagger-php and serialises as `"default": null`.
- **`$ref` early-return ✓ + revert-probed** — moving the `applyAttributeDefault` call above the `is_defined($property->ref)` early-return fails the grafts test (and migration-wins). Placement is load-bearing and correct.
- **Falsy-`0` happy path non-vacuous ✓** — neutering the fill fails the `summary`/`priority: 0` test; `priority => 0` exercises the falsy branch (no `?:` truthiness bug).
- **Indent-only bulk confirmed ✓** — `git diff -w` shows the eight per-keyword column writes are pure re-indent under the new `if ($column !== null)` wrapper; the only real logic deltas are the call-site arg, the signature, the `return`-on-null-column → fall-through, and the necessary `$column?->description` nullsafe in the `$ref` branch.
- **Deviation (test-4 skipped) justified ✓** — grepped every `$property->default =` write in `src/`: only the migration write and this `$attributes` write exist. No cast/`@property`/timestamp/accessor path ever seeds `default`, so a non-migration source outranking `$attributes` is unconstructable. The skip is correct, not a coverage hole.
- **Bookkeeping ✓** — CHANGELOG `[Unreleased]` bullet + `docs/auto-derivation.md` step 6 both present and accurate (lowest-precedence, explicit-null, `$ref`-clean).

**Finding 1 — missing test for the column-less `$attributes` fill (the common real-world shape).** #436 returned early when `$column === null`; this PR changes that to fall through to `applyAttributeDefault`, so a model with a static `$attributes` default but **no** migration column now fills `default`. Every fixture `$attributes` key also has a `posts` migration column, so the genuinely column-less branch is never asserted (the `readMigrationColumns: false` test is on `Widget`, not the `$attributes` model). I probed it manually and it works, but it should be pinned. Add to `tests/Unit/Support/Extraction/EloquentModelToSchemaTest.php`:

```php
it('fills the $attributes default with no migration column present', function (): void {
    $schema = buildModelSchema(AttributesDefaultArticle::class, readMigrationColumns: false);
    expect(modelProperty($schema, 'summary')->default)->toBe('No summary provided.')
        ->and(modelProperty($schema, 'state')->default)->toBe('draft')               // no migration to outrank it now
        ->and(is_undefined(modelProperty($schema, 'name')->default))->toBeTrue();      // absent entry stays clean
});
```

(Optional nit, non-blocking: `applyMigrationColumn` now does slightly more than its name says; fine given the plan kept the rename out of scope and the delegate is well-named.)

Add the one test and I approve. Everything else is correct, tested, and conventional.

Radiergummi:
🤖 **[lead]** review round 1: reviewer-2 approved unconditionally; reviewer-1 approves pending 1 test-add (column-less $attributes fill — the branch this PR newly enables by falling through #436's $column===null early-return). → coder round 1/3: add the test reviewer-1 specified, re-push.

Radiergummi:
🤖 **[coder]** addressed round 1: added the column-less test gap (Finding 1); pushed 0bc87cb2.

Added your exact test verbatim — `it("fills the $attributes default with no migration column present")` — building the schema with `readMigrationColumns: false` so the `$column === null` fall-through path is exercised: `summary` → `No summary provided.`, `state` → `draft` (no migration `->default()` to outrank it when the migration is unread, reinforcing the precedence story), and `name` (absent from `$attributes`) stays default-free.

Full suite 3202 green, Pint + PHPStan L8 clean, all CI matrix jobs pass (incl. swagger-php 5.8 + Snapshot). Ready for re-review.

Radiergummi:
🤖 **[lead]** coder-1 added the column-less $attributes test (0bc87cb2, non-vacuous), CI green → back to in-review for reviewer-1's confirm. reviewer-2's approval stands (test-only delta).

Radiergummi:
🤖 **[reviewer-1]** review round 2: approved, no concerns.

Re-reviewed `0bc87cb2` on the worktree. Finding 1 resolved: the column-less `$attributes`-fill test was added exactly as specified (`it('fills the $attributes default with no migration column present')`, `readMigrationColumns: false`).

**Revert-probed the new test** — wrapping the `applyAttributeDefault` call in `if ($column !== null)` (restoring #436's column-less early-return semantics) fails *only* this new test and nothing else. So it genuinely pins the `$column === null` fall-through this PR introduced.

Full gate green on `0bc87cb2`: **3202 tests pass**, Pint clean, PHPStan L8 no errors, working tree clean. Test-only commit (no src change), as expected.

Approved.

Radiergummi:
🤖 **[lead]** reviewer-1 approved round 2 (revert-probed); reviewer-2 approved the production code (0bc87cb2 is a test-only delta). → survey gate in parallel with reviewer-2's final-commit confirm. Expected: a real $attributes default may surface on a corpus model (correct, not a regression) — 0 schema-count drops = pass.

Radiergummi:
🤖 **[reviewer-2]** re-review: approve (test-only delta, production unchanged) — verified 0bc87cb2 touches only `tests/Unit/Support/Extraction/EloquentModelToSchemaTest.php` (+11, one added test); my a6f70a28 production-code approval carries to 0bc87cb2.

Radiergummi:

Compared PR branch `0bc87cb2` against a **freshly regenerated** baseline on current `main` `6f5bc71b` (the cached baseline was stale at `c6c91c30`, 5 commits behind). Corpus pins identical between the two trees; both runs installed the matching library into every app vendor (baseline=6f5bc71b, candidate=0bc87cb2 across all 11).

| metric | result |
|--------|--------|
| apps compared | 11 / 11 |
| new crashes | 0 (all `gen_exit=0` both runs) |
| regressions | 0 |
| metric deltas | **0** — paths/operations/apiOperations/responseSchemas/requestBodies/componentSchemas all flat |

Per-app paths (baseline == candidate): AdvisingApp 523, AureusERP 111, Bagisto 406, BookStack 245, Coolify 304, InvoiceNinja 529, Koel 181, Lychee 213, Pelican 215, SpeedtestTracker 37, Vito 329.

No corpus model surfaced a new `$attributes` default — expected and correct: the Tier-0 `$attributes` read fills `default` only where cast/`@property`/attribute/migration `->default()` all left it undefined, and no corpus model declares an otherwise-uncovered static `$attributes` default. Flat counts are an explicit pass condition for this change; a gain would have been additive, not a regression. (The Coolify bootstrap-phase `functions.php` fatal appeared symmetrically in both runs as a benign pre-relink transient; Coolify generation itself succeeded with `gen_exit=0` both times.)

Clear to merge from the survey gate.

Radiergummi:
🤖 **[lead]** non-controversial + all gates green (2/2 independent approvals on final commit 0bc87cb2 incl. reviewer-1's revert-probed test confirmation, full CI matrix incl. swagger-php 5.8 + Snapshot, survey Layer A 0-regression/0-crash flat delta as expected, src diff 111 lines mostly re-indent, no config/contracts/signature change) → squash-merged, closed #439, branch + worktree removed.

PR #455, opened 2026-06-23
feat(cli): diagnose installed-but-disabled plugins

Radiergummi:
## Implementation plan — diagnose installed-but-disabled plugins

Closes #444

### Goal

When a known integration package is installed but its plugin is **absent** from the active plugin
list, emit a one-line advisory hint during `openapi:generate`. Advisory only: never auto-enables a
plugin, never mutates the host app (respects the spec-only scope boundary).

### Tier

**Tier-0.** Detection is pure reflection: `class_exists()` on a package marker class + a membership
check against the configured plugin list. No body parsing, no runtime mutation. This is the lowest
tier that captures the idiom; nothing higher is warranted.

### Resolved open questions

- **Surface location → generate-time note.** The generate command already emits diagnostics through
  the Laravel console `$this->components->*` helpers (`info` on success, `error` on failure) and the
  `--explain` flag writes per-route lines to STDERR. The hint follows this existing pattern:
  `$this->components->warn(...)` (or `->info`, matching tone — see "Surface" below), printed once per
  applicable package **before** the per-spec generation loop in `GenerateCommand::handle()`. **No new
  `openapi:doctor` command** — rejected per the issue's lean and the lead's instruction.
- **Suppressibility → not built.** The existing generate diagnostics (`components->info/error`,
  `--explain`) have **no** suppression flag or config toggle. Per "lean minimal — only add suppression
  if the existing diagnostic notes already have one", we add none. Noted as a possible follow-up
  (`--no-plugin-hints` / `openapi.plugin_hints` config) if it ever proves noisy. **Flagged for the
  lead** only as a deliberate non-decision, not controversial.

### The package → plugin map (single source of truth, tiny and explicit)

Confirmed marker classes (all loadable in the test env; verified via the project autoloader):

| Marker class (any-of)                                            | Package                          | Plugin class             |
|-----------------------------------------------------------------|----------------------------------|--------------------------|
| `League\Fractal\Manager`                                         | `league/fractal`                 | `FractalPlugin::class`   |
| `Spatie\Fractal\Fractal`                                         | `spatie/laravel-fractal`         | `FractalPlugin::class`   |
| `Spatie\QueryBuilder\QueryBuilder`                              | `spatie/laravel-query-builder`   | `QueryBuilderPlugin::class` |

Scope stays exactly to the two integrations the issue names. Fortify/SwaggerPhp are **out of scope**
for this map (Fortify documents a fixed contract table rather than recovering otherwise-lost coverage;
[...]

PR #456, opened 2026-06-23
feat(lint): operation.action-method-missing — flag routes whose action method does not exist

### LINKED ISSUE #448: feat(lint): operation.action-method-missing — flag routes whose controller action method does not exist
## Problem

Apps routinely over-register resourceful routes — `Route::apiResource('albums', AlbumController::class)`
registers all 7 actions, but the controller may implement only a few. The remaining routes point at
**methods that do not exist** and would throw at runtime, yet the generator documents them as valid
operations (a bare `200 OK` with the path params), with no signal that they are dead.

Corpus evidence (re-baseline 2026-06-23) — every one verified as a genuinely-absent method:
- **Koel: 38 ops.** `Route::apiResource('albums', AlbumController::class)` but `AlbumController` implements
  only `index`/`show`/`update` → `DELETE /api/albums/{album}` (`albums.destroy`) and the rest map to a
  missing `destroy`/`store`. Same across `albums.songs`, `artists.albums`, `podcasts`, `playlists`, etc.
- **InvoiceNinja: 3 ops.** `TaskSchedulerController::edit`, `QuickbooksController::settings`,
  `RecurringQuoteController::upload` — all "Method … does not exist".

The generator already tolerates these without crashing; the gap is that it emits them silently as if real.

## Proposal (Tier-0, reflection)

Add a lint rule **`operation.action-method-missing`**: for each operation, resolve the route's action to
`Controller@method` (or an `__invoke` for an invokable/closure-less controller) and, via reflection, flag
it when the method does not exist on the controller. Deterministic, no body parsing.

Value:
- Surfaces a real defect in the **consumer app** (a route that 404s/500s at runtime).
- Explains the empty operations these produce in coverage measurement.

## Open questions (for the thread)
- Severity: warning (advisory) seems right — it's the app's bug, not a doc-authoring error.
- Should generation additionally **skip** such routes, or keep documenting the route table faithfully and
  rely on the lint signal? Lean: keep emitting (faithful to the route table) + lint-warn; don't silently drop.
- Scope to non-closure controller actions; closures/invokables resolve differently.

Found while root-causing the `reflection-failed` bucket in the #413 re-baseline (41 ops were all missing
methods, not classifier noise).

Radiergummi:
## Fast-path scope — `operation.action-method-missing` lint rule

Closes #448

Adds a lint rule that flags operations whose route action resolves to a controller
method that **does not exist** (via reflection). Deterministic Tier-0, no body parsing.

### Resolved open questions (author leans adopted)
- **Severity:** `warning` (advisory) — it's the consumer app's bug, not a doc-authoring error.
- **Generation behaviour:** keep emitting the operation faithfully (mirror the route table);
  do **not** silently drop the route. The lint warning is the signal.
- **Scope:** non-closure controller actions only. Resolve `Controller@method` and `__invoke`
  for invokable controllers; skip closures (no method to reflect).

### Shape
- New rule `src/Lint/Rules/<...>.php` with stable ID `operation.action-method-missing`,
  default severity `warning`, implementing the operation-visitor interface.
- Resolve the action from the route/operation; reflect the controller; flag a missing method.
- Register in the baseline lint rules (`BaselineRegistration`).
- Tests: missing method → finding; existing method → none; `__invoke` controller present/absent;
  closure route → skipped (no false positive).
- `docs/linting.md` catalog row (hand-maintained, between the markers).
- `CHANGELOG.md [Unreleased]` entry (own line at end of section).

Coder implements directly against this branch (fast-path). The exact rule class name,
visitor interface, and action-resolution helper to reuse are the coder's call after reading
the existing `src/Lint/Rules/` rules and `SpecTreeBuilder`/`SpecTreeWalker`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,6 +108,7 @@ All notable changes to this project are documented here.
 - Inferred response headers (Tier-0): a `201` response gets a `Location` header (`string`, `uri-reference`), and `throttle` middleware adds `X-RateLimit-Limit` / `X-RateLimit-Remaining` (`integer`) to the success response. An authored `#[ResponseHeader]` of the same name wins. (#420)
 - `openapi:generate` emits an advisory hint on stderr when an integration package (`league/fractal`, `spatie/laravel-fractal`, `spatie/laravel-query-builder`) is installed but its plugin is not enabled, pointing at the `config/openapi.php` line that would let it infer schemas and parameters. Advisory only: never auto-enables, never pollutes the `--output=-` document. (#444)
 - Inferred array query parameters (`ids[]` from a scalar-list validation rule) now carry `style: form` and `explode: true`, matching PHP's `name[]` repeated-pair wire format and removing the ambiguous serialisation that tripped the library's own `parameter.query-array-no-explode` lint rule. Scalar query parameters and explicit `#[QueryParam]` arrays are unchanged. (#454)
+- New `operation.action-method-missing` lint rule (level 1) flags routes whose action resolves to a controller method that does not exist (e.g. an over-registered resourceful route): the operation is documented but would fault at runtime. Skips closures and routes whose controller class is itself absent. (#448)
 
 ## [0.1.0] - 2026-05-18
 
--- a/docs/linting.md
+++ b/docs/linting.md
@@ -601,6 +601,7 @@ is enabled.
 | `header.invalid-name` | 1 | Header name contains invalid characters. |
 | `link.invalid-operation` | 1 | Link references an operationId that doesn't exist in the document. |
 | `multipart.file-without-multipart` | 1 | Data class carries a file property but a #[RequestBody] override forces a non-multipart body — the spec contradicts the code. |
+| `operation.action-method-missing` | 1 | Route action resolves to a controller method that does not exist; the operation is documented but would fault at runtime. |
 | `operation.id-invalid-chars` | 1 | operationId is not a codegen-safe identifier. |
 | `operation.id-missing` | 1 | Operation has no operationId. |
 | `operation.security-missing` | 1 | Route enforces auth middleware but the operation declares no security, implying the endpoint is public. |
--- a/src/Lint/LintRouteFilter.php
+++ b/src/Lint/LintRouteFilter.php
@@ -7,6 +7,7 @@ namespace Radiergummi\OpenApi\Lint;
 use Illuminate\Container\Attributes\Scoped;
 use Illuminate\Support\Str;
 use Radiergummi\OpenApi\Routing\ActionDescriptor;
+use ReflectionClass;
 use Symfony\Component\Process\Exception\LogicException;
 use Symfony\Component\Process\Exception\ProcessSignaledException;
 use Symfony\Component\Process\Exception\ProcessStartFailedException;
@@ -18,6 +19,7 @@ use function array_filter;
 use function array_map;
 use function array_values;
 use function base_path;
+use function class_exists;
 use function config_path;
 use function explode;
 use function fnmatch;
@@ -111,7 +113,10 @@ class LintRouteFilter
     private static function isVendorOrUnresolvable(ActionDescriptor $descriptor): bool
     {
         if ($descriptor->controller === null && $descriptor->method === null) {
-            return true;
+            // The introspector nulls both reflectors when the action method does not exist on an
+            // otherwise-resolvable first-party controller. Keep those so the missing-method lint
+            // rule can flag them; still drop closures (no controller class) and vendor controllers.
+            return self::isVendorOrUnresolvableMissingMethod($descriptor);
         }
 
         $file = $descriptor->method?->getFileName()
@@ -122,6 +127,23 @@ class LintRouteFilter
         }
 
         return str_contains($file, '/vendor/');
+    }
+
+    /**
+     * Whether a both-reflectors-null descriptor is a closure/vendor route rather than a first-party
+     * controller with a missing action method.
+     */
+    private static function isVendorOrUnresolvableMissingMethod(ActionDescriptor $descriptor): bool
+    {
+        $controllerClass = $descriptor->route->getControllerClass();
+
+        if ($controllerClass === null || !class_exists($controllerClass)) {
+            return true;
+        }
+
+        $file = new ReflectionClass($controllerClass)->getFileName();
+
+        return $file === false || str_contains($file, '/vendor/');
     }
 
     /**
--- a/src/Lint/LintRouteFilter.php
+++ b/src/Lint/LintRouteFilter.php
@@ -42,6 +42,11 @@ class LintRouteFilter
     /**
      * Apply --uri glob, --path, and --diff filters. A config change widens scope to all routes.
      *
+     * Drops closures, vendor controllers, and first-party routes whose action method does not
+     * exist: the tree-walk scope mirrors the generated document, which never emits an operation
+     * for a missing-method route. {@see filterForRouteRules()} keeps the latter so the
+     * missing-method lint rule can still flag them.
+     *
      * @param list<ActionDescriptor> $descriptors
      * @param list<string>           $files       Explicit source files (`--path`), absolute or base-relative.
      * @param ?DiffScope             $diff        The `--diff` scope, or null when not requested.
@@ -56,8 +61,54 @@ class LintRouteFilter
      */
     public function filter(array $descriptors, ?string $uriGlob, array $files, ?DiffScope $diff): array
     {
-        $descriptors = $this->dropClosureAndVendorRoutes($descriptors);
+        return $this->narrow(
+            $this->dropClosureAndVendorRoutes($descriptors, keepMissingMethod: false),
+            $uriGlob,
+            $files,
+            $diff,
+        );
+    }
 
+    /**
+     * The same scope narrowing as {@see filter()}, but keeping first-party missing-method routes so
+     * the per-route lint pass can flag them. These routes are deliberately excluded from the
+     * tree-walk scope (no operation is generated for them), so they must not re-enter it.
+     *
+     * @param list<ActionDescriptor> $descriptors
+     * @param list<string>           $files
+     *
+     * @return list<ActionDescriptor>
+     *
+     * @throws LogicException
+     * @throws ProcessSignaledException
+     * @throws ProcessStartFailedException
+     * @throws ProcessTimedOutException
+     * @throws RuntimeException
+     */
+    public function filterForRouteRules(array $descriptors, ?string $uriGlob, array $files, ?DiffScope $diff): array
+    {
+        return $this->narrow(
+            $this->dropClosureAndVendorRoutes($descriptors, keepMissingMethod: true),
+            $uriGlob,
+            $files,
+            $diff,
+        );
+    }
+
+    /**
+     * @param list<ActionDescriptor> $descriptors
+     * @param list<string>           $files
+     *
+     * @return list<ActionDescriptor>
+     *
+     * @throws LogicException
+     * @throws ProcessSignaledException
+     * @throws ProcessStartFailedException
+     * @throws ProcessTimedOutException
+     * @throws RuntimeException
+     */
+    private function narrow(array $descriptors, ?string $uriGlob, array $files, ?DiffScope $diff): array
+    {
         if (is_string($uriGlob) && $uriGlob !== '') {
             $descriptors = array_values(
                 array_filter(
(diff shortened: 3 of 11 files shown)


Radiergummi:
🤖 **[lead]** fast-path (area:lint). Scaffolded branch + draft PR; handing to coder to implement directly. One reviewer pass after; new lint rule → will escalate for human gate (default-severity decision) in this supervised batch.

Radiergummi:
🤖 **[coder]** deviation: the rule needed a small change to a shared filter to work end-to-end.

`RouteIntrospector` degrades a route whose action method does not exist to a descriptor with `controller: null, method: null` (it cannot build a `ReflectionMethod`). `LintRouteFilter::isVendorOrUnresolvable()` dropped **every** both-reflectors-null descriptor as a closure/vendor/unresolvable route, so the missing-method routes never reached any lint rule. I narrowed that check: a both-null descriptor whose raw route resolves to an **existing, non-vendor** controller class is now kept (closures — no controller class — and vendor controllers are still dropped). Covered by two new `LintRouteFilterTest` cases (keeps first-party missing-method route; still drops closure). The rule reads `Controller@method` from the raw `Route` (`getControllerClass()`/`getActionMethod()`), not the nulled reflectors.

Also dropped the planned "invokable controller with no __invoke" test case: Laravel's `RouteAction::makeInvokable()` throws `UnexpectedValueException` at route construction for an invokable class lacking `__invoke`, so that route shape cannot exist at lint time. All real missing-method cases (Koel `destroy`/`store`, InvoiceNinja `edit`/`settings`/`upload`) are `Controller@method` form, which is fully covered.

Radiergummi:
🤖 **[coder]** implemented; local gates green (Pint, PHPStan level 8, full Pest suite: 3213 passed — the known `DiffConfigCommandTest` case did not surface here); pushed 49829dc7. Ready for review.

Radiergummi:
🤖 **[lead]** CI fully green. → code-review. Reviewer: scrutinize the DEVIATION modifying shared src/Lint/LintRouteFilter.php (narrowed isVendorOrUnresolvable to keep both-null descriptors) — assess blast radius on OTHER lint rules now seeing these routes, and whether a less-invasive resolution (rule resolves controller from raw route itself) is warranted. Also: Degraded severity choice (renders as 'warning'), and the dropped 'invokable missing __invoke' test (justified: Laravel throws at route construction).

Radiergummi:
🤖 **[reviewer]** review round 1 — **CHANGES REQUESTED** (1 blocking, 1 minor)

Verified by checking out the PR head (`49829dc7`) in its worktree and running the linter end-to-end, not just reasoning. Tests/Pint/PHPStan all green on the changed files; the rule itself is correct. The blocker is the deviation's blast radius.

---

### 🔴 BLOCKING — the shared-filter change makes dead routes draw *other* lint findings (spurious noise)

The deviation narrows `LintRouteFilter::isVendorOrUnresolvable()` so a both-reflectors-null descriptor for a first-party missing-method route now **passes the filter**. The coder's note frames this as "so the missing-method lint rule can flag them." But the filtered descriptor set feeds **two** consumers in `LintRunner::walkSpec()`, not just the RouteRule pass:

- `src/Lint/LintRunner.php:635-637` — when descriptors are non-empty, `$document->paths` is narrowed to `descriptorUriSet($descriptors)`. Re-admitting the dead route puts its URI back in that set, so **every tree-walk rule now processes the operation the generator emitted for the dead route.**

Empirically confirmed (kernel `openapi:lint --uri <missing> --format json`, controller with `index` but no `destroy`):

- **On `main`:** EXIT=0, `total_operations: 0`, **0 findings** — the route is dropped, never reaches any rule, isn't even in the spec.
- **On this PR:** EXIT=1, `total_operations: 1`, **2 findings** — `operation.action-method-missing` *and* `response.no-error` ("Operation DELETE /blast/missing has no error response").

So a dead route that 404s/500s at runtime now also collects `response.no-error` — and would collect any other operation/response tree rule that applies to a bare operation (`operation.id-missing`, summary/description rules, etc.). On a Koel-style app (38 missing-method routes) this multiplies the noise per dead route. Telling a user their dead route "has no error response" is misleading: the fix is to delete the route, not document it. The intent was *one* finding ("this method doesn't exist"); the realized behavior is one + a pile-on.

This is exactly the less-invasive question the plan raised. The minimal correct change is to keep the dead route **out of the tree-walk scope** while still letting the new rule see it. Concretely, the new rule needs the *raw route*, which it already reads via `$descriptor->route->getControllerClass()/getActionMethod()` — it does **not** need the descriptor to survive `dropClosureAndVendorRoutes()`. Options, least-invasive first:

1. **Run the RouteRule pass over a descriptor set that includes the dead routes, but keep the tree-walk path-narrowing on the existing (dead-routes-dropped) set.** This severs the unintended coupling without changing the shared filter's tree-walk semantics. Requires plumbing two sets through `walkSpec`, but contains the blast radius to zero.
2. If a single set must be kept, have the introspector carry a discriminator (it *already* distinguishes missing-method at `src/Support/Routing/RouteIntrospector.php:123` from missing-class at :104) and have `inScopePathItems`/tree-walk skip dead-route operations — but that's more surface than option 1.

Either way: please show, with the same `--uri <missing>` probe, that the dead route yields **exactly one** finding (`operation.action-method-missing`) and no tree-walk findings, and add a regression test asserting that (a feature test that lints a missing-method route at default level and asserts the only finding is this rule).

### 🟡 MINOR — `routeUri` location inconsistency

The new rule's finding reports `routeUri: "blast/missing"` (no leading slash, inherited from the descriptor route URI via `FindingLocation::fromDescriptor`), while sibling tree-walk findings on the same route report `/blast/missing`. Cosmetic (scope-matching uses `ltrim`, `LintRunner.php:355`), but two findings for one route showing different URIs is confusing in output. Worth normalizing if cheap; non-blocking.

---

### ✅ Verified-correct (no change needed)

- **Severity = `Degraded` → "warning".** Confirmed real: `CliFormatter.php:60-61` (`Degraded => 'warning'`), `LoggingFindingsCollector.php:35` (`Degraded => warning`). `Degraded` semantics ("misrepresents the API / misbehaves in tooling") is the right tier for a documented-but-dead route; not `Broken` (the document still validates).
- **Dropped "invokable without `__invoke`" test.** Verified: `Route::get('x', NoInvokeController::class)` throws `UnexpectedValueException: Invalid route action` at construction. That shape cannot exist at lint time, so the test is correctly omitted; the `__invoke`-present branch is still covered by `InvokableActionController`.
- **Other RouteRule consumers are unaffected.** `VisibilityAttributeNoOp` / `HideExposeConflict` only emit when `#[Hide]`/`#[Expose]` are present; on a both-null descriptor `actionAttributes()`/`controllerAttributes()` return `[]` (guarded in `ActionDescriptor`), so no spurious visibility findings. `FindingLocation::fromDescriptor` is null-safe on both-null descriptors.
- **Rule correctness.** `Controller@method` missing → exactly one finding naming `Controller::method()`; existing method → none; `__invoke` present → none; closure → skipped; missing controller *class* → skipped (dropped by the filter's `class_exists` guard, matching the issue scope). Tests pass (26 passed across the three test files).
- **Bookkeeping.** CHANGELOG `[Unreleased]` entry present; `docs/linting.md` catalog row correctly placed within the `lint-rule-catalog` markers and alphabetically sorted; registered in `BaselineRegistration`. (New-rule default severity is a human-gate item per the run plan.)

Bottom line: rule logic is sound and well-tested; **the filter deviation leaks dead routes into the tree walk and must be scoped to the RouteRule pass only** before merge.

Radiergummi:
🤖 **[lead]** review round 1: CHANGES REQUESTED (1 blocking). The LintRouteFilter narrowing re-admits dead routes into the tree-walk path (LintRunner.php:635-637), so tree-walk rules fire on them — a dead route now emits total_operations:1 + a spurious response.no-error (exit 0→1). Fix: keep dead routes scoped to the RouteRule pass ONLY, out of tree-walk scope; add a regression test proving a dead route yields exactly ONE finding. Back to coder (round 1/3).

Radiergummi:
🤖 **[coder]** addressed round-1 blocking finding; pushed `02ad7e8f`. Ready for re-review.

**What I did (reviewer option 1 — sever the unintended coupling):**

Split `LintRouteFilter` into two scopes that share the same `--uri`/`--path`/`--diff` narrowing:
- `filter()` — drops first-party missing-method routes (tree-walk scope; mirrors the generated document, which the generator *does* still emit an operation into).
- `filterForRouteRules()` — keeps them (per-route pass scope, a superset).

`LintRunner::runWithCollector()` now computes both sets from one `collectGlobalFiltered()` discovery, and `walkSpec()` takes both: the tree walk + path-narrowing run over `$descriptors` (dead-routes-dropped), the RouteRule pass runs over `$routeRuleDescriptors`. Findings are URI-gated by the superset so the per-route finding survives, while the dead route's operation stays out of the tree walk.

One extra fix the reviewer's repro surfaced: the tree-walk path narrowing was guarded by `$descriptors !== []`, so a `--uri` matching **only** a dead route (empty tree-walk set) skipped narrowing and walked the *whole* document. The narrowing now applies to an empty scoped set too (nothing left to walk), with the unscoped path unchanged.

**Probe (same fixture: controller with `index`, route registered for missing `destroy`; `--uri <missing>` at level 1, no `--only`):**

```
pre-fix:  exit=1  total_operations=2  findings=[response.no-error, operation.action-method-missing]
post-fix: exit=1  total_operations=1  findings=[operation.action-method-missing]
```

**Regression test:** `tests/Feature/Lint/ActionMethodMissingTest.php` → "does not draw tree-walk findings onto the dead route" lints the missing route at the default level (no `--only`) and asserts the rule-id list is exactly `[operation.action-method-missing]`. Verified it **fails** on the pre-fix commit (`[response.no-error, operation.action-method-missing]`) and passes here. Also updated the two `LintRouteFilterTest` missing-method cases to assert the new split (`filter()` drops, `filterForRouteRules()` keeps).

**MINOR (leading-slash `routeUri`) — skipped, with reason:** the inconsistency lives in `FindingLocation::fromDescriptor` (raw `route->uri()`, no slash) vs `fromOperation` (`pathUri`, with slash), and is shared by *all* RouteRule findings, not just this one. Normalizing `fromDescriptor` globally would flip existing assertions both ways (e.g. `LintCommandTest` asserts the no-slash form at :314/:321 and the slash form at :404/:412), so it is not a trivial, in-scope change. Left as-is.

Gate: `composer check` green locally (Pint clean, PHPStan level 8 no errors, 3214 tests passed).

Radiergummi:
🤖 **[lead]** round-1 blocking finding addressed (02ad7e8f): LintRouteFilter split into filter()/filterForRouteRules() so dead routes feed only the RouteRule pass, not the tree walk; regression test added (fails pre-fix); latent --uri narrowing hole also fixed. → round-2 re-review. (NB: this PR adds a new lint-rule default severity, a human-gate item — after re-review confirms the fix, lead escalates to needs-human for severity sign-off, not auto-merge.)

Radiergummi:
🤖 **[reviewer]** review round 2 — **APPROVED** — round-1 blocking finding genuinely resolved; no regressions; latent narrowing hole closed and tested.

Re-reviewed `02ad7e8f` against the tree (worktree), ran the linter end-to-end, and proved the regression test non-vacuous — not reason-only.

### ✅ Blocking finding (dead routes leaking into the tree walk) — resolved

The split is exactly reviewer option 1: `LintRouteFilter::filter()` (tree-walk scope, drops first-party missing-method routes) vs `filterForRouteRules()` (per-route scope, keeps them), both sharing one `narrow()`. `LintRunner::runWithCollector()` computes both from a single `collectGlobalFiltered()` discovery; the tree walk + path-narrowing run over the dead-routes-dropped set (`$treeWalkUris`), the RouteRule pass over the superset, and findings are URI-gated by the superset so the per-route finding survives while the dead route's operation stays out of the tree walk. The coupling at `LintRunner.php:635-637` (old) is severed.

**Non-vacuousness proven, not assumed:** I reverted `src/Lint/LintRunner.php` + `src/Lint/LintRouteFilter.php` to the pre-fix commit `49829dc7` (keeping the new test) and ran `tests/Feature/Lint/ActionMethodMissingTest.php` → "does not draw tree-walk findings onto the dead route". It **failed** with exactly:
```
- 0 => operation.action-method-missing
+ 0 => response.no-error
+ 1 => operation.action-method-missing
```
Restoring the fix → passes. So the test fails for the exact reason the round-1 blocker described, and the diff under review is what fixes it.

### ✅ Latent `--uri`-narrowing hole — fixed and covered

The old `$descriptors !== []` guard skipped tree-walk narrowing when the scoped set was empty (a `--uri` matching only a dead route walked the *whole* document). Now narrowing runs whenever `$treeWalkUris !== null`, even for an empty set. Verified `inScopePathItems($document, [])` returns `[]` (every `isset` lookup misses → zero paths walked), so a scoped-to-only-dead-route run narrows to nothing. The new feature test exercises precisely this path (its `--uri` matches only the missing route).

### ✅ Live routes unaffected

Traced all three scope paths: unscoped narrows identically to pre-fix (`elseif $descriptors !== []`); scoped-to-live narrows the tree walk to live routes via `$treeWalkUris`; scoped-to-dead narrows to nothing while the RouteRule pass still emits the one finding. Full Pest suite **3214 passed**; Pint clean; PHPStan level 8 no errors.

### ✅ Tests + conventions

`LintRouteFilterTest` split cases correctly assert `filter()` drops / `filterForRouteRules()` keeps (missing-method) and both drop (closure). New docblocks accurate; `@throws` PHPStan-validated; new method signatures match the existing in-file style (the over-100-char lines are soft-limit and consistent with the pre-existing `filter()` signature). CHANGELOG `[Unreleased]`, `docs/linting.md` catalog row, and `BaselineRegistration` registration all intact.

### ✅ Skipped leading-slash `routeUri` — reasonable

Confirmed it lives in shared `FindingLocation::fromDescriptor` (raw `route->uri()`, no slash) vs `fromOperation` (`pathUri`, with slash), shared by all RouteRule findings; normalizing globally would flip existing `LintCommandTest` assertions both ways. Correctly deferred, not in scope here.

No new findings. Approved. (NB: new-rule default-severity sign-off is the human gate, per the run plan — not auto-merge.)

Radiergummi:
🤖 **[lead]** Technically ready, green, two-pass clean → **escalating for the new-lint-rule default-severity decision** (a human-gate item per the run plan; auto-merge withheld).

**What's ready:**
- Round-1 blocking finding (dead routes leaking into the tree walk) fixed via the `filter()`/`filterForRouteRules()` split; non-vacuous regression test (proven to fail pre-fix). Two independent reviews: round-1 (rule correctness) + round-2 (fix correctness) both APPROVED.
- Rebased onto current `main` (`c62d3e1d`, includes #455); CHANGELOG keeps both entries. **CI fully green** across the whole matrix (PHP 8.4/8.5 × Laravel 12/13, swagger-php 5.8, Snapshot, quality, Coverage).

**Decision you own:** the new `operation.action-method-missing` rule ships at default severity **`Degraded` ("warning")** — rationale: a documented-but-dead route misrepresents the API/misbehaves in tooling, but the document still validates, so not `Broken`. Please confirm `Degraded` is the right default tier (vs `Broken`/`Notice`) before merge. PR left ready and unmerged, worktree in place. Squash-merge when you're happy with the severity.

Radiergummi:
🤖 **[lead]** rebased onto current `main` (resolved the expected CHANGELOG conflict by keeping both entries) and re-gated locally green (`composer check`: PHPStan no errors, full Pest suite passed). Force-pushed; CI re-running. Still parked at needs-human for your decision — this only refreshes the branch so it's merge-ready when you rule.

PR #458, opened 2026-06-26
feat(responses): read API Resource toArray() returning a $data variable, and recognize $this->unless()

### LINKED ISSUE #446: feat(responses): read API Resource toArray() that returns a $data variable (assigned an array literal), and recognize $this->unless()
## Problem

`ResourceToArrayReader` only reads a **single inline** `return [...]` literal (`SingleReturnArrayLiteralFinder`
requires `$return->expr instanceof Array_`). The very common idiom of building the array in a variable and
returning it is refused → the resource emits an **empty component** (`{type: object, properties: {}}`),
which is non-substantive.

Corpus evidence (re-baseline 2026-06-23): Koel `SongResource`, `SongFileResource`, `PodcastResource`,
`LiveEventResource` and Bagisto resources all resolve to empty components for this reason. `SongResource`
backs many song/episode endpoints, so the route impact is well beyond the component count.

```php
// Koel App\Http\Resources\SongResource::toArray() — refused today
public function toArray(Request $request): array
{
    $data = [
        'type' => Str::plural($this->song->type->value),
        'id' => $this->song->id,
        'title' => $this->song->title,
        'lyrics' => $this->unless($embedding, $this->song->lyrics),
        // …
    ];

    if ($this->song->isEpisode()) {
        $data += [ 'episode_description' => …, /* … */ ];
    }

    return $data;            // <- variable return, not a literal → whole resource refused
}
```

## Proposal (Tier-1, never-wrong; bounded)

1. **Resolve a returned variable to its single unconditional array-literal assignment.** When the method's
   one top-level `return $var;` has exactly one unconditional `$var = [ … ];` assignment, read that literal.
   This mirrors the existing `resolveVarExpr` already used for field *values*, lifted to the top-level
   return. Refuse (as today) when the variable is assigned more than once or conditionally.
2. **Recognize `$this->unless($cond, $value)` as an optional-field wrapper** — the inverse of the already-
   handled `$this->when(...)`. `resolveConditionalWrapper` keys on `str_starts_with($name, 'when')`; add
   `unless` so the field resolves to its value argument and is marked optional.

## Out of scope / composes with other gaps
- The conditional `$data += [ … ]` merge block stays unread by (1) (the base `$data` literal is read; the
  `+=` additions are optional fields — a possible follow-on).
- `SongResource` fields read `$this->song->field` where `song` is a typed wrapper property pointing at an
  Eloquent **model** — the leaf needs resolution via the model's metadata (casts/columns), a sibling of
  #411 (which handles typed *value-object* property chains, not model-magic leaves). So fixing this issue
  makes the *structure* readable; full field typing for model-backed chains is the related follow-on.

## Acceptance
- A resource whose `toArray()` is `$data = [ literal ]; return $data;` emits the same schema as the inline
  form. `$this->unless(...)` fields appear, marked optional. No regression on inline-literal resources.

Found while root-causing the genuinely-missing `resource::make`/`resource::collection` buckets (#413
re-baseline). Relates to #411.

Radiergummi:
## Plan — read API Resource `toArray()` that returns a `$data` variable, and recognize `$this->unless()`

Closes #446

### Tier rationale — Tier 1 (bounded AST whitelist), never-wrong

Both changes are strictly additive reads of a bounded, well-known shape with the **same
single-unconditional-assignment guard already proven** in the sibling
`ReturnExpressionResourceReader::expressionAssignedTo()`. No variable tracking across calls, no
dataflow — when the bounded shape does not hold, both paths degrade to today's behaviour (return
`null` / unconstrained). This is the lowest tier that captures the `$data = [...]; return $data;`
idiom; it is not an attribute's job because the structure *is* expressible from a single read.

### Design

Two independent changes.

**(1) Resolve a returned variable to its single unconditional array-literal assignment.**

The single gate for "is this a readable literal" is `SingleReturnArrayLiteralFinder::find()` — it
returns the `Array_` only when `$return->expr instanceof Array_`. Extend it: when `$return->expr`
is a `PhpParser\Node\Expr\Variable`, resolve through the variable's single unconditional `Assign`
whose RHS is an `Array_`.

Placement decision: this lives in the **shared finder** (`Support/MethodBody/`), not inside
`ResourceToArrayReader`, because the finder already encapsulates the scanner and is the one place
that decides "literal or not". Re-implementing the scan in `ResourceToArrayReader` would duplicate
that. The finder's own docblock already names both consumers ("API Resource `toArray()` **and
Fractal `transform()`**"), so Fractal `transform()` gains the same `$data = [...]; return $data;`
support for free — a consistent, additive win, not a regression. **Controversy flag below.** (The
issue text says "in `ResourceToArrayReader`"; I'm placing it one layer down at the shared gate and
will record that deviation as a PR comment per the workflow.)

Mechanics (mirror `expressionAssignedTo`): the finder uses a `StatementNodeFinder`, **instantiated
internally in its constructor body** (`$this->statementNodeFinder = new StatementNodeFinder();`) —
exactly as `ReturnExpressionResourceReader` does, which news it up rather than injecting it. The
public constructor signature stays `(MethodBodyScanner $scanner)`, so **no construction site
changes** (production autowiring and all seven manual `new SingleReturnArrayLiteralFinder(new
MethodBodyScanner())` calls across the test suite keep compiling), and there is no speculative DI.
For a `Variable` return with a string name, find assignments to that variable name:
- exactly one assignment under `IncludeConditionalContexts` **and** exactly one under
  `SkipConditionalContexts` (i.e. assigned once, unconditionally) — else `null`;
- the unconditional assignment's RHS must be an `Array_` — else `null`.

Refuse (return `null`, unchanged behaviour) for: dynamically-named variable, zero or ≥2
assignments, a conditional assignment, an RHS that is not an array literal (call/merge/spread/
ternary), or a returned variable that is a parameter (never assigned). Note `$data += [...]` is an
`AssignOp\Plus`, not an `Assign`, so a base `$data = [literal]` followed by a conditional `$data +=
[...]` still resolves the base literal (the `+=` additions stay unread — issue out-of-scope), and
remains never-wrong (a subset of genuinely-present, conditionally-added fields).

**(2) Recognize `$this->unless($cond, $value)` as the inverse of `$this->when()`.**

In `ResourceToArrayReader`, `resolveConditionalWrapper()` keys on
`str_starts_with($methodName, self::WHEN)`. `unless` does not start with `when`, so add a
`private const string UNLESS = 'unless';` and a small predicate
`isConditionalWrapperName(string $name): bool => str_starts_with($name, self::WHEN) || $name === self::UNLESS`.
Use it (a) in the `resolveConditionalWrapper()` guard so `unless` enters, with a branch
`($methodName === self::WHEN || $methodName === self::UNLESS) && isset($arguments[1])` resolving
the value argument (index 1) as **optional** — `when($cond,$value,$default)` and
`unless($cond,$value,$default)` both carry the value at index 1; and (b) in
`resolveNestedResource()` where the first-argument `$conditional` check currently only matches
`when*`, so `new X($this->unless($c,$y))` marks the nested field optional too, consistent with
`when`.

### Files

- **Modify** `src/Support/MethodBody/SingleReturnArrayLiteralFinder.php` — instantiate
  `StatementNodeFinder` internally (constructor body, signature unchanged); when the single
  top-level return's expr is a `Variable`, resolve it to its single unconditional array-literal
  assignment (new private helper, e.g. `arrayLiteralAssignedTo(Variable, list<Stmt>): ?Array_`).
  Literal returns unchanged.
- **Modify** `src/Plugins/ApiResources/Support/ResourceToArrayReader.php` — add `UNLESS` const +
  `isConditionalWrapperName()`; wire `unless` into `resolveConditionalWrapper()` and
  `resolveNestedResource()`. Update the two docblocks that enumerate `when()/whenLoaded()/
  whenCounted()` to include `unless()`.
- **Modify** `tests/Fixtures/ReturnLiteralFixture.php` — keep `variableReturn` (now a *positive*
  case) and add refuse fixtures (`conditionalVariableAssignment`, `variableAssignedTwice`,
  `variableAssignedNonArray`).
- **Add** a fixture resource (e.g. `tests/Fixtures/Resources/VariableReturnResource.php` and a
  matching `UnlessFieldResource.php`, or fold `unless` into an existing conditional fixture) +
  controller actions in `ResourceToArrayInferenceTest.php`.
- **Modify** tests below.
- **Modify** `docs/auto-derivation.md` line 16 + `CHANGELOG.md`.

### Tests (TDD — write/flip first, watch them fail, then implement)

`tests/Unit/Support/MethodBody/SingleReturnArrayLiteralFinderTest.php`:
- **Flip** `it('refuses a return of a variable')` → `it('resolves a return of a variable assigned a single array literal')`, expecting an `Array_` with the right item count (uses existing `variableReturn` fixture).
- **Add** refuses-when: variable assigned inside a conditional (`if`) branch; variable assigned twice (even if both arrays); variable assigned a non-array RHS (`array_merge(...)` / a call); returned variable that is an unassigned parameter. Each → `toBeNull()`.
- Confirm no regression: the existing literal/closure/early-return/merged/limit/no-return cases still pass. (The `returnLiteralFinder()` helper and the other six manual finder constructions across the suite are untouched — the constructor signature does not change.)

`tests/Feature/Plugins/ApiResources/ResourceToArrayInferenceTest.php`:
- **Add** a resource fixture whose `toArray()` is `$data = [ <literal incl. model + nested fields> ]; return $data;` → asserts the **same** schema (same properties, types, required set) as the equivalent inline-literal fixture (`LiteralOnlyResource`/`InferredArticleResource` shape). Proves parity.
- **Add** an `$this->unless($cond, $this->resource->field)` field → present in `properties`, marked **optional** (not in `required`). Pair with a `$this->when(...)` field in the same fixture to assert symmetry.
- **Add** (consistency) a nested resource wrapped in `$this->unless(...)` → optional nested `$ref` field.
- **Add** a negative: a resource returning a *conditionally*-assigned `$data` (assigned in two `if` branches) still falls back to the wrapped-model schema (today's refuse behaviour) — guards against over-reading.

### Docs + CHANGELOG

- `docs/auto-derivation.md` line 16: extend "single-`return [...]` `toArray()` literal" → "...literal **(or a variable assigned one such literal once, unconditionally)**", and "`when*` wrappers optional" → "`when*`/`unless` wrappers optional".
- `CHANGELOG.md` `[Unreleased] > ### Added`: one bullet —
  "API Resource `toArray()` (and Fractal `transform()`) whose body assigns the array to a variable and returns it (`$data = [...]; return $data;`) is now read like the inline-literal form; `$this->unless(…)` is recognized as the optional-field inverse of `$this->when(…)`. (#446)"

No new lint rule. No new config key. No new public/`Contracts` surface.

### Controversy flags (human gate)

- **Shared-finder placement touches Fractal.** Extending `SingleReturnArrayLiteralFinder` (shared
  `Support/`) means Fractal `transform()` also gains variable-return support — additive and
  same-guard, but it is a behaviour change beyond `area:responses`/ApiResources. The alternative
  (resolve inside `ResourceToArrayReader` only) duplicates the scan and leaves Fractal inconsistent.
  Recommend the shared placement; flag for reviewer confirmation. **A surveyor run (Layer A across
  the 11 apps) is warranted** since this widens reads on two plugins (Koel/Bagisto are the cited
  beneficiaries — expect *added* paths/schemas, zero removals).
- **Existing unit test flips** (`refuses a return of a variable` → resolves). Intentional and
  called out so the reviewer does not read it as an accidental regression.
- ~~Constructor-signature change~~ — **dropped** (plan-review amendment). `StatementNodeFinder` is
  newed internally in the constructor body, mirroring `ReturnExpressionResourceReader`; the public
  signature is unchanged, so no construction site (production or the seven manual test calls) is
  affected, and no `@internal`/public-surface impact remains.

Not controversial: no `src/Contracts/**`, no stage-order change, no config key, no public signature.


Radiergummi:

Tier-1, never-wrong (same single-unconditional-assignment guard as `ReturnExpressionResourceReader::expressionAssignedTo`). Two additive reads: (1) variable-return resolution placed in the **shared** `SingleReturnArrayLiteralFinder` (so Fractal `transform()` benefits too — flagged for confirmation), (2) `$this->unless()` as the optional-field inverse of `$this->when()` in `ResourceToArrayReader`. One existing unit test intentionally **flips** (refuses-variable → resolves). Survey (Layer A) warranted: expect added paths/schemas on Koel/Bagisto, zero removals.

Radiergummi:
🤖 **[lead]** planning done → plan-review. Confirm: (1) shared SingleReturnArrayLiteralFinder placement (widens variable-return to Fractal transform() too — additive, same guard) is the right call vs. duplicating in ResourceToArrayReader; (2) the intentional test-flip + @internal ctor-signature change are acceptable. Survey will run (generation-affecting).

Radiergummi:
🤖 **[reviewer]** plan-review: **approved** (one amendment applied to the plan body).

Verified against the code:
- **Tier-1, never-wrong, lowest viable.** Both reads are bounded single-unconditional-assignment with no dataflow; the variable resolution faithfully mirrors `ReturnExpressionResourceReader::expressionAssignedTo()` (`StatementNodeFinder::findAll` under both `Include`/`SkipConditionalContexts`, exactly-one each). Refuse set is complete; `$data += [...]` is `AssignOp\Plus` (not `Assign`) so the Koel base-literal still resolves and stays a never-wrong subset.
- **Shared-finder placement is correct.** `SingleReturnArrayLiteralFinder` lives in `Support/MethodBody/` and already serves both `ResourceToArrayReader` and Fractal's `TransformerTransformReader`; extending it there (vs. duplicating the scan in `ResourceToArrayReader`) is additive, same-guard, and keeps `Support/` plugin-agnostic (`StatementNodeFinder` is also `Support/`). Fractal's variable-return win is consistent, not a regression. Surveyor Layer-A run on confirm is the right gate.
- **`$this->unless()` mirrors `$this->when()`.** Value at index 1, marked optional, inverse semantics; wired into both `resolveConditionalWrapper()` and `resolveNestedResource()`. Correct.
- **Proving tests / docs / CHANGELOG all named**, including the intentional flip of `it('refuses a return of a variable')` and the conditional-assignment negative.

**Amendment I made to the plan (controversy flag #3 dropped):** the plan proposed adding a `StatementNodeFinder` *constructor parameter* to `SingleReturnArrayLiteralFinder` and claimed "only the test helper constructs it manually." That is inaccurate — there are **seven** manual `new SingleReturnArrayLiteralFinder(new MethodBodyScanner())` sites (the unit helper, `tests/Support/SchemaFromResourceFactory.php`, and five Fractal tests), so a signature change would leave the suite uncompilable until all seven are touched. The precedent the plan cites, `ReturnExpressionResourceReader`, does **not** inject `StatementNodeFinder` — it news it internally in the constructor body. The plan now does the same: signature stays `(MethodBodyScanner $scanner)`, `StatementNodeFinder` is instantiated internally. This keeps the change surgical, touches zero construction sites, and matches the "no speculative DI" convention.

Radiergummi:
🤖 **[lead]** plan-review approved (incl. reviewer's fix: instantiate StatementNodeFinder internally, no ctor-signature change — keeps all 7 construction sites compiling). → coding. Implement per the amended plan via TDD.

Radiergummi:
🤖 **[coder]** deviation / notes (shared-finder placement, as planned):

- Implemented the variable-return read in the shared `SingleReturnArrayLiteralFinder` (constructor signature unchanged; `StatementNodeFinder` newed internally), so **Fractal `transform()` gains the same `$data = […]; return $data;` support** — the additive win the plan flagged. Added a positive Fractal reader test (`VariableReturnTransformer`) pinning it.
- Two pre-existing "dynamic/refused" fixtures happened to use the *now-readable* variable-return shape and had to be re-pointed to genuinely-dynamic bodies so their refusal tests keep their intent: `DynamicUnmappedResource` (ApiResources) and `DynamicBodyTransformer` (Fractal) now assign a non-literal (`array_merge(…)`) instead of `$x = [literal]; return $x;`. No behaviour assertion was weakened; both still degrade as before.
- The existing unit test `it("refuses a return of a variable")` was intentionally flipped to `resolves …` (called out in the plan).
- `$this->unless($cond, $value)` wired into both `resolveConditionalWrapper()` (value at arg index 1, optional) and `resolveNestedResource()` (optional nested `$ref`).

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed deda87be. Ready for review.

Radiergummi:
🤖 **[lead]** coder done, remote CI fully green on deda87be (8.4/8.5 × L12/13, swagger-php 5.8, snapshot, quality, coverage). → review. Dispatching two independent code reviewers + a Layer A survey (generation-affecting; widens reads on ApiResources + Fractal).

Radiergummi:
🤖 **[reviewer]** review round 1 — 4 findings (none blocking-critical; **F1** worth addressing before merge)

Verified locally on head `deda87be`: affected tests green (56 passed), Pint clean, PHPStan (changed files) clean. Confirmed Laravel core `ConditionallyLoadsAttributes::unless($condition,$value,$default=new MissingValue)` exists (value at index 1) — the inverse-of-`when` premise and the exact-match `unless` gate are correct and safe. swagger-php 5.8 boundary: no new shapes — variable-return yields the *same* `Array_` an inline literal would, and `unless` routes through the identical `resolveValue(..., optional:true)` path as `when` (optional = omitted from `required`), so no raw nested-array schema is newly emitted. Good.

**F1 (medium, design / regression) — unconditional augmentation after the base literal silently under-documents and, for `@mixin`ed ApiResources, regresses vs. the model fallback.**
`SingleReturnArrayLiteralFinder::arrayLiteralAssignedTo()` (`src/Support/MethodBody/SingleReturnArrayLiteralFinder.php:87-119`) counts only `Assign`-to-`Variable`. It ignores `$data['k'] = v` (`ArrayDimFetch` target) **and** an unconditional `$data += [...]`. So:
```php
$data = ['id' => $this->id];
$data['title'] = $this->title;   // always present, dropped
return $data;
```
now resolves to `{id}` only. Because `read()` takes the literal as authoritative and does **not** merge the model (proven by the `ConditionalVariableReturnResource` test: refusal → `$ref: Article`), this *suppresses* the previously-emitted full model schema. The issue's "never-wrong subset" rationale only holds for **conditional** additions (Koel's `if(...) $data += [...]`, which is correctly base-resolved). Unconditional augmentation is always-present, so this is fidelity-reducing, not subset-safe — it bends the PR's "additive / no-regression" claim. It does not emit invalid OpenAPI (under-documentation, not mis-documentation), hence medium not blocker. Recommend: refuse (return `null` → model fallback) when the **unconditional** path carries any *additional* write to the same variable (an `ArrayDimFetch`-targeted `Assign` or an `AssignOp` such as `+=`). That still preserves the Koel conditional-`+=` win, since those live inside an `if` and are not on the unconditional path.

**F2 (medium, test gap) — the headline issue scenario is untested.** The central motivation (`$data = [literal]; if(...){ $data += [...]; } return $data;` → base literal resolves, conditional `+=` keys stay unread) is asserted only in a code comment; no fixture exercises it. Every new test covers either a plain variable-return (no `+=`) or a *conditional reassignment* (refused). Add a resource (or finder) fixture with a conditional `$data += [...]` block to lock the Koel `SongResource` behavior the issue is built on.

**F3 (low, test gap) — uncovered `unless`/literal edges:** `$this->unless($cond, null)` (null value → optional unconstrained) and an empty variable literal (`$data = []; return $data;`). Minor; name them for completeness.

**F4 (low, observation) — fixture re-pointing is legitimate but removes coverage tied to F1.** Re-pointing `DynamicBodyTransformer` and `DynamicUnmappedResource` to `array_merge(...)` is *necessary* (their old variable-then-return bodies now resolve, so the "refuses/empty" assertions would otherwise flip) — not masking a bug in those tests. But it also deletes the only coverage of the `$data['x'] = ...` dynamic shape without a replacement, which is exactly the shape F1 now partial-resolves. Resolving F1 + F2 closes this.

Bookkeeping ✓: CHANGELOG `[Unreleased]`, `docs/auto-derivation.md` and `docs/plugins.md` (both ApiResources **and** Fractal sections) updated; `strict_types` on all new fixtures; no issue numbers in code comments; the `final readonly class` + ctor-body-assigned `StatementNodeFinder` mirrors the sibling `StatementNodeFinder`'s own pattern (legal on 8.4).

Radiergummi:
🤖 **[reviewer]** review round 1 — approved

Reviewed adversarially against #446, head `deda87be`. Walked the assignment guard branch-by-branch and ran all gates on the branch.

**Gates (clean checkout):** full suite **3223 passed**, `vendor/bin/pint --test` passed, PHPStan level 8 **no errors**. (An early local run showed spurious failures — the shared `vendor` autoloader's PSR-4 `$baseDir` was pinned to an unrelated worktree; a `composer dump-autoload` in this checkout cleared it. Not a PR issue.)

**Correctness of the single-unconditional-assignment guard — solid.** The two-pass `IncludeConditionalContexts` (count must be 1) vs `SkipConditionalContexts` (count must be 1) check refuses correctly on every input I tried:
- assignment inside `if`/`foreach`/`try`/`while`/closure/arrow/ternary/`match` arm → Include=1, Skip=0 → refuse;
- ≥2 assignments (incl. a closure-nested second assignment, which the Include pass catches) → refuse;
- `$data = match(...)` / non-array RHS → resolves the single Assign but RHS not `Array_` → null;
- `$data += [...]` (`AssignOp\Plus`), `$data[] =` (`ArrayDimFetch` var), list-destructuring (`Array_` var) → not matched as *the* assignment → base literal read, never-wrong subset;
- `return $param` (unassigned) → 0 assignments → null.

**`unless` semantics — correct.** Value read from arg index 1, marked `optional: true`, wired at both call sites via `isConditionalWrapperName()` (flat `resolveConditionalWrapper` + nested `resolveNestedResource`). Narrow: only exact `'unless'` added; `mergeWhen`/user methods unaffected. Tested (scalar + nested `$ref`, both absent from `required`).

**Boundary clean** — finder change adds only php-parser nodes + the already-`Support` `StatementNodeFinder`; no plugin/third-party dep enters `Support/`.

**Re-pointed fixtures verified** — `DynamicUnmappedResource`/`DynamicBodyTransformer` switched to `array_merge(...)` bodies so they still refuse (FuncCall RHS, not `Array_`); their assertions (empty `properties` + generation-log note / empty schema) are unchanged and still meaningful. No hidden regression.

**TDD / docs / changelog** — flipped unit test + 4 refuse cases + feature parity/optional/negative-fallback tests + Fractal positive test; `CHANGELOG [Unreleased]`, `docs/auto-derivation.md`, `docs/plugins.md` all updated.

Non-blocking notes (no change required):
1. Reference aliasing (`$ref = &$data; $ref = [...]`) would let the base literal be read while the runtime value differs — genuine Tier-2 dataflow, not a realistic `toArray()` idiom, and identical to the sibling `expressionAssignedTo` boundary. Acceptable.
2. Optional: a `$data = [];` (empty-array) parity case would round out coverage; behaviour already matches inline `return []`.

Approving. The shared-finder placement (Fractal also gains variable-return support) is the right call — additive, same guard, and the surveyor gate covers the widened reads.

Radiergummi:
🤖 **[lead]** review round 1 — split verdict, routing to coder. reviewer-458a **approved**; reviewer-458b found **F1 (medium, must-fix)** + F2 + F3.

**Adjudication:** F1 stands. Resolving only the base literal when the unconditional path *also* does `$data['k']=v` / `$data += [...]` drops always-present fields and — because the resolved literal is authoritative (no model merge) — **suppresses the fuller `@mixin` model fallback** the previously-refused shape emitted. That's a fidelity regression at the (@mixin × unconditional-additional-write) intersection, not merely an acceptable subset.

**Coder, address this round:**
- **F1 (must):** refuse → model fallback when the *unconditional* path has any additional write to the returned var (ArrayDimFetch `Assign` or `AssignOp` e.g. `+=`). Preserve the Koel win: *conditional* `if(...){ $data += [...] }` writes (inside a branch) must still resolve the base literal as optional-additions. Net: base-literal read stays for the conditional case; unconditional extra writes now refuse instead of under-documenting.
- **F2 (must):** add a real test for the headline Koel `SongResource` scenario (`$data=[lit]; if(...){$data+=[...];} return $data;` base-resolves) — currently only asserted in a comment.
- **F3 (low, nice-to-have):** add `$this->unless($c, null)` and empty `$data=[]; return $data;` parity cases.

After the fix both reviewers re-verify the new sha (458a's approval was for deda87be). **Heads-up:** run `composer dump-autoload` in your worktree before trusting any red run — the shared vendor PSR-4 $baseDir can be pinned to another worktree and cause spurious failures (per 458a).

Radiergummi:
🤖 **[lead]** resuming run → moving to coding for round 2. Coder: address F1 (must) + F2 (must) + F3 (nice) per the adjudication comment above.

Radiergummi:
🤖 **[coder]** addressed round 2: F1 — the variable-return finder now refuses when the *unconditional* path carries any additional write to the returned variable (an `ArrayDimFetch`-targeted `Assign` like `$data['k'] = …` or an `AssignOp` like `$data += [...]`), degrading to the wrapped-model fallback so always-present fields are never dropped; *conditional* writes inside an `if (…)` branch still leave the base literal as a never-wrong subset (Koel `SongResource` win preserved). F2 — added the headline resource-level test (`ConditionalMergeArticleResource`: `$data=[lit]; if(…){$data+=[...];} return $data;` base-resolves to `{id,title}`) plus a resource-level F1 fallback test (`UnconditionalMergeResource` → `$ref: Article`). F3 — added `$this->unless($cond, null)` (null value → optional unconstrained `{}`) and empty-literal (`$data=[]; return $data;`) parity cases. CHANGELOG + `docs/plugins.md` (ApiResources & Fractal) refined. Local gates green (Pint clean, PHPStan L8 no errors, 3229 tests pass). Pushed de71286b; watching remote CI.

Radiergummi:
🤖 **[reviewer]** review round 2 — approved (F1 + F2 + F3 all resolved); one low non-blocking note

Re-verified the new sha `de71286b` in the worktree (`composer dump-autoload` first, per the PSR-4 heads-up). I wrote round-1 F1, so I reconstructed the adversarial inputs against the actual scan semantics.

**Gates (worktree):** finder unit suite **16 passed**, `ResourceToArrayInferenceTest` + Fractal `TransformerTransformReaderTest` **36 passed (167 assertions)**; `vendor/bin/pint --test` clean; PHPStan level 8 on both changed `src/` files **no errors**.

**F1 (must) — RESOLVED.** The new `mutatesVariable()` scan (`SingleReturnArrayLiteralFinder.php:122-156`) runs under `SkipConditionalContexts`, and `StatementNodeFinder` only visits top-level `Expression`/`Return_` exprs (never descends into `if`/`foreach`/`while`/`try`), so the conditional-vs-unconditional split is exactly right. Traced each requested input:
- `$data=['a'=>1]; $data['b']=2; return $data;` → `Assign` with `ArrayDimFetch` target on the unconditional path → **refuses → `$ref` model**. ✔ (asserted: `unconditionalArrayDimWrite`, `UnconditionalMergeResource` → `$ref: Article`, resource component absent)
- `$data=['a'=>1]; $data += ['b'=>2]; return $data;` → `AssignOp\Plus` unconditional → **refuses**. ✔ (`unconditionalMergeWrite`)
- `$data=['a'=>1]; if($c){ $data += ['b'=>2]; } return $data;` → `+=` lives inside `If_`, skipped by the unconditional scan → **base resolves to `{a}`** (Koel win preserved). ✔ (`conditionalMergeAfterLiteral`, `ConditionalMergeArticleResource` → `{id,title}`, `subtitle` absent, `required:[id,title]`)
- Nested dim `$data['a']['b']=`, `$data[]=` (append) → `ArrayDimFetch`-rooted `Assign`, unwrap loop bottoms out at the `Variable` → refuse. Closure-wrapped writes / conditional reassignment / list-destructuring → caught by the existing `allAssignments`/`unconditionalAssignments` count gates (Include-pass finds the closure/conditional second assignment; destructuring is `Assign`→`Array_` so not counted, count 0 → refuse). The `@mixin`-model fallback the old refused shape emitted is restored — the fidelity regression is closed.

**F2 (must) — RESOLVED.** The headline Koel `SongResource` scenario is now a real, full-generation test (`ConditionalMergeArticleResource`, `@mixin Article`), asserting base-literal resolution with the conditionally-merged key omitted — no longer a bare comment. Paired with the `UnconditionalMergeResource` fallback test that pins the F1 fidelity guarantee end-to-end.

**F3 (nice) — RESOLVED.** `$this->unless($c, null)` (`UnlessArticleResource.draft_note` → unconstrained optional `{}`) and the empty-literal `$data=[]; return $data;` (`emptyVariableReturn` → 0-item `Array_`) parity cases are present.

**Conventions / boundaries / bookkeeping:** strict_types on all new fixtures; no issue numbers in code comments; finder stays in `Support/` depending only on php-parser + the `Support` `StatementNodeFinder` (boundary intact); CHANGELOG `[Unreleased]` and `docs/auto-derivation.md` + `docs/plugins.md` (ApiResources **and** Fractal) updated and accurate to the conditional-vs-unconditional behaviour.

**Low, non-blocking (no change required):** the refusal scan covers assignment-shaped mutations (`$data['k']=`, `$data[]=`, `$data += / .=`), but not *function-call* mutations of the returned variable (`data_set($data,…)`, `Arr::set($data,…)`, `array_push($data,…)`). Those would under-document (resolve the base literal and suppress the model fallback) the same way F1 did, but for far rarer idioms and squarely within bounded-Tier-1 graceful degradation. Worth a follow-up note at most; not a merge blocker.

APPROVE. — reviewer 458b

Radiergummi:
🤖 **[reviewer]** review round 2 — approved (reviewer 458a), F1/F2/F3 all verified resolved.

Re-verified adversarially on new HEAD `de71286b`. Ran `composer dump-autoload` first (per the $baseDir heads-up), then the gates: full suite **3229 passed**, `vendor/bin/pint --test` clean, PHPStan level 8 **no errors**.

**F1 (must) — RESOLVED.** `SingleReturnArrayLiteralFinder::arrayLiteralAssignedTo()` now runs a third pass (`mutatesVariable` under `SkipConditionalContexts`) and refuses when the *unconditional* path carries any extra write to the returned variable. I drove a PoC against the live finder to pin both directions:
- `$data=[...]; $data[]=2; return $data;` → **REFUSE** (ArrayDimFetch `Assign`); `$data["meta"]["x"]=2` (nested dim) → **REFUSE**; unconditional `$data += [...]` → **REFUSE** → all degrade to the `@mixin` model fallback. Resource-level `UnconditionalMergeResource` confirms the `$ref: Article` fallback end-to-end.
- Koel direction preserved: `if($c){ $data += [...]; }` and `if($c){ $data["x"]=2; }` both still **RESOLVE the base literal** (the `if`-block is a `Stmt\If_`, which `collectUnconditional` skips entirely since it only descends top-level `Expression`/`Return_` exprs). Confirmed via PoC + the `conditionalMergeAfterLiteral` unit test.

**F2 (must) — RESOLVED.** Real fixture `ConditionalMergeArticleResource` (`$data=[lit]; if(verbose){ $data += [...]; } return $data;`) with a feature assertion: properties `{id,title}`, `subtitle` absent, `required=[id,title]`. The headline Koel `SongResource` shape is now locked by a test, not a comment.

**F3 (nice) — RESOLVED.** `UnlessArticleResource` asserts `$this->unless($c, null)` → `draft_note` stays unconstrained (`[]`); empty-literal parity covered by `emptyVariableReturn` unit test (resolves to a 0-item `Array_`).

**Other checks:** `unless` wiring is exact-match (`isConditionalWrapperName`: any `when*` OR exactly `unless`), value at arg index 1, wired at both flat + nested call sites — no `whenX`/user-method regression. Boundary clean (finder stays in `Support/`, only php-parser + the already-`Support` `StatementNodeFinder`). Fractal parity fixture `VariableReturnTransformer` pins the shared-finder win; `DynamicBodyTransformer`/`DynamicUnmappedResource` re-pointed to `array_merge(...)` still refuse. CHANGELOG `[Unreleased]` + `docs/auto-derivation.md` + `docs/plugins.md` (ApiResources & Fractal) all updated; strict_types on every new fixture; no issue numbers in new code comments.

**Non-blocking observation (no change required):** an unconditional *by-reference* mutation via a callee (`sort($data)`, `array_push($data, …)`, `data_set($data, …)`) is **not** detected — the base literal still resolves and under-documents. This is genuine Tier-2 dataflow (requires knowing the callee mutates by-ref), out of the adjudicated F1 scope (ArrayDimFetch `Assign` / `AssignOp`), and identical to the reference-aliasing boundary 458b accepted in round 1. Acceptable.

Approving.

Radiergummi:
🤖 **[lead]** review round 2 — **both reviewers approved independently** on de71286b (458a + 458b, F1/F2/F3 all verified resolved; F1 fix confirmed by adversarial PoC in both directions). Docs updated inline + reviewer-verified → no docs sub-phase. → survey (area:responses generation-affecting).

Radiergummi:

Baseline `main` @ `7c21268a` vs candidate `de71286b`. All 11 corpus apps generate cleanly (gen_exit=0); no deterministic metric (paths/operations/apiOperations/responseSchemas/documentedResponses/requestBodies/componentSchemas) dropped on any app, and no lint rule rose.

Intended effect confirmed — response-schema enrichment from reading API Resource `toArray()`:
- Koel `responseSchemas` **42 → 56** (+14)
- `resource.fields-undeclared` lint findings **20 → 5** (no other rule changed)

Note: on the first corpus pass, InvoiceNinja/Koel/Lychee hit a transient `blocked-compat` bootstrap flake (composer path-repo link assertion) — re-running each with `--only` booted and generated cleanly (paths 529/181/213, all matching baseline). Not attributable to this change.

Clear to merge.

Radiergummi:
🤖 **[lead]** non-controversial + all gates green → squash-merged, closing #446.

- ✅ Full CI matrix green on `de71286b` (8.4/8.5 × L12/13, swagger-php 5.8, snapshot, quality, coverage).
- ✅ Two independent reviewer approvals on `de71286b` (round 2; F1/F2/F3 all verified resolved, F1 fix confirmed by adversarial PoC in both directions).
- ✅ Survey Layer A pass — 0 regressions across 11 apps; intended lift confirmed (Koel responseSchemas 42→56, resource.fields-undeclared 20→5).
- ✅ Non-controversial: 135 `src/` lines, no `Contracts/`, no config key, no public-signature change, Support/Plugins boundary clean.

Worktree removed.

PR #459, opened 2026-06-26
feat(requests): statically read FormRequest::rules() array literal as a fallback when invoking it throws

### LINKED ISSUE #449: feat(requests): statically read FormRequest::rules() array literal as a fallback when invoking it throws
## Problem

The FormRequest request-body extractor **invokes** `rules()` (instantiate + call) to read the validation
rules. Many real `rules()` methods read runtime state at the top (`$this->all()`, `$this->headers`,
`auth()->user()->id`, `request()->…`), so invocation throws during generation and the body **degrades to
empty** — even though the rules themselves are a static array literal.

Corpus evidence (re-baseline 2026-06-23): `request-body.schema-degraded` fires **102×** across 5 apps
(InvoiceNinja 69, Pelican 11, Lychee 10, Koel 9, Bagisto 3). Representative messages: "Attempt to read
property \"id\" on null", "Typed property Request::\$headers must not be accessed before initialization".

The rules are usually right there as a literal:

```php
// InvoiceNinja BulkActionQuoteRequest::rules() — invocation throws on $this->all(), but the literal is static
public function rules()
{
    $input = $this->all();                          // <- throws during generation
    $rules = [
        'action' => 'sometimes|in:template,convert_to_invoice,…',
        'ids' => 'required|array',
        'template' => 'sometimes|string',
        'send_email' => 'sometimes|bool',
        'email_type' => 'sometimes|in:quote,reminder1,…',
    ];
    if (in_array($input['action'], ['convert','convert_to_invoice'])) {
        $rules['action'] = [new ConvertableQuoteRule()];   // conditional tweak — optional
    }
    return $rules;
}
```

When invoking `rules()` throws, **fall back to a static read of the method body's array literal** — the same
machinery already used for controller inline `validate([...])` and a controller `$rules` property / `rules()`
method (per the auto-derivation docs), applied to the FormRequest class. Read the `$rules = [ … ]; return
$rules;` (and bare `return [ … ];`) literal through `ValidationRulesToSchema`; ignore conditional
`$rules[...] = …` tweaks (the base literal entries are correct, never-wrong; conditional overrides are an
optional follow-on). Keep invocation as the primary path (it resolves dynamic rules) — static read is the
graceful fallback that replaces today's empty degradation.

## Acceptance
- A FormRequest whose `rules()` throws on invocation but assigns `$rules = [ literal ]` (or `return [ … ]`)
  emits a request body from the literal. `request-body.schema-degraded` stops firing for those.
- No change for FormRequests whose `rules()` invokes cleanly (primary path unchanged).
- A genuinely dynamic `rules()` with no readable literal still degrades (now with a logged note).

Found while scanning corpus lint output in the #413 re-baseline (request-body.schema-degraded, 102 ops).

Radiergummi:
## Plan — static `FormRequest::rules()` literal fallback

Closes #449.

### Goal

When invoking a `FormRequest`'s `rules()` throws during generation (runtime-state reads at the
top: `$this->all()`, `$this->headers`, `auth()->user()->id`), fall back to a **static AST read**
of the method body's array literal instead of degrading to an empty body. Invocation stays the
**primary** path; the static read is a graceful fallback that recovers the base literal.

### Tier — Tier 1 (bounded AST whitelist)

Parse only the `rules()` body, match a small whitelist of shapes in the first N top-level
statements, no cross-call / interprocedural tracking:

- bare `return [ <literal map> ];`
- `$rules = [ <literal map> ]; … return $rules;` — a **single same-method** assignment of the
  returned variable to an array literal (not dataflow: the assignment and the return live in the
  same body, matched positionally).

Conditional `$rules['x'] = …` tweaks are **ignored** (the base literal entries are never-wrong;
overrides are an optional follow-on). Anything dynamic degrades gracefully **and logs** — exactly
the Tier-1 contract in `CLAUDE.md`. This is the same machinery already used for controller inline
`validate([...])` (`InlineValidatorRulesReader`), reused here. **No escalation.**

### Files

**New — `src/Plugins/Core/Support/FormRequestStaticRulesReader.php`** (`@internal`, `#[Scoped]`,
`final readonly`)
- Entry: `read(ReflectionMethod $rulesMethod): ?array` → returns `array<string, array<int,
  mixed>|string>` (the same raw rule shape `rules()` returns, which `ValidationRulesToSchema`
  already consumes) or `null` when no readable literal.
- Uses `Support\MethodBody\MethodBodyScanner::firstStatements($method, LIMIT)`.
- **Return resolution** — reuse the **already-merged** `SingleReturnArrayLiteralFinder`
  (it lives on `main` today, used by Fractal + ApiResources — the "#458 lands it concurrently"
  note below was stale):
  1. Call `SingleReturnArrayLiteralFinder::find($rulesMethod)` first. It already encapsulates the
     single-top-level-return + not-nested-in-a-conditional guard and the bare `return [ literal ];`
     case. Non-null ⇒ that `Array_`, done.
  2. Only when `find()` returns `null`, run the **genuinely-new** variable-return scan over
     `firstStatements`: require exactly one top-level `return Variable $v` (reuse the same
     single-return / no-conditional guard shape — note `find()` returns null for a `return $v`
     because the expr isn't an `Array_`, so this branch is reachable for the variable case);
     find the **first** top-level `$v = Array_` assignment among the preceding statements; ignore
     later `ArrayDimFetch` assignments to `$v` (the conditional tweaks). A non-array or absent base
     assignment ⇒ `null`.
- **Array_ → rules mapping**: own a small outer loop over `$node->items` (skip spread/unkeyed,
  evaluate each literal key via `AstLiteralEvaluator`) and delegate **each field's value** to the
  shared per-field mapper below. FormRequest's policy is **recover-what-you-can**: skip an
  anomalous entry, keep the literal rest; empty result ⇒ `null`.

**New — `src/Support/MethodBody/RuleFieldLiteralMapper.php`** (`@internal`, `final readonly`)
- Extracts **only** the per-field tolerance — `InlineValidatorRulesReader::fieldRulesOf`: literal
  strings pass through, string-list constants pass through, array literals keep their literal
  elements and drop dynamic ones (`new Rule()`, `Rule::unique()`), fully-dynamic ⇒ `null`. This is
  the subtle, genuinely-shared bit; the **only** part worth not duplicating.
- `map(Expr $fieldValue): null|array<int,mixed>|string`. Pure, depends on nothing but
  `AstLiteralEvaluator`. No scanner, no file, no degrade-reason VO needed.
- **Why per-field, not whole-array (corrects the original plan):** the two callers have
  **different outer-loop degrade policies**. `InlineValidatorRulesReader::rulesFromArrayLiteral` is
  **all-or-nothing** — a spread / unkeyed / dynamic-key / non-string-key entry degrades the *whole*
  result (`rules === null`). FormRequest wants **recover-what-you-can**. A whole-array
  `map(Array_)` returning "empty = no literal" therefore **cannot** preserve the inline reader's
  existing all-or-nothing behaviour, and the suite would not catch the regression (only the
  `'neither an array literal'` reason — raised in `read()`, outside the extracted method — is
  test-pinned). Extracting only `fieldRulesOf` keeps each caller's outer loop and exact degrade
  policy intact. Trailing-comment descriptions stay in each caller's outer loop (they already hold
  the `$file` + `MethodBodyScanner`).

**Modified — `src/Plugins/Core/Support/InlineValidatorRulesReader.php`**
- **Surgical**: keep `rulesFromArrayLiteral`'s outer loop and every degrade branch byte-identical;
  replace only the `fieldRulesOf($item->value)` call with `RuleFieldLiteralMapper::map(...)` and
  delete the now-duplicated private `fieldRulesOf`. **Truly behaviour-preserving** — pinned by the
  existing `InlineValidatorRulesReaderTest` (the outer loop, spread/key degrades, and skip-field
  semantics are untouched). Does **not** touch `ResourceToArrayReader`.

**Modified — `src/Plugins/Core/Support/FormRequestRulesReader.php`**
- Inject the new reader with a **constructor default** so the existing `new
  FormRequestRulesReader()` call sites in tests keep working (8 sites across 5 files:
  `FormRequestRulesReaderTest`, `CoreQueryParameterResolverTest`,
  `FormRequestRequestSchemaResolverTest`, `SchemaFromFormRequestTest`,
  `FakerSynthesisIntegrationTest`): `private FormRequestStaticRulesReader $staticReader = new
  FormRequestStaticRulesReader()`. (Legal: promoted `readonly` param with a `new ...()` default on
  PHP 8.4 — same pattern as `MethodBodyScanner`'s default `NullLogger`.)
- In the `catch (Throwable $exception)` block: attempt
  `$static = $this->staticReader->read(new ReflectionMethod($formRequestClass, 'rules'))`. If
  non-null ⇒ `FormRequestRulesResult::recovered($static)`; else ⇒ `degraded(...)` as today, with
  the reason noting the static fallback also found no readable literal (the "logged note").
- Primary path (`$instance->rules()` succeeds) is untouched.

`SchemaFromFormRequest` and the `request-body.schema-degraded` finding need **no change**: it
already logs a warning + emits the finding when `rulesResult->rules === null`. Recovery simply
means that branch no longer fires for these requests.

New fixtures under `tests/Fixtures/FormRequestStatic/` (FormRequest stubs, matching the existing
`tests/Fixtures/` stub style):

| # | Fixture / shape | Expectation |
|---|---|---|
| 1 | `rules()` reads `$this->all()` (throws), then `$rules = [ literal ]; return $rules;` | recovered from literal; **primary corpus shape** (`BulkActionQuoteRequest`) |
| 2 | `rules()` throws, bare `return [ literal ];` | recovered |
| 3 | `rules()` throws, `$rules = [...]; if (...) { $rules['x'] = [...]; } return $rules;` | recovers **base** literal; conditional tweak ignored |
| 4 | field value with a dynamic element in the base literal (`'foo' => ['required', new Rule()]`) | per-field tolerance: literal rules kept, dynamic element dropped |
| 5 | `rules()` **invokes cleanly** | primary path unchanged; static reader not consulted; same schema as today (no regression) |
| 6 | `rules()` throws, `return $this->buildRules();` / `$rules = $this->base();` (no literal) | **still degrades**: `request-body.schema-degraded` still fires + warning logged |
| 7 | `rules()` throws, two top-level returns or a return inside a conditional | degrades (don't guess) |
| 8 | `rules()` not declared | unchanged (already degrades) |

- Unit: `FormRequestStaticRulesReaderTest` (cases 1–4, 6, 7), extend `FormRequestRulesReaderTest`
  (fallback wiring: throwing-but-static-literal ⇒ recovered; throwing-and-dynamic ⇒ degraded).
- Behaviour-preserving: existing `InlineValidatorRulesReaderTest`,
  `SchemaFromFormRequestTest`, `FormRequestRequestSchemaResolverTest`,
  `CoreQueryParameterResolverTest` must stay green after the mapper extraction.
- Gates: `composer test`, `vendor/bin/pint --test`, `composer lint` (PHPStan L8).

### Docs & changelog

- `docs/auto-derivation.md` — note that `rules()` is statically read as a fallback when invocation
  throws (one sentence in the FormRequest derivation section). Possibly a line in
  `docs/troubleshooting.md` where the degrade is described.
- `CHANGELOG.md` `[Unreleased] / Added` — "FormRequest request bodies fall back to a static read
  of the `rules()` array literal when invoking `rules()` throws on runtime state. (#449)"
- No lint-rule change (`request-body.schema-degraded` is unchanged; it just fires less).

### Controversy / coordination flags

- **Not controversial for auto-merge**: no `Contracts\` change, no stage-order change, no config
  key, no public-signature change (both readers are `@internal`).
- **area:requests ⇒ generation-affecting ⇒ survey required.** Expected corpus effect: the 102×
  `request-body.schema-degraded` (InvoiceNinja 69, Pelican 11, Lychee 10, Koel 9, Bagisto 3)
  **drops**, request-body schemas **gain**; Layer-A must show **no regression** elsewhere.
- **Refactor note (reviewer-amended)**: extraction is **narrowed to the per-field
  `RuleFieldLiteralMapper`** — the one genuinely-shared, subtle bit. The original whole-array
  `RulesArrayLiteralMapper` was dropped: it could not preserve `InlineValidatorRulesReader`'s
  all-or-nothing degrade policy and the suite would not have caught the regression. The narrowed
  extraction leaves the inline reader's outer loop byte-identical (only the `fieldRulesOf` call is
  delegated), so "behaviour-preserving, pinned by tests" now holds literally. Duplicating
  `fieldRulesOf` locally was the zero-touch alternative — rejected: that 30-line per-field
  tolerance must stay in lockstep across two readers, so a divergence smell.
- **#446 overlap (coordination, not a blocker)**: the `$var = [...]; return $var;` variable-return
  scan is new here; `SingleReturnArrayLiteralFinder` (bare-return case) is **already on main**.
  #446's variable-return work and this one can dedupe into a shared `Support\MethodBody` finder in
  a follow-up. No file conflict (different files).


Radiergummi:

Tier-1 bounded AST fallback in the FormRequest rules reader: when `rules()` invocation throws, statically read the `$rules = [literal]; return $rules;` / bare `return [...];` array literal and run it through the existing `ValidationRulesToSchema`. Primary invocation path unchanged.

Flags for the reviewer/lead:
- **survey required** (area:requests, generation-affecting) — expect the 102× `request-body.schema-degraded` to drop, no regression elsewhere.
- **refactor note**: extracts `RulesArrayLiteralMapper` out of `InlineValidatorRulesReader` for honest reuse (behaviour-preserving, pinned by tests). Alt: duplicate locally for a zero-touch diff.
- **#446 coordination**: shares the `$var = […]; return $var;` variable-return primitive; kept local here, dedupe in a follow-up. No file conflict.
- Not otherwise controversial (no Contracts/stage-order/config/public-signature change).

Radiergummi:
🤖 **[lead]** planning done → plan-review. Scrutinize: (1) is the InlineValidatorRulesReader→RulesArrayLiteralMapper extraction justified, or is the noted zero-touch alternative cleaner/more surgical per CLAUDE.md? (2) confirm the new FormRequestRulesReader dep uses a new() ctor default so the 6 construction sites keep working; (3) #446/#449 share the var-return primitive but different files — dedup is an accepted follow-up, not a blocker. Survey will run.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved with two amendments applied to the plan body.**

Sound on tier (Tier-1, lowest viable — invocation stays primary, static read replaces the empty degrade), surgical scope, constructor `new ...()` default (8 call sites confirmed compile-safe), Core/Support boundary, the proving tests (incl. still-degrades-when-dynamic #6, ignored conditional `$rules[...]=` #7, clean-invocation-unchanged #5, multiple/nested-return degrade), docs, and CHANGELOG. No file conflict with #446.

Two refinements I wrote directly into the plan:

1. **Narrowed the extraction from whole-array to per-field (the key call).** The original `RulesArrayLiteralMapper::map(Array_)` returning "empty rules = no literal" is **not** behaviour-preserving: `InlineValidatorRulesReader::rulesFromArrayLiteral` is **all-or-nothing** — a spread / unkeyed / dynamic-key / non-string-key entry degrades the *whole* result, it never emits partial rules — and the suite would not catch the regression (only `'neither an array literal'`, raised in `read()` *outside* the extracted method, is test-pinned). The two callers also *want* different outer-loop policies (inline = all-or-nothing; FormRequest = recover-what-you-can). So extract only `fieldRulesOf` → `Support\MethodBody\RuleFieldLiteralMapper` (pure, `AstLiteralEvaluator` only). The inline reader's outer loop stays byte-identical (only the `fieldRulesOf` call delegates), so "behaviour-preserving, pinned by tests" now holds literally. Duplicating the 30-line per-field tolerance was the zero-touch alt — rightly rejected as a lockstep divergence smell.

2. **Reuse the already-merged `SingleReturnArrayLiteralFinder`.** It's on `main` today (used by Fractal + ApiResources); the "#458 lands it concurrently" note was stale. Call `find()` first for the bare `return [ literal ];` case; only the `$rules = [...]; return $rules;` variable-return scan is genuinely new. Shrinks the new code and largely dissolves the dedup follow-up.

No blockers remain; coder can proceed against the amended body.

Radiergummi:
🤖 **[lead]** plan-review approved (amended: narrow extraction to per-field RuleFieldLiteralMapper — behaviour-preserving; reuse already-merged SingleReturnArrayLiteralFinder::find() for bare-return, only the var-return scan is new). → coding. NB #458 concurrently extends the shared finder additively; #459 only consumes its existing find() so no conflict. Implement per amended plan via TDD.

Radiergummi:
🤖 **[coder]** implemented; local gates green (pint/phpstan/pest, 3225 passed); pushed 8efc6662. Full CI matrix green incl. PHP 8.5 + swagger-php 5.8. Ready for review.

- New `FormRequestStaticRulesReader` (Core) recovers the `rules()` array literal (bare `return [ … ];` and `$rules = [ … ]; … return $rules;`) when invocation throws; conditional `$rules[…] = …` tweaks ignored; genuinely-dynamic still degrades with a log note.
- Per-field tolerance extracted into pure `Support\MethodBody\RuleFieldLiteralMapper`; `InlineValidatorRulesReader` outer loop is behaviour-preserving (delegates the per-field map only).
- `FormRequestRulesReader` gains a `new ...()`-default static-reader dep so existing `new FormRequestRulesReader()` sites still compile.

Note: the variable-return scan here is the new code the plan called for; my branch base does **not** carry the #458 finder enhancement (still only on the concurrent sibling worktree, not `origin/main`). When #458/#446 merges first, this can dedupe into the shared finder on rebase, as the plan's overlap note anticipates.

Radiergummi:
🤖 **[lead]** coder done, full CI matrix green on 8efc6662 (8.4/8.5 × L12/13, swagger-php 5.8). → review. Dispatching two independent code reviewers + Layer A survey (generation-affecting; expect request-body.schema-degraded 102× to drop, zero regressions).

Radiergummi:
🤖 **[reviewer]** review round 1 — approved (3 low-severity nits, none blocking)

Independent second review, head `8efc6662`. Verified locally in the worktree: PHPStan L8 clean (1339 files), `pint --test` passed, `FormRequestStaticRulesReaderTest` + `FormRequestRulesReaderTest` + `InlineValidatorRulesReaderTest` (31) and the broader FormRequest/request-body/query-param/Faker feature suite (108) all green.

**Verdict: APPROVED.** The five review lenses hold up:
- **5.8 / schema correctness:** the recovered rules feed `ValidationRulesToSchema::process()` at `SchemaFromFormRequest.php:129` exactly as the invocation path does, returning the identical `array<string, string|array>` shape. No new schema shape reaches swagger-php — the 5.8 boundary is structurally untouched.
- **Never-wrong subset:** a field whose value is not a pure literal (ternary, concat-with-call, `new Rule()`) is *skipped*, never guessed (`RuleFieldLiteralMapper::map` → null → `continue`); conditional `$rules['x']=…` tweaks are ignored by design; an empty base literal degrades rather than emitting a partial body. So the fallback under-documents at worst, never disagrees.
- **Logging:** reuses the existing `logger->warning` in `SchemaFromFormRequest` with an augmented `degradeReason`; silent on the recovery path. Consistent, not noisy.
- **Refactor:** `RuleFieldLiteralMapper::map` is byte-identical to the deleted `fieldRulesOf`; inline reader outer loop untouched; orphaned `array_all`/`array_is_list` imports removed, `Expr` still used. Pinned by the 22 inline-reader cases.
- **Conventions:** strict_types, `@internal`, CHANGELOG + `docs/request-bodies.md` updated, no issue-#s in comments, ≤100 cols.

**Nits (non-blocking):**
1. `FormRequestStaticRulesReader::variableReturnLiteral()` returns the **first** top-level `$var = [literal]` assignment. In the (implausible, dead-code) case of two unconditional top-level full-literal reassignments to the same variable, it emits the dead first one rather than the effective last. Last-wins, or refusing on multiple top-level literal assignments, would be strictly never-wrong. Practically never occurs in corpus.
2. `collectReturns()` + the single-return/in-top-level guard duplicate `SingleReturnArrayLiteralFinder`'s private copy — already acknowledged in the PR body as a deferred `Support\MethodBody` dedup overlapping #446.
3. `read()` parses the body twice (`bareReturnFinder->find()` then `variableReturnLiteral()`, each calling `firstStatements()`). Off the hot path (only on invocation-throw), negligible.

Survey still required (area:requests, generation-affecting) before merge.

Radiergummi:
🤖 **[reviewer]** review round 1 — approved

Independent adversarial review of head sha `8efc6662`. Verified locally in the worktree: `composer lint` (PHPStan L8) clean, Pint clean on all changed files, new `FormRequestStaticRulesReaderTest` (7) + updated `FormRequestRulesReaderTest` (7) + `InlineValidatorRulesReaderTest` (17, behaviour-preservation pin) all green.

**Adversarial checks — all pass:**

1. **Catch-only fallback** (`FormRequestRulesReader::read` `src/Plugins/Core/Support/FormRequestRulesReader.php:45-65`): the static read fires strictly inside `catch (Throwable)`. Clean invocation returns `recovered($instance->rules())` untouched; a clean *empty/partial* result is authoritative and correctly does **not** trigger fallback. `Throwable` (not `Exception`) catches the corpus `Attempt to read property "id" on null` / uninitialized-typed-property `Error`s. `readStaticRules` wraps `new ReflectionMethod` in `catch (ReflectionException)`, and `MethodBodyScanner` swallows+logs parse failures (returns `null`), so the fallback cannot crash a generation that previously degraded.

2. **Variable-return correctness** (`variableReturnLiteral` `FormRequestStaticRulesReader.php:164-212`): single top-level `return $v` enforced via deep `collectReturns` + `count===1` + `in_array(..., strict)`; base literal read from the **first** top-level `Variable = Array_` assignment. Conditional `$rules[x]=…` tweaks are nested (not top-level) and are `ArrayDimFetch` not `Variable`, so they cannot be picked up — pinned by `ConditionalTweakFormRequest`. `MultiReturnFormRequest` (2 returns) correctly refuses.

3. **Refactor behaviour-preserving:** `fieldRulesOf` → `RuleFieldLiteralMapper::map` is a byte-identical move; the **all-or-nothing** outer-loop degrade (spread/unkeyed/dynamic-key/non-string-key) stays in `InlineValidatorRulesReader::rulesFromArrayLiteral:238-274`, distinct from the static readers recover-what-you-can `continue`. Removed `array_all`/`array_is_list` imports are unused elsewhere; `AstLiteralEvaluator`/`NonLiteralValueException` retained where still referenced. 17 inline tests green.

4. **Still-degrades:** `DynamicRulesFormRequest` (interprocedural `buildRules()`) → `null`, degrade reason appended `(no static rules() literal to fall back to)`. This degrade-reason note is the "logged note" surfaced through the existing `request-body.schema-degraded` path.

5. **Boundary/conventions:** `RuleFieldLiteralMapper` in plugin-agnostic `Support\MethodBody`; `FormRequestStaticRulesReader` in `Plugins\Core\Support` (FormRequest is a Laravel convention). `strict_types`, no abbreviations, ≤100 cols, no issue-#s in comments, `@internal` tags present. CHANGELOG `[Unreleased]` + `docs/request-bodies.md` both updated.

**Non-blocking nit (1):** the single-top-level-return guard (`collectReturns` + `count===1` + `in_array` top-level check) is duplicated between `SingleReturnArrayLiteralFinder` and `FormRequestStaticRulesReader::variableReturnLiteral`. A `findSingleTopLevelReturn(): ?Return_` helper on the finder would let the variable-return branch reuse it. Pure cleanup; not merge-blocking.

Verdict: **APPROVED.**

Radiergummi:

Baseline `main` @ `7c21268a` vs candidate `8efc6662`. All 11 corpus apps generate cleanly (gen_exit=0) on both; no deterministic metric (paths/operations/apiOperations/responseSchemas/documentedResponses/requestBodies/componentSchemas) dropped on any app.

Intended effect confirmed: `request-body.schema-degraded` lint findings dropped **102 → 47** with no other lint rule rising —

| app | base | cand |
|-----|------|------|
| Bagisto | 3 | 2 |
| InvoiceNinja | 69 | 38 |
| Koel | 9 | 0 |
| Lychee | 10 | 0 |
| Pelican | 11 | 7 |

Clear to merge.

Radiergummi:
🤖 **[lead]** all quality gates green — **escalating for human merge confirmation on diff size only.**

- ✅ Full CI matrix green (8.4/8.5 × L12/13, swagger-php 5.8, snapshot, quality, coverage) on `8efc6662`.
- ✅ Two independent reviewer approvals (round 1), no open concerns.
- ✅ Survey Layer A pass — 0 regressions across 11 apps (baseline `7c21268a` vs candidate `8efc6662`).
- ⚠️ **Sole reason for escalation:** the production diff is ~290 added / 53 deleted `src/` lines (new `FormRequestStaticRulesReader` +186, extracted `RuleFieldLiteralMapper` +71, net-reduced `InlineValidatorRulesReader`), exceeding the autonomous-merge controversy threshold (~150 `src/` lines). No `Contracts/`, no config key, no public-signature change, boundary clean.

PR is **ready, green, and dual-approved** — left unmerged with its worktree in place for a quick human merge confirmation. @radiergummi

Radiergummi:
🤖 **[lead]** rebased onto `main` after #458 merged (resolved the expected `CHANGELOG.md [Unreleased]` conflict by keeping both entries). New head `d3b03594`; local `composer check` green (3239 tests). CI re-running on the new sha. PR remains ready for human merge — no further agent action needed.

Radiergummi:
🤖 **[lead]** Merged today — clearing stale agent:needs-human label so resume-scan no longer counts it as in-flight.

Issue #453, opened 2026-06-23
feat(responses): read the OO new JsonResponse([...], status) form for parity with response()->json([...])


Radiergummi:
## Problem

`InlineJsonResponseResolver` reads a literal `response()->json([...])` / `response()->noContent()`, but
explicitly excludes the **OO form** `new JsonResponse([...], $status)` (documented "out of scope by design"
at `InlineJsonResponseResolver.php:41`). That's an inconsistency: the two forms are the same AST shape (an
array-literal first arg + an optional status), differing only in `New_` vs a `response()->json` call.

Corpus evidence (re-baseline 2026-06-23): the **Pterodactyl/Pelican lineage uses the OO form pervasively**
— 40 genuinely-missing Pelican ops are `new JsonResponse(...)` (20 POST, 5 PUT, 2 GET, 13 DELETE):

```php
// Pelican MountController — no-content form (should be 204 correctly-empty)
public function delete(DeleteMountRequest $request, Mount $mount): JsonResponse
{
    // …
    return new JsonResponse([], JsonResponse::HTTP_NO_CONTENT);
}
```

The 13 DELETEs are `new JsonResponse([], 204)` → genuinely no-content (currently neither emitted as 204 nor
classified as such); the 27 non-DELETE ops are candidates for a real body.

## Proposal (Tier-1, parity, never-wrong)

Extend the inline-response reader to recognize `new JsonResponse(<arg>, <status?>)` (and
`new \Illuminate\Http\JsonResponse(...)`) with the same handling as `response()->json(<arg>, <status?>)`:
- literal array arg → object schema at the literal/explicit status;
- empty array + a 204/no-content status → a contentless 2xx;
- a `JsonResource`/`Data`/model arg → defer to the existing resource/data resolution;
- non-literal arg → degrade as today.

Same bounded first-N-statements scan, same status resolution (literal or `JsonResponse::HTTP_*` constant).

## Note
This reverses a deliberate exclusion — filing with corpus evidence so the call can be re-made. The OO form
is idiomatic in framework-style codebases (Pterodactyl/Pelican), so the exclusion leaves a whole app class
uncovered for a shape that is no harder to read than the helper form.

## Acceptance
- `return new JsonResponse(['x' => …]);` emits the same response as `return response()->json(['x' => …]);`.
- `new JsonResponse([], 204)` is treated as a contentless 2xx (correctly-empty), not genuinely-missing.
- Non-literal args degrade with a note.

[...]

PR #464, opened 2026-06-26
fix(responses): a body-scanned response with content beats a 204 destroy convention

### LINKED ISSUE #452: bug(responses): conventional resourceful status (204 for destroy) overrides a literal response()->json([...]), emitting the wrong status + dropping the body
## Problem

For a resourceful `destroy` action the generator assigns the conventional success status **204 No Content**.
But when the controller actually returns a literal `response()->json([...])` (a 200 with a body), that
convention **overrides** the body-scanned response — the spec documents `204 No Content` for an endpoint
that really returns `200 {message: string}`. Both the **status** (204 vs 200) and the **body** are wrong.

Verified on two corpus apps (re-baseline 2026-06-23):

```php
// AureusERP IndustryController::destroy  — returns 200 {message}
public function destroy(string $id)
{
    $industry = Industry::findOrFail($id);
    Gate::authorize('delete', $industry);
    $industry->delete();
    return response()->json(['message' => 'Industry deleted successfully.']);
}
// generated spec → "204": { "description": "No Content" }   ❌ (should be 200 {message: string})
```

```php
// InvoiceNinja DocumentController::destroy
return response()->json(['message' => ctrans('texts.success')]);
// generated spec → 2xx codes: ["204"]   ❌
```

This is the dominant cause of the `response()->json([array literal])` genuinely-missing bucket (≈20 corpus
ops, mostly `destroy`): the literal body is correctly recognized but discarded in favour of the convention.

## Expected

A statically-read explicit response in the body (a literal `response()->json([...])` with its status) is
**stronger evidence than a status convention** and should win. The conventional 204 (or 200) should apply
only as the fallback when the body carries no readable explicit response.

## Fix direction

Make the resourceful-success-status convention yield to a body-scanned primary response: when the body-scan
resolves a literal `response()->json([...])` (status + body), use its status and schema instead of the
conventional code. Preserve 204 only when the action is genuinely no-content (`noContent()`, void, no body).

## Acceptance
- `destroy` returning `response()->json(['message' => …])` emits `200` with `{message: string}`, not `204`.
- A `destroy` that is genuinely void / `noContent()` still emits `204`.
- No regression on non-resourceful actions.

Found while root-causing the `response()->json([array literal])` genuinely-missing bucket (#413 re-baseline).

Radiergummi:
Closes #452

## Summary

A resourceful `destroy` action whose body returns a literal `response()->json([...])` (a 200 with
a body) gets clobbered by the conventional **204 No Content**: the spec emits `204` and discards the
real body. The body scan already reads the literal correctly — it is the precedence step that throws
it away.

Fix: a **204 (no-content) convention must yield to a body-scanned primary response that carries
content**. When a content-bearing response was resolved, keep its status (200) and schema; apply the
204 only as the fallback for genuinely no-content actions. Non-204 conventions (201 `store`, 200
`index`/`show`/`update`) keep relabeling as today — they are coherent with a body, 204 is not.

## Tier

**Tier-1.** No new parsing: the bounded-AST body scan (`InlineJsonResponseResolver`) already resolves
the literal `response()->json([...])` into a 200 + content. This is purely a precedence correction in
`OperationBuilder`.

## Root cause

`src/Support/Generator/OperationBuilder.php::build()`:

- **line 232** applies the convention status whenever there is no `#[Response]` 2xx override *and* no
  body-scanned **explicit** status. `response()->json(['message' => …])` (no status arg, no chained
  `->setStatusCode()`) resolves a 200 **with content** but is marked **non-explicit**, so the guard
  fires and the convention wins.
- **line 821** `applyConventionStatus()` for `204` returns a bare `new OA\Response(['response' =>
  '204', 'description' => 'No Content'])`, discarding the resolved body.

Why the other destroy shapes are already correct (and must stay so):
- `response()->noContent()` → resolver returns a **204 marked explicit** → the existing
  `!$autoStatusIsExplicit` guard already skips the convention → stays `204`.
- void / `new JsonResponse()` (no readable body) → `$autoPrimaryResponse` is `null` → default `200 OK`
  placeholder (no content) → `204` applied → bare `204`. Correct (the OO `new JsonResponse([...])`
  form is a separate, unbuilt case, #453).

So the **only** broken case is: a 204 convention plus a **content-bearing** body-scanned response.

## Fix

**File: `src/Support/Generator/OperationBuilder.php`** (`build()`, the line-232 guard).

Suppress a 204 convention when the auto-resolved primary response carries content:

```php
$bodyScanHasContent = $autoPrimaryResponse !== null
    && is_array($autoPrimaryResponse->content)
    && $autoPrimaryResponse->content !== [];

// A 204 forbids a body; a content-bearing body-scanned response is stronger evidence of the real
// status, so it keeps its status and schema instead of being discarded.
$conventionSuppressedByBody =
    $convention?->successStatusCode === 204 && $bodyScanHasContent;

if (
    $primaryOverride === null
    && !$autoStatusIsExplicit
    && $convention?->successStatusCode !== null
    && !$conventionSuppressedByBody
) {
    $primaryResponse = $this->applyConventionStatus($primaryResponse, $convention->successStatusCode);
}
```

Keep the content check inline (single use). A tiny private `responseHasContent(OA\Response): bool`
helper is acceptable if the coder prefers, but do not add speculative abstraction.

**Provenance (same file, `statusProvenance()` ~line 1191).** When the 204 convention is suppressed,
the emitted status now comes from the body but `$bodyScannedExplicit` is `false`, so the current
`match` would mis-attribute it to the convention. Thread the new flag through and add a `match` arm
*before* the convention arm:

```php
$conventionSuppressedByBody => ['response body', 'implicit body status (204 convention forbids a body)', []],
```

This only changes the **private** `statusProvenance` signature (add one `bool` param) — not a public
or `Contracts` surface.

## Tests (TDD — write first, watch them fail)

**Fixture:** add a `destroy` and a void destroy to
`tests/Fixtures/ResourceConventionLiteralController.php`:

```php
// DELETE destroy → convention says 204, but the literal json body (200) must win.
public function destroy(string $widget): JsonResponse
{
    return response()->json(['message' => 'Deleted.']);
}

// DELETE destroy with no readable body → genuine 204 No Content.
public function destroyVoid(string $widget): SymfonyResponse
{
    return response()->noContent();
}
```

**`tests/Unit/Support/Generator/OperationBuilderConventionStatusTest.php`** — new cases (the existing
`buildResourceOperation('delete', …, 'destroy')` helper already gates the DELETE verb):

1. **happy path** — `destroy` returning `response()->json(['message' => …])` ⇒ primary status `200`
   **and** the response has `content` with a `message` string property (assert the body survived, not
   just the status).
2. **no-content regression** — `destroyVoid` (`noContent()`) ⇒ `204`, no `content` key.
3. **non-204 convention untouched** — the existing `ResourceConventionDefaultController::store`
   (`json([])`, no status) ⇒ still `201` (proves the fix is scoped to 204; should already pass).

**Edge cases to cover / decide:**
- explicit status on a destroy — `response()->json([...], 202)` ⇒ `202` (explicit marker already
  wins; add a quick assert).
- empty-array literal — `response()->json([])` on a destroy: verify what `InlineJsonResponseResolver`
  emits; if it produces an `application/json` content entry, it counts as content ⇒ `200`. Assert the
  actual behavior and note the decision in a code comment if non-obvious.
- non-resourceful DELETE (no resourceful method name) ⇒ no convention, unaffected — covered by
  absence of regression; no new test needed.

**Regression guard:** existing `tests/Feature/ResourceConventionTest.php` "emits a body-less 204 for
the destroy action" (`WidgetController::destroy` returns bare `new JsonResponse()`, no readable body
⇒ `$autoPrimaryResponse === null`) must stay green.

## Docs / CHANGELOG

- **`CHANGELOG.md` `[Unreleased]` → Fixed:** "A resourceful `destroy` that returns
  `response()->json([...])` now documents the real `200` and its body instead of a bogus `204 No
  Content`."
- **`docs/` responses page** (locate the resourceful-convention / 204 paragraph, likely
  `docs/responses.md`): add one sentence that an explicit body response in the handler overrides the
  conventional status. Trivial, inline by the coder.

## Controversy / merge gate

- **area:core — touches response-resolution precedence.** The change is to the convention-vs-body
  precedence guard inside `OperationBuilder::build()`. It does **not** change `SpecStage` ordering,
  does **not** touch `Contracts/**`, and adds **no config key**. It changes one **private** method
  signature (`statusProvenance`). Localized and well-covered, but flag for the lead per the area:core
  precedence rule.
- **area:responses → run the `[survey]` gate.** Generation-affecting. Expected to *improve* the
  `response()->json([array literal])` genuinely-missing bucket (~20 corpus `destroy` ops:
  AureusERP `IndustryController::destroy`, InvoiceNinja `DocumentController::destroy`) with **zero
  regressions**. Surveyor confirms 0 regressions and, ideally, measures the lift.


Radiergummi:

Radiergummi:
🤖 **[lead]** plan posted (PR #464). → plan-review. Planner flags an area:core response-precedence change (private-signature only, no Contracts/stage-order/config) → expect a human merge gate.

Radiergummi:
🤖 **[reviewer]** plan-review: APPROVED with 2 required corrections (fold into the plan before coding). Tier and scope are right; verified the cited code locations.

**Verified against source:**
- Bug confirmed: `OperationBuilder::build()` line 232 guard fires for a non-explicit body-scan, and `applyConventionStatus()` line 824 returns a bare `204`, discarding `$autoPrimaryResponse->content`. The fix (suppress the 204 convention when the body-scan carries content) is a **precedence-only** correction — no new parsing. Lowest viable tier confirmed; not over-built.
- Guard idiom is correct: `is_array($autoPrimaryResponse->content) && … !== []` properly excludes swagger-php’s `UNDEFINED` string sentinel (content-less responses leave `->content` as the sentinel string, not `[]`).
- 204-preserving paths all stay correct: `noContent()` → marked-explicit 204 (skipped by `!$autoStatusIsExplicit`, content-less anyway); void / bare `new JsonResponse()` → `$autoPrimaryResponse === null` → 204 applies; `json([...], 204)` → resolver returns a content-less 204. None trip `bodyScanHasContent`.
- Scope is surgical and correctly limited to `successStatusCode === 204`. 201 `store` / 200 `index`/`show`/`update` keep relabeling via the non-204 branch (which preserves content) — out of scope for #452, correctly left untouched.
- Provenance threading is **necessary and correct**: a 204-suppressed status comes from the body with `$bodyScannedExplicit === false`, so todays `match` would misattribute it to the convention resolver. The new `$conventionSuppressedByBody` arm before the convention arm fixes it. One `bool` param on the **private** `statusProvenance` — no public/Contracts surface.

**Required corrections:**

1. **Docs target is wrong — `docs/responses.md` does not exist.** The correct page is **`docs/auto-derivation.md`, "Resource action conventions"** (~line 596), which already states the convention’s status *"defers to a status read from … the call"* for **explicit** statuses. Extend that sentence to also cover a **content-bearing body with an implicit (helper-default 200) status**: such a body keeps its `200` + schema, and the conventional `204` applies only to genuinely body-less actions. (Lines 352-363 are the adjacent `store`/status paragraph for cross-reference.)

2. **Empty-array edge case: the plan’s hypothesized "⇒ 200" is incorrect.** `InlineJsonCallReader::readBody()` returns `bodySchema: null` for `[]` (empty `$definition === []` → `null`), so `InlineJsonResponseResolver::responseFromCall()` returns `null` → `$autoPrimaryResponse === null` → `bodyScanHasContent` is false → the **204 convention applies**. So `response()->json([])` on a `destroy` **stays 204**, not 200. Assert the actual behavior (stays `204`, no `content`) and drop the "⇒ 200" branch from the plan; a one-line code comment noting empty-literal → no content → 204 is coherent is fine.

Neither correction changes the fix approach — only the docs file and one test’s expected value. Test matrix (200+body survives with `message` string asserted, `noContent()` → 204, `store` → 201, explicit `json([...],202)` → 202, empty `json([])` → 204, Feature bare-`JsonResponse` 204 regression guard) is otherwise solid.

**Controversy:** I concur with the planner. area:core response-**precedence** change, but it touches **only a private signature** (`statusProvenance`) — no `Contracts/**`, no `SpecStage` ordering, no config key. Warrants the area:core **human merge gate** flag and the `[survey]` gate (generation-affecting; expected to improve the `response()->json([array literal])` genuinely-missing bucket with zero regressions). Not auto-mergeable.

Radiergummi:
🤖 **[lead]** plan-review approved (2 corrections folded: docs page = docs/auto-derivation.md 'Resource action conventions'; `json([])` on destroy stays 204 — drop the ⇒200 branch). → coding.

Radiergummi:
🤖 **[coder]** deviation: the plan named `destroy`/`destroyVoid`/`destroyEmptyJson` methods on a single fixture, but `ResourceConventionResolver` only fires on the **exact** resource method names (`index`/`show`/`store`/`update`/`destroy`). A method named `destroyVoid` never triggers the 204 convention, so it would not exercise the fix path. To test each destroy/store body-shape *under* the convention I added dedicated single-action fixtures, each exposing a real `destroy`/`store`: `ResourceDestroyNoContentController` (noContent → 204), `ResourceDestroyEmptyJsonController` (json([]) → 204), `ResourceDestroyExplicitStatusController` (json([...],202) → 202), `ResourceStoreEmptyJsonController` (json([]) → 201). The content-bearing happy path stays on `ResourceConventionLiteralController::destroy`. Same intent and matrix as the plan, just faithful to the convention's method-name gate.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed d2963ba. Ready for review.

Remote CI all green (PHP 8.4/8.5 × Laravel 12/13, swagger-php 5.8, quality, snapshot, coverage). 6 new convention-status cases + the existing body-less 204 Feature guard pass; PHPStan clean.

Radiergummi:
🤖 **[lead]** coder done, full CI matrix green on d2963ba4 (incl. swagger-php 5.8 + PHP 8.5). → review. Dispatching two independent code reviewers. (area:core precedence change → human merge gate regardless; review + survey still run for a complete record.)

Radiergummi:
🤖 **[reviewer]** review round 1 — approved, no concerns. (464b, independent boundary review)

Adversarial pass on the *boundaries*, not the correctness table:

- **Guard scoped to 204.** `$conventionSuppressedByBody` is gated on `$convention?->successStatusCode === 204`, so resourceful **non-destroy** content-bearing returns are untouched: `applyConventionStatus()` (OperationBuilder.php:842-843) only rewrites `->response`/`->description` for non-204 and **preserves `->content`**; only the 204 branch (:838) discards it. index/show/update keep relabeling to 200 + body, store to 201 + body, exactly as before. `ResourceStoreEmptyJsonController` (still 201) pins the scoping.
- **Non-resourceful DELETE** — `$convention` is null → neither the suppression nor the apply block fires. Unaffected.
- **`destroy` + explicit `#[Response]`** — `$primaryOverride !== null` short-circuits the apply block and is the first `statusProvenance` match arm, so the attribute wins over both convention and body scan even though `$conventionSuppressedByBody` may also be true. No mislabel.
- **Provenance arm ordering audited (:1214-1228).** override → bodyScannedExplicit → conventionSuppressedByBody → convention → default. The `json([...],202)` destroy sets **both** `bodyScannedExplicit` and `conventionSuppressedByBody` true; `bodyScannedExplicit` precedes, so it correctly attributes to `response body / explicit status in handler body` and the apply block is skipped via `!$autoStatusIsExplicit`. Status 202, asserted by test. No pre-existing attribution regressed.
- **swagger-php 5.8** — the 200+content path emits the same content the body scan already produced (no new shape); the dedicated 5.8 CI job is green.
- **Content sentinel** — `is_array(...->content) && ... !== []` correctly excludes both swagger-php`s `UNDEFINED` string sentinel and the empty `json([])` case (resolver returns null → `$autoPrimaryResponse === null`). `ResourceDestroyEmptyJsonController` (stays 204, no content) pins it.

**Branch/coverage:** every changed branch is exercised by the 6 new cases + the bare-`JsonResponse` 204 Feature guard; suite green locally (19 passed), Pint + PHPStan clean on the touched file. One untested-but-correct generalization: a `destroy` returning a content-bearing response from a *non*-inline-json resolver (API Resource / paginator) would now also keep 200 + schema instead of a bogus 204 — a strict improvement consistent with the issue intent, not a regression; no test needed.

**Conventions/boundaries/bookkeeping:** strict_types on all 4 new fixtures, docblocks present, no abbreviations, comments are why-focused with no em-dashes/issue-numbers; change lives in `src/Support/Generator` (plugin-agnostic), touches only the **private** `statusProvenance` signature, no `Contracts`/stage-order/config. CHANGELOG `[Unreleased] → Fixed` entry present; `docs/auto-derivation.md` "Resource action conventions" updated at the correct spot.

Controversy: concur with planner/464a — area:core response-**precedence** change (private signature only) warrants the human merge gate + the `[survey]` gate. Not auto-mergeable, but the diff is correct and well-covered.

Radiergummi:
🤖 **[reviewer]** review round 1 — approved, no concerns. (reviewer 464a)

Reviewed adversarially in a fresh worktree of `d2963ba4` (`composer dump-autoload` first). All gates verified locally: 19/19 tests green (`OperationBuilderConventionStatusTest` + `ResourceConventionTest`), Pint passes on every touched file, codecov 100% of modified lines.

**Correctness — constructed the breaking inputs, none break it:**
- content-bearing `destroy json(['message'=>…])` ⇒ `200` + `{message: string}` body survives (`conventionSuppressedByBody` true, convention skipped). ✓
- body-less `noContent()` ⇒ `204`, no `content` — `$autoStatusIsExplicit` short-circuits the guard and the `UNDEFINED` content sentinel never trips `is_array`. ✓
- bare `new JsonResponse()` ⇒ `$autoPrimaryResponse === null` ⇒ bare `204` (Feature regression test green and still meaningful). ✓
- explicit `json([...], 202)` ⇒ `202`; provenance resolves to the `bodyScannedExplicit` arm, which **precedes** the new `conventionSuppressedByBody` arm, so the order is right and the new arm never mis-fires on an explicit status. ✓
- empty `json([])` on `destroy` ⇒ stays `204` (empty literal → resolver `null` → no content). ✓
- empty `json([])` on `store` ⇒ `201` — suppression is scoped to `successStatusCode === 204`, non-204 conventions keep relabeling. ✓

**Sentinel idiom confirmed:** `OA\Response::$content` defaults to `Generator::UNDEFINED` (a string); `is_array(...) && !== []` cleanly excludes both the sentinel and an empty array. (Matches my plan-review finding.)

**Coder deviation is legitimate:** `ResourceConventionResolver` keys its convention map on the **exact** method names (`index/show/store/update/destroy`), so `destroyVoid`/`destroyEmptyJson` would never trigger the convention. The dedicated single-action fixtures (`ResourceDestroyNoContent/EmptyJson/ExplicitStatus`, `ResourceStoreEmptyJson`) each expose a real `destroy`/`store` and exercise the identical matrix under a firing convention. Not masking a gap.

**Bookkeeping/boundaries:** CHANGELOG `Fixed` entry present (end of section, #452); `docs/auto-derivation.md` "Resource action conventions" extended for the content-bearing-body case; change lives in `src/Support/Generator` with no plugin/Contracts dependency, no config key, no stage-order change. Single call site of `statusProvenance` updated.

Minor (non-blocking, no change requested): the new `conventionSuppressedByBody` provenance arm is executed by the happy-path test but its source/reason wording isn't asserted — consistent with the existing status-provenance arms, which are also unpinned by value, so I'm not requiring it.

**Controversy:** area:core response-**precedence** change, but it touches **only a private signature** (`statusProvenance`, one new `bool` param) — no `Contracts/**`, no `SpecStage` ordering, no config key. Per the area:core precedence rule this warrants the **human merge gate** and the `[survey]` gate (generation-affecting; expected to improve the `response()->json([array literal])` genuinely-missing bucket with zero regressions). Not auto-mergeable.

Radiergummi:
🤖 **[lead]** review round 1 — **both reviewers approved independently** (464a + 464b) on d2963ba4; adversarial passes confirm the 204-suppression guard, provenance arm ordering, and no body-less-204 regression. → survey (generation-affecting). Will escalate to human merge gate after (area:core precedence).

Radiergummi:

Baseline regenerated on current `main` (`629236e0`) — the prior cache was stale (`de71286b`). Candidate tested at `d2963ba4`. All 11 apps boot + generate (gen_exit=0); paths/operations/apiOperations/documentedResponses/requestBodies/componentSchemas held equal everywhere. The only deltas are improvements in `responseSchemas`, exactly the destroy-bucket the fix targets:

| app | metric | baseline | candidate | Δ |
|------|--------|----------|-----------|---|
| AureusERP | responseSchemas | 66 | 78 | +12 |
| InvoiceNinja | responseSchemas | 67 | 69 | +2 |
| BookStack | responseSchemas | 3 | 4 | +1 |

Content-bearing `2xx` DELETE ops now surface with a 200 + schema instead of being clobbered to a bare 204 (AureusERP: 19 such ops, InvoiceNinja: 4; e.g. `DELETE /api/v1/companies/{company}` → 200 with body). No regressions, no new crashes. Clear to merge.

Radiergummi:
🤖 **[lead]** all quality gates green — **escalating for human merge confirmation (area:core response-precedence change).**

- ✅ Full CI matrix green on `d2963ba4` (incl. swagger-php 5.8 + PHP 8.5).
- ✅ Two independent reviewer approvals (464a + 464b); both adversarially confirmed the 204-suppression guard, provenance arm ordering, and no body-less-204 regression.
- ✅ Survey Layer A pass — 0 regressions across 11 apps; +responseSchemas in the exact target bucket (AureusERP +12, InvoiceNinja +2, BookStack +1; content-bearing 2xx DELETE ops now emit 200+schema instead of a bogus bare 204).
- ⚠️ **Reason for escalation:** changes response-resolution **precedence** in `OperationBuilder` (private signature only — no Contracts/stage-order/config). Per the area:core rule this warrants a human gate.

Ready, green, dual-approved + survey-clean — left unmerged with worktree in place. @radiergummi

PR #510, opened 2026-07-01
fix(core): parse each controller once — stop bypassing the scoped MethodBodyScanner

### LINKED ISSUE #485: fix(core): MethodBodyScanner scoped AST cache defeated by inline new defaults
## Problem

`Support\MethodBody\MethodBodyScanner` carries a per-instance AST cache and is annotated `#[Scoped]`, but it is **never explicitly bound** in the provider — the shared-instance contract relies entirely on attribute auto-caching when transitively resolved. Three production paths bypass it and construct fresh, empty-cache instances:

- `FormRequestStaticRulesReader.php:46` — `MethodBodyScanner $scanner = new MethodBodyScanner()` as a **constructor default** (caller `FormRequestRulesReader.php:31` also defaults `new FormRequestStaticRulesReader()`),
- `ReturnExpressionResourceReader::create()` (`:114`),
- `DataReturnExpressionReader::create()` (`:66`).

Net effect: one generation run can parse the same controller file up to ~4×. The scoped-singleton lifecycle contract exists and is silently defeated by convenience constructors.

## Proposed change

1. Bind `MethodBodyScanner` explicitly in the provider; remove the inline `new` defaults and `::create()` short-circuits so all consumers receive the shared scoped instance.
2. Adopt the general rule (arch test or Pest architecture check): **stateful scoped services must be explicitly bound and never `new`ed as parameter defaults.**

## Why

Straight perf fix (files parse once per run, as designed) and hardens the lifecycle model the whole pipeline depends on. Worth verifying with a before/after parse count on a real app (dogfood harness).

Radiergummi:
Closes #485

## Implementation plan — #485

**Tier:** N/A (not an inference change). This is a **service-lifecycle / performance fix**: it makes
the `#[Scoped]` `MethodBodyScanner` genuinely one-per-run instead of being re-`new`ed with an empty
cache by three convenience constructors. No AST shapes, resolvers, or output change.

### Root cause (verified)

`MethodBodyScanner` is `#[Scoped]` and holds a per-file `$astCache` + `$sourceCache`, but is never
explicitly bound. Its shared-instance guarantee relies entirely on the container auto-caching the
`#[Scoped]` attribute when the scanner is resolved **transitively**. All the plugin resolvers/
contributors are registered as class-strings and resolved through the container, so they *do* share
one scanner — **except** for three constructors that hand-roll a fresh, empty-cache instance:

- `FormRequestStaticRulesReader.php:46` — `MethodBodyScanner $scanner = new MethodBodyScanner()`
  parameter default. Its own caller `FormRequestRulesReader.php:31` likewise defaults
  `new FormRequestStaticRulesReader()`, so when `FormRequestRulesReader` is container-resolved it
  autowires the reader (good), but any `new FormRequestRulesReader()` (tests, and the default-chain)
  builds a private scanner.
- `ReturnExpressionResourceReader::create()` (`:111-119`) — `scanner: new MethodBodyScanner()`.
  Reached via `ResourceClassLocator::create()` (`:45-48`), which is **not used in `src/`** (the
  provider binds `ResourceTargetLocator → ResourceClassLocator::class`, container-autowired), but
  **is used widely in tests**.
- `DataReturnExpressionReader::create()` (`:64-67`) — `new self(new MethodBodyScanner())`. **Not
  used in `src/`** (the SpatieData plugin registers `DataResponseResolver` as a class-string,
  container-autowired); **used in one test**.

Net effect at runtime today: the production request-body path can build a second scanner (the
FormRequest default chain), so a controller file that hosts both a FormRequest and, say, a resource
return gets parsed more than once per run.

### Change (surgical)

1. **`FormRequestStaticRulesReader`** — drop the `= new MethodBodyScanner()` default; make
   `MethodBodyScanner $scanner` a required promoted ctor param (keep `readonly`, keep `#[Scoped]`).
2. **`FormRequestRulesReader`** — drop the `= new FormRequestStaticRulesReader()` default; require
   the dependency. Container-resolution of `FormRequestRulesReader` (its only production entry, via
   `CoreQueryParameterResolver` and `SchemaFromFormRequest`, both class-string-registered) then
   autowires the whole chain onto the one scoped scanner.
3. **`ReturnExpressionResourceReader`** — remove the `new MethodBodyScanner()` inside `create()`.
   Because `create()` still needs to build `DocBlockParser`/`TypeNodeResolver` for **test** use, the
   cleanest surgical move is to keep `create()` but have it take the scanner explicitly (see
   "Deciding the arch rule" below); the *production* path never touches `create()`.
4. **`DataReturnExpressionReader`** — same treatment for `create()`.

The precise final shape of the `create()` factories (delete vs. require-scanner) is coupled to the
arch rule and to the test fallout (~15 test sites call these factories / the inline defaults), so
this PR resolves it as follows, keeping the diff minimal:

- **`FormRequest*` inline defaults: delete.** The tests that do `new FormRequestRulesReader()` are
  updated to pass a scanner-backed reader via a tiny shared test helper (mirrors the existing
  `new PaginatorCallReader(new MethodBodyScanner())` pattern already used across the unit suite).
- **`ReturnExpressionResourceReader::create()` / `DataReturnExpressionReader::create()`: keep, but
  accept an optional pre-built `MethodBodyScanner`** so tests share one within a case. The
  production paths do not use these factories, so this is test-ergonomics only. Default stays
  `new MethodBodyScanner()` **inside `create()` only** (a test-only factory), which the arch rule
  below explicitly scopes out.

### Deciding the arch rule (issue's item 2)

The issue proposes: *"stateful scoped services must be explicitly bound and never `new`ed as
parameter defaults."* Two facts constrain this:

- Explicit `$app->scoped(MethodBodyScanner::class, …)` is **not required** — the `#[Scoped]`
  attribute already makes the container cache it; the bug is purely the `new`-as-default bypass.
  So the plan does **not** add a provider binding (it would be redundant per the documented
  "reset() methods are redundant under the scoped lifecycle" note). *If plan-review prefers an
  explicit binding for defensiveness, that is a one-line add — flagged for the reviewer.*
- Tests legitimately construct these services directly. A blanket "never `new` a `#[Scoped]`
  class" arch rule would fail the whole unit suite.

**Proposed rule (narrow, production-only):** a Pest arch test asserting that **no `src/` class
declares a constructor parameter defaulting to `new MethodBodyScanner()`** (i.e. no
`= new MethodBodyScanner()` in `src/`). This captures exactly the defeat mechanism the issue names,
leaves test factories free, and needs no new abstraction. Implemented as a small static check over
`src/**` sources (grep-style assertion in `tests/Arch/`), since Pest's `arch()` DSL cannot see
default-value expressions. Rule lives in `tests/Arch/ConventionsTest.php`.

1. **Pinned shared-instance test** (`tests/Unit/Support/MethodBody/…` or a small
   `tests/Unit/…SharedScannerTest`): resolve `FormRequestRulesReader` (and the ApiResources /
   SpatieData resolver chains) from the container **twice within one scope** and assert the
   `MethodBodyScanner` reached is the **same object** (`===`). Today the FormRequest default-chain
   fails this. Verify via a container `resolving`/reflection probe on the shared scanner identity.
2. **Arch test** (item above): fails while any `= new MethodBodyScanner()` default remains in
   `src/`; passes once the three are removed.
3. Update the ~4 test sites that construct `new FormRequestRulesReader()` / `new
   FormRequestStaticRulesReader()` to inject a scanner (shared helper). The `::create()` callers
   keep working unchanged (factories retained).

Edge cases to keep green: Octane scoped-reset semantics (a *new* scope must get a *fresh* scanner —
the shared-instance assertion is **within** one scope, not across); the FormRequest static-fallback
path (`FormRequestRulesReaderTest`) still recovers literals.

### Observable behaviour / docs / changelog

- No spec output change (survey must show **0 regressions** — this is the gate that proves the perf
  fix is behaviour-preserving; optionally include a before/after parse-count note from the dogfood
  harness in the PR, per the issue).
- `docs/`: no user-facing page changes. The service-lifecycle note in `CLAUDE.md` already states the
  scoped contract; no edit needed unless plan-review wants the arch rule mentioned.
- `CHANGELOG.md [Unreleased]`: one line under a "Fixed"/"Performance" heading — "Controller source
  files are now parsed once per generation run; the shared `MethodBodyScanner` is no longer bypassed
  by convenience constructors."

### Controversial? 

No public surface: `MethodBodyScanner`, the readers, and the factories are all `@internal`. No
`Contracts/**`, no stage order, no config key, no public signature change. Fast-ish, but it **is
`area:core` and generation-adjacent**, so the survey gate (0-regression) applies. Not a human gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,6 +86,7 @@ All notable changes to this project are documented here.
 - `path.parameter-undeclared` no longer false-positives on custom-key route bindings (`{param:field}`); URI-placeholder extraction is now shared between the generator and the linter. (#426)
 - A resourceful `destroy` that returns `response()->json([...])` now documents the real `200` and its body instead of a bogus `204 No Content`; the conventional `204` applies only to genuinely body-less actions (`noContent()`, a bare `new JsonResponse()`, or an empty `response()->json([])`). (#452)
 - Survey tooling: `tools/survey/metrics.php` resolves its Composer autoloader from the main checkout when run from a git worktree (which has no `vendor/` of its own), so parsing a YAML published spec no longer fatals and false-FAILs those apps. (#468)
+- Each controller source file is now parsed once per generation run: the FormRequest rules readers no longer hand-roll a private `MethodBodyScanner` via constructor defaults, so every production consumer autowires the one `#[Scoped]` scanner and shares its per-run AST cache. (#485)
 
 ### Documentation
 - README refreshed for the current feature set (Tier-1 method-body inference, SwaggerPhp plugin) and trimmed to a gentle overview that defers detail to `docs/`.
--- a/src/Plugins/Core/Support/FormRequestRulesReader.php
+++ b/src/Plugins/Core/Support/FormRequestRulesReader.php
@@ -28,7 +28,7 @@ use function sprintf;
 final readonly class FormRequestRulesReader
 {
     public function __construct(
-        private FormRequestStaticRulesReader $staticRulesReader = new FormRequestStaticRulesReader(),
+        private FormRequestStaticRulesReader $staticRulesReader,
     ) {}
 
     /**
--- a/src/Plugins/Core/Support/FormRequestStaticRulesReader.php
+++ b/src/Plugins/Core/Support/FormRequestStaticRulesReader.php
@@ -43,7 +43,7 @@ final readonly class FormRequestStaticRulesReader
     private SingleReturnArrayLiteralFinder $bareReturnFinder;
 
     public function __construct(
-        private MethodBodyScanner $scanner = new MethodBodyScanner(),
+        private MethodBodyScanner $scanner,
     ) {
         $this->bareReturnFinder = new SingleReturnArrayLiteralFinder($this->scanner);
     }
--- a/tests/Arch/ConventionsTest.php
+++ b/tests/Arch/ConventionsTest.php
@@ -33,6 +33,25 @@ foreach ($plugins as $plugin) {
         );
 }
 
+// Scoped-scanner integrity — the #[Scoped] MethodBodyScanner holds a per-run AST cache, so a
+// constructor parameter defaulting to `= new MethodBodyScanner()` would hand any hand-constructed
+// consumer a private, empty-cache instance and defeat the shared cache. Production consumers must
+// require the dependency and let the container autowire the one scoped instance. Scoped narrowly to
+// ctor-parameter defaults: factory bodies (`create()`) legitimately build a scanner for test use.
+test('no src class defaults a constructor parameter to a new MethodBodyScanner', function (): void {
+    $offenders = [];
+
+    foreach (phpFilesUnder(srcPath()) as $file) {
+        $contents = file_get_contents($file);
+
+        if ($contents !== false && preg_match('/=\s*new\s+MethodBodyScanner\s*\(/', $contents) === 1) {
+            $offenders[] = $file;
+        }
+    }
+
+    expect($offenders)->toBe([]);
+});
+
 // @internal isolation — the documented public extension surface (Contracts/) must not leak
 // implementation types tagged @internal. Reflection-based rather than a namespace denylist:
 // Contracts legitimately uses non-@internal classes that live alongside @internal ones (e.g.
--- a/tests/Feature/Examples/FakerSynthesisIntegrationTest.php
+++ b/tests/Feature/Examples/FakerSynthesisIntegrationTest.php
@@ -8,7 +8,6 @@ use OpenApi\Generator;
 use Psr\Log\NullLogger;
 use Radiergummi\OpenApi\Attributes\RequestField;
 use Radiergummi\OpenApi\Lint\ArrayFindingsCollector;
-use Radiergummi\OpenApi\Plugins\Core\Support\FormRequestRulesReader;
 use Radiergummi\OpenApi\Plugins\Core\Support\SchemaFromFormRequest;
 use Radiergummi\OpenApi\Support\Extraction\FakerExampleSynthesiser;
 use Radiergummi\OpenApi\Support\Extraction\ValidationRulesToSchema;
@@ -44,7 +43,7 @@ function fakerSynthesisBuildSchema(bool $synthesise): OA\Schema
         synthesiser: new FakerExampleSynthesiser(enabled: $synthesise, seed: 1234),
         findings: new ArrayFindingsCollector(),
         explicitSchema: new ExplicitClassSchema(new NullLogger()),
-        rulesReader: new FormRequestRulesReader(),
+        rulesReader: formRequestRulesReader(),
     );
 
     $builder->build(fakerSynthesisMakeEmailFormRequest());
@@ -134,7 +133,7 @@ it('does not overwrite an authored example from a #[RequestField] attribute', fu
         synthesiser: new FakerExampleSynthesiser(enabled: true, seed: 1234),
         findings: new ArrayFindingsCollector(),
         explicitSchema: new ExplicitClassSchema(new NullLogger()),
-        rulesReader: new FormRequestRulesReader(),
+        rulesReader: formRequestRulesReader(),
     );
 
     $builder->build($request::class);
--- a/tests/Pest.php
+++ b/tests/Pest.php
@@ -6,7 +6,10 @@ use Pest\Expectation;
 use PHPUnit\Framework\Assert;
 use Psr\Log\AbstractLogger;
 use Radiergummi\OpenApi\Lint\Finding;
+use Radiergummi\OpenApi\Plugins\Core\Support\FormRequestRulesReader;
+use Radiergummi\OpenApi\Plugins\Core\Support\FormRequestStaticRulesReader;
 use Radiergummi\OpenApi\Support\Generator\OpenApiGenerator;
+use Radiergummi\OpenApi\Support\MethodBody\MethodBodyScanner;
 use Radiergummi\OpenApi\Support\Spec\SpecRegistry;
 use Radiergummi\OpenApi\Tests\TestCase;
 use Symfony\Component\Yaml\Yaml;
@@ -87,6 +90,16 @@ function generateSpec(?string $specName = null, string $environment = 'testing')
     $env = $environment !== 'testing' ? $environment : app()->environment();
 
     return Yaml::parse(app(OpenApiGenerator::class)->generate($spec, $env)->toYaml());
+}
+
+/**
+ * A {@see FormRequestRulesReader} backed by a fresh scanner, for tests that construct the reader
+ * directly. Production wires the whole chain onto the one scoped {@see MethodBodyScanner}; tests
+ * that build it by hand supply their own scanner here.
+ */
+function formRequestRulesReader(): FormRequestRulesReader
+{
+    return new FormRequestRulesReader(new FormRequestStaticRulesReader(new MethodBodyScanner()));
 }
 
 /**
--- a/tests/Unit/Core/Extractors/FormRequestRequestSchemaResolverTest.php
+++ b/tests/Unit/Core/Extractors/FormRequestRequestSchemaResolverTest.php
@@ -7,7 +7,6 @@ use Psr\Log\NullLogger;
 use Radiergummi\OpenApi\Enums\MediaType;
 use Radiergummi\OpenApi\Lint\ArrayFindingsCollector;
 use Radiergummi\OpenApi\Plugins\Core\Resolvers\FormRequestRequestSchemaResolver;
-use Radiergummi\OpenApi\Plugins\Core\Support\FormRequestRulesReader;
 use Radiergummi\OpenApi\Plugins\Core\Support\SchemaFromFormRequest;
 use Radiergummi\OpenApi\Routing\ActionDescriptor;
 use Radiergummi\OpenApi\Support\Extraction\FakerExampleSynthesiser;
@@ -37,7 +36,7 @@ function makeFormRequestResolver(): FormRequestRequestSchemaResolver
         synthesiser: new FakerExampleSynthesiser(enabled: false),
         findings: new ArrayFindingsCollector(),
         explicitSchema: new ExplicitClassSchema(new NullLogger()),
-        rulesReader: new FormRequestRulesReader(),
+        rulesReader: formRequestRulesReader(),
     );
 
     return new FormRequestRequestSchemaResolver(
@@ -105,7 +104,7 @@ it('registers the FormRequest schema in the component registry', function (): vo
         synthesiser: new FakerExampleSynthesiser(enabled: false),
         findings: new ArrayFindingsCollector(),
         explicitSchema: new ExplicitClassSchema(new NullLogger()),
-        rulesReader: new FormRequestRulesReader(),
+        rulesReader: formRequestRulesReader(),
     );
     $resolver = new FormRequestRequestSchemaResolver(
         schemaBuilder: $builder,
--- a/tests/Unit/Core/Extractors/SchemaFromFormRequestTest.php
+++ b/tests/Unit/Core/Extractors/SchemaFromFormRequestTest.php
@@ -8,7 +8,6 @@ use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use Radiergummi\OpenApi\Contracts\Lint\Severity;
 use Radiergummi\OpenApi\Lint\ArrayFindingsCollector;
-use Radiergummi\OpenApi\Plugins\Core\Support\FormRequestRulesReader;
 use Radiergummi\OpenApi\Plugins\Core\Support\SchemaFromFormRequest;
 use Radiergummi\OpenApi\Support\Extraction\FakerExampleSynthesiser;
 use Radiergummi\OpenApi\Support\Extraction\ValidationRulesToSchema;
@@ -32,7 +31,7 @@ beforeEach(function (): void {
         synthesiser: new FakerExampleSynthesiser(enabled: false),
         findings: $this->findings,
         explicitSchema: new ExplicitClassSchema(new NullLogger()),
-        rulesReader: new FormRequestRulesReader(),
+        rulesReader: formRequestRulesReader(),
     );
 });
 
@@ -179,7 +178,7 @@ it('registers a placeholder schema, logs a warning, and emits a finding when rul
         synthesiser: new FakerExampleSynthesiser(enabled: false),
         findings: $findings,
         explicitSchema: new ExplicitClassSchema(new NullLogger()),
-        rulesReader: new FormRequestRulesReader(),
+        rulesReader: formRequestRulesReader(),
     );
 
     $brokenClass = new class () extends FormRequest {
--- a/tests/Unit/Core/Resolvers/CoreQueryParameterResolverTest.php
+++ b/tests/Unit/Core/Resolvers/CoreQueryParameterResolverTest.php
@@ -7,7 +7,6 @@ use OpenApi\Generator;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use Radiergummi\OpenApi\Plugins\Core\Resolvers\CoreQueryParameterResolver;
-use Radiergummi\OpenApi\Plugins\Core\Support\FormRequestRulesReader;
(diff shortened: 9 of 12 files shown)


Radiergummi:

Key call-outs for the reviewer: this is a lifecycle/perf fix, no spec output change (survey must show 0 regressions). The plan does **not** add an explicit `$app->scoped(MethodBodyScanner::class, …)` binding — `#[Scoped]` already caches it; the bug is purely the three `= new MethodBodyScanner()` bypasses. Arch rule is scoped narrowly to "no `= new MethodBodyScanner()` default in `src/`" so test factories stay free. If plan-review prefers a defensive explicit binding, thats a one-line add.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Reviewer to assess: scanner-cache dedup, arch rule scope, and the flagged option of an explicit provider binding.

Radiergummi:
🤖 **[reviewer]** plan-review: approved with two required amendments.

Verified against source: all three defeat sites exist and match the plan —
`FormRequestStaticRulesReader.php:46` (`= new MethodBodyScanner()` ctor default),
`ReturnExpressionResourceReader.php:114` (`scanner: new MethodBodyScanner()` in `create()`),
`DataReturnExpressionReader.php:66` (`new self(new MethodBodyScanner())` in `create()`).
`MethodBodyScanner` is `#[Scoped]` with a mutable `$astCache`/`$sourceCache` (`:44,:47`) and no
explicit binding, so the auto-cache contract is real and the `new`-as-default genuinely bypasses it.

**On the lead's question (explicit provider binding vs `#[Scoped]` auto-cache): the plan is right to
skip the binding.** The `#[Scoped]` attribute already makes the container cache the instance per
scope on transitive resolution; an explicit `$app->scoped(MethodBodyScanner::class, …)` would be
redundant, and CLAUDE.md's service-lifecycle note explicitly calls the equivalent `reset()` machinery
redundant under the scoped lifecycle. The bug is purely the `new`-as-default bypass, so fixing the
bypass is the whole fix. Adding a defensive one-line binding would be belt-and-suspenders that
contradicts the documented lifecycle model — **don't add it.** Removing the three `new` defaults so
every production consumer autowires the one scoped instance is the correct and minimal change.

**Required amendments (fold in before/while coding):**

1. **Scanner-identity assertion must probe the actual `$scanner` on the resolved chain, not just
   count instances.** `MethodBodyScanner` is `final` (not readonly; carries the AST cache), so the
   pinned test should resolve `FormRequestRulesReader` from the container twice within one scope and
   assert (via reflection on the private `$scanner` property down the chain, or a container
   `resolving` probe) that the reached scanner is `===` the same object. Also add the equivalent for
   the ApiResources and SpatieData resolver chains as the plan promises — today only the FormRequest
   default-chain regresses, but pinning all three prevents a future `new` default reintroducing the
   bug elsewhere. Keep a **separate-scope** case asserting a fresh scanner (Octane reset semantics).

2. **Arch-rule scope is correct as written — keep it narrow.** Assert no `= new MethodBodyScanner()`
   default expression in `src/**`. Do **not** widen it to "no `new` of any `#[Scoped]` class": that
   would false-positive on legitimate stateless defaults (e.g. `MethodBodyScanner`'s own
   `LoggerInterface $logger = new NullLogger()` default at `:53`) and break test factories. The
   narrow rule captures exactly the #485 defeat mechanism.

Scope is otherwise surgical: no `Contracts/**`, no stage order, no config, all touched classes
`@internal`. Naming the CHANGELOG line under "Fixed"/"Performance" and the survey 0-regression gate
(area:core, generation-adjacent) is correct — the survey is the proof this is behaviour-preserving.
The plan's own note that the `create()` factories stay test-only (retaining an internal
`new MethodBodyScanner()` inside `create()`) is acceptable since production never routes through them;
the arch rule intentionally scopes to *ctor parameter defaults*, not factory bodies — make sure the
regex/assertion reflects that (match `= new MethodBodyScanner()` as a default, not every occurrence,
or the retained `create()` bodies will trip it).

No blocking concerns. Approved to proceed once amendments 1 and 2 are folded in.

Radiergummi:
🤖 **[lead]** plan-review approved → coding. Coder must fold in the two amendments from plan-review: (1) identity test probes the resolved `$scanner` chain with `===` across all three resolver chains + separate-scope freshness case; (2) arch rule matches ctor defaults `= new MethodBodyScanner()` only. No explicit provider binding (ruled redundant under #[Scoped]).

Radiergummi:
🤖 **[coder]** implemented; local gates green (Pint pass, PHPStan no errors, 3310 tests pass). Pushed c47f17ac. Ready for review.

**The fix (surgical):** dropped the two `= new MethodBodyScanner()` / `= new FormRequestStaticRulesReader()` constructor-parameter defaults in `FormRequestStaticRulesReader` and `FormRequestRulesReader`, so every production consumer autowires the one `#[Scoped]` scanner and shares its per-run AST cache. No explicit provider binding added (redundant under `#[Scoped]`, per plan-review).

**Amendment 1 — scanner-identity test** (`tests/Unit/Support/MethodBody/SharedMethodBodyScannerTest.php`): resolves each production chain from the container and walks its private-property graph to the reached `MethodBodyScanner`, asserting `===` the shared `app(MethodBodyScanner::class)` across all four entrypoints (`CoreQueryParameterResolver`, `SchemaFromFormRequest`, `ResourceTargetLocator`, `DataResponseResolver`), a cross-chain single-scope case, and a separate-scope freshness case (`forgetScopedInstances()` → not `===`). Verified red before the fix on the FormRequest chains, green after.

**Amendment 2 — arch rule** (`tests/Arch/ConventionsTest.php`): asserts no `src/` file declares a ctor param defaulting to `= new MethodBodyScanner()`. Matches the default-expression only (`/=\s*new\s+MethodBodyScanner\s*\(/`), so the retained test-only `create()` factory bodies (`scanner: new …`, `new self(new …)`) and `MethodBodyScanner`'s own `NullLogger` default do not trip it. Verified it flagged `FormRequestStaticRulesReader.php` before the fix.

**Test-site fallout:** added a `formRequestRulesReader()` helper in `tests/Pest.php` (builds the chain on a fresh scanner) and pointed the ~6 hand-construction sites at it; dropped the now-orphaned imports.

Note: `create()` factories left untouched (production never routes through them; kept minimal per surgical-change guidance) — the plan floated making them accept an optional scanner for ergonomics, but no test needed it, so I skipped that to keep the diff tight.

`area:core` + generation-adjacent → survey 0-regression gate applies.

Radiergummi:
🤖 **[lead]** implementation pushed (c47f17ac), local gates green → review. Round 1: reviewer-1. Note the coder's recorded deviation (create() factories left untouched — production never routes through them); reviewer to confirm that's sound.

Radiergummi:
🤖 **[reviewer]** review round 1: approved, no blocking concerns.

I re-ran the suite locally against c47f17ac (applied to a clean main), including a deliberate
red-before check. Both plan-review amendments landed as specified and the fix is correct and minimal.

**Correctness — verified.** The two constructor-parameter defaults are dropped
(`FormRequestRulesReader.php:31`, `FormRequestStaticRulesReader.php:46`), so
`CoreQueryParameterResolver` and `SchemaFromFormRequest` autowire the whole chain onto the one
`#[Scoped]` scanner. No explicit provider binding added — correct per the lifecycle model (the
`#[Scoped]` attribute already caches; a binding would be redundant). No `Contracts/**`, no stage
order, no config, no public signature change; all touched classes `@internal`.

**Amendment 1 (identity test) — landed and load-bearing.** `SharedMethodBodyScannerTest` walks each
production chain's private-property graph to the reached `MethodBodyScanner` and asserts `===` the
shared `app(MethodBodyScanner::class)` across all four entrypoints (FormRequest query-param,
FormRequest request-body, ApiResources `ResourceTargetLocator`, SpatieData `DataResponseResolver`),
plus a cross-chain single-scope case and a `forgetScopedInstances()` freshness case. I confirmed
**red-before/green-after**: with the two defaults reverted (tests kept), the FormRequest per-chain and
cross-chain cases fail with "Failed asserting that two variables reference the same object"; with the
fix, all 6 cases pass (13 assertions). The property-graph walk is a sound, resilient probe.

**Amendment 2 (arch rule) — landed and narrow.** `ConventionsTest` matches
`/=\s*new\s+MethodBodyScanner\s*\(/` over `src/`, i.e. ctor-parameter defaults only. Verified it
flags `FormRequestStaticRulesReader` before the fix and passes after, and that it does **not**
false-positive on the retained `create()` factory bodies (`scanner: new …`, `new self(new …)`) or on
`MethodBodyScanner`'s own `NullLogger` default — exactly the scope I asked for.

**On the recorded deviation (`create()` factories left untouched) — sound, one wording nit.** The
*conclusion* is correct: the ApiResources and SpatieData production chains reach the shared scanner,
which I confirmed both by the passing identity test and by tracing the wiring —
`$app->scoped(ResourceTargetLocator::class, ResourceClassLocator::class)`
(`OpenApiServiceProvider.php:577-580`) is a plain class-string binding, so the container autowires
`ResourceClassLocator::__construct` → `ReturnExpressionResourceReader::__construct(MethodBodyScanner …)`
and never invokes the `create()` factory. So the empty-cache `create()` path is dead in production.

Minor accuracy note (PR-comment only, not code): the comment says "production never routes through
them," but `ReturnExpressionResourceReader::create()` **is** referenced in production source at
`ResourceClassLocator.php:47`. The precise statement is that the container binding autowires the
constructor and does not call `create()`. Worth correcting in the record so a future reader doesn't
grep, find the call site, and think the note is wrong. **Not a defect** — the identity test covers
this chain empirically.

Re the re-introduction vector the lead raised: the arch rule intentionally doesn't cover `create()`
factory bodies, so a *future* refactor that bound `ResourceTargetLocator` via a `create()`-calling
closure wouldn't be caught by the arch rule — but the identity test's ApiResources/SpatieData cases
**would** fail in that scenario. Between the two guards the current surface is covered; widening the
arch rule to factory bodies isn't warranted now (would need the test-ergonomics exception the plan
deliberately kept). No action required.

**Tests / gates — all green locally.**
- Affected suites (`tests/Unit/Core`, `tests/Arch`, `tests/Unit/Support/MethodBody`, the Faker
  integration test): 396 passed, 920 assertions.
- Pint `--test` on the changed files: passed.
- PHPStan (full): no errors.
- Test-site fallout is clean and mechanical: the `formRequestRulesReader()` helper in `tests/Pest.php`
  builds the chain on a fresh scanner; the ~6 hand-construction sites point at it; orphaned imports
  dropped; the single `FormRequestStaticRulesReaderTest` site passes an explicit `new
  MethodBodyScanner()`. Consistent with the "shared test helper" pattern the plan named.

**Bookkeeping.** CHANGELOG `[Unreleased]` "Fixed" entry present and accurate (#485). Docs: no
user-facing surface change — all touched classes `@internal`; the lifecycle contract already lives in
CLAUDE.md; no page edit needed. No new lint rule → no catalog row.

**Test-suite gap check (my standing duty).** The pinned behaviour (one scanner per scope; fresh per
new scope; no `new` default in `src`) is covered on all four chains plus the boundary cases. Nothing
consumer-visible is left uncovered — this is internal lifecycle plumbing with no new public surface.

Approved. `area:core` + generation-adjacent → the survey 0-regression gate still applies before merge
(this fix should be byte-neutral on output; the survey is the proof).

Radiergummi:
🤖 **[reviewer]** independent review (reviewer-2): approved, no concerns.

Formed my own view from #485, the PR plan, and the diff at `c47f17ac` before reconciling with reviewer-1; we converge.

**Correctness / wiring (traced independently).** The `#[Scoped]` `MethodBodyScanner` is *not* explicitly bound (the plan deliberately relies on attribute auto-caching — correct, and consistent with the "reset() is redundant under scoped lifecycle" note). All four production chains autowire onto the one instance: `SchemaFromFormRequest`/`CoreQueryParameterResolver` → required `FormRequestRulesReader` → required `FormRequestStaticRulesReader` → required scanner; `ResourceTargetLocator` bound as class-string → `ResourceClassLocator::__construct(ReturnExpressionResourceReader)` → required scanner; `DataResponseResolver` class-string → `DataReturnExpressionReader` required scanner. The two retained `create()` factories (`ReturnExpressionResourceReader`, `DataReturnExpressionReader`) are test-only and never on the container path — confirmed.

**TDD (verified red-before/green-after myself).** Reverted both defaults on disk, kept the tests: the arch rule flags `FormRequestStaticRulesReader` and the identity test fails ("two variables reference the same object") — 3 failed. Restored: identity test 6/6 (13 assertions), arch 9/9, full affected set 77 passed. Then restored working tree clean.

**Edge cases a first pass tends to miss.**
- *Graph-walk brittleness:* `reachableMethodBodyScanner()` BFS-walks initialized object properties. I checked every chain holds its deps as direct promoted object properties (no arrays/closures hiding the scanner), so the walk reaches it and `->not->toBeNull()` is safe. `$seen` is `spl_object_id`-keyed and checked after the early return — no cycle/infinite-loop risk. Coupling to the private graph is acceptable here: the one-per-scope invariant is otherwise unobservable.
- *Octane scoped semantics:* `forgetScopedInstances()` exists in the bundled container; the fresh-scope case asserts a *new* instance post-reset, correctly pinning "shared within a scope, fresh across scopes."
- *PHP 8.4:* `new ReflectionObject(...)->getProperties()` (paren-free new) is 8.4+; project floor is 8.4. Fine.

**Scope / conventions / bookkeeping.** Diff stays within plan: two default removals in `src/`, arch rule, identity test, mechanical test-site updates via the `formRequestRulesReader()` helper + one explicit-scanner site. No `Contracts/` or stage-order touch; all classes `@internal`. Strict-types present; comments concrete. CHANGELOG `[Unreleased]` entry accurate; no new lint rule; no user-facing docs surface.

**Gates (this worktree, `c47f17ac`).** Pint `--test` passed; PHPStan no errors; affected suites green.

Agreeing with reviewer-1's wording nit: the recorded-deviation note's "production never routes through them" is imprecise (`ResourceClassLocator.php:47` *does* reference `create()`); the precise statement is that the class-string binding autowires the constructor and never invokes `create()`. PR-comment record only, not a code defect.

Second independent approval. `area:core` + generation-adjacent → the survey 0-regression gate is the remaining proof of byte-neutrality before merge.

Radiergummi:
🤖 **[lead]** two independent approvals → survey. Expectation: perf-only fix, spec output byte-neutral across the corpus; any delta is a regression.

Radiergummi:
🤖 **[surveyor]** Layer A pass — byte-neutral across 11 apps.

Ran a **fresh `main` baseline** (417ee8eb — the cached one from Jun 26 was stale after #459/#466/#464/#465) and the candidate (c47f17ac) over the pinned 11-app corpus on PHP 8.4. Compared **aggregate metrics AND byte-level spec identity** per app.

**Deterministic metrics: 0 regressions, 0 changes.** paths / operations / apiOperations / responseSchemas / documentedResponses / requestBodies / maxRequestProperties / componentSchemas / completenessPercent are identical on all 11 apps. All booted, `generateExit=0`, no crashes.

**Byte-level specs: 8/11 byte-identical outright.** The 3 that differ do so *only* in Laravel/Livewire per-boot random tokens, not in any library-produced content:

| app | what differs | cause |
|-----|--------------|-------|
| AureusERP | `/livewire-<hash>/…` paths + opIds | Livewire per-boot asset-URL hash |
| SpeedtestTracker | `/livewire-<hash>/…` paths + opIds | Livewire per-boot asset-URL hash |
| InvoiceNinja | `<prefix>.generated_<rand>` operationIds | Laravel auto-names unnamed routes `generated::<Str::random>` per boot |

After masking those runtime tokens, all three are **semantically identical** to baseline (sorted-key JSON equal). Decisive tell: **all 92** `*.generated_` operationIds in InvoiceNinja flip between runs — a library change could never uniformly rewrite 100% of them; only a per-boot RNG does. These tokens vary every boot regardless of the library and appear on both baseline and candidate.

Consistent with the fix being pure DI plumbing (drop `new MethodBodyScanner()` / `new FormRequestStaticRulesReader()` defaults so the `#[Scoped]` cached scanner is injected). The scanner’s per-method output is deterministic whether cached or fresh, so spec output is unchanged — confirmed empirically.

Perf claim (parse-once) is validated by the design + the byte-neutral output; did not add the parse-count instrumentation (would require patching the library under test, which the gate forbids). Clear to merge.

Radiergummi:
🤖 **[lead]** all gates green (2 independent approvals, full CI matrix incl. swagger-php 5.8 leg, survey Layer A byte-neutral across 11 apps) + non-controversial (internal-only, surgical diff) → squash-merged. Closes #485; worktree removed.

PR #511, opened 2026-07-01
fix(core): consolidate Tier-1 return-resolution + fix Data stale-mutation bug

### LINKED ISSUE #486: refactor(core): consolidate Tier-1 matcher plumbing — shared ReturnExpressionResolver with mutation check
## Problem

The Tier-1 substrate is half-centralized: leaf primitives (`MethodBodyScanner`, `StatementNodeFinder`, `AstLiteralEvaluator`, `SchemaDefinitionFromLiteral`) are shared, but matcher plumbing is copied per consumer:

1. **Duplicated return-assignment resolution with a behavioral divergence (latent bug):** `SingleReturnArrayLiteralFinder::arrayLiteralAssignedTo` (`:94`) and `DataReturnExpressionReader::expressionAssignedTo` (`:173`) implement the same "resolve `return $var` through its single unconditional assignment" dance — but **only the former checks for later mutation of the variable** (`mutatesVariable`, `:143`). The Data path can attribute a stale literal to the return. Both also hand-roll the "collect method-level returns skipping nested scopes" walk.
2. `new StatementNodeFinder()` duplicated in **14 classes** (it's stateless — should be one shared readonly service).
3. The "unqualified helper not shadowed by a same-namespace function" guard is duplicated verbatim (`AbortErrorContributor::helperName` `:109`, `InlineJsonCallReader::isResponseHelperCall` `:85`).
4. No "register a matcher" seam: a genuinely new whitelisted shape means a new resolver class + hand-rolled `instanceof` matcher + `CorePlugin` registration + possibly a provider binding.

## Proposed change

- Extract one shared `ReturnExpressionResolver` (with the mutation check) used by both readers; delete the duplicates.
- Share a single `StatementNodeFinder`; extract the helper-shadowing guard into a matcher helper.
- Add a small call-shape matcher utility so new shapes are closer to data (name map + arg spec) than plumbing — extending `AbortErrorContributor` for `abort_if` was a one-line map entry; that should be the norm, not the exception.

## Why

Keeps the bounded-AST charter intact while making each new Tier-1 shape cheap and consistent. Explicitly **not** in scope: rebuilding on PHPStan's collector API — rejected in the epic (wrong unit of work; none of the ~12 shapes needs type inference; corpus says the unlockable band is Tier-2-dynamic).

Radiergummi:
Closes #486

## Implementation plan — #486

**Tier:** stays entirely within **Tier-1** (bounded AST whitelist). This is consolidation +
**one latent-bug fix**, not new inference. Explicitly **not** a PHPStan-collector rebuild (rejected
in the epic; recorded in the issue).

### The load-bearing part first: the latent bug (must land as a pinned test)

Three classes independently implement "resolve `return $var` through its single unconditional
assignment", and they **diverge**:

| class | method | mutation check? | return |
|---|---|---|---|
| `SingleReturnArrayLiteralFinder` (Support) | `arrayLiteralAssignedTo` (:94) | **yes** (`mutatesVariable`, :143) | `?Array_` |
| `DataReturnExpressionReader` (SpatieData) | `expressionAssignedTo` (:173) | **no** | `?Expr` |
| `ReturnExpressionResourceReader` (ApiResources) | `expressionAssignedTo` (:489) | **no** | `?Expr` (+ `note()` logging) |

The Data reader can therefore attribute a **stale literal** to the return when the variable is
mutated after its single assignment, e.g.:

```php
public function index(): Collection {
    $data = FooData::collect($foos);
    $data = $data->concat($bars);   // unconditional reassignment / mutation after the literal
    return $data;
}
```

`SingleReturnArrayLiteralFinder` refuses this (its richer fallback is the honest answer);
`DataReturnExpressionReader` does not, so it emits a `DataReturnTarget` for a value the code has
since changed. **This is a bug fix, and the mutation check must be covered by a test that fails
before and passes after** (per the lead's note).

Note the two lax readers return arbitrary `?Expr` (a `new X()` / `X::collect()`), not just an
`Array_`, so the mutation guard must be lifted to the *general* case: "the variable is assigned
exactly once unconditionally **and never mutated afterward** on the unconditional path" — the same
`mutatesVariable` semantics (`ArrayDimFetch` element writes + any `AssignOp`), independent of what
the assigned expression is.

### Proposed change (surgical, in dependency order)

**1. Shared `ReturnExpressionResolver` (Support\MethodBody), with the mutation check.**

New `@internal readonly` service in `src/Support/MethodBody/`. One public method resolving a
returned variable to its single unconditional assignment expression, refusing on: dynamically-named
variable, zero/multiple assignments (any-context vs unconditional counts must both be 1, as today),
and **any post-assignment unconditional mutation** (the `mutatesVariable` guard hoisted verbatim
from `SingleReturnArrayLiteralFinder`). Returns `?Expr`.

Also fold in the shared "collect method-level returns, skipping nested scopes" walk (currently
hand-rolled 4×): a method returning `list<Return_>` for the method's own returns. Reconcile the two
skip-predicates — `SingleReturnArrayLiteralFinder`/`FormRequestStaticRulesReader` skip
`FunctionLike`; the other two skip `Closure|ArrowFunction|ClassLike`. At method-body top-level these
are behaviourally equivalent (no nested `ClassMethod`/`Function_` decls appear in a controller body),
but to be safe the shared version skips **`FunctionLike` OR `ClassLike`** (superset of both), and
this equivalence is asserted by keeping every existing per-consumer test green.

Consumers refactored onto it, **deleting the duplicates**:
- `SingleReturnArrayLiteralFinder::arrayLiteralAssignedTo` → call the resolver, then narrow the
  result to `Array_` (`return $expr instanceof Array_ ? $expr : null`). Keep its `find()` shape.
- `DataReturnExpressionReader::expressionAssignedTo` → **deleted**; call the resolver. **This is
  where the bug is fixed** (it now inherits the mutation guard).
- `ReturnExpressionResourceReader::expressionAssignedTo` → replace body with a resolver call. It
  keeps its `note()` logging by mapping the resolver's null outcome to the existing notice(s); to
  preserve today's exact log wording (dynamically-named vs not-assigned-once), the resolver returns
  enough to distinguish, OR (simpler, preferred) the reader keeps a thin wrapper that logs then
  delegates. Chosen: **thin logging wrapper** so log messages are byte-identical and the resolver
  stays log-free (matches the Support-layer "no plugin/log policy" convention). Flag: this reader
  currently does **not** have the mutation check, so tightening it may change resource resolution
  on a mutated-variable body — the survey gate must confirm this is a net-neutral/positive change,
  not a regression. If the survey shows resource-count drops, we scope the mutation-guard tightening
  to the Data path only and leave ApiResources unchanged (documented decision in a PR comment).

**2. Single shared `StatementNodeFinder` instance.**

It is `final readonly` and stateless. Currently `new StatementNodeFinder()` appears in **14
classes** across `src/` (Core resolvers/contributors, ApiResources, Fractal, QueryBuilder,
SpatieData, Support). Two options:
- (a) inject it via the container (`#[Scoped]` or plain — it's stateless, a plain shared singleton
  is fine) and replace the 14 `new`s with a promoted ctor param;
- (b) leave construction but centralise.

Chosen: **(a) constructor injection** for the classes already container-resolved; for the handful of
Support/leaf classes constructed by hand in tests, keep a no-arg-default *only if* it does not
reintroduce the #485 anti-pattern — since `StatementNodeFinder` is **stateless**, a
`= new StatementNodeFinder()` default is harmless (no cache to defeat) and the #485 arch rule is
scoped to `MethodBodyScanner` specifically. This keeps the 14-site change mechanical and test-safe.
*(If plan-review prefers a hard "no `new` defaults at all" rule, that widens the test churn; flagged.)*

**3. Extract the helper-shadowing guard.**

`AbortErrorContributor::helperName` (:109) and `InlineJsonCallReader::isResponseHelperCall` (:85)
both implement verbatim: "an unqualified helper name is skipped when a same-namespace function of
that name exists (PHP would call that instead)". Extract to a small `@internal` static helper in
`Support\MethodBody` (e.g. `UnqualifiedHelperCall::isShadowed(Name $name): bool` or
`resolvesToGlobalHelper(...)`), used by both. The two call sites differ slightly (`FuncCall` name vs
`response()` receiver name), so the helper takes the `PhpParser\Node\Name` and returns whether it
resolves to the global helper (fully-qualified ⇒ yes; unqualified ⇒ yes unless a namespaced function
exists). Both call sites keep their surrounding shape.

**4. Call-shape matcher utility (scoped down).**

The issue asks for a "register a matcher" seam so a new whitelisted shape is closer to data than
plumbing, citing `abort_if` being a one-line map entry in `AbortErrorContributor` as the ideal.
`AbortErrorContributor` **already** demonstrates the data-driven pattern (`HELPER_PARAMETERS` map).
Building a *general* matcher-registration seam risks a speculative abstraction (CLAUDE.md §Simplicity;
no second consumer yet). **Recommendation: descope item 4 from this PR.** The concrete, non-
speculative wins are items 1–3 (a real bug fix + verbatim-duplication removal). A matcher-registry
seam should be its own `spec` issue, designed against ≥2 real new-shape requests, not built ahead of
demand. **This is the main scope decision for plan-review to confirm** — if the lead wants item 4
kept, it needs its own design pass and likely a human gate (new internal extension surface).

### Files

- **New:** `src/Support/MethodBody/ReturnExpressionResolver.php`; helper for item 3 (e.g.
  `src/Support/MethodBody/UnqualifiedHelperCall.php`).
- **Modified:** `SingleReturnArrayLiteralFinder`, `DataReturnExpressionReader`,
  `ReturnExpressionResourceReader`, `FormRequestStaticRulesReader` (its `collectReturns` can move to
  the shared walk too), `AbortErrorContributor`, `InlineJsonCallReader`, + the 14
  `new StatementNodeFinder()` sites (mechanical) if item 2 uses injection.

1. **Mutation-check regression, Data path** (`DataReturnExpressionReaderTest`): a fixture whose
   `index()` assigns `$data = FooData::collect(...)` then unconditionally mutates `$data`
   (reassign / `$data[] =` / `+=`) before `return $data;`. Assert `read()` returns **null**
   (refuses). **Fails on main** (stale literal attributed), **passes after**. This is the pinned bug
   test the lead requires.
2. **Resolver unit tests** (`tests/Unit/Support/MethodBody/ReturnExpressionResolverTest`): single
   unconditional assignment ⇒ resolves; conditional assignment ⇒ null; multiple assignments ⇒ null;
   dynamically-named var ⇒ null; post-assignment element write / compound assign ⇒ null;
   no matching assignment ⇒ null.
3. **Helper-shadowing guard** (item 3): keep/relocate the existing behaviour under a shared test;
   assert both `abort` and `response` unqualified-with-shadow cases refuse, fully-qualified pass.
4. **Green-preservation:** the full existing `SingleReturnArrayLiteralFinderTest`,
   `ReturnExpressionResourceReaderTest`, `ResourceClassLocatorTest`, `AbortErrorContributorTest`,
   and the inline-json/response tests must stay green (proves the consolidation is behaviour-
   preserving apart from the intended Data-path tightening).

### Observable behaviour / docs / changelog

- The **only** intended output change is the Data path refusing a mutated-variable return (was:
  stale schema; now: degrades to the honest fallback). Survey gate: `area:core` + `area:plugins`
  (SpatieData/ApiResources) ⇒ **survey required**; expect Data-path corpus deltas to be neutral or
  positive (fewer wrong schemas), and ApiResources to be unchanged (or, if the mutation-guard
  tightening there causes drops, scope it out per item 1's fallback).
- `docs/`: no user-facing surface change (all `@internal`). No page edit unless item 4 is kept
  (it isn't, per the recommendation).
- `CHANGELOG.md [Unreleased]` — "Fixed": Spatie Data response inference no longer attributes a
  stale value to a return variable that is mutated after its assignment.

### Controversial?

- No `Contracts/**`, no stage order, no config key, no public signature change — all touched classes
  are `@internal`. **Not** inherently a human gate.
- Two judgment calls for plan-review: **(a) descoping item 4** (matcher-registry seam → separate
  spec issue) and **(b) whether to tighten the ApiResources reader with the mutation guard** or
  scope that guard to the Data path. Both are called out so the reviewer can decide before coding.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -117,6 +117,7 @@ All notable changes to this project are documented here.
 - `openapi:generate` emits an advisory hint on stderr when an integration package (`league/fractal`, `spatie/laravel-fractal`, `spatie/laravel-query-builder`) is installed but its plugin is not enabled, pointing at the `config/openapi.php` line that would let it infer schemas and parameters. Advisory only: never auto-enables, never pollutes the `--output=-` document. (#444)
 - Inferred array query parameters (`ids[]` from a scalar-list validation rule) now carry `style: form` and `explode: true`, matching PHP's `name[]` repeated-pair wire format and removing the ambiguous serialisation that tripped the library's own `parameter.query-array-no-explode` lint rule. Scalar query parameters and explicit `#[QueryParam]` arrays are unchanged. (#454)
 - New `operation.action-method-missing` lint rule (level 1) flags routes whose action resolves to a controller method that does not exist (e.g. an over-registered resourceful route): the operation is documented but would fault at runtime. Skips closures and routes whose controller class is itself absent. (#448)
+- Fixed: the Spatie Data and API Resource return-expression readers no longer attribute a stale value to a returned variable that is mutated after its single unconditional assignment (`$data['k'] = …`, `$data += …`); such a body now degrades to the honest fallback, matching the array-literal finder. Consolidates the three divergent return-resolution copies onto a shared `ReturnExpressionResolver`. (#486)
 
 ## [0.1.0] - 2026-05-18
 
--- a/src/Plugins/ApiResources/Support/ReturnExpressionResourceReader.php
+++ b/src/Plugins/ApiResources/Support/ReturnExpressionResourceReader.php
@@ -9,12 +9,8 @@ use Illuminate\Database\Eloquent\Attributes\UseResource;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Http\Resources\Json\JsonResource;
 use Illuminate\Http\Resources\Json\ResourceCollection;
-use PhpParser\Node;
 use PhpParser\Node\Expr;
-use PhpParser\Node\Expr\ArrowFunction;
-use PhpParser\Node\Expr\Assign;
 use PhpParser\Node\Expr\ClassConstFetch;
-use PhpParser\Node\Expr\Closure as ClosureExpression;
 use PhpParser\Node\Expr\ConstFetch;
 use PhpParser\Node\Expr\MethodCall;
 use PhpParser\Node\Expr\New_ as NewExpression;
@@ -23,15 +19,14 @@ use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Identifier;
 use PhpParser\Node\Name;
 use PhpParser\Node\Stmt;
-use PhpParser\Node\Stmt\ClassLike;
 use PhpParser\Node\Stmt\Return_;
 use Psr\Log\LoggerInterface;
 use Psr\Log\NullLogger;
 use Radiergummi\OpenApi\Enums\PaginatorKind;
 use Radiergummi\OpenApi\Routing\ResourceTarget;
-use Radiergummi\OpenApi\Support\MethodBody\ConditionalContextPolicy;
 use Radiergummi\OpenApi\Support\MethodBody\MethodBodyScanner;
-use Radiergummi\OpenApi\Support\MethodBody\StatementNodeFinder;
+use Radiergummi\OpenApi\Support\MethodBody\ReturnExpressionResolver;
+use Radiergummi\OpenApi\Support\MethodBody\ReturnVariableRefusal;
 use Radiergummi\OpenApi\Support\PhpDoc\DocBlockParser;
 use Radiergummi\OpenApi\Support\Types\TypeNodeResolver;
 use ReflectionClass;
@@ -44,7 +39,6 @@ use function class_exists;
 use function count;
 use function in_array;
 use function is_a;
-use function is_array;
 use function is_string;
 use function method_exists;
 use function sprintf;
@@ -97,16 +91,13 @@ final class ReturnExpressionResourceReader
      */
     private array $cache = [];
 
-    private readonly StatementNodeFinder $statementNodeFinder;
-
     public function __construct(
         private readonly MethodBodyScanner $scanner,
         private readonly DocBlockParser $docBlockParser,
         private readonly TypeNodeResolver $typeNodeResolver,
         private readonly LoggerInterface $logger,
-    ) {
-        $this->statementNodeFinder = new StatementNodeFinder();
-    }
+        private readonly ReturnExpressionResolver $returnExpressionResolver = new ReturnExpressionResolver(),
+    ) {}
 
     public static function create(?LoggerInterface $logger = null): self
     {
@@ -438,13 +429,7 @@ final class ReturnExpressionResourceReader
      */
     private function methodLevelReturns(array $statements): array
     {
-        $found = [];
-
-        foreach ($statements as $statement) {
-            $this->collectMethodLevelReturns($statement, $found);
-        }
-
-        return $found;
+        return $this->returnExpressionResolver->methodLevelReturns($statements);
     }
 
     // endregion
@@ -452,37 +437,9 @@ final class ReturnExpressionResourceReader
     // region Call-shape matching
 
     /**
-     * @param list<Return_> $found
-     */
-    private function collectMethodLevelReturns(Node $node, array &$found): void
-    {
-        if (
-            $node instanceof ClosureExpression
-            || $node instanceof ArrowFunction
-            || $node instanceof ClassLike
-        ) {
-            return;
-        }
-
-        if ($node instanceof Return_) {
-            $found[] = $node;
-        }
-
-        foreach ($node->getSubNodeNames() as $subNodeName) {
-            /** @var mixed $children */
-            $children = $node->{$subNodeName};
-
-            foreach (is_array($children) ? $children : [$children] as $child) {
-                if ($child instanceof Node) {
-                    $this->collectMethodLevelReturns($child, $found);
-                }
-            }
-        }
-    }
-
-    /**
-     * Resolves `return $variable;` through the single unconditional assignment to that variable.
-     * A conditional reassignment makes the type a guess. Pass `log: false` to suppress the notice.
+     * Resolves `return $variable;` through the single unconditional assignment to that variable,
+     * refused when the variable is dynamically named, not assigned exactly once on the
+     * unconditional path, or mutated after its assignment. Pass `log: false` to suppress the notice.
      *
      * @param list<Stmt> $statements
      */
@@ -492,50 +449,35 @@ final class ReturnExpressionResourceReader
         ReflectionMethod $method,
         bool $log = true,
     ): ?Expr {
-        $variableName = $variable->name;
+        $resolution = $this->returnExpressionResolver->resolveVariable($variable, $statements);
 
-        if (!is_string($variableName)) {
-            if ($log) {
-                $this->note($method, 'returns a dynamically-named variable');
-            }
-
-            return null;
+        if ($resolution->expression !== null) {
+            return $resolution->expression;
         }
 
-        $isAssignmentToVariable = static fn(Node $node): bool
-            => $node instanceof Assign
-            && $node->var instanceof Variable
-            && $node->var->name === $variableName;
-
-        $allAssignments = $this->statementNodeFinder->findAll(
-            $statements,
-            ConditionalContextPolicy::IncludeConditionalContexts,
-            $isAssignmentToVariable,
-        );
-        $unconditionalAssignments = $this->statementNodeFinder->findAll(
-            $statements,
-            ConditionalContextPolicy::SkipConditionalContexts,
-            $isAssignmentToVariable,
-        );
-
-        if (count($allAssignments) !== 1 || count($unconditionalAssignments) !== 1) {
-            if ($log) {
-                $this->note(
-                    $method,
-                    sprintf(
-                        'returns $%s, which is not assigned exactly once on the unconditional path',
-                        $variableName,
-                    ),
-                );
-            }
-
-            return null;
+        if ($log) {
+            $this->note($method, $this->refusalReason($resolution->refusal, $variable));
         }
 
-        /** @var Assign $assignment */
-        $assignment = $unconditionalAssignments[0];
+        return null;
+    }
 
-        return $assignment->expr;
+    /**
+     * The existing per-refusal notice wording, preserved verbatim so the log surface is unchanged.
+     */
+    private function refusalReason(?ReturnVariableRefusal $refusal, Variable $variable): string
+    {
+        return match ($refusal) {
+            ReturnVariableRefusal::DynamicallyNamedVariable => 'returns a dynamically-named variable',
+            ReturnVariableRefusal::MutatedAfterAssignment => sprintf(
+                'returns $%s, which is mutated after its single unconditional assignment',
+                is_string($variable->name) ? $variable->name : '{dynamic}',
+            ),
+            default => sprintf(
+                'returns $%s, which is not assigned exactly once on the unconditional path',
+                is_string($variable->name) ? $variable->name : '{dynamic}',
+            ),
+        };
     }
 
     /**
--- a/src/Plugins/Core/ErrorContributors/AbortErrorContributor.php
+++ b/src/Plugins/Core/ErrorContributors/AbortErrorContributor.php
@@ -20,6 +20,7 @@ use Radiergummi\OpenApi\Support\MethodBody\ConditionalContextPolicy;
 use Radiergummi\OpenApi\Support\MethodBody\MethodBodyScanner;
 use Radiergummi\OpenApi\Support\MethodBody\NonLiteralValueException;
 use Radiergummi\OpenApi\Support\MethodBody\StatementNodeFinder;
+use Radiergummi\OpenApi\Support\MethodBody\UnqualifiedHelperCall;
 use ReflectionMethod;
 use Symfony\Component\HttpFoundation\Response as HttpFoundationResponse;
 use Symfony\Component\HttpKernel\Exception\HttpException;
@@ -27,7 +28,6 @@ use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
 
 use function array_key_exists;
 use function array_search;
-use function function_exists;
 use function is_int;
 use function is_string;
 use function sprintf;
@@ -118,17 +118,7 @@ final readonly class AbortErrorContributor implements ErrorResponseContributor
             return null;
         }
 
-        if ($call->name->isFullyQualified()) {
-            return $name;
-        }
-
-        $namespacedName = $call->name->getAttribute('namespacedName');
-
-        if ($namespacedName instanceof Name && function_exists($namespacedName->toString())) {
-            return null;
-        }
-
-        return $name;
+        return UnqualifiedHelperCall::resolvesToGlobalHelper($call->name) ? $name : null;
     }
 
     // endregion
--- a/src/Plugins/Core/Support/FormRequestStaticRulesReader.php
+++ b/src/Plugins/Core/Support/FormRequestStaticRulesReader.php
@@ -5,23 +5,20 @@ declare(strict_types=1);
 namespace Radiergummi\OpenApi\Plugins\Core\Support;
 
 use Illuminate\Container\Attributes\Scoped;
-use PhpParser\Node;
 use PhpParser\Node\Expr\Array_;
 use PhpParser\Node\Expr\Assign;
 use PhpParser\Node\Expr\Variable;
-use PhpParser\Node\FunctionLike;
 use PhpParser\Node\Stmt\Expression;
-use PhpParser\Node\Stmt\Return_;
 use Radiergummi\OpenApi\Support\MethodBody\AstLiteralEvaluator;
 use Radiergummi\OpenApi\Support\MethodBody\MethodBodyScanner;
 use Radiergummi\OpenApi\Support\MethodBody\NonLiteralValueException;
+use Radiergummi\OpenApi\Support\MethodBody\ReturnExpressionResolver;
 use Radiergummi\OpenApi\Support\MethodBody\RuleFieldLiteralMapper;
 use Radiergummi\OpenApi\Support\MethodBody\SingleReturnArrayLiteralFinder;
 use ReflectionMethod;
 
 use function count;
 use function in_array;
-use function is_array;
 use function is_string;
 
 /**
@@ -44,8 +41,12 @@ final readonly class FormRequestStaticRulesReader
 
     public function __construct(
         private MethodBodyScanner $scanner,
+        private ReturnExpressionResolver $returnExpressionResolver = new ReturnExpressionResolver(),
     ) {
-        $this->bareReturnFinder = new SingleReturnArrayLiteralFinder($this->scanner);
+        $this->bareReturnFinder = new SingleReturnArrayLiteralFinder(
+            $this->scanner,
+            $this->returnExpressionResolver,
+        );
     }
 
     /**
@@ -76,12 +77,7 @@ final readonly class FormRequestStaticRulesReader
             return null;
         }
 
-        /** @var list<Return_> $returns */
-        $returns = [];
-
-        foreach ($statements as $statement) {
-            $this->collectReturns($statement, $returns);
-        }
+        $returns = $this->returnExpressionResolver->methodLevelReturns($statements);
 
         if (count($returns) !== 1) {
             return null;
@@ -155,32 +151,5 @@ final readonly class FormRequestStaticRulesReader
         }
 
         return $rules === [] ? null : $rules;
-    }
-
-    /**
-     * Depth-first return collection, skipping nested function-like scopes (closures, anon classes).
-     *
-     * @param list<Return_> $returns
-     */
-    private function collectReturns(Node $node, array &$returns): void
-    {
-        if ($node instanceof FunctionLike) {
-            return;
-        }
-
-        if ($node instanceof Return_) {
-            $returns[] = $node;
-        }
-
-        foreach ($node->getSubNodeNames() as $subNodeName) {
-            /** @var mixed $children */
-            $children = $node->{$subNodeName};
-
-            foreach (is_array($children) ? $children : [$children] as $child) {
-                if ($child instanceof Node) {
-                    $this->collectReturns($child, $returns);
-                }
-            }
-        }
     }
 }
--- a/src/Plugins/Core/Support/InlineJsonCallReader.php
+++ b/src/Plugins/Core/Support/InlineJsonCallReader.php
@@ -21,9 +21,9 @@ use Radiergummi\OpenApi\Support\MethodBody\ConditionalContextPolicy;
 use Radiergummi\OpenApi\Support\MethodBody\NonLiteralValueException;
 use Radiergummi\OpenApi\Support\MethodBody\SchemaDefinitionFromLiteral;
 use Radiergummi\OpenApi\Support\MethodBody\StatementNodeFinder;
+use Radiergummi\OpenApi\Support\MethodBody\UnqualifiedHelperCall;
 
 use function array_find;
-use function function_exists;
 use function in_array;
 use function is_int;
 use function sprintf;
@@ -94,13 +94,7 @@ final readonly class InlineJsonCallReader
             return false;
         }
 
-        if ($receiver->name->isFullyQualified()) {
-            return true;
-        }
-
-        $namespacedName = $receiver->name->getAttribute('namespacedName');
-
-        return !($namespacedName instanceof Name && function_exists($namespacedName->toString()));
+        return UnqualifiedHelperCall::resolvesToGlobalHelper($receiver->name);
     }
 
     // endregion
--- a/src/Plugins/SpatieData/Support/DataReturnExpressionReader.php
+++ b/src/Plugins/SpatieData/Support/DataReturnExpressionReader.php
@@ -5,22 +5,16 @@ declare(strict_types=1);
 namespace Radiergummi\OpenApi\Plugins\SpatieData\Support;
 
 use Illuminate\Container\Attributes\Scoped;
-use PhpParser\Node;
 use PhpParser\Node\Expr;
-use PhpParser\Node\Expr\ArrowFunction;
-use PhpParser\Node\Expr\Assign;
-use PhpParser\Node\Expr\Closure as ClosureExpression;
 use PhpParser\Node\Expr\New_ as NewExpression;
 use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Identifier;
 use PhpParser\Node\Name;
 use PhpParser\Node\Stmt;
-use PhpParser\Node\Stmt\ClassLike;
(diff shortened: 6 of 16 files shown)


Radiergummi:

Two scope decisions flagged for the reviewer before coding starts:
1. **Descope item 4** (the matcher-registration seam) to a separate `spec` issue — no second consumer exists yet, building it now is speculative abstraction. The concrete wins are items 1–3 (the Data stale-mutation **bug fix** + verbatim-duplication removal).
2. **The ApiResources reader** currently lacks the mutation check too. Plan tightens it via the shared resolver, but flags a fallback: if the survey shows resource-count drops, scope the mutation guard to the Data path only. 

The pinned bug test (Data path refusing a mutated-variable return; fails on main, passes after) is specified per the lead’s note.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Reviewer: please rule on the two flagged scope decisions — (1) descoping the matcher-registration seam to a follow-up issue (lead leans agree: no second consumer = speculative); (2) whether the mutation-check guard extends to the ApiResources reader now or is scoped to the Data path pending survey evidence.

Radiergummi:
🤖 **[reviewer]** plan-review: approved with rulings on both flagged decisions + one required amendment.

Verified the latent bug against source. The divergence is real and exactly as described:
`SingleReturnArrayLiteralFinder::arrayLiteralAssignedTo` (`:94`) runs the `mutatesVariable`
unconditional-mutation guard (`:122-130`, `:143-156`); `DataReturnExpressionReader::expressionAssignedTo`
(`:173-205`) checks only the assignment counts and returns `$assignment->expr` with **no mutation
check** — so a `$data = FooData::collect(...); $data = $data->concat($bars); return $data;` body
attributes the stale `collect()` literal. `ReturnExpressionResourceReader::expressionAssignedTo`
(`:489-539`) is the same lax shape, plus `note()` logging (`:498`, `:522`). Confirmed the
helper-shadowing guard is byte-duplicated: the `namespacedName` + `function_exists` check appears
verbatim in `AbortErrorContributor::helperName` and `InlineJsonCallReader::isResponseHelperCall`.
`new StatementNodeFinder()` is in 14 `src/` classes as claimed, and it's `final readonly` stateless.

**Ruling on (a) — descope item 4 (matcher-registration seam): AGREE, descope it.** No second
consumer exists; `AbortErrorContributor`'s `HELPER_PARAMETERS` map already demonstrates the
data-driven pattern the issue idealizes. Building a general "register a matcher" seam now is a
speculative internal extension surface — against CLAUDE.md §Simplicity and the project's
"seams need a real consumer / no speculative DI" convention. File it as its own `spec` issue, to be
designed against ≥2 concrete new-shape requests. Items 1–3 (the real bug fix + verbatim-dup removal)
are the non-speculative wins and stand on their own.

**Ruling on (b) — apply the mutation guard to the ApiResources reader too, now (not Data-only).**
Reason: the whole point of #486 is to kill the three-way divergence in this exact dance; leaving
ApiResources lax reintroduces the same class of stale-value bug the PR exists to fix, just in a
different plugin. A mutated variable's base literal is as wrong for a resource `toArray()` as it is
for a Data return. Tightening it is a correctness improvement, not a regression. **Keep the plan's
survey-driven fallback as the safety valve:** if the survey shows an ApiResources resource-count
*drop* that traces to this guard, scope the guard to the Data path for this PR and record the
decision + the specific corpus delta as a PR comment, then file the ApiResources tightening as a
follow-up. That gives the honest-correctness default while the survey gate (area:core + area:plugins)
protects against an unexpected corpus hit. Note for the surveyor: a *drop in wrong/stale schemas* is
the expected, desirable delta here — distinguish "fewer schemas because we stopped emitting a stale
one" (good) from "fewer schemas because we now refuse a case we used to handle correctly" (bad); only
the latter is a regression.

**Required amendment:**

1. **Preserve the ApiResources log wording exactly, and keep the resolver log-free.** The plan's
   "thin logging wrapper in the reader, resolver stays log-free" is the right call (Support layer must
   not carry plugin/log policy). Make the wrapper reproduce **both** existing notices byte-for-byte —
   the dynamically-named-variable notice (`:499`) and the "not assigned exactly once on the
   unconditional path" notice (`:526`) — and add a notice for the **new** mutation-refusal branch so a
   consumer isn't left with a silent null where they previously got a schema. Assert the exact
   messages in the test (item 3), and keep `ReturnExpressionResourceReaderTest` green to prove the
   wrapper is behaviour-preserving apart from the intended tightening.

Everything else is sound: the mutation guard must be lifted to the *general* `?Expr` case (not just
`Array_`) since the lax readers return `new X()`/`X::collect()`, and the shared walk skipping
`FunctionLike OR ClassLike` as a superset of both existing skip-predicates is correct (equivalence
proven by keeping every per-consumer test green). Leaving the 14 `new StatementNodeFinder()` defaults
as harmless stateless defaults (the #485 arch rule is `MethodBodyScanner`-specific) is fine — no need
to widen churn. The pinned mutation-regression test (fails on main, passes after) is the load-bearing
artifact; make sure it lands first. CHANGELOG "Fixed" line and the survey requirement are correctly
identified.

No blocking concerns. Approved to proceed: descope item 4, apply the guard to both paths with the
survey fallback, and fold in amendment 1.

Radiergummi:
🤖 **[lead]** plan-review approved → coding. Rulings: (a) item 4 (matcher-registration seam) descoped to a follow-up issue; (b) mutation guard applies to the ApiResources reader too, with the survey fallback (scope to Data-only + follow-up issue only if the survey shows a traceable resource-count drop). Amendment: ApiResources logging wrapper reproduces both existing notices byte-for-byte + adds one for the mutation-refusal branch, exact-message assertions; resolver stays log-free.

Radiergummi:
🤖 **[coder]** implemented; local gates green (Pint, PHPStan level 8, 3319 tests). Pushed 94164314. Ready for review.

**What landed (items 1–3; item 4 descoped per plan-review):**

1. **Shared `Support\MethodBody\ReturnExpressionResolver`** (`@internal readonly`, log-free) owning both hand-rolled concerns:
   - `methodLevelReturns()` — the method's own returns, skipping nested scopes via `FunctionLike || ClassLike` (superset of both prior skip-predicates).
   - `resolveVariable()` — returns a `ReturnVariableResolution` (`?Expr` + `?ReturnVariableRefusal`), refusing dynamically-named / not-assigned-once / **mutated-after-assignment**. The mutation guard (`$v['k'] = …`, `$v += …`) is hoisted verbatim from `SingleReturnArrayLiteralFinder`.
2. **The latent bug is fixed on both plugin paths.** `DataReturnExpressionReader` and `ReturnExpressionResourceReader` inherit the mutation guard they previously lacked. Pinned regression test lands first: `DataReturnExpressionReaderTest` → `elementWriteAfterAssignment` (fails on parent, passes now).
3. **Consumers refactored, duplicates deleted:** `SingleReturnArrayLiteralFinder`, `DataReturnExpressionReader`, `ReturnExpressionResourceReader`, and `FormRequestStaticRulesReader` (its `collectReturns` copy removed) all delegate to the resolver.
4. **Helper-shadowing guard** extracted to `Support\MethodBody\UnqualifiedHelperCall::resolvesToGlobalHelper(Name)`; `AbortErrorContributor` + `InlineJsonCallReader` both call it.

**Amendment satisfied:** the ApiResources logging wrapper (`expressionAssignedTo` → `refusalReason`) reproduces both existing notices **byte-for-byte** (dynamically-named-variable; not-assigned-once) and adds one for the mutation-refusal branch. Exact-message assertions in `ReturnExpressionResourceReaderTest`: `conditionallyAssignedVariable` (not-assigned-once), `dynamicallyNamedVariable` (dynamic), `mutatedAfterAssignment` (mutation). The resolver itself stays log-free.

**New tests:** `ReturnExpressionResolverTest` (8 cases: single/conditional/multiple/element-write/compound-assign/no-match/dynamic/closure-not-descended), `UnqualifiedHelperCallTest` (4: fully-qualified, unqualified-no-shadow, unqualified-shadowed, no-namespacedName-attribute), plus the three ApiResources exact-message tests and the pinned Data bug test.

**Deviations / notes:**
- **Mutation-guard example nuance:** the plan's example `$data = $data->concat($bars)` is a plain *reassignment* → already refused by the assignment-count check in both readers on parent. The genuine gap the guard closes is **element writes and compound assigns**, which are not counted as assignments-to-variable. The pinned tests use those forms.
- **survey fallback:** the guard is applied to the ApiResources path too (reviewer ruling b). If the survey shows a traceable ApiResources resource-count *drop* attributable to this guard, I'll scope it to the Data path + record the corpus delta here.
- **Item 4 (matcher-registration seam):** descoped per plan-review. I attempted to file the follow-up `spec` issue but issue-creation was blocked in this session — flagging to the lead to queue it. Draft title: *"spec: call-shape matcher-registration seam for Tier-1 body scanners"* — design against ≥2 concrete new-shape requests; may close as won't-fix if the per-reader data-driven maps already absorb them.

Survey required (`area:core` + `area:plugins`: SpatieData/ApiResources). Expected delta: fewer wrong/stale schemas (good), not fewer correctly-handled cases (bad).

Radiergummi:
🤖 **[lead]** implementation pushed (94164314), local gates green → review. Round 1: reviewer-1. Descoped item 4 is tracked as #512. Survey to follow review; ApiResources-drop fallback stands as agreed.

Radiergummi:
🤖 **[reviewer]** review round 1: approved, no blocking concerns.

I re-ran the suite locally against 94164314 (applied to a clean main), including deliberate
red-before checks on both plugin paths. The consolidation is behaviour-preserving, the latent bug is
genuinely fixed, and both plan-review rulings landed.

**The pinned bug — verified red-before/green-after.** `DataReturnExpressionReaderTest` →
`elementWriteAfterAssignment` (`$data = ScalarOnlyData::collect(...); $data['extra'] = …; return
$data;`) **fails on main's reader** (the stale `collect()` value is attributed) and **passes with the
fix**. The coder's nuance note is correct and important: a plain reassignment
(`$data = $data->concat(...)`) was already refused by the assignment-count check, so the genuine gap
is **element writes and compound assigns**, which are not counted as assignments-to-variable — and
that is exactly what both the Data and the ApiResources pinned tests exercise.

**Mutation-guard semantics — verbatim, no widening/narrowing.** `ReturnExpressionResolver::
mutatesVariable` is byte-identical to the `SingleReturnArrayLiteralFinder` original: `AssignOp` (any
compound assign) or `Assign` whose target is an `ArrayDimFetch` rooted at the variable, walking
nested `ArrayDimFetch` to the base. Same `SkipConditionalContexts` policy for the mutation scan, so a
conditionally-guarded write still doesn't refuse. Coverage matches the array-literal finder exactly.

**Both plugin paths get the guard (ruling b) — verified.** With main's `ReturnExpressionResource
Reader` restored (guard absent), the ApiResources `mutatedAfterAssignment` test **fails** while the
other 29 cases pass; with the fix all pass. So the guard is genuinely applied to ApiResources, and
the pre-existing behaviour is otherwise untouched.

**Amendment (ApiResources logging wrapper) — satisfied.** `refusalReason()` maps the resolver's
`ReturnVariableRefusal` enum back onto the notices: `DynamicallyNamedVariable` →
`'returns a dynamically-named variable'` and `NotAssignedOnce` (via `default`) →
`'returns $%s, which is not assigned exactly once on the unconditional path'` — both **byte-identical**
to the pre-refactor wording — plus a new `MutatedAfterAssignment` notice. The three exact-message
tests assert the full composed notice strings (`conditionallyAssignedVariable`,
`dynamicallyNamedVariable`, `mutatedAfterAssignment`). The shared resolver is log-free, keeping log
policy out of the Support layer as the plan required. Minor robustness note (non-blocking): the
`default` arm also covers a `null` refusal, which is unreachable here (a null `expression` always
carries a refusal) — harmless, and it keeps the match total.

**Item 4 stayed descoped — confirmed.** No matcher-registration seam was introduced. The
`UnqualifiedHelperCall::resolvesToGlobalHelper(Name)` extraction is behaviour-identical at both call
sites: `AbortErrorContributor::helperName` and `InlineJsonCallReader::isResponseHelperCall` now
delegate, and the extracted logic (fully-qualified ⇒ true; unqualified ⇒ true unless a same-namespace
function shadows it, via the `namespacedName` attribute) matches the originals line-for-line.

**Consolidation is clean.** All four hand-rolled copies deleted (`SingleReturnArrayLiteralFinder`,
`DataReturnExpressionReader`, `ReturnExpressionResourceReader`, `FormRequestStaticRulesReader`), each
delegating to the resolver. `SingleReturnArrayLiteralFinder` narrows the resolver result to `Array_`,
preserving its exact contract (it already had the guard, so zero behaviour change). The shared
skip-predicate is `FunctionLike || ClassLike` (superset of the two prior predicates); the
`FormRequestStaticRulesReaderTest` and the array-literal finder tests stay green, confirming the
widening is behaviour-preserving at method-body top level as the plan claimed.

**Conventions.** strict-types headers present; `@internal` on the resolver, both VOs, and the helper;
modern PHP (readonly, private-ctor factory VOs, backed-enum refusal reasons, `match`); no
abbreviations; comments concise and why-focused with no issue numbers in code. `Support\MethodBody`
placement is correct — plugin-agnostic, no plugin/third-party dependency, no log policy.

**Tests / gates — all green locally.**
- `tests/Unit/Support/MethodBody`, both plugin reader tests, `tests/Unit/Core`: 424 passed, 937
  assertions. `ReturnExpressionResolverTest` has all 8 cases (single / conditional / multiple /
  element-write / compound-assign / no-match / dynamic / closure-not-descended);
  `UnqualifiedHelperCallTest` has 4 (fully-qualified, unqualified-no-shadow, shadowed, no-attribute).
- PHPStan (full, level 8): no errors. Pint `--test` on all changed files: passed.

**Test-suite gap check (standing duty).** The refused branches (dynamic / not-assigned-once /
mutated), the resolved branch, the nested-scope skip, and both helper-shadowing outcomes are all
covered, on both plugin paths where they surface, with exact-message assertions where a log is
consumer-observable. Nothing material left uncovered.

**Bookkeeping.** CHANGELOG `[Unreleased]` "Fixed" entry present and accurate (names both paths, both
mutation forms, the honest-fallback outcome, the consolidation) (#486). Docs: all touched classes
`@internal`; no user-facing surface, no page edit needed. No new lint rule → no catalog row.

**To relay to the lead (not a code issue):** the coder reported item-4 follow-up issue creation was
blocked this session — the `spec` issue *"call-shape matcher-registration seam for Tier-1 body
scanners"* still needs to be filed so the descope is tracked.

Approved. `area:core` + `area:plugins` (SpatieData/ApiResources) → survey required before merge.
Expected delta per ruling (b): **fewer wrong/stale schemas** (good), not fewer correctly-handled
cases; if the survey shows an ApiResources resource-count drop traceable to this guard, the coder's
documented fallback is to scope the guard to the Data path and record the corpus delta.

Radiergummi:
🤖 **[lead]** review round 1 approved (reviewer-1). Item-4 descope follow-up is tracked as #512. Queued for the independent second review (reviewer-2), then survey (with the fewer-stale-schemas vs. genuine-regression distinction flagged).

Radiergummi:
🤖 **[reviewer]** independent review (reviewer-2): approved, no concerns.

Formed my own view from #486, the plan (incl. plan-review rulings), and the diff at `94164314` before reconciling with reviewer-1; we converge on every load-bearing point.

**Latent bug — verified red-before/green-after myself.** I reverted *both* readers to their `main` versions in the branch worktree (keeping the new tests) and re-ran the two mutation cases: `DataReturnExpressionReaderTest::elementWriteAfterAssignment` and `ReturnExpressionResourceReaderTest::mutatedAfterAssignment` **fail on main** and pass with the fix. The failure output makes the bug concrete: main attributes the stale `NestedAuthorResource::collection(...)` value and emits a `ResourceTarget(isCollection: true)` for a variable the code has since mutated. Confirmed the coder's/reviewer-1's nuance: a plain reassignment was already refused by the assignment-count check, so the genuine gap is **element writes (`$v['k']=…`) and compound assigns (`$v+=…`)** — exactly what the pinned tests exercise.

**Mutation guard — byte-identical, no widening/narrowing.** `ReturnExpressionResolver::mutatesVariable` matches the deleted `SingleReturnArrayLiteralFinder` original line-for-line (`AssignOp`, or `Assign` whose target is an `ArrayDimFetch` rooted at the variable, walking nested dims to the base). The mutation scan uses `SkipConditionalContexts`, so a conditionally-guarded write still does not refuse — the base value stays a never-wrong subset. Same policy as before.

**Return-collection consolidation — behaviour-preserving.** The shared skip-predicate is `FunctionLike || ClassLike` (superset of the two prior predicates: `FunctionLike` vs `Closure|ArrowFunction|ClassLike`), equivalent at method-body top level. All pre-existing `SingleReturnArrayLiteralFinderTest` (65 total in that file set) and `FormRequestStaticRulesReaderTest` stay green, and the new closure-not-descended resolver test passes.

**ApiResources logging wrapper — notices byte-identical.** Checked `refusalReason()` against `main`'s inline `note()` strings: `DynamicallyNamedVariable → 'returns a dynamically-named variable'` and `NotAssignedOnce (default) → 'returns $%s, which is not assigned exactly once on the unconditional path'` are verbatim; the new `MutatedAfterAssignment` notice is covered by an exact-message test. The three composed-notice tests assert the full strings. Resolver stays log-free (Support-layer convention upheld). Independently spotted the same minor non-blocking point reviewer-1 raised: the `default`/`{dynamic}` fallback also covers the unreachable `null`-refusal + non-string-name case — dead but harmless, keeps the `match` total.

**Helper-shadowing extraction — verbatim.** `UnqualifiedHelperCall::resolvesToGlobalHelper(Name)` reproduces both deleted guards exactly (FQ ⇒ true; unqualified ⇒ true unless a same-namespace function shadows via the `namespacedName` attribute). `AbortErrorContributorTest` + `InlineJson*` suites (101 tests) pass unchanged.

**Scope / deviations — all sanctioned.** Item 4 (matcher-registration seam) descoped → tracked as #512. ApiResources mutation guard applied to *both* paths per plan-review ruling (b), with the documented survey fallback (scope to Data-only + record delta if a traceable ApiResources resource-count drop appears). Item 2 (share one `StatementNodeFinder` across 14 sites) is only partially applied — the 3 readers that fold into the resolver lose their local `new StatementNodeFinder()`; 12 remain — which plan-review explicitly ruled acceptable (the finder is stateless `final readonly`; the #485 anti-pattern is `MethodBodyScanner`-specific). Confirmed `ReturnExpressionResolver` is itself stateless, so its `= new …()` ctor defaults are harmless.

**Conventions / bookkeeping.** strict-types + `@internal` on the resolver, both VOs, and the helper; modern PHP (readonly, private-ctor factory VOs, backed enum, `match`); no abbreviations; comments why-focused, no issue numbers in code; `Support\MethodBody` placement is plugin-agnostic and log-free. CHANGELOG `[Unreleased]` "Fixed" entry accurate; all touched classes `@internal` → no docs page, no lint-catalog row.

**Gates (branch worktree, `94164314`).** Pint `--test` passed; PHPStan level 8 no errors; 166 relevant tests pass across resolver/helper/both-plugin/error-contributor/ResourceClassLocator suites.

Second independent approval. `area:core` + `area:plugins` → survey required before merge; expected delta is **fewer wrong/stale schemas**, not fewer correctly-handled cases (the ApiResources-drop fallback stands).

Radiergummi:
🤖 **[lead]** two independent approvals → survey. Expected delta: fewer wrong/stale schemas (the readers now refuse mutated return variables) — that is acceptable and must be itemized. A drop in cases we previously handled RIGHT (esp. ApiResources resource counts) is a regression and triggers the documented Data-path-only fallback.

Radiergummi:
🤖 **[coder]** rebased onto main (post-#510 merge); all remote CI green. Pushed 5dd61935.

**Conflict resolved** in `FormRequestStaticRulesReader` ctor — exactly the predicted territory. #510 dropped the `= new MethodBodyScanner()` ctor default (arch rule `no src class defaults a constructor parameter to a new MethodBodyScanner`); my side added the `ReturnExpressionResolver` param. Kept **both** intents: `$scanner` is now required (autowires the one `#[Scoped]` scanner, no cache bypass), and `$returnExpressionResolver` keeps its harmless `= new ReturnExpressionResolver()` default — the resolver is stateless (only a `StatementNodeFinder`, no scanner, no cache), so it cannot reintroduce the scoped-cache bypass, and it does not trip the arch rule. The other two readers (Data, ApiResources) already had `$scanner` required in my original commit, so they rebased clean. CHANGELOG auto-merged keeping both entries (#510 at the Fixed subsection, #486 at end of section).

**Local gate** (full `composer check`): Pint passed, PHPStan level 8 clean, **3326 tests** (was 3319 + #510's 7 new pinned scanner tests, now compatible with my change). The `no src class defaults ... new MethodBodyScanner` arch test passes.

**Remote CI:** all green on 5dd61935 — quality, coverage, snapshot, PHP 8.4/8.5 × Laravel 12/13, and the swagger-php 5.8 job.

Prior review verdict + the surveyor's pending verdict (ran against pre-rebase 94164314) remain content-valid: the rebase changed only the ctor conflict resolution, no behavioural change to the readers.

Radiergummi:
🤖 **[surveyor]** Layer A pass — 0 deltas across 11 apps (byte-neutral on this corpus).

**Note on run validity:** my first candidate run hit unresolved git conflict markers in `FormRequestStaticRulesReader.php` (leftover from the mid-rebase state after #510 merged) — a PHP parse error that made the library fail to load and crashed Koel + Lychee. That run was **discarded**. This result is the clean re-run against the current PR head **5dd61935** (working tree clean, no conflict markers, rebased on current main d421601f incl. merged #510).

**Comparison** (fresh `main` baseline 417ee8eb, proven byte-neutral to d421601f by the #510 gate):

- **Aggregate metrics: identical on all 11 apps** — paths / operations / apiOperations / responseSchemas / documentedResponses / requestBodies / maxRequestProperties / componentSchemas / completenessPercent unchanged. All booted, `generateExit=0`, no crashes.
- **Semantic specs: identical on all 11 apps** after masking per-boot runtime tokens (Livewire URL hash, Laravel `generated::<random>` route names).

**Classification:** no deltas to classify. The corpus contains no route where the Data/ApiResources readers were previously attributing a *mutated-after-assignment* return variable, so the tightened refusal (element writes / compound assigns) changed no output here. That means neither the acceptable case (a) "stale/wrong schema no longer emitted" nor the regression case (b) "previously-correct case lost / ApiResources resource-count drop" occurred in this corpus. Resource counts (componentSchemas, responseSchemas) are unchanged on every ApiResources-using app (Lychee 128/248, Koel 56/93, InvoiceNinja 69/238, Vito 114/87 — all equal to baseline).

The bug fix is covered by the new unit tests in the PR (`DataReturnExpressionReaderTest`, `ReturnExpressionResolverTest`); this corpus simply does not exercise that shape at runtime. Clear to merge.

Radiergummi:
🤖 **[lead]** all gates green (2 independent approvals, full CI matrix on rebased head, survey Layer A byte-neutral — corpus doesn't exercise the mutated-return shape at runtime; the bug fix is pinned by unit tests) + non-controversial per run directive (internal-only, no public contract) → squash-merged. Closes #486; worktree removed. Follow-up #512 tracks the descoped matcher seam.

Issue #481, opened 2026-07-01
feat(core): shared return-shape descriptor service (container kind, item type, nullability, unions)


Radiergummi:
## Problem

Every response-side plugin independently re-answers the same structural questions, each hitting the same wall (reflection erases generics) and each reinventing the reflect → `@return`-generic → AST-scan fallback:

- collection-vs-single detection in four places (`PaginatorResponseResolver:129`, `ResourceClassLocator::isCollectionType:133`, `DataResponseResolver:91` + `SchemaFromDataClass:426`, `FractalResponseResolver:83`),
- paginator detection in three,
- union→oneOf twice — hand-rolled on raw reflection in `DataResponseResolver::resolveUnion:173` vs proper in `JsonSchemaFromType:60`,
- nullability via three different mechanisms.

## Proposed change

One shared **return-shape descriptor service** (Support, plugin-agnostic): ingests a `Reflector`, merges signature + `@return` docblock, and returns a normalized descriptor:

```
{ container: single|list|paginated|cursor, itemType: Type, nullable: bool, unionMembers: list<Type> }
```

Plugins then answer only their plugin-specific question ("is this item class one *I* own?") instead of re-deriving structure. `ReturnTypeExtractor`'s generic-unwrap logic becomes part of this service instead of being wired into only 2 callers.

## Why

Union support, generic unwrapping (`MyWrapper<UserResource>`), and nullability become uniform capabilities every plugin gets for free, instead of per-plugin accidents. Still strictly Tier 0/1: signature + docblock only; the semantic reads (toArray() AST, transform(), cast semantics) correctly stay in plugin code. Boundary rule: **structural type shape = shared; "what does this convention mean" = plugin.**

Radiergummi:
🤖 **[lead]** admitted → planning. Type-resolution chain resumed: #480 merged (PR #513), this is the next link; #479 follows. Survey gate required before merge.

Issue #479, opened 2026-07-01
feat(responses): baseline typed-return response resolver — typed DTOs, array shapes, typed collections


Radiergummi:
## Problem

A statically-typed controller whose return is a plain typed DTO, a typed collection of DTOs, or a documented `@return array{id: int}` gets **no response schema** — just the default `200 OK` (`OperationBuilder.php:229`). Every registered `PrimaryResponseResolver` is convention-shaped: Core's paginator/Eloquent/find-scan/inline-JSON resolvers, plus Resource/Data/Fractal per plugin. A typed return that matches none of them falls through entirely.

Array-shape resolution *exists* (`Support\Types\TypeNodeToSchema` handles `ArrayShapeNode`, `array<K,V>`, `list<T>`) but only runs as a leaf inside other builders (Eloquent `@property`, Data properties). There is no top-level "typed return → schema" path.

## Proposed change

Add a **baseline** `TypedReturnResponseResolver` (registered by `BaselineRegistration`, *not* by Core — a typed return is language-level, not a Laravel convention, so it must work with Core disabled). Registered last in the primary-response chain so convention plugins keep winning where they apply. Coverage:

- plain typed DTO return → component schema via the unified type engine (see engine-unification issue),
- `@return array{...}` / `array<K,V>` / `list<T>` docblock shapes,
- typed collections of DTOs (generic unwrap via the return-shape descriptor service).

Degrades to `null` exactly like existing resolvers — no invented schemas.

## Why

Tier 0 by definition (signatures + PHPDoc). This is the single highest-leverage coverage fix for the priority segment: the user who typed their returns and today gets nothing. Directly serves the "well-typed apps → almost automatic" positioning.

## Dependencies

Best built on top of the docblock-engine unification and the return-shape descriptor service; can start with the DTO + array-shape subset independently.

PR #513, opened 2026-07-01
refactor(core): unify the two type→schema engines on symfony/type-info Type

### LINKED ISSUE #480: refactor(core): unify TypeNodeToSchema and JsonSchemaFromType on one canonical type representation
## Problem

Types flow through **four representations** — `ReflectionType` → phpstan `TypeNode` → symfony/type-info `Type` → `OA\Schema` — with lossy bridges, and two parallel docblock→schema engines that self-describe as "companions":

- `Support\Types\TypeNodeToSchema` (phpstan type nodes; handles array shapes; **drops enum cases**),
- `Support\Generator\JsonSchemaFromType` (symfony `Type` tree; carries enums; no array shapes).

The scalar-keyword→JSON-type map is **triplicated** (`TypeNodeToSchema::scalarDefinition()` :292, `EloquentModelToSchema::scalarKeywordToDefinition()` :465, `JsonSchemaFromType::fromBuiltinType()` :230), and date/uuid formatting is duplicated. `TypeNodeResolver::resolveClass()` round-trips phpstan node → symfony Type → class-string, discarding everything but the FQCN.

## Proposed change

Make **symfony/type-info `Type` the canonical internal representation** (it already carries enum cases and is where FQCN resolution terminates). Teach the symfony-side path array shapes, collapse `TypeNodeToSchema` into `JsonSchemaFromType` (one engine), and delete the duplicated scalar/format maps. phpstan/phpdoc-parser stays as the *parser*; its nodes convert to symfony `Type` once, at the boundary.

## Why

Removes a whole class of divergence bugs (enum cases surviving through one engine but not the other), deletes code net, and is the substrate the typed-return resolver and return-shape descriptor sit on. Pure Tier-0 infrastructure; no behavior change intended except where the two engines currently disagree — those become test cases.

Radiergummi:
Closes #480

## Implementation plan — #480 (amended per plan-review)

**Tier: 0** (signatures + PHPDoc; no body parsing). Pure type-representation infrastructure.
**Behavior-preserving by mandate:** every place the two engines disagree today becomes a **pinned
characterization test**, not a silent pick. No new output except at the enumerated divergence
points, each a deliberate, reviewed decision.

> **Merge routing: this PR goes to `needs-human` after review + a mandatory survey — NOT auto-merge.**
> Confirmed by lead + reviewer. It touches no `Contracts/**`, but it (a) is the **head of the
> #480 → #481 → #479 chain**, (b) **deletes an internal engine**, (c) adds a callback to a
> **widely-consumed signature**, and (d) has **intended additive output deltas** at the divergence
> points. Moritz signs off on the two irreversible decisions — `Type`-as-canonical and the
> divergence-catalog verdicts — before merge. Flow: coder lands → reviewer + **survey Layer A
> (0 regressions; additions itemized)** → lead routes to `needs-human`.

### What was verified during planning (grounds the whole plan)

- **Two engines, complementary gaps** (read both in full):
  - `Support\Types\TypeNodeToSchema` (phpstan `TypeNode`): handles **array shapes** (`array{…}`),
    `array<K,V>` maps, `list<T>`/`T[]`, collection generics (`Collection<T>`), `true`/`false` →
    boolean; delegates leaf classes to a caller `callable(string $fqcn): ?OA\Schema`. **Does not
    carry enum cases**, no DateTime/UUID/UrlRoutable formatting, and **returns `null` on unmapped**
    (caller degrades).
  - `Support\Generator\JsonSchemaFromType` (symfony `Type`): handles **backed-enum → `$ref`
    component with case descriptions**, unit-enum note, **union → `oneOf`**, nullable wrapping,
    DateTime/UUID/UrlRoutable formats, builtin scalars + `array`→`{type:array,items:{}}`. **No array
    shapes, no map/list element typing, no collection generics**, and **emits a stub
    `{type:string, description:"Unmapped…"}` schema** (not null) on unmapped types.
- **Decisive finding: symfony/type-info already parses every structural shape the phpstan path
  handles.** `StringTypeResolver` yields `ArrayShapeType` for `array{id:int}`, `CollectionType`
  (with `isList()`, `getCollectionKeyType()`, `getCollectionValueType()`) for `list<T>` / `array<K,V>`
  / `T[]`, and `BuiltinType` for `true`/`false`. So unification is a **schema-mapping gap in
  `JsonSchemaFromType`** (missing `ArrayShapeType`, `CollectionType`, `GenericType`, `ObjectShapeType`,
  `IntersectionType`, `TemplateType` cases), **not a parser rewrite**. This is what makes symfony
  `Type` the viable canonical representation.
- **Boundary constraint:** resolving a docblock string to a symfony `Type` needs a **`TypeContext`**
  for relative/imported class names — `list<App\User>` **throws** under a bare `StringTypeResolver`,
  but resolves under the reflection-derived context that `TypeNodeResolver::resolveClass()` already
  builds (with a context-free retry). The phpstan→symfony bridge therefore routes through
  `TypeNodeResolver`, **not** a naked `StringTypeResolver->resolve((string) $node)`.
- **Triplicated scalar map confirmed:** `TypeNodeToSchema::scalarDefinition()` (:292),
  `EloquentModelToSchema::scalarKeywordToDefinition()` (:465), `JsonSchemaFromType::fromBuiltinType()`
  (:228). The first two include `true`/`false`; the third does not (they hit `unmappedBuiltin`).

### Version-floor requirement (AMENDMENT 1 — most important; do not skip)

`composer.json` allows **`symfony/type-info: ^7.3 || ^8.0`** — the locked-local 8.1 is **not** the
floor. Every accessor this PR leans on (`ArrayShapeType`, `CollectionType::isList()/getCollectionKeyType()
/getCollectionValueType()`, `ArrayShapeType::getShape()/getExtraKeyType()/getExtraValueType()`,
`GenericType`) must exist and behave identically on **7.3**. The characterization harness (or at
minimum its structural cases) **must run against the 7.x floor**, not just the locked version — this
is exactly the kind of API drift that passes locally and fails a CI leg. If 7.3 lacks or differs on
any accessor, tighten the composer constraint in this PR and record why. Add a CI note / harness
matrix entry for the floor.

### Step 0 (MANDATORY, precedes any consolidation): characterization harness + divergence catalog

Feed a shared corpus of type strings through **both** engines and snapshot each serialized
`OA\Schema`. The harness output is the **divergence catalog**, attached as a PR comment. Corpus
covers the union of both engines' input space and every catalog class below.

**Divergence catalog — all classes the harness MUST cover and pin (canonical output chosen per
entry). Classes 1–6 pre-identified; 7–13 added in plan-review.**

1. **Enum cases.** phpstan path routes a leaf enum through the `classSchema` callback; symfony path:
   backed enum → `$ref` component with `enum:[…]` + case descriptions. Canonical keeps the
   case-bearing behavior (the headline bug: "enum cases surviving one engine but not the other").
2. **`true`/`false` pseudo-types.** phpstan → `{type:boolean}`; symfony `fromBuiltinType()` does
   **not** map them (hits `unmappedBuiltin`). Canonical: `{type:boolean}`.
3. **Object formatting (DateTime/UUID/UrlRoutable).** Only the symfony path formats these; phpstan
   routes them through the `classSchema` callback. Canonical treatment pinned against the leaf-callback
   precedence (see below).
4. **Map element typing.** phpstan types `array<string,T>` as `{type:object, additionalProperties:<T>}`;
   symfony path has no map handling today.
5. **Unmapped fallback — NULL vs STUB (behavioral, not cosmetic).** `TypeNodeToSchema` returns
   **`null`** on an unmapped/unresolvable type (caller degrades to its own fallback);
   `JsonSchemaFromType` emits a **stub `{type:string, description:"Unmapped…"}`** schema. This
   materially changes what a caller like `EloquentModelToSchema` does on an unresolvable leaf. Pin
   the canonical choice **and enumerate which callers relied on the `null`-degrade** (audit every
   `TypeNodeToSchema` consumer + the 6 `JsonSchemaFromType` consumers). Survey must confirm the chosen
   behavior is net-neutral/additive, not a regression.
6. **Nullable double-wrap.** Both use `NullableSchema::wrap`; ensure the unified engine wraps exactly
   once and audit every caller for a now-redundant wrap.
7. **Multi-class union `A|B` (no null).** `TypeNodeToSchema::resolveNullable` (:107-127) returns
   **null** for a non-nullable multi-class union; `JsonSchemaFromType` emits **`oneOf`** (:60-67).
   Divergence — pin canonical (`oneOf`, matching the symfony path) and flag for survey.
8. **`array<string, mixed>`.** Map with `additionalProperties: true` (the Spatie Data parity case).
   Pin.
9. **`iterable<T>`.** Resolves via the array-like generic path; pin its list mapping.
10. **`ObjectShapeType` / `IntersectionType` / `TemplateType`.** All exist as `Type` subclasses;
    neither engine models them today. Route to the chosen unmapped fallback — the harness must
    **assert each lands on the fallback rather than crashing** (no unhandled `Type` subclass).
11. **`self` / `static` and generics-in-collections.** Resolve to a class name via the **context-aware
    boundary**. Add a harness case proving the `TypeContext` path actually resolves imported/relative
    names (this is the one place the boundary can throw — the `list<App\User>`-without-context risk).
12. **`ArrayShapeType` property ordering (ksort).** Verified in vendor: `ArrayShapeType`'s constructor
    **`ksort`s** its shape, so `array{b:int, a:string}` yields properties in **sorted** order —
    whereas `TypeNodeToSchema::objectFromShape` (:140-150) iterates in **source order**. Real
    divergence. Canonical = the sorted symfony outcome, but it **changes today's phpstan output** —
    pin it and **flag explicitly for the survey** to confirm net-neutral.
13. **(covered above)** unmapped-fallback wording — the description string differs; pick one, pin it
    (subsumed by class 5's behavioral decision).

Each catalog entry becomes a named test asserting the **chosen** canonical output. Any entry that
changes a real consumer's emitted schema is called out in the PR body and must pass the survey — a
behavior-preserving PR shows **0 regressions**; intended changes are *additions* (previously-unmapped
shape now mapped) or the explicitly-flagged ordering/fallback decisions, never silent regressions.

### Consolidation design

**Canonical representation: symfony/type-info `Type`.** One engine: `JsonSchemaFromType` grows the
missing structural cases; `TypeNodeToSchema` is deleted.

1. **Boundary: phpstan `TypeNode` → symfony `Type`.** Add a single conversion entry point (extend
   `TypeNodeResolver`, which already owns phpstan→symfony resolution) —
   `toType(TypeNode $node, Reflector $context): ?Type` resolving via the **context-aware**
   `StringTypeResolver` path (reusing `contextFor()` + the context-free retry). This is the *only*
   phpstan→symfony bridge; everything downstream is symfony `Type`.

2. **Teach `JsonSchemaFromType` the structural cases** (translating `TypeNodeToSchema`'s logic):
   - **instanceof ordering (AMENDMENT 4):** `ArrayShapeType extends CollectionType`, so `fromType()`
     **must test `ArrayShapeType` before `CollectionType`** — the same discipline as the existing
     `NullableType`-before-`UnionType` / `BackedEnum`-before-`Object` rules the class documents
     (:42-43). Add the new checks in the correct order.
   - **`ArrayShapeType` → object schema (AMENDMENT 3 — corrected API):** `getShape()` returns
     `array<key, array{type: Type, optional: bool}>` **keyed by the shape key** (NOT a `list` of
     `ArrayShapeTypeItem` objects — that name does not exist and would not compile). Iterate
     `foreach (getShape() as $key => ['type' => $t, 'optional' => $opt])`, build `properties` +
     `required`; honor `getExtraKeyType()/getExtraValueType()` for the open-shape /
     `additionalProperties` case (`isSealed()` distinguishes). **Ordering is already `ksort`ed by the
     constructor** (class 12) — do not attempt to restore source order.
   - **`CollectionType`:** `isList()` → `{type:array, items:<value schema>}`; keyed (`!isList()`) with
     a string key → `{type:object, additionalProperties:<value schema>}` (map), `mixed` value →
     `additionalProperties: true` (class 8); integer key → list. Use `getCollectionKeyType()` /
     `getCollectionValueType()`.
   - **`GenericType`** for collection classes (`Collection<T>` etc.) → the list mapping matching
     today's `COLLECTION_GENERICS`; a `GenericType` wrapping a non-collection class unwraps to base.
   - **`BuiltinType` `true`/`false`** → `{type:boolean}` (class 2).
   - **`ObjectShapeType` / `IntersectionType` / `TemplateType`** → chosen unmapped fallback, asserted
     non-crashing (class 10).

3. **Leaf-class policy + optional callback (AMENDMENT 5 — the delicate part; guardrail folded in).**
   `TypeNodeToSchema` externalized leaf-class → schema via `callable(string $fqcn): ?OA\Schema`;
   `JsonSchemaFromType` hard-codes leaf handling (enum→$ref, DateTime/UUID/UrlRoutable→format, else
   fallback). The unified `fromType()` gains an **optional leaf-class resolver parameter with a
   `null` default**:
   - **When absent (null default): the built-in path is byte-identical to today.** This is the
     contract that protects `JsonSchemaFromType`'s **6+ existing `src` consumers** — `SchemaFromDataClass`
     (SpatieData), `UriParametersExtractor`, `PublicPropertyTypeReader`, `ValidationRulesToSchema`,
     `EloquentModelToSchema`, and the Fractal path — none of which pass a callback.
   - **When supplied and it returns non-null, it wins** (consumer ref-vs-inline policy); returns
     null/absent → built-in enum/format/fallback.
   - **Required tests:** precedence for **both** states (callback non-null → wins; null/absent →
     built-in enum/format/warning), **plus a green-guard** asserting the 6 existing callers' output
     is **byte-identical** (they pass no callback). This keeps `EloquentModelToSchema::classTagSchema`
     (model `$ref` policy) intact while giving the Data/return paths the enum-component behavior.

4. **Delete the triplicated scalar map.** Route `TypeNodeToSchema::scalarDefinition()` and
   `EloquentModelToSchema::scalarKeywordToDefinition()` call sites through the boundary +
   `fromBuiltinType()` (extended with `true`/`false`). `EloquentModelToSchema`'s cast/`@property`
   leaf paths (:449-455, :465) build a `Type` (via the boundary or `Type::int()` etc.) and call the
   unified engine. Net: one scalar map.

5. **`TypeNodeResolver::resolveClass()` round-trip.** Once the boundary returns `Type`, callers
   needing only the FQCN keep `resolveClassName()`; callers needing schema use the `Type`. **Do NOT
   delete `resolveClassName`/`genericValueClass`/`listValueType`/`throwsClasses`** — many consumers
   (`@throws` mapping, generic-value unwrap for resources) and #481 will generalize them. Scope this
   PR to *adding* the `Type`-returning boundary and *removing* `TypeNodeToSchema`; broader
   `TypeNodeResolver` slimming is follow-up.

### Files

- **Delete:** `src/Support/Types/TypeNodeToSchema.php` (its test cases port into the unified engine's
  test as characterization cases). Sole `src` consumer is `EloquentModelToSchema` (verified).
- **Modify:** `src/Support/Generator/JsonSchemaFromType.php` (structural cases in correct instanceof
  order + optional leaf callback), `src/Support/Types/TypeNodeResolver.php` (add `toType()` boundary),
  `src/Support/Extraction/EloquentModelToSchema.php` (drop local scalar map + `TypeNodeToSchema` dep;
  route `@property`/cast leaf + list/shape resolution through boundary→unified engine),
  `src/OpenApiServiceProvider.php` (drop the `TypeNodeToSchema` binding; verify `JsonSchemaFromType`
  wiring).

### Tests (TDD — divergence-first)

1. **Characterization catalog** (Step 0): shared-corpus dual-engine snapshot → pinned assertions on
   the **canonical** output for all 13 classes, **run against the 7.x floor** (amendment 1).
2. **Structural parity:** port every `TypeNodeToSchemaTest` case (array shapes required/optional/
   nested, list, map, `array<string,mixed>`, `iterable<T>`, collection generics, `true`/`false`,
   nullable, unresolvable→chosen fallback).
3. **Enum parity:** leaf backed enum with and without a leaf callback yields the `$ref` component +
   case descriptions (class 1 pinned).
4. **Leaf-callback precedence + green-guard:** both callback states pinned; 6 existing callers'
   output asserted byte-identical (amendment 5).
5. **Property-ordering (class 12) and null-vs-stub fallback (class 5):** explicit pinned tests, and
   the survey must confirm each is net-neutral/additive.
6. **Regression guards kept green:** `JsonSchemaFromTypeTest`, `EloquentModelToSchemaTest`,
   `EloquentModelPropertyLookupTest`, `PublicPropertyTypeReaderTest`, `ValidationRulesToSchemaTest`,
   `UriParametersExtractorTest`, `EnumCaseDescriptionTest`, `DeprecatedFieldTest`, and the Fractal
   suite.

- **Intended output changes are additive or explicitly-flagged** (previously-unmapped array/collection
  shapes now map; `true`/`false`→boolean; sorted array-shape properties; the chosen null-vs-stub
  fallback). **Survey Layer A is mandatory and must show 0 regressions**; additions and the flagged
  ordering/fallback changes are itemized and confirmed net-neutral.
- `docs/`: engines are `@internal`; no user page changes expected. Note any observable
  unmapped-fallback wording change (unlikely user-visible).
- `CHANGELOG.md [Unreleased]` — "Changed" (internal): "Unified the two docblock/type→schema engines on
  symfony/type-info `Type`; removed the triplicated scalar-keyword map. No spec output change except
  previously-unmapped array/collection shapes on the symfony path now resolve, `true`/`false` map to
  boolean, and array-shape properties serialize in sorted key order."

### Not in scope (protects #481/#479 — verified against both issue bodies by reviewer)

- **#481 (return-shape descriptor)** wants `{container, itemType: Type, nullable, unionMembers:
  list<Type>}`. The substrate it needs is the **`toType()` boundary** this PR adds — not the schema
  engine. Keeping `Type` (not `OA\Schema`) as boundary currency and preserving
  `genericValueClass`/`listValueType`/`resolveClassName` leaves #481 room to generalize them. The
  descriptor service + container/paginator detection are **not** built here.
- **#479 (typed-return resolver)** sits on both #480 and #481 — untouched here.

### Controversial? Yes → `needs-human` (see header)

No `Contracts/**`, no stage order, no config key. But chain-head + engine deletion + widely-consumed
signature gaining a callback + intended additive deltas ⇒ **routes to `needs-human` after review +
survey** for Moritz's sign-off on `Type`-as-canonical and the divergence-catalog verdicts.


Radiergummi:

Decisive planning finding: **symfony/type-info v8.1 already parses every structural shape** the phpstan engine handles (`ArrayShapeType`, `CollectionType` with `isList()`/key/value accessors, `true`/`false` as `BuiltinType`). So this is a **schema-mapping gap in `JsonSchemaFromType`**, not a parser problem — which is what makes `Type`-as-canonical viable. Boundary constraint: docblock→`Type` resolution needs the reflection-derived `TypeContext` (`list<App\User>` throws context-free), so the bridge routes through `TypeNodeResolver`, not a bare `StringTypeResolver`.

Per the run directive, **Step 0 is a characterization harness** that runs the shared corpus through both engines and emits a divergence catalog before any consolidation. Six divergence classes already identified and pinned in the plan (enum cases, `true`/`false`, object formats, map element typing, unmapped-fallback wording, nullable double-wrap).

Two flags for the reviewer:
1. The **optional leaf-class callback** on the unified `fromType()` (preserving `EloquentModelToSchema`s model-`$ref` policy vs. built-in enum/format handling) is the delicate part — precedence must be pinned per the divergence catalog.
2. Recommend a **human gate** after review+survey: no formal contract is touched, but it deletes an internal engine, changes a signature, has intended output deltas, and is the **head of the #480→#481→#479 chain** — Moritz should sign off on the divergence decisions + `Type`-as-canonical before the downstream issues build on it. I kept `genericValueClass`/`listValueType`/`resolveClassName` intact specifically so #481 is not painted into a corner.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Lead notes: (1) the behavior-preserving mandate is satisfied via the step-0 characterization harness + six pinned divergence classes — reviewer, please pressure-test that catalog for completeness; (2) leaning to ACCEPT the planner's recommendation to route this PR to needs-human after review + survey (chain-head risk, engine deletion, intended additive deltas) — final call at merge time; (3) confirm the Type-as-canonical boundary doesn't corner #481/#479.

Radiergummi:
🤖 **[reviewer]** plan-review: approved with required amendments (catalog gaps + version-floor check); concur on needs-human before merge.

I verified the plan's core claims against source and vendor before ruling. The divergence catalog is
accurate: all six pre-identified classes are real (checked line-by-line):

- **#1 enum cases** — `TypeNodeToSchema::fromIdentifier` (:269-287) routes leaf classes through the
  `$classSchema` callback; `JsonSchemaFromType` has native `BackedEnumType`→`$ref`+case-descriptions
  (:69-74, :135-158). Real.
- **#2 `true`/`false`** — mapped to boolean in `TypeNodeToSchema::scalarDefinition` (:298) but
  **absent** from `fromBuiltinType`'s match (:230-237), so they fall to `unmappedBuiltin`. Real.
- **#3 object formats** — DateTime/UUID/UrlRoutable only in `fromObjectType` (:202-225). Real.
- **#4 map additionalProperties** — only `TypeNodeToSchema::fromGeneric` (:204-221). Real.
- **#5 unmapped-fallback wording** — and note this is **not just wording**: `TypeNodeToSchema`
  returns **`null`** on unmapped (caller degrades), while `JsonSchemaFromType` emits a **stub
  `{type:string, description:"Unmapped..."}` schema** (:96-101, :219-224, :241-249). That null-vs-stub
  difference is a *behavioral* divergence, not cosmetic — the canonical choice materially changes what
  a caller like `EloquentModelToSchema` does on an unresolvable leaf. Pin it explicitly and confirm
  which callers relied on the `null`-degrade.
- **#6 nullable double-wrap** — both use `NullableSchema::wrap`; confirmed.

**Chain safety (verified against both issue bodies): OK.** #481's descriptor wants `itemType: Type`
and `unionMembers: list<Type>`; the substrate it needs is the phpstan→symfony bridge, which is exactly
the `toType()` boundary this PR adds — not the schema engine. Keeping `Type` (not `OA\Schema`) as the
boundary currency and preserving `genericValueClass`/`listValueType`/`resolveClassName` genuinely
leaves #481 room to generalize them rather than forcing a redesign. #479 sits on both and is untouched
here. The "not in scope" section is correctly drawn.

**Leaf-callback design (issue's delicate part): coherent, with one guardrail.** `JsonSchemaFromType`
has **6+ real `src` consumers** (SpatieData `SchemaFromDataClass`, `UriParametersExtractor`,
`PublicPropertyTypeReader`, `ValidationRulesToSchema`, `EloquentModelToSchema`), all of which today
get the built-in enum/format handling. The optional-callback-defaulting-to-null design preserves that
for every existing caller **only if the parameter is genuinely optional with a null default** and the
built-in path is unchanged when it's absent. Required: pin the precedence per #1/#3 with tests for
*both* states (callback returns non-null → wins; null/absent → built-in enum/format/warning), and add
a green-guard asserting the 6 existing callers' output is byte-identical (they pass no callback).

**Required amendments — catalog gaps the harness MUST cover (found by inspecting vendor/symfony
/type-info 8.1):**

1. **Version-floor risk (most important).** The plan asserts "symfony v8.1 already parses every
   structural shape," but composer allows `^7.3.8 || ^7.4.1 || ^8.0 || ^8.1`. `ArrayShapeType`,
   `CollectionType`, `GenericType` and the `getShape()/isList()/getCollectionKeyType()` API this PR
   leans on must exist and behave identically on the **lowest supported** type-info, not just the
   8.1 that happens to be locked locally. Add to the plan: run the characterization harness (or at
   least the structural cases) against the 7.x floor, or tighten the constraint if 7.x lacks the API.
   This is the kind of thing that passes locally and fails a CI leg.
2. **`getShape()` shape is misdescribed.** On 8.1 `ArrayShapeType::getShape()` returns
   `array<key, array{type: Type, optional: bool}>` (keyed by the shape key) — **not** a `list` of
   `ArrayShapeTypeItem` objects as the plan states. The coder must read the actual return shape;
   the plan's API names would not compile.
3. **Property ordering divergence (new catalog entry).** `ArrayShapeType` **`ksort`s** its shape in
   the constructor, so `array{b:int, a:string}` yields properties in **sorted** order — whereas
   `TypeNodeToSchema::objectFromShape` (:140-150) iterates `array{}` in **source order**. That is a
   real, currently-unlisted divergence; add it to the catalog and pin the canonical ordering
   (sorted is the symfony-canonical outcome, but it *is* a change from today's phpstan path — flag in
   the survey).
4. **instanceof ordering constraint.** `ArrayShapeType extends CollectionType`, so the unified
   `fromType()` must test `ArrayShapeType` **before** `CollectionType`, exactly like the existing
   NullableType-before-UnionType / BackedEnum-before-Object rules the class already documents
   (:42-43). Call this out so it isn't discovered as a bug.
5. **Catalog classes to add beyond the six** (the lead asked to hunt for missed ones): multi-class
   **union** `A|B` (today `TypeNodeToSchema::resolveNullable` returns **null** for non-nullable
   multi-class unions at :107-127, while `JsonSchemaFromType` emits `oneOf` at :60-67 — divergence),
   **`array<string,mixed>`** (map with `additionalProperties: true`, the Spatie parity case),
   **`iterable<T>`**, and **`ObjectShapeType`**/**`IntersectionType`**/**`TemplateType`** (all present
   as Type subclasses; the plan routes ObjectShape to the unmapped fallback — acceptable, but the
   harness must *assert* each lands on the chosen fallback rather than crashing). `self`/`static`/
   generics-in-collections resolve to a class name via the context-aware boundary; add a harness case
   to confirm the `TypeContext` path (the plan's noted `list<App\User>`-throws-without-context risk)
   actually resolves imported/relative names, since that's the one place the boundary can throw.

**Tier & scope:** correctly Tier-0 (signatures/PHPDoc, no body parsing). Deleting `TypeNodeToSchema`
(one real `src` consumer: `EloquentModelToSchema`, confirmed) and adding the optional callback is
surgical. The plan wisely defers broader `TypeNodeResolver` slimming — keep that discipline.

**On the needs-human recommendation: I concur.** Even though it touches no `Contracts/**`, the
combination of (a) chain-head for #481/#479, (b) engine deletion, (c) a widely-consumed signature
gaining a callback, and (d) intended additive output deltas at the divergence points warrants
Moritz's sign-off on the two irreversible decisions — `Type`-as-canonical and the divergence-catalog
verdicts — before merge. Coder lands it → reviewer + **mandatory** survey Layer A (0 regressions;
additions itemized) → lead routes to `needs-human`. The survey must specifically confirm the
property-ordering change (#3) and the null-vs-stub fallback change (#5) are net-neutral/additive, not
regressions.

No blocking objection to the approach. Approved to proceed once the harness scope is extended to
cover amendments 1-5 and the leaf-callback guardrail is folded in.

Radiergummi:
🤖 **[lead]** plan-review approved with amendments (see reviewer comment): add divergence classes (null-vs-stub fallback semantics, multi-class unions, array<string,mixed>, iterable<T>, ObjectShape/Intersection/Template fallback, self/static, ArrayShapeType ksort ordering), run structural cases against the lowest supported symfony/type-info (7.x floor), correct getShape() shape + ArrayShapeType-before-CollectionType instanceof order, leaf-callback null-default + 6-caller green-guard. Human gate CONFIRMED: this PR routes to needs-human after review + survey, not auto-merge. Holding in plan-review until a coding slot frees (2-PR cap); planner folds amendments into the PR body meanwhile.

Radiergummi:
🤖 **[planner]** plan amended per plan-review.

Folded all five required amendments + the leaf-callback guardrail into the PR body:
1. **Version-floor** — corrected: composer allows `^7.3 || ^8.0` (floor is 7.3, not the locked 8.1). Added a requirement that the characterization harness runs structural cases against the 7.x floor, tighten-or-record if any accessor differs.
2. **`getShape()` shape corrected** — verified in vendor: returns `array<key, array{type: Type, optional: bool}>` (keyed), not `ArrayShapeTypeItem` objects. Iteration guidance updated so it compiles.
3. **instanceof order** — `ArrayShapeType extends CollectionType` (confirmed in vendor), so `fromType()` must test `ArrayShapeType` before `CollectionType`; documented alongside the existing Nullable-before-Union rule.
4. **Leaf-callback** — null-default optional param, built-in path byte-identical when absent, precedence tests for both states, + a green-guard over the 6 existing `JsonSchemaFromType` consumers.
5. **Catalog expanded from 6 to 13 classes** — added: null-vs-stub unmapped fallback (flagged as *behavioral*, not cosmetic, per your note), multi-class union `A|B`→oneOf, `array<string,mixed>`, `iterable<T>`, ObjectShape/Intersection/Template must-not-crash fallback, self/static context-aware boundary resolution, and the **ArrayShapeType ksort property-ordering** divergence (verified in vendor; canonical = sorted, flagged for survey).

Also pinned in the header: **routes to needs-human after review + survey** (confirmed by lead + reviewer). Survey Layer A must specifically confirm the ksort ordering (class 12) and null-vs-stub fallback (class 5) are net-neutral/additive. Plan is coder-ready for cold-start; holding until a coding slot frees per the 2-PR cap.

Radiergummi:
🤖 **[lead]** coding slot freed by #510 merge → coding. Implement from the amended PR body (13 pinned divergence classes, 7.3 type-info floor, leaf-callback guardrail). Reminder: this PR routes to needs-human after review + survey.

Radiergummi:
🤖 **[coder]** implemented; local gates green (Pint pass, PHPStan no errors, 3316 tests). Pushed dc8ccfee. Ready for review. **Routes to needs-human after review + survey — do not auto-merge.**

## Step 0 — divergence catalog (canonical verdicts pinned)

Harness lives in `tests/Unit/Support/Types/TypeSchemaEngineCharacterizationTest.php` (drives the production path `toType → fromType`); structural/enum/leaf ports in `TypeSchemaStructuralParityTest.php`. Each class below is a pinned test asserting the **chosen** canonical output.

| # | Class | Former phpstan | Former symfony | **Canonical (pinned)** |
|---|-------|----------------|----------------|------------------------|
| 1 | enum leaf | via callback | `$ref` + case descriptions | **`$ref` component + case descriptions** |
| 2 | `true`/`false` | boolean | unmapped stub | **boolean** |
| 3 | DateTime/UUID/UrlRoutable | via callback | string+format | **string+format** (leaf callback may override) |
| 4 | `array<string,T>` map | map | (none) | **`{object, additionalProperties:<T>}`** |
| 5 | unresolvable leaf | partial recovery (empty schema) | whole resolve throws | **degrade to null** — the boundary catches and the caller falls back. Behaviour change: a shape with one bogus class no longer partially recovers. Real `@property` types resolve against model reflection, so this is a test-only edge in practice. |
| 6 | nullable | wrap once | wrap once | **wrap once** (`?int` → `['integer','null']`) |
| 7 | multi-class union `A\|B` | null | oneOf | **oneOf** |
| 8 | `array<string,mixed>` | permissive map | (none) | **`additionalProperties: true`** |
| 9 | `iterable<T>` | list | (none) | **list** |
| 10 | Intersection/ObjectShape/Template | null | — | **non-crashing fallback** (asserted) |
| 11 | `self`/`static`, imported names | resolved | throws w/o context | **resolved via context-aware boundary** |
| 12 | array-shape property order | source order | ksort | **sorted key order (ksort)** — changes phpstan output; flagged for survey |
| 13 | unmapped-fallback wording / bare `array` | null / — | stub | bare `array` → **`{array, items:{}}`**; `mixed` leaf → **untyped `{}`** (preserves the old present-but-untyped property) |

**Amendment 4 (instanceof order):** `ArrayShapeType` is matched before `CollectionType` (it extends it); `NullableType` before `UnionType`; `BackedEnumType` before `EnumType`/`ObjectType`.
**Amendment 5 (leaf callback):** optional `?callable` with a `null` default; when null/declining, the built-in path is byte-identical (guarded — the 5 existing `JsonSchemaFromType` `src` consumers pass no callback and their suites stay green). Precedence pinned for both states.

## Two decisions for sign-off

**1. Version floor (amendment 1) — the declared `^7.3` floor is not reachable in this dependency graph.** `symfony/property-info` (transitive via webauthn/fortify + swagger-php) requires `symfony/type-info ^7.4.7|^8.0.7`, and `radebatz/type-info-extras` (swagger-php 6.2) requires `^7.3.8`. So the effective installable 7.x floor today is **7.4.7**, not 7.3. I verified the full suite (including the type-engine tests) **green on 7.4.7** by forcing that version locally. The accessors this PR uses (`getShape`, `isList`, `getCollectionKeyType/Value`, `getExtraValueType`, `isSealed`, `isIdentifiedBy`) all exist and behave identically on 7.4.7. I left the `composer.json` constraint at `^7.3 || ^8.0` (the accessors are 7.3-compatible; only unrelated transitive deps lift the real floor). **Decision for Moritz: tighten to `^7.4 || ^8.0` to match reality, or keep `^7.3`?** Happy to tighten in a follow-up push if preferred.

**2. Scalar-map dedup (plan step 4) — partially done, deviation noted.** The genuinely triplicated map is reduced by one: `TypeNodeToSchema::scalarDefinition()` is gone with the class, and `JsonSchemaFromType::fromBuiltinType()` now covers `true`/`false`. I **did not** collapse `EloquentModelToSchema::scalarKeywordToDefinition()` into the engine: it is entangled with cast-only keyword handling (`decimal`/`date`/`encrypted`/`real`) and returns an array to spread into `OA\Property`, so routing it through `fromType()` (which returns `OA\Schema`) would add indirection for zero behaviour change at real regression risk on a behaviour-preserving, needs-human PR. Flagging per the deviation protocol — can do it as a follow-up if the reviewer wants the full collapse.

## Scope / not-in-scope
No `Contracts/**`, stage order, or config touched. `Type` stays the boundary currency; `resolveClassName`/`genericValueClass`/`listValueType`/`throwsClasses` are all preserved for #481/#479. The descriptor service (#481) and typed-return resolver (#479) are not built here.

Radiergummi:
🤖 **[lead]** implementation pushed (dc8ccfee), local gates green → review. Round 1: reviewer-1. Two items flagged for Moritz's needs-human sign-off (7.4 floor question; scalar-map dedup deviation) — reviewers should assess both but final call is human. Survey Layer A follows review.

Radiergummi:
🤖 **[reviewer]** review round 1: approved, no blocking concerns. Routes to needs-human as planned (not auto-merge).

I re-ran the full suite locally against dc8ccfee (applied to a clean main), plus a red-before spot
check and a source trace of the bridge. This is the run's highest-risk PR; I went through all six
priorities and it holds up.

**1. No silent picks — every catalog class is pinned.** `TypeSchemaEngineCharacterizationTest` drives
the production path (`toType → fromType`) and pins classes 2–13 as explicit verdicts, including the
two behaviour changes I flagged at plan-review: **class 12** (array-shape property order is `ksort`ed,
`array{b,a}` → `['a','b']`) and **class 5** (an unresolvable leaf degrades the *whole* type to null,
including `array{known:int, mystery:BogusClass}` → null, rather than the old partial recovery). Class
1 (enum + case descriptions) is pinned in `TypeSchemaStructuralParityTest` with the actual case-doc
text asserted. The harness exercises both former engines' input spaces (scalars, `true`/`false`,
`?T`, multi-class union, `array{…}`, `array<K,V>`, `list`/`T[]`/`iterable`, collection generics, leaf
classes/enums/DateTime, `mixed`, intersection). I confirmed the tests genuinely pin behaviour by
reverting the `true`/`false`→boolean mapping: class 2 fails with "Unmapped builtin type: true" and
passes again once restored — real assertions, not accidents of implementation.

**2. Green-guard — verified.** The 5 existing `JsonSchemaFromType` `src` consumers pass no callback,
and the whole Eloquent/Fractal/SpatieData/Faker/UriParameters/PublicProperty/ValidationRules surface
stays green (540 passed in the type+consumer suites; 3344 in the full suite). The
`EloquentModelToSchema` migration preserves the model-`$ref` policy and precedence exactly:
`classTagSchema` is now passed as the leaf callback to `fromType()`, and its body is unchanged
(is-a-Model → pooled `$ref`; else `fromType(Type::object(...))`). Leaf-callback precedence is pinned
both ways (`TypeSchemaStructuralParityTest`: callback wins over the built-in enum path; callback
returning null falls back to it) — amendment 5 satisfied. The unmapped-fallback default path is
byte-identical when the callback is absent.

**3. Bridge correctness — sound, with one discovery worth calling out.** `TypeNodeResolver::toType()`
resolves through the reflection-derived `TypeContext` with a context-free retry, and catches
`Throwable` to degrade to null (so a bad type string never aborts a run). Context-awareness
(`self`/`static`, imported/relative names) is covered (characterization class 11 + `TypeNodeResolverTest`).
The coder discovered a **symfony/type-info quirk not in the plan**: a bare same-namespace class name
(no `use` import) is wrapped in a spurious `CollectionType`, so `normalizeSameNamespaceClass()`
unwraps a `CollectionType` whose wrapped type is not a `GenericType`. This is a real heuristic worth
a careful eye, but it is (a) narrowly scoped (only fires on a CollectionType wrapping a
non-GenericType, which a genuine `array<…>`/`Collection<…>` never is — those wrap a GenericType), (b)
nullable-aware (recurses through `NullableType`), and (c) directly tested
(`TypeNodeResolverTest:117` with a dedicated same-namespace fixture), and every real
collection/map case still resolves correctly (structural parity suite green). I'm satisfied it's
correct; flagging it so Moritz sees it at needs-human since it's an undocumented upstream workaround.

**4. instanceof order + floor API — correct.** `ArrayShapeType` is matched before `CollectionType`
(it extends it), `NullableType` before `UnionType`, `BackedEnumType` before `EnumType`/`ObjectType` —
all as the class docblock now states. The accessors used (`getShape`, `isList`,
`getCollectionKeyType/ValueType`, `getExtraValueType`, `isSealed`, `isIdentifiedBy`) all exist on the
installed 8.1; the coder reports verifying them green on a forced 7.4.7 (see item 5). `getShape()` is
correctly consumed as `array<key, array{type: Type, optional: bool}>` (my plan-review correction was
folded in — no `ArrayShapeTypeItem` object assumption).

**5. My recommendation on the two flagged items (both go to Moritz):**
   - **Version floor — recommend tightening to `^7.4 || ^8.0`.** The coder verified the effective
     installable 7.x floor is 7.4.7 (forced transitively by `symfony/property-info` via
     webauthn/fortify + swagger-php), so the declared `^7.3` names a version that can never resolve in
     this graph — a false promise. Tightening is honest and costs nothing (nothing can install 7.3
     anyway). One-line change; fine to fold in now or as Moritz prefers.
   - **Scalar-map dedup left partial — recommend accepting as-is.** The genuine triplication is
     reduced (the `true`/`false`-bearing map is gone with `TypeNodeToSchema`; `fromBuiltinType` now
     covers `true`/`false`). `EloquentModelToSchema::scalarKeywordToDefinition()` is entangled with
     cast-only keywords (`decimal`/`date`/`encrypted`/`real`) and returns an array to spread into
     `OA\Property`; routing it through `fromType()` (which returns `OA\Schema`) would add indirection
     for zero behaviour change and real regression risk on a behaviour-preserving PR. Forcing the full
     collapse on the highest-risk PR of the run is exactly the kind of non-surgical change CLAUDE.md
     warns against. Leave it; file the cast-keyword collapse as a follow-up if wanted.

**6. Deleted-engine check — clean.** No code, provider binding, doc, or config references
`TypeNodeToSchema` anymore (the provider binding is removed; `EloquentModelToSchema` no longer depends
on it; `TypeNodeToSchemaTest` is deleted and its cases ported into the parity suite). The only
remaining mentions are descriptive comments/docblocks in the new tests naming the former engine —
harmless.

**Conventions / bookkeeping.** strict-types headers; `@internal` retained; modern PHP (`match`,
readonly, first-class-callable `classTagSchema(...)`); no abbreviations; comments concise/why-focused,
no issue numbers in code. CHANGELOG `[Unreleased]` "Changed" entry present, accurate, and honest about
the additive-only deltas. Docs: all engines `@internal`; no user page change needed.

**Test-gap check (standing duty).** Refused/degrade branches, both leaf-callback states, the
nested-scope/quirk paths, and the exotic-Type fallback are all covered. The one thing that is a
**process gate, not a test gap**: the intended deltas (class 12 ksort ordering, class 5 null-degrade)
must be confirmed net-neutral/additive by the survey — they are the only observable output changes,
and the survey Layer A (0 regressions; additions itemized) is the proof. That's already the plan.

Approved. Correct call to route to **needs-human** after the survey: chain-head for #481/#479, engine
deletion, a widely-consumed signature gaining the callback, the undocumented type-info quirk workaround,
and the two sign-off decisions all warrant Moritz's eyes before merge.

Radiergummi:
🤖 **[reviewer]** independent review (reviewer-2): approved, **1 non-blocking finding** to surface at needs-human. Routes to needs-human as planned.

I formed my own view from #480, the amended plan, the divergence catalog, and the diff at `dc8ccfee` before reconciling with reviewer-1. We converge on the substrate — but my adversarial probe of `normalizeSameNamespaceClass()` (the lead's flagged angle) turned up one real edge reviewer-1's pass did not.

### Finding 1 (non-blocking, correctness edge) — `normalizeSameNamespaceClass()` over-unwraps bare `iterable`

`normalizeSameNamespaceClass()` (`TypeNodeResolver.php`) unwraps any `CollectionType` whose wrapped type is **not** a `GenericType`. I probed symfony/type-info directly on the locked version:

- bare `array` → `CollectionType(BuiltinType "array")` → unwrap → `BuiltinType(array)`
- bare `iterable` → `CollectionType(BuiltinType "iterable")` → unwrap → `BuiltinType(iterable)`
- same-namespace class (`TypeContext`) → `CollectionType(ObjectType)` → unwrap → `ObjectType` (the intended fix)
- every genuine generic (`array<K,V>`, `list<T>`, `T[]`, `iterable<T>`, `Collection<T>`) → `CollectionType(GenericType)` → **not** unwrapped ✓

So the guard fires on bare `array`/`iterable` too, not just same-namespace classes. Downstream effect through the production `toType → fromType` path:

- bare `array` → `{type: array, items: {}}` — **benign** (coincides with the CollectionType path; pinned by "class 13").
- bare `iterable` → `{type: string, description: "Unmapped builtin type: iterable"}` — a **stub**. I confirmed that **without** the unwrap, `fromType(CollectionType(BuiltinType iterable))` yields `{type: array, items: {}}`; the unwrap is what degrades it. `fromBuiltinType()` maps `ARRAY` but has no `ITERABLE` case, so the unwrapped builtin falls to `unmappedBuiltin`.

Versus `main`: `TypeNodeToSchema` returned **null** for a bare `iterable` identifier (caller degrades to an empty property). The PR emits a misleading `{type: string}` for it. Low blast radius — it only bites a bare `iterable` reaching the **docblock** path (`toType`), e.g. `@property iterable $x` on a model or an `iterable`-typed return; reflection-typed consumers go via `TypeResolver` and never touch this heuristic. There is **no test for bare `iterable`** (the suite tests `iterable<T>`, not the bare form), so it slips the characterization net.

**Clean fix (one line):** tighten the guard to skip `BuiltinType` as well as `GenericType`, e.g. `if ($type instanceof CollectionType && !$type->getWrappedType() instanceof GenericType && !$type->getWrappedType() instanceof BuiltinType)`. A same-namespace class wraps an `ObjectType` (neither), so the intended fix is preserved; bare `array`/`iterable` keep their `CollectionType` and map correctly. Add a bare-`iterable` characterization case. **Not a merge-blocker** given the rarity, the survey gate, and the needs-human stop — but worth Moritz's eyes since it's in the exact undocumented upstream workaround this PR introduces.

### Everything else — verified, matches reviewer-1

- **No silent picks.** `TypeSchemaEngineCharacterizationTest` drives the production `toType → fromType` path and pins classes 2–13; class 1 (enum + case-doc text) in the parity suite. I confirmed real assertions by reverting `true`/`false`→boolean: class 2 fails "Unmapped builtin type: true", passes on restore.
- **Callback-absent byte-identity across ALL call sites (I grepped every one, not just 6).** `fromType` consumers: `SchemaFromDataClass` (×3), `UriParametersExtractor`, `PublicPropertyTypeReader`, `ValidationRulesToSchema`, and `EloquentModelToSchema` (×3). Only `EloquentModelToSchema` passes a callback (`classTagSchema(...)`, model-`$ref` policy preserved verbatim); all others pass none and hit the built-in path unchanged. Consumer suites green (483 in type+extraction+fractal).
- **Throwable→null degrade in `toType()`** with a context-free retry is correct; `EloquentModelToSchema::propertyFromNode` maps the null to its existing empty-fallback + warning, so no caller regresses on unresolvable input.
- **Deleted-engine test porting — no *silent* loss.** 29 deleted cases → 23 parity + 12 characterization. The deleted "present-but-untyped shape value" / "unreadable map value = empty schema" cases were **intentionally superseded** by the class-5 verdict (unresolvable leaf degrades the whole type to null), which is documented in the catalog and coder deviation note and pinned at characterization line 88. Not lost — deliberately re-decided (and flagged for survey).
- **instanceof order** (`ArrayShapeType` before `CollectionType`, `Nullable` before `Union`, `BackedEnum` before `Enum`/`Object`) correct; `getShape()` consumed as `array<key, array{type, optional}>` (no `ArrayShapeTypeItem` assumption).
- **Chain safety.** `Type` stays the boundary currency; `resolveClassName`/`genericValueClass`/`listValueType`/`throwsClasses` preserved for #481/#479.
- **Gates (branch worktree `dc8ccfee`):** Pint passed, PHPStan level-8 no errors, 483 relevant tests pass.

### On the two sign-off items
- **Version floor:** agree with reviewer-1 — tighten to `^7.4 || ^8.0` (the effective installable floor is 7.4.7, forced transitively; `^7.3` names an unreachable version). Moritz's call.
- **Scalar-map partial dedup:** agree — accept as-is; forcing the `EloquentModelToSchema` cast-keyword collapse (returns an array to spread, not an `OA\Schema`) onto the highest-risk PR is non-surgical for zero behaviour change. Follow-up.

**Verdict: approve.** The unification is sound and behaviour-preserving apart from the catalogued verdicts. Finding 1 is a genuine correctness edge in the flagged heuristic with a one-line fix; I'm not blocking on it given the needs-human gate, but it should be fixed here or tracked. Survey Layer A must still confirm the class-12 (ksort) and class-5 (null-degrade) deltas are net-neutral/additive.

Radiergummi:
🤖 **[lead]** round 2 dispatched to the coder: fold in reviewer-2's iterable-guard fix (exclude BuiltinType in the unwrap guard + pinned bare-iterable/bare-array tests) and tighten the type-info constraint to ^7.4 || ^8.0 per both reviewers. Scalar-map dedup stays as-is (accepted deviation). Survey runs after the round-2 delta check so it measures the final commit.

Radiergummi:
🤖 **[coder]** addressed round 2; pushed c04eb9d5. Local gates green (Pint, PHPStan, 3317 tests).

**Finding (reviewer-2 — bare `iterable` degraded to the unmapped-string stub):** fixed. Root cause was the `normalizeSameNamespaceClass()` guard unwrapping *any* `CollectionType` wrapping a non-`GenericType` — which also unwrapped bare `array`/`iterable` (both resolve to `CollectionType(BuiltinType)`). Bare `array` stayed correct by coincidence; bare `iterable` unwrapped to `BuiltinType(iterable)` → `fromBuiltinType` default → `{type:string, "Unmapped builtin type: iterable"}`.

I went one notch tighter than "exclude `BuiltinType`": the guard now unwraps **only when the CollectionType wraps an `ObjectType`** — the actual same-namespace-class quirk always wraps an ObjectType (verified: same-namespace bare *enums* resolve straight to `BackedEnumType`, no quirk; genuine arrays/collections wrap a `GenericType` or a bare `BuiltinType`). This is strictly more precise than the non-`BuiltinType` form and matches the documented intent. Pinned tests added for bare `array` and bare `iterable` on the `toType → fromType` path (both → `{type:array, items:{}}`).

**Version floor:** tightened `symfony/type-info` to `^7.4 || ^8.0` in `composer.json` (+ the CHANGELOG dependency line), since the `^7.3` floor is unreachable — `symfony/property-info` transitively forces `^7.4.7`. `composer.lock` is re-hashed (content-hash only; no package version changed; installed stays 8.1). Verified green on the 7.4.7 floor earlier.

**Scalar-map dedup:** left as-is per the accepted deviation (follow-up).

Ready for reviewer-2's delta check.

Radiergummi:
🤖 **[reviewer]** delta-check (reviewer-2) on c04eb9d5 (ff on dc8ccfee): **finding resolved, delta clean — approved.**

`git diff dc8ccfee..c04eb9d5` touches exactly 5 files, all in scope: the `normalizeSameNamespaceClass` guard + its docblock, one new parity test, and the `^7.4` constraint (composer.json + CHANGELOG wording + lockfile hash). Nothing else moved.

**The bare-iterable fix — verified better than what I suggested.** The coder used an **ObjectType-positive** guard (`$type->getWrappedType() instanceof ObjectType`) rather than my subtractive BuiltinType-exclusion. This is the stronger choice: it unwraps only the exact quirk shape `CollectionType(ObjectType)` as a positive whitelist, so no future `CollectionType(SomethingElse)` can slip through. I re-probed symfony/type-info through the production `toType → fromType` path:
- bare `iterable` → **`{type: array, items: {}}`** (was the `{type: string}` stub) ✓ fixed
- bare `array` → `{type: array, items: {}}` (unchanged) ✓
- `array<int>` / `list<T>` / `Collection<T>` → stay `CollectionType(GenericType)`, unaffected ✓
- same-namespace generic value class (`Collection<int, SameNamespaceValue>`) → still unwraps to the real class (`genericValueClass` returns `SameNamespaceValue`), the quirk this workaround exists for ✓

**Guard can't miss a legitimate quirk case.** I confirmed the lead's premise empirically: a bare same-namespace backed enum (`MediaType` resolved against an `Enums`-namespace context) resolves **directly to `BackedEnumType`** and never reaches the CollectionType wrapper, so the ObjectType-only guard correctly needn't handle it. Nullable wrapping is handled by the existing `NullableType` recursion (verified `?MediaType` → `NullableType(BackedEnumType)`). The quirk empirically only produces `CollectionType(ObjectType)` for a plain same-namespace object, which the guard still unwraps.

**Pinned tests assert the correct canonical output.** New parity case pins both bare `array` and bare `iterable` to `['type' => 'array', 'items' => []]` — consistent with the class-13 bare-array verdict; the `TypeNodeResolverTest` same-namespace case still green.

**Constraint + lockfile clean.** `composer.json` and the CHANGELOG line both now read `^7.4 || ^8.0`; `composer.lock` changed **only** its `content-hash` (no package add/remove/version drift — diffed and confirmed), and installed `symfony/type-info` stays v8.1.0, satisfying the tightened range.

**Gates (c04eb9d5):** full Types suite 61 passed; Extraction + Plugins 668 passed; Pint `--test` passed; PHPStan level-8 no errors.

Delta approved. My round-1 concern is closed. PR still routes to needs-human after survey; the class-12 (ksort) and class-5 (null-degrade) deltas remain the survey's to confirm net-neutral/additive.

Radiergummi:
🤖 **[lead]** two independent approvals (incl. round-2 delta check) → survey. Intended deltas to confirm net-neutral/additive: class-12 ksort property ordering, class-5 null-degrade (unresolvable leaf), plus additive coverage (array shapes, list<T>, true/false, bare iterable). After survey: needs-human (Moritz sign-off on Type-as-canonical, catalog verdicts, the type-info quirk workaround, ^7.4 floor, scalar-map deviation).

Radiergummi:
🤖 **[coder]** rebased onto post-#511 `main`; pushed 924a2486. **Remote CI fully green on the final sha.**

- Clean rebase — **no conflicts** (my CHANGELOG entry sits at the end of `### Changed`; #511's landed in other sections, and it touched the shared `ReturnExpressionResolver`/reader files which my diff doesn't overlap). `sync-branch`'s `composer check` passed (3333 tests).
- CI re-triggered automatically on the force-push (no close/reopen needed). All matrix cells pass: PHP 8.4/8.5 × Laravel 12/13, the **swagger-php 5.8** leg, `quality` (Pint + PHPStan), **Snapshot (byte-exact)**, survey tooling, and codecov.
- The byte-exact Snapshot passing confirms the rebase is spec-output-neutral.

My diff vs new main is unchanged (22 files, +768/-615). Ready for the surveyor's Layer A verdict and reviewer-2's delta check to land against 924a2486.

Radiergummi:
🤖 **[surveyor]** Layer A pass — 1 additive improvement, 0 regressions across 11 apps.

Ran candidate c04eb9d5 (on current main, incl. merged #510) vs the fresh `main` baseline 417ee8eb (byte-neutral to d421601f via the #510 gate). All 11 apps booted, `generateExit=0`, no crashes.

**Aggregate coverage — no regression (identical on every app):**
- Total responseSchemas **500 → 500**, componentSchemas **1009 → 1009**.
- completenessPercent, apiOperations, paths, operations, requestBodies unchanged on all 11 apps.

**Semantic spec diff (per-boot tokens masked): 10/11 apps byte-identical.** One app has exactly one delta:

| app | JSON path | baseline | candidate | class |
|-----|-----------|----------|-----------|-------|
| AdvisingApp | `components.schemas.Media.properties.model_id` | `{}` (empty/any) | `oneOf: [{integer},{string}]` | **(b) ADDITIVE** |

Classification: **ACCEPTABLE — additive.** Main emitted an empty schema where the unified `symfony/type-info` engine now resolves the real shape — precisely the documented change ("previously-unmapped shapes on the symfony path now resolve"). `model_id` is a polymorphic morph key (`int|string`), so `oneOf[integer, string]` is correct and strictly more specific than `{}`. Verified nothing else in the `Media` schema moved: all 23 properties, `required`, and `type` are identical; only `model_id` gained detail.

No occurrences of the other documented cases in this corpus (no `true`/`false`→boolean sites, no ksort-visible reorderings that changed masked bytes) and — importantly — **no class-5 degradations, no wrong types, no dropped enum cases, no dropped `$ref`s, no previously-correct schema made worse or removed.** Clear to merge.

_(Note: the branch has since rebased to 924a2486, reported content-identical to c04eb9d5; this result stands for that content.)_

Radiergummi:
🤖 **[lead]** ALL automated gates green → needs-human (as planned from plan-review, not a failure). Ready for Moritz's sign-off, then merge.

**Why human-gated:** chain-head for #481/#479; deletes TypeNodeToSchema; changes the widely-consumed JsonSchemaFromType signature; adopts symfony Type as the canonical representation; adds an undocumented-upstream-quirk workaround (normalizeSameNamespaceClass, ObjectType-positive guard).

**Evidence:** 2 independent reviewer approvals (round 1 + adversarial second pass + round-2 delta check); full CI matrix green on final sha 924a2486; survey Layer A PASS — 10/11 apps byte-identical, 1 additive improvement (AdvisingApp Media.model_id {} → oneOf[integer,string]), coverage aggregates unchanged (500/500 response schemas, 1009/1009 component schemas).

**Decisions awaiting sign-off:** (1) Type-as-canonical + the 13 pinned divergence-catalog verdicts (esp. class-5 null-degrade, class-12 ksort ordering); (2) symfony/type-info floor tightened to ^7.4||^8.0; (3) scalar-map dedup left partial in EloquentModelToSchema (accepted deviation, follow-up candidate); (4) the upstream-quirk workaround. Merge unblocks #481 → #479.

Issue #493, opened 2026-07-01
feat(requests): field-attribute overlay seam — separate decorating an inferred field from declaring a missed one


Radiergummi:
## Problem

The `FieldAttribute` family (`#[PathParam]`, `#[QueryParam]`, `#[RequestField]`, `#[ResponseField]`) is dual-role: the same attribute at the same call site simultaneously (a) **decorates** an already-inferred field with human copy (description/example on a param whose type route-binding or reflection already produced — `PathParam` exposes no `type` at all), and (b) **declares** a field inference can't reach (e.g. `RequestField`'s method-level repeatable form for bodies validated in an Action/service — genuinely Tier-2-refused territory).

Because no seam separates the two roles:

- lint cannot distinguish healthy description-decoration from a smell that restates a reflected type,
- each consumer re-decides merge semantics — `QueryParam` alone is read in ~6 sites with no shared reader, and `FieldDefault::Unset` (the only "did the author speak?" primitive) exists only inside this family's `descriptor()` merge.

## Proposed change

A post-extraction **metadata overlay** keyed by field name: after a field is inferred, attribute-supplied description/example/format/deprecation are applied as an overlay through the shared `ResolvedField` merge primitive (single documented precedence, provenance recorded). The declaring role remains for fields inference misses — but the decorating role stops carrying type/constraint arguments for already-inferred fields, and a lint rule can then flag an attribute that restates what reflection produced (the "attribute smell" from the inference philosophy, currently undetectable).

## Why

Resolves the one structural gap in the otherwise-healthy attribute surface (24/~30 attributes audited as legitimately irreducible). Depends on the ResolvedField merge primitive.

Radiergummi:
🤖 **[lead]** admitted → planning. Provenance chain: #482 merged (its ResolvedField merge primitive is the substrate this seam builds on). #483 Part A (PR #520) and #481 (PR #519) are both escalated-ready pending Moritz — plan against main; note any dependency on their merges explicitly.

Issue #484, opened 2026-07-01
refactor(lint): emit inference-failure findings at generation time with reasons; make FieldProvenance lint-readable


Radiergummi:
## Problem

Several lint rules are generation diagnostics in disguise: they observe an *absence* in the finished spec and re-derive by re-reflection what the pipeline knew precisely when it bailed. ~6 rules construct fresh reflectors, ~12 re-read attributes — e.g. `ActionMethodMissing.php:49` does `new ReflectionClass(...)` on a controller the introspector already touched; `ActionMissingReturnType.php:71` re-reads `getReturnType()`; `ThrowsTransitiveMissing` re-runs the `@throws` walk error-response inference already did.

The correct channel **already exists and is proven**: `SchemaFromFormRequest` emits `request-body.schema-degraded` findings with the exact degrade reason at the failure site (`SchemaFromFormRequest.php:117,199-223`), with `RequestBodySchemaDegraded` as a registration stub for ID/suppression/severity. Six generation-side classes already write into the container-bound `FindingsCollector`. The re-reflecting rules simply never adopted the pattern.

Separately, `FieldProvenance` is documented as a side channel "that never reaches the document" (`OperationDescriptor.php:30-33`) and `src/Lint/` never references it.

## Proposed change

1. Convert the inference-failure rules (`operation.return-type-missing`, `throws.transitive-missing`, `operation.action-method-missing`, "schema missing/empty" observations) to the `request-body.schema-degraded` pattern: the code that bailed emits the finding with its reason; the rule class becomes a registration stub.
2. Make `FieldProvenance` lint-readable (cheap once the ResolvedField merge primitive makes it total), retiring re-reflection in the origin-derivation rules.

## Why

"Schema missing *because `rules()` takes a runtime argument*" is actionable; "schema missing" is not. Findings gain the fix-hint quality the FormRequest path already demonstrates, and lint stops paying reflection cost to reconstruct facts the pipeline discarded. Keep lint downstream of the finished document for spec-shape rules — this enriches the intermediate, it does not move lint upstream.

Radiergummi:
[...]

Issue #483, opened 2026-07-01
feat(core): record producer and confidence in ComponentSchemaRegistry; retire NeedsInferenceDocument double generation


Radiergummi:
## Problem

Nothing records **which resolver/plugin produced a component schema, or at what confidence**. `ComponentSchemaRegistry::componentClassMap()` keeps only component-key → class-string. Two consequences:

- `buildOnce()` is first-write-wins keyed by class (`ComponentSchemaRegistry.php:178,348`): a class reachable as a leaf from two plugin owners is built by whoever got there first — deterministic but implicit, with **no warning when two builders would have disagreed**.
- The SwaggerPhp migration rules (`NeedsInferenceDocument`) approximate "was this inferred?" by running a **second full generation** with stages excluded and diffing (`LintRunner:276`) — the most expensive workaround in the codebase, standing in for a provenance record never kept.

## Proposed change

Record `{ producer: class-string, degraded: bool, reason: ?string }` alongside each entry at registration time in `ComponentSchemaRegistry` (the single funnel every builder already goes through). Then:

- migration rules read the record instead of double-generating — delete the `inferenceOnly()` second pass,
- emit a diagnostic when a second builder attempts a *different* schema for an already-owned class,
- `openapi:why` can answer "who produced this component."

## Why

One field at the existing choke point removes an entire duplicate generation per lint run and makes cross-plugin schema ownership observable instead of accidental. Low cost, reuses existing infrastructure.

## Requirements ported from #203 and #208 (backlog audit 2026-07-01)

This issue supersedes #203, which tracked the same goal and carried two constraints that MUST hold here:

1. **Gating.** Ordinary `openapi:generate` runs must not pay for provenance-over-authored-classes when nobody is linting — the inferred-alongside-authored view is built only when migration linting requests it.
2. **The harvester must become non-destructive (or provenance-recording).** Producer tags alone are NOT sufficient to retire the double generation: as #208's investigation established, origin-provenance answers "who produced this," never the **counterfactual** "what would inference have produced?" — because authored schemas are realized upstream of (and suppress) the inference path for the same class. To replace the second generation, the registry must retain the **inferred view alongside the authored winner** for classes where both exist; the migration rules' redundancy verdict then reads that pair off the single document. `SchemaEquivalence`, the dangling-`$ref` guard, and provenance-by-class stay as-is; only the oracle changes.

[...]

PR #514, opened 2026-07-02
refactor(core): ResolvedField merge primitive + total provenance in OperationBuilder

### LINKED ISSUE #482: refactor(core): ResolvedField merge primitive — declared field precedence and total provenance in OperationBuilder
## Problem

`OperationBuilder` (1,391 lines) is an undeclared merge engine: every operation field has a bespoke inline precedence chain — status (`OperationBuilder.php:224-256`: explicit `#[Response(2xx)]` → body-scanned marker → 204-suppressed-by-content → convention → 200 default), summary (`:1162-1199`), tags (`:269-291`), request body (`applyRequestBodyOverride`). Three different precedence disciplines exist across consumers (attribute-wins, inference-wins, highest-specificity), and `FieldDefault::Unset` — the only principled "did the author speak?" primitive — is used by exactly one attribute family.

Consequences:

- **Provenance is anecdotal**: `FieldProvenance` covers only summary/status/tags, because recording it requires hand-writing it per bespoke chain.
- The body-scanned status is smuggled through a **transient vendor-extension sentinel** (`EXPLICIT_STATUS_EXTENSION`, `OperationBuilder.php:80`) because `PrimaryResponseResolver` can only return a bare `OA\Response`.

## Proposed change

1. A small `ResolvedField { value, source, reason }` merge primitive; every operation field resolves through it with a declared precedence list. `FieldProvenance` becomes **total** — every field, for free.
2. Extract the mechanical attribute-reading/`OA\*` construction bulk (~60% of the class: headers, cookies, examples, links, response headers) into per-attribute applier classes. No behavior change.
3. Widen `PrimaryResponseResolver`'s return to a small result object (response + explicit-status flag) and delete the vendor-extension sentinel.

## Why

Makes precedence a documented model instead of loop-and-break folklore (scribe documents its per-concern merge semantics; ours are implicit). Total provenance is the prerequisite for the lint/provenance issues and turns `openapi:why` from a 3-field demo into a real explainer. Keep per-operation building — do **not** decompose into per-concern passes across all operations; `ActionDescriptor` locality is load-bearing.

Radiergummi:
Closes #482

## Implementation plan — #482

**Tier: N/A** (no inference change). Internal refactor of `OperationBuilder`'s field-merge logic +
one contract widening. **Behavior-preserving except for total provenance** (new provenance entries
are additive — `openapi:why --fields` gains fields; the emitted OpenAPI document is byte-identical).

### What was verified (grounds the plan)

- `OperationBuilder` (1,391 lines) resolves each operation field via a **bespoke inline precedence
  chain**: status (`:208-256`), summary (`:1156-1199`), tags (`:269-291` + `:1240-1263`), request
  body (`applyRequestBodyOverride :466`). `FieldProvenance` (`Support/Provenance/`) is emitted for
  **only** summary/status/tags because each entry is hand-written per chain (`:1160-1263`).
- `FieldDefault::Unset` is the only "did the author speak?" primitive and is used solely by the
  `FieldAttribute` family (confirmed: 8 attribute files reference it; none in `OperationBuilder`'s
  own field merges).
- The **body-scanned explicit status** is smuggled through a transient vendor extension
  `EXPLICIT_STATUS_EXTENSION` (`:80`): `InlineJsonResponseResolver::markStatusExplicit` (:277-284)
  writes `$response->x[...]=true`, `OperationBuilder::takeExplicitStatusMarker` (:818-833) reads +
  strips it. It is the **only** writer of the sentinel (verified).
- `PrimaryResponseResolver` (`Contracts/Registry/`) returns a bare `?OA\Response`. **8 concrete
  implementations** across Core (Inline/Paginator/EloquentModel/FindReturnModel), Fortify,
  ApiResources, Fractal (Helper/Resolver/EntityTransformer), SpatieData.
- The **mechanical applier bulk** (~60% of the class): 12 `apply*`/`build*` private methods for
  headers, cookies, examples, example-files, response-headers, links, request-examples, request-body
  override.

### Design — three pieces, in dependency order

**Piece 1 — `ResolvedField` merge primitive + total provenance (the load-bearing part).**

New `@internal` value object in `Support/Provenance/` (or `Support/Generator/`):

```
ResolvedField { mixed value; string source; string reason; list<string> supersededBy }
```

with a small resolver that takes an **ordered list of candidates** (each a `{value, source, reason}`
where value may be "absent") and returns the first present one as the winner, recording every
lower-precedence *present* candidate it beat into `supersededBy`. `FieldProvenance` is then derived
directly from a `ResolvedField` (they are near-identical today — `FieldProvenance` becomes the
render-time projection of a `ResolvedField`, or `ResolvedField` replaces it; **decide in favor of
keeping `FieldProvenance` as the public/why-facing type** and making `ResolvedField` the internal
merge mechanic, to avoid churning `WhyCommand`/`OperationDescriptor` signatures — flag for review).

Convert each bespoke chain to declare its candidate list through this primitive:
- **status** (`:208-256`): candidates in the documented order — explicit `#[Response(2xx)]` →
  body-scanned explicit → content-bearing-body-over-204 → convention → `200` default.
- **summary** (`:1162-1199`): method `#[Summary]` → method `#[Operation(summary)]` → docblock →
  class `#[Summary]` → class `#[Operation(summary)]` → convention.
- **tags** (`:269-291`, `:1240-1263`): the replace/merge/controller-derived chain.
- **request body**: `applyRequestBodyOverride` precedence.
- Fields that today have **no** provenance (description, deprecated, operationId, externalDocs,
  parameters, security) gain a `ResolvedField` resolution so `FieldProvenance` becomes **total** —
  this is the issue's headline deliverable. Each is a small candidate list; the emitted value must
  be **identical** to today (pinned by the byte-identical document guard).

**Key constraint (from the run directive + #483/#493 dependency): the precedence must be a declared,
inspectable model, not loop-and-break.** The candidate list per field is the documented model. Keep
it as data (an ordered array of candidates with source/reason strings) so #483's provenance record
and #493's overlay seam can consume the same precedence vocabulary.

**Piece 2 — extract the mechanical applier bulk into per-attribute applier classes (no behavior
change).** Move the 12 `apply*`/`build*` methods (headers, cookies, examples, example-files,
response-headers, links, request-examples, request-body-override) into small `@internal` applier
classes in `Support/Generator/` (e.g. `ResponseHeaderApplier`, `ExampleApplier`, …). `OperationBuilder`
orchestrates; each applier owns one attribute-family's `OA\*` construction. **Pure move** — asserted
by the byte-identical document guard. This is the ~60% size reduction; keep per-operation building
(the issue is explicit: do **not** decompose into per-concern passes across operations —
`ActionDescriptor` locality is load-bearing).

**Piece 3 — widen `PrimaryResponseResolver` return + delete the sentinel.** Replace the bare
`?OA\Response` with a small result object:

```
PrimaryResponse { OA\Response response; bool statusIsExplicit }   // @internal, Contracts/Registry
```

`resolvePrimaryResponse(): ?PrimaryResponse` (null still = pass to next). `InlineJsonResponseResolver`
sets `statusIsExplicit: true` instead of writing the vendor extension; `OperationBuilder` reads the
flag off the result instead of `takeExplicitStatusMarker`. Delete `EXPLICIT_STATUS_EXTENSION`,
`markStatusExplicit`, `takeExplicitStatusMarker`. **All 8 implementations update** to return the
result object (7 of them trivially wrap their `OA\Response` with `statusIsExplicit: false`).

### Chain safety — do NOT paint #483/#493/#484 into a corner (read all three)

- **#483** (producer/confidence in `ComponentSchemaRegistry`) is a *different choke point*
  (component-schema registration), not `OperationBuilder`. It reuses the **provenance vocabulary**
  (source/reason/producer) this PR formalizes but does not touch `ResolvedField`. Load-bearing note
  from the run directive: #483 needs the registry to keep the **inferred view alongside the authored
  winner**, gated so ordinary generate runs pay nothing. **That is #483's work, not #482's** — but
  #482 must not make provenance *generation-time-only-when-linting*: keep `FieldProvenance`
  emission cheap and unconditional here (it is already always-on and document-free), so #483 can add
  its own gated component-level record without fighting a gate #482 introduced. **Do not add a
  linting gate to `OperationBuilder` provenance.**
- **#493** (field-attribute overlay seam) explicitly *"Depends on the ResolvedField merge
  primitive."* It will apply attribute description/example/format as an **overlay through
  `ResolvedField`**. So the primitive must be usable **outside** `OperationBuilder` (field-name-keyed
  overlay on already-inferred fields) — design `ResolvedField` as a standalone merge mechanic
  (candidate list in, winner+provenance out), **not** a private `OperationBuilder` method. This is
  the one forward-compat requirement that shapes the API; everything else stays internal to #482.
- **#484** (inference-failure findings + lint-readable `FieldProvenance`) needs `FieldProvenance`
  to become **total and lint-readable**. #482 delivers *total*; #484 makes it lint-readable. Keep
  `FieldProvenance` on `OperationDescriptor` as it is (already carried), so #484 can read it.

**Scope: only #482's work.** No component-registry changes (#483), no overlay seam (#493), no
lint-rule conversion (#484).

### Files

- **New:** `Support/Provenance/ResolvedField.php` (+ its merge resolver, possibly
  `ResolvedFieldMerger`); per-attribute applier classes in `Support/Generator/`
  (`ResponseHeaderApplier`, `ExampleApplier`, `ResponseExampleFileApplier`, `LinkApplier`,
  `HeaderParameterApplier`, `CookieParameterApplier`, `RequestExampleApplier`,
  `RequestBodyOverrideApplier` — final split decided during coding to keep each cohesive);
  `Contracts/Registry/PrimaryResponse.php` (result object).
- **Modify:** `Support/Generator/OperationBuilder.php` (orchestrate via primitive + appliers; delete
  sentinel plumbing), `Contracts/Registry/PrimaryResponseResolver.php` (widen return),
  the **8** `PrimaryResponseResolver` implementations, `Plugins/Core/Resolvers/InlineJsonResponseResolver.php`
  (flag instead of extension). `WhyCommand`/`OperationDescriptor` only if `FieldProvenance` shape
  changes (goal: it does not).

### Tests (TDD)

1. **Byte-identical document guard (the safety net for Pieces 1+2):** snapshot the generated
   OpenAPI document for a representative fixture set **before** the refactor; assert byte-identical
   **after**. Pieces 1 and 2 must not move a single output byte. (Survey Layer A is the corpus-scale
   version of this — mandatory, 0 regressions.)
2. **`ResolvedField` unit tests:** winner selection across an ordered candidate list; `supersededBy`
   accumulation (only *present* lower candidates); all-absent → no field; single-candidate → no
   supersede; the `FieldDefault::Unset`-style "author spoke vs. absent" distinction.
3. **Total-provenance tests:** every operation field now yields a `FieldProvenance` entry with the
   correct source/reason for each precedence branch — including the fields that had **none** before
   (description, deprecated, operationId, externalDocs, parameters, security). Cover each branch of
   status/summary/tags exactly as the current hand-written provenance tests do (port them).
4. **Sentinel-deletion tests:** `InlineJsonResponseResolver` explicit-status still wins over the
   resource convention (the behavior the extension protected), now via the result flag; assert no
   `x-laravel-openapi-explicit-status` appears in any output (it never did, but pin it).
5. **Resolver-widening green-guard:** all 8 implementations compile against the new return type and
   their existing response tests stay green.

- **Emitted document: unchanged.** New provenance is additive (why-facing only). Survey Layer A must
  show **0 regressions**.
- `docs/`: `openapi:why` gains total field coverage — update the `why` page if it enumerates covered
  fields. The precedence model becomes documentable (the issue notes scribe documents its merge
  semantics; a short "field precedence" doc section is in scope if cheap, else defer to #484/#493).
- `CHANGELOG.md [Unreleased]` — "Changed" (internal): "`OperationBuilder` field precedence now runs
  through a single `ResolvedField` merge primitive; `openapi:why --fields` reports provenance for
  every operation field. Removed the transient explicit-status vendor extension. No spec output
  change."

### Controversial? YES — recommend human gate

- **`PrimaryResponseResolver` is in `Contracts/Registry/`** — the documented plugin extension
  surface. Widening its return type is a **breaking contract change for third-party resolvers** and
  touches 8 in-tree implementations. Per CLAUDE.md + team protocol, a `Contracts/**` change is a
  human gate. Even though the package is pre-1.0 (no back-compat obligation), the lead should route
  this to **needs-human** for Moritz to approve the new contract shape (`PrimaryResponse` result
  object) before merge.
- Also: it is the **head of the provenance block** (#483/#484/#493 build on the `ResolvedField`
  vocabulary), so the primitive's public-ish shape wants a sign-off.
- Suggested flow: coder lands Pieces 1-3 → reviewer + **mandatory survey Layer A (0 regressions)** →
  lead routes to `needs-human` for the contract-change approval. The applier extraction (Piece 2)
  and the primitive (Piece 1) are non-controversial on their own; the contract widening (Piece 3)
  is the gated part — if the lead prefers, Piece 3 could be split to a follow-up PR, but the sentinel
  deletion is cleanest done together (flag this split option for the lead).


Radiergummi:

Three pieces: (1) `ResolvedField` merge primitive turning every field's bespoke precedence chain into a declared candidate list, making `FieldProvenance` **total** (the headline); (2) pure extraction of the ~60% mechanical applier bulk into per-attribute applier classes (byte-identical); (3) widen `PrimaryResponseResolver` to a `PrimaryResponse {response, statusIsExplicit}` result object and delete the `EXPLICIT_STATUS_EXTENSION` vendor-extension sentinel.

Verified: the sentinel has exactly one writer (`InlineJsonResponseResolver`), and the return-widening touches **8** resolver implementations across Core + Fortify/ApiResources/Fractal/SpatieData.

Two flags for the reviewer:
1. **Human gate recommended** — Piece 3 changes `PrimaryResponseResolver`, which lives in `Contracts/Registry/` (the plugin extension surface). A contract change + head-of-provenance-block ⇒ Moritz should approve the `PrimaryResponse` shape. (Split option noted in the body if the lead prefers Piece 3 as a follow-up, but the sentinel deletion is cleanest done with it.)
2. **Chain safety:** I designed `ResolvedField` as a standalone merge mechanic (candidate-list in, winner+provenance out) usable **outside** `OperationBuilder`, because #493's overlay seam depends on it; and I explicitly did **not** add any linting gate to `OperationBuilder` provenance, so #483 can add its own gated component-level record without fighting a gate introduced here. Scoped strictly to #482 — no #483 registry work, no #493 overlay, no #484 lint conversion.

Safety net for Pieces 1+2 is a byte-identical-document guard; survey Layer A is the corpus-scale version (mandatory, 0 regressions — provenance additions are why-facing only).

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Lead lean on the flagged decision: SPLIT Piece 3 (PrimaryResponseResolver contract widening + sentinel deletion) into its own small follow-up PR — it touches Contracts/Registry/ and therefore needs Moritz regardless, and splitting keeps Pieces 1+2 (byte-identical mechanics) auto-mergeable so the provenance chain doesn't stall behind a human gate. Reviewer to confirm or argue the planner's cleanest-together case.

Radiergummi:
🤖 **[reviewer]** plan-review: approved with a required scope split (Piece 3 + sentinel → follow-up). Verified all four grounding claims.

I checked the plan's factual basis against source:
- **Sentinel** is exactly as described: `EXPLICIT_STATUS_EXTENSION` (`OperationBuilder:80`), written only
  by `InlineJsonResponseResolver::markStatusExplicit` (3 call sites), read+stripped by
  `takeExplicitStatusMarker` (`:818`). `InlineJsonResponseResolver` is the sole writer — confirmed.
- **Contract:** `PrimaryResponseResolver::resolvePrimaryResponse(ActionDescriptor): ?OA\Response` is a
  bare return in `Contracts/Registry/` — widening it is a genuine contract change (human gate).
- **FieldProvenance totality:** already carries `field, value, source, reason, supersededBy` — the
  exact shape a `ResolvedField` would project. The plan's call to keep `FieldProvenance` as the
  why-facing type and make `ResolvedField` the internal mechanic is correct and avoids churning
  `WhyCommand`/`OperationDescriptor`. The totality claim (add `ResolvedField` resolutions for the
  fields that have no provenance today — description/deprecated/operationId/externalDocs/parameters/
  security) is coherent; the byte-identical guard is what proves each added resolution keeps today's
  emitted value.
- **12-applier extraction:** the plan is explicit it is a pure move, guarded byte-identical, and keeps
  per-operation building (no per-concern passes; `ActionDescriptor` locality preserved). No hidden
  precedence change — the appliers own `OA\*` construction, not field precedence, which stays in the
  `ResolvedField` chains.

**Byte-identical guard — real and testable.** Snapshot the generated document for the fixture set
before, assert byte-identical after Pieces 1+2. This is a legitimate, mechanical safety net for a
pure-internal refactor, and the corpus-scale version is the mandatory survey Layer A (0 regressions).
Pieces 1+2 move zero output bytes by construction.

**Chain safety — verified against both downstream issues.**
- **#493** (overlay seam) states outright that attribute description/example/format is applied "as an
  overlay **through the shared `ResolvedField` merge primitive**" — so it *depends* on `ResolvedField`
  being usable **outside** `OperationBuilder`, field-name-keyed. The plan's one forward-compat
  requirement (design `ResolvedField` as a standalone merge mechanic: candidate list in, winner +
  provenance out; **not** a private `OperationBuilder` method) is exactly right and is the load-bearing
  API decision here. Confirm the coder honors it — it's what makes #493 buildable without a redesign.
- **#483** (producer/confidence) works on `ComponentSchemaRegistry`, a different choke point; it reuses
  the provenance *vocabulary* but not `ResolvedField`, and needs provenance to stay cheap and
  unconditional. The plan's directive "do not add a linting gate to `OperationBuilder` provenance"
  correctly keeps #483 unobstructed. Room genuinely left for both.

**Required amendment — the split (ruling on the lead's question).** Split **Piece 3 (the
`PrimaryResponseResolver` widening + `PrimaryResponse` result object) AND the `EXPLICIT_STATUS_EXTENSION`
sentinel deletion** into a follow-up PR; keep **Pieces 1+2** (the `ResolvedField` primitive + total
provenance + applier extraction) in this PR as auto-mergeable behind the byte-identical guard.

On the planner's "sentinel deletion is cleanest with Piece 3" argument: it's correct, and it *supports*
the split rather than opposing it. The sentinel and Piece 3 are the **same** flow — both concern how
the body-scanned explicit-status signal travels from `InlineJsonResponseResolver` through the
`PrimaryResponseResolver` boundary into `OperationBuilder`. Deleting the sentinel requires the widened
return to carry `statusIsExplicit`, so the sentinel deletion is *coupled to* Piece 3, not to Pieces 1+2.
Therefore the sentinel deletion should travel **with** Piece 3 into the follow-up. That yields two
internally-coherent PRs:
  - **This PR (Pieces 1+2):** pure-internal, byte-identical, touches no `Contracts/**`, no sentinel
    plumbing — cleanly auto-mergeable after review + survey.
  - **Follow-up (Piece 3 + sentinel):** the `Contracts/Registry` widening, its 8 implementations, the
    `InlineJsonResponseResolver` flag change, and the sentinel deletion — one coherent human-gated unit.

Trying to keep the sentinel deletion in Pieces 1+2 would be the *worst* split: it would drag the
`InlineJsonResponseResolver`/`OperationBuilder` explicit-status rewrite into the "pure-internal" PR
without the widened contract that makes it clean, forcing a temporary half-measure. Keep them together
in the follow-up.

**Bookkeeping for this PR (Pieces 1+2):** CHANGELOG "Changed" (internal) — the `ResolvedField`
primitive + total `openapi:why --fields` coverage (drop the sentence about removing the vendor
extension; that moves to the follow-up). Docs: update the `why` page if it enumerates covered fields.
Survey Layer A mandatory (0 regressions). No `Contracts/**` in this PR ⇒ **not** a human gate; the
follow-up (Piece 3 + sentinel) **is** the human gate.

Approved for Pieces 1+2 with the split folded in. If the lead prefers to keep all three pieces in one
PR, that's defensible too — but then the whole PR is human-gated and loses the auto-mergeable Pieces
1+2, which is the weaker outcome. Recommend the split.

Radiergummi:
🤖 **[lead]** plan-review approved → coding, WITH the scope split as plan of record: this PR = Pieces 1+2 only (ResolvedField primitive + total provenance + applier extraction; byte-identical guard). Piece 3 + EXPLICIT_STATUS_EXTENSION sentinel deletion move together to a human-gated follow-up PR. Load-bearing API requirement for the coder: ResolvedField must be usable OUTSIDE OperationBuilder (field-name-keyed) — #493 depends on it.

Radiergummi:
🤖 **[coder]** implemented Pieces 1+2 (Piece 3 + sentinel deletion descoped to the follow-up per plan-review); local gates green, remote CI green. Pushed adeb7a3a.

**Piece 1 — `ResolvedField` merge primitive + total provenance.**
- New `Support/Provenance/`: `FieldCandidate` (present/absent, keyed on `FieldDefault::Unset`, optional superseded-label), `ResolvedField::merge(field, candidates)` → winner + `supersededBy`, `->toProvenance($displayValue)` projecting to the existing `FieldProvenance`.
- **Standalone / field-name-keyed** as required for #493's overlay seam: `ResolvedField::merge` takes a field name + candidate list and is not a private `OperationBuilder` method — usable to overlay attribute values on already-resolved fields.
- Converted the bespoke summary / status / tags / description chains to declared candidate lists, and made provenance **total** across the single-valued fields that had none before: `description`, `deprecated`, `operationId`, `externalDocs`, `security`.
- Provenance stays **cheap and unconditional** — no linting gate added to `OperationBuilder` (leaves #483 room).

**Piece 2 — mechanical applier extraction (pure move).**
- Extracted request-parameter / response-header / example / link `OA\*` construction into `Support/Generator/Appliers/` (`RequestParameterApplier`, `ResponseHeaderApplier`, `ExampleApplier`, `LinkApplier`). `OperationBuilder` orchestrates and keeps field precedence; per-operation building preserved (no per-concern passes). `OperationBuilder` 1391 → 1127 lines despite Piece 1 adding provenance code.

**Byte-identical guard:** the existing `ExamplesTest` snapshot suite (8 committed `examples/*/openapi.yaml` across vanilla / api-resources / form-requests / fractal / query-builder / spatie-data / swagger-php / combined) is the representative-fixture byte-identical guard — green before and after. No output byte moved; provenance is why-facing only.

**Tests:** `ResolvedFieldTest` (11 cases: winner selection, present-only supersede accumulation, all-absent → null, single-candidate → no supersede, native-value preservation, projection). Ported/extended `OperationProvenanceTest` for the new total fields with exact source/reason assertions. Appliers are covered end-to-end by the existing header/cookie/example/link feature tests + the snapshot guard, so no redundant private-glue unit tests. Full suite 3341 pass; Pint + PHPStan clean; all remote CI legs (incl. swagger-php 5.8) green.

**Scope note (deviation):** `parameters` provenance is deliberately **not** added — a single `FieldProvenance` entry can't faithfully represent a list of N parameters, and the plan's "each is a small candidate list" doesn't hold for a multi-valued field. Per-parameter provenance is left for a future issue (noted in the docs). All other single-valued fields are covered, so `FieldProvenance` is total over the fields it can meaningfully describe.

Not marking ready / not relabeling (per lead). This is Pieces 1+2 only — no `Contracts/**` touched, sentinel plumbing untouched. Survey Layer A applies (area:core; expect 0 regressions — snapshots already prove byte-identical).

Radiergummi:
🤖 **[lead]** implementation pushed (adeb7a3a), local + remote gates green → review. Round 1: reviewer-1. Open ruling for review: the coder deliberately skipped provenance for the multi-valued `parameters` field (single FieldProvenance entry can't represent N parameters) — assess whether that deviation stands (follow-up) or needs a representation decision now.

Radiergummi:
🤖 **[reviewer]** review round 1: approved, no blocking concerns. One required record-correction (accuracy, not code) + I rule on the parameters deviation.

I re-ran the full suite locally against adeb7a3a (applied to a clean checkout), plus a scope/precedence
trace of the new `OperationBuilder`.

**Scope — the split held exactly.** `git diff --name-only` touches no `Contracts/**`, no
`InlineJsonResponseResolver`, no `PrimaryResponse`. The `EXPLICIT_STATUS_EXTENSION` sentinel and
`takeExplicitStatusMarker` are still present and untouched in `OperationBuilder`/`InlineJsonResponseResolver`
— Piece 3 + sentinel correctly deferred to the follow-up, as agreed.

**ResolvedField — meets the #493 load-bearing requirement precisely.** It is a standalone
`ResolvedField::merge(string $field, list<FieldCandidate>): ?self` — field-name-keyed, a static
mechanic explicitly decoupled from `OperationBuilder` (the docblock calls this out), with a
`toProvenance($displayValue)` projection that keeps `FieldProvenance` as the render-facing type (no
`WhyCommand`/`OperationDescriptor` churn). `FieldCandidate` uses `FieldDefault::Unset` as the
present/absent primitive with an optional `supersededLabel`. This is exactly the "candidate list in,
winner + provenance out" API #493's overlay seam needs — a future field-name-keyed overlay can call
`merge()` directly. Good.

**Precedence hunt (the class of bug this refactor risks) — clean, with one architectural nuance the
record should state plainly.** There are **two** patterns in the new `OperationBuilder`, and the
distinction matters for #493's implementer:
- **summary / description are genuinely value-driven by the merge** — `$summary = $resolved->value`
  is taken from the winning candidate. The candidate list *is* the precedence, and I checked the order
  against main: method `#[Summary]` → docblock → class `#[Summary]` → convention (unchanged).
- **status / tags / security / deprecated are provenance-only.** Their emitted *value* is still
  computed by the pre-existing inline logic (`$primaryResponse`/`applyConventionStatus`, `mergeTags`,
  `resolveSecurity`), and the `*Provenance()` methods build a candidate list where exactly one branch
  is `present($value,…)` and the rest are `absent`, purely to *record* which branch won (guarded by
  `assert($resolved !== null)` + a "exactly one branch present" invariant). The value logic is
  untouched.

This provenance-only treatment for status/tags is the **safer** realization of the plan — because the
value logic isn't recomputed from candidates, the "candidate order silently changes the emitted value"
bug is *impossible* for those fields (only summary/description carry that risk, and their order is
verified). It does **not** corner #493, which needs the standalone primitive (delivered), not that
every OperationBuilder field value route through it.

**Required record-correction (not a code change):** the coder comment says "Converted the bespoke
summary / **status / tags** / description chains to declared candidate lists." For status/tags/security
that overstates it — only *provenance* was moved to the primitive; the *value* resolution stays inline.
Please correct this in the PR record so #493's implementer knows status/tags have a **parallel**
provenance model that must be kept in sync with the inline value logic (the `*Provenance()` restatement
is not the source of truth for those values). This is the one maintenance hazard the refactor
introduces, and it should be documented, not discovered later.

**Byte-identical guard — adequate here, survey is the real net.** The `ExamplesTest` snapshot suite
byte-matches 8 committed `examples/*/openapi.yaml` flavors (vanilla / api-resources / form-requests /
fractal / query-builder / spatie-data / swagger-php / combined) — a genuine
`->toBe(file_get_contents($snapshot))` guard, green here. Given status/tags/security value logic is
untouched (see above), the residual output risk is concentrated in summary/description (value-driven)
and the applier extraction (pure move), both well-exercised by the 8 flavors + the header/cookie/
example/link feature suites. I judged a dedicated before/after fixture **not** required beyond this,
**but** since this is `area:core` and a 264-net-line OperationBuilder reduction, the mandatory survey
Layer A (corpus-scale byte comparison, 0 regressions) is the load-bearing proof — the 8 example apps
don't cover every field combination in the wild. Keep the survey gate.

**Provenance totality — verified complete over single-valued fields.** The `array_filter` provenance
block now emits summary, description, status, tags, deprecated, operationId, externalDocs, and
security. I confirmed each single-valued field the lead listed (description, deprecated, operationId,
externalDocs, security) is present and pinned in `OperationProvenanceTest` with exact source/reason
assertions. None missed.

**Ruling on the flagged `parameters` deviation: accept + follow-up issue (agree with the lead).** A
single `FieldProvenance{field, value: string, source, reason, supersededBy}` cannot faithfully
represent a list of N parameters, and the plan's "each is a small candidate list" premise doesn't hold
for a multi-valued field. Forcing a representation now would likely require changing `FieldProvenance`'s
shape — which would corner #484 (which *reads* `FieldProvenance`). The coder's reasoning is sound and
documented in the comment + docs. Per-parameter provenance is genuinely its own design problem; file it
as a follow-up. `FieldProvenance` is total over the fields it can meaningfully describe, which is the
right scope for #482.

**Conventions / gates.** strict-types; `@internal` on `ResolvedField`/`FieldCandidate`; modern PHP
(readonly, static factories, `match`, first-class-callable convention reasons); no abbreviations;
concise why-focused docblocks; no issue numbers in code. Full suite 3341 passed / 8151 assertions;
PHPStan clean; Pint clean. CHANGELOG `[Unreleased]` "Changed" entry present (correctly scoped to the
primitive + total `openapi:why` coverage; no mention of the sentinel, which is right — that's the
follow-up). Docs: `multi-spec.md`/`troubleshooting.md` touched; the parameters-provenance follow-up is
noted. No `Contracts/**` ⇒ not a human gate for this PR (the follow-up Piece 3 is).

Approved. Please fold in the record-correction (status/tags are provenance-only, value logic
unchanged) as a PR comment so the parallel-model hazard is on the record for #493/#484, and confirm the
parameters follow-up issue is filed. Survey Layer A before merge.

Radiergummi:
🤖 **[reviewer]** independent review (reviewer-2): approved, **1 non-blocking finding** (status provenance `supersededBy` divergence).

Formed my own view from #482, the amended plan (Pieces 1+2, split confirmed), and the diff at `adeb7a3a` before reconciling with reviewer-1. We converge on the architecture — and I share reviewer-1's two-pattern read (summary/description are value-driven by the merge; status/tags/security/deprecated are provenance-only, value stays inline). But the lead asked me to sweep the provenance-only fields, and that sweep surfaced one real divergence reviewer-1's pass concluded was "impossible."

### Finding 1 (non-blocking) — status `supersededBy` changed for `#[Response(2xx)]` on a resourceful route

`statusProvenance()` computes `$conventional = $convention?->convention->successStatusCode !== null` **independently of whether a higher-precedence branch already won** (unlike tags/security, whose branch booleans are each gated `!higher && …`, and unlike deprecated, a newly-total field). So for a resourceful action carrying an explicit `#[Response(2xx)]` (e.g. the suite's own `flights.update`), **both** the `#[Response] (method)` candidate and the `convention` candidate are `present`, and `ResolvedField::merge` records the convention in `supersededBy`.

I proved it directly by driving the exact candidate list `statusProvenance()` builds for that case:
```
source=#[Response] (method)  supersededBy=["ResourceConventionResolver"]
```
`main`'s `statusProvenance` (`OperationBuilder.php:1213`) is a `match(true)` that stops at the first true branch and hard-codes `$superseded = []` in **every** arm — so main recorded `supersededBy: []` here.

Consequences:
- It's a change to `openapi:why --fields` output (status now shows a superseded convention where it didn't). **Not** a document change — provenance is why-facing only, so the byte-exact snapshot guard and survey Layer A won't catch it. I confirmed the document is byte-identical (snapshot group 9/9, 926 generator/feature tests green).
- The method's own docblock is now inaccurate: *"Status records no superseded candidates (matching today's provenance), which holds because every non-deciding branch … is absent."* The convention branch is a non-deciding branch that is **not** absent in this case.
- It contradicts the PR's "new provenance entries are additive" mandate: status already had provenance, so this is a change to an existing field's provenance, not an addition. It also slips the test net — `it('reports an explicit #[Response] status as the source…')` (`OperationProvenanceTest:158`) covers `flights.update` but only asserts the source string, not `supersededBy`.

Arguably the new output is *more* informative (the convention genuinely lost). So the honest fix is a **decision**, not obviously a bug-revert: either (a) gate `$conventional` (and the `$default` computation already does gate) on the higher branches being absent, restoring `supersededBy: []` to match main + the docblock; or (b) keep the richer output but fix the docblock and add a `supersededBy` assertion to the `flights.update` test so it's pinned intent, not accident. I lean (a) for a "behaviour-preserving except total provenance" PR, but either is fine if chosen deliberately. Since this PR is not human-gated, worth resolving before merge or tracking as a fast follow.

### Everything else — verified, matches reviewer-1

- **Scope split held:** no `Contracts/**`, no `InlineJsonResponseResolver`, no `PrimaryResponse`; `EXPLICIT_STATUS_EXTENSION` + `takeExplicitStatusMarker` still present (Piece 3 + sentinel correctly deferred).
- **`ResolvedField` meets the #493 requirement:** standalone `merge(field, list<FieldCandidate>): ?self`, field-name-keyed, decoupled from `OperationBuilder`, `toProvenance()` projection keeps `FieldProvenance` as the render type (no `WhyCommand`/`OperationDescriptor` churn). `merge()` logic is correct (first present wins; lower present recorded; all-absent → null). Note `isPresent()` treats a literal `null` value as present (only `FieldDefault::Unset` is absent) — correct and consistent with the `candidate()` helper that maps `null`→absent for the value-driven fields.
- **Value-driven fields (summary/description) precedence verified against main:** method `#[Summary]` → docblock → class `#[Summary]` → convention, each a genuine value candidate (`candidate($value,…)` present only when non-null), so `supersededBy` there is truthful — this is the correct total-provenance behaviour and matches the `flights.show` superseded-convention test.
- **Applier extraction fidelity:** spot-diffed `ExampleApplier::applyResponseExamples` against main's method — byte-identical move (same try/catch, scaffolding, ordering). The 4 appliers shape the emitted document, so the snapshot byte-exact guard + header/cookie/example/link feature suites (926 tests) are a strong fidelity net; no dropped conditional or reordered write survives that.
- **Provenance totality:** description, deprecated, operationId, externalDocs, security all present and pinned with exact source/reason in `OperationProvenanceTest`. Agree with the accept+follow-up ruling on `parameters` (multi-valued field can't fit the single-value `FieldProvenance` without cornering #484).
- **PHP 8.4 / conventions:** `readonly` VOs with promoted readonly deps; `@internal` on `ResolvedField`/`FieldCandidate`/appliers; strict-types; no abbreviations; `match` used well. PHPStan level-8 clean, Pint passed.

**Verdict: approve.** The refactor is sound and the document is byte-identical. Finding 1 is a why-facing-only provenance divergence with a one-line fix (or a deliberate keep + docblock/test correction) — I'm not blocking, but it should be resolved or tracked, and the coder's "converted status/tags chains" record-correction reviewer-1 already required should mention it. Survey Layer A remains the corpus-scale byte-identity proof.

Radiergummi:
🤖 **[coder]** record correction (per reviewer-1): what the `ResolvedField` merge actually drives per field.

To keep #493 (overlay seam) and #484 (lint-readable provenance) implementers from mis-reading this PR, the precise distinction between the two kinds of field conversion here:

**Value-driven through the candidate list (the merge picks the emitted value):**
- **`summary`**, **`description`** — the candidate list *is* the precedence chain. Each candidate carries a distinct value (method attr / docblock / class attr / convention); `ResolvedField::merge` selects the winning value, and that value is what the operation emits. Change the candidate order and the emitted value changes.

**Provenance-only (the emitted value still comes from the original inline logic; the merge only records which branch won):**
- **`status`**, **`tags`**, **`security`**, **`deprecated`** — the value is computed by the pre-existing inline logic (status resolution with `applyConventionStatus`, tag `mergeTags`, `resolveSecurity`, deprecation read). The candidate list is fed the *already-resolved* value and mutually-exclusive branch booleans; the merge selects only the **source/reason** (and, deliberately, an empty `supersededBy`). It does **not** decide the value.

**Keep-in-sync hazard (important for #493/#484):** because the provenance-only fields feed the resolved value into every candidate, the candidate booleans must stay in lockstep with the inline value logic. If a future change alters *how* status/tags/security/deprecated pick their value, the matching candidate branch in `*Provenance()` must be updated too — otherwise `openapi:why` will report a source that no longer matches the emitted value. The value logic and the provenance branch logic are two views of the same decision and must not drift. (This PR only records the decision; it does not unify the two — that unification, if wanted, is #484's scope.)

Radiergummi:
🤖 **[coder]** addressed round 2 (reviewer-2 status-provenance drift); pushed 60da5e16.

**Fix (lead ruling — behavior-preserving option a):** in `statusProvenance()`, the convention candidate is now gated on no higher-precedence branch having won (`$higherBranchWon = hasResponseOverride || bodyScannedExplicit || conventionSuppressedByBody`), the same `!higher && …` pattern `tagProvenance`/`securityProvenance` already use. So an explicit `#[Response(2xx)]` (or a body-scanned / 204-suppressing status) again records `supersededBy: []`, matching main. The richer superseded record is genuinely better but belongs to #484's provenance work, not this byte-identical PR — noted in the code comment.

**Regression pin:** the `flights.update` provenance test now asserts the `--fields` output contains no `superseded` line, so the merge order can't drift silently again.

Local gate green: Pint, PHPStan, 3341 tests (8152 assertions). Snapshots still byte-identical. Watching remote CI now.

Radiergummi:
🤖 **[surveyor]** Layer A pass — byte-neutral across 11 apps.

Ran candidate adeb7a3a (pure refactor: ResolvedField merge primitive + extracted appliers) vs the fresh `main` baseline 417ee8eb (byte-neutral to d421601f via the #510 gate). All 11 apps booted, `generateExit=0`, no crashes.

- **Aggregate metrics: identical on all 11 apps** — paths / operations / apiOperations / responseSchemas / documentedResponses / requestBodies / maxRequestProperties / componentSchemas / completenessPercent all unchanged.
- **Semantic specs: identical on all 11 apps** after masking per-boot runtime tokens (Livewire URL hash, Laravel `generated::<random>` route names).

Zero deltas — consistent with a mechanical extraction that preserves generated output, matching the fixture-suite snapshot proof. Clear to merge on the survey gate.

_(Ran before reviewer-2 concludes, per lead sanction; the label stays `agent:in-review` until then. Lead notes the in-flight round-2 fix is why-facing only [provenance for `openapi:why`], document output untouched, so this result stays content-valid for the new head; will confirm the delta diff when it lands.)_

Radiergummi:
🤖 **[reviewer]** delta-check (reviewer-2) on 60da5e16 (on adeb7a3a): **finding resolved via option (a), delta clean — approved.**

`git diff adeb7a3a..60da5e16` touches exactly 2 files: the `statusProvenance` gate + its docblock, and the `flights.update` test. Nothing else moved.

**Gate matches the tags/security pattern exactly.** The fix introduces `$higherBranchWon = $hasResponseOverride || $bodyScannedExplicit || $conventionSuppressedByBody`, then gates `$conventional = !$higherBranchWon && $convention?->…successStatusCode !== null` and `$default = !$higherBranchWon && !$conventional`. That is the same discipline the other provenance-only fields already use — security gates its lower branches `!$isPublic && …` / `!$isPublic && !$hasAttribute`, tags gate `!$merges && !$replaces && …`. Exactly one candidate is ever present, so `supersededBy` is `[]` for all three, consistently.

**My original repro now yields `supersededBy: []`.** Drove the exact fixed candidate list for the resourceful `#[Response(201)]` case:
```
source=#[Response] (method)   supersededBy=[]      (was ["ResourceConventionResolver"])
pure-convention (store):      source=ResourceConventionResolver  supersededBy=[]   (unchanged)
```
The override case is restored to main's behaviour and the genuine-convention case is untouched.

**Docblock now tells the truth.** It states the convention candidate is absent "when a higher-precedence branch (an explicit #[Response], a body-scanned status, or a content-bearing body suppressing a 204) already won" — which is now accurate — and correctly defers a richer superseded record to #484.

**Test pins it against silent drift.** The `flights.update` case adds `->and(str_contains($output, 'superseded'))->toBeFalse()`; the pre-fix output rendered a "superseded:" line, so this assertion genuinely fails on the old logic and guards the merge order going forward.

**Gates (60da5e16):** provenance suite 24 passed (62 assertions, +1 for the new pin); snapshot byte-exact guard 9/9 (document still byte-identical); Pint passed; PHPStan level-8 no errors.

Delta approved. My round-1 finding is closed; the document remains byte-identical and survey Layer A stays the corpus-scale proof.

Radiergummi:
🤖 **[lead]** all gates green (2 independent approvals + round-2 delta check, full CI matrix on 60da5e16, survey Layer A byte-neutral across 11 apps, snapshot suite byte-exact) + non-controversial per run directive (internal-only, no public contract; Piece 3 deferred to human-gated #516) → squash-merged. Closes #482; worktree removed. Follow-ups: #516 (Piece 3 + sentinel), #518 (parameters provenance).

PR #515, opened 2026-07-02
feat(responses): bounded return-expression read for Resource construction — retire most #[ResponseResource]

### LINKED ISSUE #487: feat(responses): bounded return-expression read for Resource construction — retire most #[ResponseResource]
## Problem

`#[ResponseResource]` is the clearest attribute compensating for unbuilt inference. Its documented reason to exist: the return type is too loose (`JsonResponse`, base `JsonResource`, `AnonymousResourceCollection`). The Tier-0 half is built — `ResourceClassLocator` reads concrete return types, `#[Collects]`, `$collects` — and `ReturnExpressionResourceReader` exists as an AST fallback, but its shape coverage leaves common loose-return controllers uncovered, so the attribute keeps doing Tier-0/1-derivable work.

## Proposed change

Extend the bounded return-expression read to the standard construction shapes, as a whitelisted Tier-1 scan (no variable tracking across calls, unconditional statements only, consistent with `ConditionalContextPolicy`):

- `return UserResource::make($x);` / `return new UserResource($x);`
- `return UserResource::collection($xs);`
- `return response()->json(UserResource::make($x), 201);` (composition with the existing inline-JSON reader)
- one-hop `$resource = UserResource::make($x); return $resource;` via the shared `ReturnExpressionResolver`

Resolution: shape → resource class → existing `SchemaFromResource` path. Degrade gracefully + log when unmatched, per the Tier-1 charter.

## Why

Direct authoring-attribute reduction on a common idiom — the strongest "architecture gap" example from the attribute audit. This is the middle band scramble wins today via full inference; the bounded read captures it without the Tier-2 correctness risk (cf. dedoc/scramble#520). Depends on the Tier-1 consolidation issue for the shared return resolver.

Radiergummi:
Closes #487

## Implementation plan — #487

**Tier: 1** (bounded AST whitelist; unconditional statements only, no variable tracking across
calls). Builds on #486 (MERGED, PR #511): reuse the shared `ReturnExpressionResolver` — **no new
return-walk plumbing**.

### What was verified (this reframes the issue — read carefully)

Most of the construction shapes the issue lists are **already built** in
`ReturnExpressionResourceReader` (post-#486, `src/Plugins/ApiResources/Support/`):

- `X::make(...)`, `new X(...)`, `X::collection(...)`, `X::collect(...)`, `->toResource(X::class)`,
  `->toResourceCollection(X::class)`, resource-preserving chains (`->additional(...)`), and the
  **one-hop `$r = X::make($x); return $r;`** variable case (via the shared
  `ReturnExpressionResolver::resolveVariable`, wrapped by `expressionAssignedTo` :446) — **all
  already recognised** (`targetFromExpression` :515, `targetFromStaticCall` :584,
  `targetFromNewExpression` :624).

So the real coverage gap is **not** the shapes in the whitelist — it is **two things**:

**GAP 1 (the big one): `ResourceClassLocator` never routes a `JsonResponse` return to the body
scan.** `ResourceClassLocator::locate` (`src/Plugins/ApiResources/Support/ResourceClassLocator.php`)
gates *when* the reader runs:
- untyped / builtin return → body scan (`:78-83`),
- generic container (`Enumerable` subclasses only — `GenericContainerReturnType::matches` checks
  `Illuminate\Support\Enumerable`) → body scan (`:87-92`),
- base `JsonResource` → body scan (`:116-123`),
- **but a `JsonResponse`-typed return is a non-builtin named type that is not a `JsonResource`, so it
  hits `:94 if (!is_a($name, JsonResource::class)) return null;` and the body scan is NEVER
  consulted.** `JsonResponse` is the *first* loose return type the issue names — today
  `return UserResource::make($x)` in a `JsonResponse`-typed action gets no schema, even though the
  reader could resolve it. This is a **gating gap**, not a shape gap.

**GAP 2: `response()->json(UserResource::make($x), 201)` composition.** `targetFromExpression`
(:515) unwraps resource-preserving *method chains* but does not look **inside** a
`response()->json(<resource>, <status>)` call. The resource wrapped as the JSON data argument is
invisible to the reader. `InlineJsonCallReader` (Core) exposes `isJsonHelperCall(Node)` /
`isFactoryMethodCall(Node, 'json')` publicly (built on the merged `UnqualifiedHelperCall`), so the
composition can be recognised without duplicating the `response()`-helper detection.

### Proposed change (surgical, two gaps)

**Change 1 — widen `ResourceClassLocator`'s gating to consult the body scan for loose, non-resource
return types (fixes GAP 1).** Replace the bare `:94 return null` with: before giving up on a
non-`JsonResource` named return type, if the type is one of the **loose response wrapper types**
(`Illuminate\Http\JsonResponse`, and its parents `Symfony\…\JsonResponse` / `Illuminate\Http\Response`
/ `Symfony\…\Response` — the "too loose to carry a resource type" set the issue's rationale names),
consult `locateFromReturnExpression($descriptor, silent: true)`; a non-resource shape reads back null
and preserves today's behaviour. Keep it **silent** on this path (like the untyped/container paths)
since these actions overwhelmingly are not resources and a per-action notice would be noise. Anything
else (an unrelated concrete class return) keeps returning null unchanged. Encapsulate the "loose
wrapper type" predicate (mirror `GenericContainerReturnType`'s style — a small
`LooseResponseReturnType::matches(?ReflectionType)` in `Support/Routing/`, plugin-agnostic, so both
the ApiResources gating and any future consumer share it).

**Change 2 — teach `ReturnExpressionResourceReader` the `response()->json(<resource>, <status>)`
shape (fixes GAP 2).** In `targetFromExpression`, when the expression is an inline-JSON call
(detected via the Core `InlineJsonCallReader::isJsonHelperCall`, injected — see cross-plugin note),
unwrap to its **first (data) argument** and recurse `targetFromExpression` on it. The status argument
is irrelevant to the *resource class* (status is Core's inline-JSON concern; this reader only
resolves the resource shape). Degrade + log on an unmatched inner shape, per charter. Cover the
`::collection(...)` variant inside `json()` too.

- **Cross-plugin dependency note:** `InlineJsonCallReader` lives in `Plugins\Core\Support`.
  `ReturnExpressionResourceReader` is in `Plugins\ApiResources`. ApiResources depending on a Core
  class is a **plugin→plugin dependency** — check the arch boundary. Two options: (a) inject
  `InlineJsonCallReader` (Core is always present as the Core Plugin; but the arch test may forbid
  ApiResources→Core), or (b) **promote the `response()->json()` call-shape detection into
  `Support/MethodBody/`** (it is generic Tier-1 call-shape matching, already half-there via
  `UnqualifiedHelperCall`), so both Core and ApiResources depend on Support, not on each other. **(b)
  is the correct move** per CLAUDE.md ("infrastructure shared across plugins goes in `src/Support/`").
  Plan: extract a minimal `ResponseJsonCall` matcher/unwrapper to `Support/MethodBody/` (detect the
  call, expose the data-argument expression), have `InlineJsonCallReader` delegate its detection to
  it (no behavior change — green-guard Core's inline-JSON tests), and have the resource reader use
  it. **Flag for plan-review:** confirm this Support extraction is wanted vs. a narrower ApiResources-
  local detector; the arch test (`tests/Arch/CoreBoundaryTest`, `Support must not depend on any
  plugin`) is the constraint.

### Bounded-AST charter compliance

- Unconditional statements only (the reader already scans the first `STATEMENT_LIMIT` and uses
  `ConditionalContextPolicy`); the new inner-unwrap adds no variable tracking across calls.
- The one-hop variable case already degrades via `ReturnExpressionResolver` (mutation-checked since
  #486) — the composition case does not add a second hop.
- Every unmatched inner shape degrades to null + one NOTICE per action/run (existing `note()`),
  `#[ResponseResource]` remains the escape hatch.

### Files

- **New:** `Support/Routing/LooseResponseReturnType.php` (loose-wrapper predicate);
  `Support/MethodBody/ResponseJsonCall.php` (generic `response()->json()` detect + data-arg unwrap,
  if the Support-extraction option is taken).
- **Modify:** `Plugins/ApiResources/Support/ResourceClassLocator.php` (Change 1 gating),
  `Plugins/ApiResources/Support/ReturnExpressionResourceReader.php` (Change 2 composition in
  `targetFromExpression` + its docblock shape list), `Plugins/Core/Support/InlineJsonCallReader.php`
  (delegate detection to the Support matcher, if extracted).

1. **GAP 1 gating** (`ResourceClassLocatorTest` / a feature test): a controller action typed
   `: JsonResponse` returning `UserResource::make($user)` / `UserResource::collection($users)` /
   `new UserResource($user)` now resolves to the resource schema (was null). Include the parent
   `Response`/`JsonResponse` variants and confirm a **non-resource** `JsonResponse` return (e.g.
   `response()->json(['ok' => true])`) still reads back null from *this* path (Core's inline-JSON
   resolver owns that) — no double-claim.
2. **GAP 2 composition** (`ReturnExpressionResourceReaderTest`): `return response()->json(
   UserResource::make($x), 201);` and the `::collection(...)` variant resolve to the resource;
   `response()->json($nonResource, 201)` degrades to null + logs.
3. **Charter/degrade:** unmatched inner shape → null + one notice; conditional/mutated variable →
   refused (inherited from `ReturnExpressionResolver`, already tested — add a resource-flavoured case).
4. **Attribute-retirement proof (the issue's headline):** take existing fixtures that today **need**
   `#[ResponseResource]` because of a loose `JsonResponse` return, drop the attribute, and assert the
   schema is now inferred identically. This is the "retire most #[ResponseResource]" evidence.
5. **Green-guards:** full `ReturnExpressionResourceReaderTest` + `ResourceClassLocatorTest` +
   `ReturnExpressionResolutionTest` stay green; if `InlineJsonCallReader` delegates detection, Core's
   inline-JSON tests stay byte-green.

- **Additive coverage:** actions with loose response return types + a standard resource-construction
  body now get a response schema where they got none. **Survey required** (`area:responses` /
  `area:plugins`) — Layer A must show **0 regressions** and should quantify the coverage lift
  (resources newly resolved across the corpus) and any `#[ResponseResource]` now made redundant.
- `docs/`: update the resources/response page + the `#[ResponseResource]` section to state that
  `JsonResponse`-returning actions with a standard resource construction are now inferred (attribute
  needed only for genuinely loose runtime shapes). Note the new recognised shape
  (`response()->json(Resource::make(...))`).
- `CHANGELOG.md [Unreleased]` — "Added": "API Resource responses are now inferred for actions that
  return `JsonResponse` (or wrap the resource in `response()->json(...)`) via the bounded return-
  expression scan, retiring most `#[ResponseResource]` uses."

### Controversial?

- No `Contracts/**` change; no stage order; no config key. Additive inference within the Tier-1
  charter. The cross-plugin arch question (Change 2's Support extraction) is the one design decision
  for plan-review, but it is resolvable within existing conventions (Support-layer matcher). 
- **Not** an automatic human gate, but it **is** generation-affecting ⇒ **survey Layer A mandatory**.

**LOAD-BEARING ordering caveat for GAP 2 (verified — the coder must resolve this first).** Core
registers **before** config plugins (`plugins: [CorePlugin::class, ...config('openapi.plugins')]`),
so `InlineJsonResponseResolver` (Core) runs **ahead of** `ResourceResponseResolver` (ApiResources)
in the first-non-null primary-response chain. For `return response()->json(UserResource::make($x),
201)`, Core's inline-JSON resolver will likely **already emit a schema** — the resource rendered as
an inline array/object shape (whatever `UserResource::make($x)` looks like as JSON data it can read),
winning before ApiResources is ever consulted. If so, **GAP 2's composition case is a no-op as
planned**: the ApiResources reader never sees that action.

Two honest options, decide in plan-review:
  - **(A) Scope GAP 2 out of this PR.** The high-value, verifiable win is GAP 1 (the `JsonResponse`
    gating fix), which is *not* subject to this ordering problem (a bare `return UserResource::make()`
    in a `JsonResponse` method is not a `response()->json()` call, so Core's inline-JSON resolver does
    not fire, and the resource path is the only claimant). Ship GAP 1 + the one-hop/`::make`/`new`/
    `::collection` coverage it unlocks; file GAP 2 separately.
  - **(B) Keep GAP 2 but make Core's inline-JSON resolver *defer* when its `json()` data argument is
    a resource construction** (so ApiResources' richer resource `$ref` wins over an inline shape).
    That is a change to a **Core resolver's** emit decision, widening scope and needing its own test
    that the inline-JSON path yields to the resource path for `response()->json(<resource>)`.
  - **Recommendation: (A).** GAP 1 alone retires most `#[ResponseResource]` on the loose-return idiom
    the issue leads with, stays surgical, and avoids reshaping a Core resolver. GAP 2 is worth a
    follow-up once GAP 1's survey lift is measured. This keeps the PR within the Tier-1 charter and
    off any Core-resolver behavior change. **Flag for the lead/reviewer to confirm the split.**

  Either way: pin a test that the resource path and the inline-JSON path do **not** both emit for the
  same `response()->json(<resource>)` action (no double response / conflicting schema).


Radiergummi:

**This plan reframes the issue after inspecting the merged code.** Most construction shapes the issue lists (`X::make`, `new X`, `X::collection/collect`, one-hop `$r = X::make(); return $r;`) are **already built** in `ReturnExpressionResourceReader` (and it already uses the merged `ReturnExpressionResolver` from #486). The genuine gaps are two:

- **GAP 1 (the real win):** `ResourceClassLocator` never routes a **`JsonResponse`-typed** return to the body scan — at `:94` a non-builtin, non-`JsonResource` named type returns null, so `return UserResource::make($x)` in a `JsonResponse` method gets no schema today. `JsonResponse` is the first loose type the issue names. Fix = a small loose-wrapper-type predicate that sends these to the existing body scan. Not subject to any resolver-ordering problem.
- **GAP 2:** `response()->json(UserResource::make($x), 201)` composition — the resource inside the JSON call is invisible to the reader.

**Load-bearing caveat I verified for GAP 2:** Core registers before plugins, so `InlineJsonResponseResolver` (Core) runs **ahead of** `ResourceResponseResolver` (ApiResources) and will likely already emit for `response()->json(<resource>)` — making GAP 2 a **no-op as planned** unless a Core resolver is changed. I recommend **scoping GAP 2 out** (option A in the body): ship GAP 1 (retires most `#[ResponseResource]` on the loose-return idiom), file GAP 2 as a follow-up. Flagged for your call.

One design decision for review: GAP 2 (if kept) needs `response()->json()` detection shared across Core + ApiResources — I recommend extracting a matcher to `Support/MethodBody/` (arch boundary: Support must not depend on a plugin; plugin→plugin dep is the thing to avoid). If GAP 2 is scoped out, this question disappears.

Generation-affecting ⇒ **survey Layer A mandatory** (quantify the coverage lift + `#[ResponseResource]` made redundant). Within the Tier-1 charter; no Contracts change.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Lead lean: AGREE with scoping GAP 2 out (ship GAP 1, file the response()->json(<resource>) composition as a follow-up issue) — Core-resolver ordering already covers most of it and the surgical win is GAP 1. Reviewer to verify the GAP-2-is-mostly-noop claim against resolver registration order before we descope on that basis.

Radiergummi:
🤖 **[reviewer]** plan-review: approved. The reframe is correct and the recommended GAP-2 descope (option A) is the right call. Verified all three points.

**(a) Construction shapes already built + GAP 1 is the real substance — confirmed.** Post-#486,
`ReturnExpressionResourceReader` already recognises `X::make`/`new X`/`X::collection`/`X::collect`/
`->toResource(...)`/`->toResourceCollection(...)`/resource-preserving chains and the one-hop
`$r = X::make(); return $r;` variable case (via the shared mutation-checked `ReturnExpressionResolver`).
GAP 1 is exactly as diagnosed: `ResourceClassLocator::locate` at **`:94`** —
`if (!is_a($name, JsonResource::class, allow_string: true)) return null;` — a `JsonResponse`-typed
return is a non-builtin named type that is not a `JsonResource`, so it hits that bare return and the
body scan is **never** consulted, even though the untyped (`:82`), generic-container (`:91`), and
base-`JsonResource` (`:118`) paths all do consult it. So the fix is a **gating** widening, not a shape
addition: route the loose-wrapper return types to `locateFromReturnExpression($descriptor, silent: true)`
before the `:94` bail. The `silent: true` choice matches the sibling loose paths (most such actions are
not resources; a per-action notice would be noise). Encapsulating the predicate as
`LooseResponseReturnType::matches()` in `Support/Routing/` mirrors the existing
`GenericContainerReturnType` and is plugin-agnostic — good.

**(b) GAP-2-is-mostly-noop — verified against registration order.** `plugins: [CorePlugin::class,
...config('openapi.plugins', [])]` (`OpenApiServiceProvider:270`), and Core registers
`InlineJsonResponseResolver` (`CorePlugin:74`) **before** ApiResources registers
`ResourceResponseResolver` (`ApiResourcesPlugin:28`). Primary-response resolution is first-non-null,
so for `return response()->json(UserResource::make($x), 201)` Core's inline-JSON resolver runs first
and will already emit a schema — ApiResources never sees the action. The GAP-2 composition case is
therefore a no-op unless a Core resolver is changed to defer, which is option (B)'s scope creep into a
Core resolver's emit decision. **Endorse the plan's recommendation (A): ship GAP 1, file GAP 2
separately.**

I'll add a reason (A) is even stronger than the plan states: **the arch boundary makes GAP 2 costly
even when it isn't a no-op.** `tests/Arch/ConventionsTest.php:14-21` forbids any plugin importing a
sibling plugin (ApiResources→Core included), and `CoreBoundaryTest` forbids Support→plugin. So GAP 2's
option (a) "inject `InlineJsonCallReader`" is arch-illegal, and the only compliant path is the plan's
option (b) — extract a `response()->json()` matcher into `Support/MethodBody/` and have Core delegate
to it. That's a real Support extraction + a Core green-guard, for a case that's a likely no-op. Not
worth it in this PR. **Descope GAP 2 and the `Support/MethodBody/ResponseJsonCall.php` extraction
along with it** — neither belongs here. The follow-up issue should carry both, and decide (A-inline-
shape-is-fine) vs (B-resource-should-win) as its central question.

**(c) Tier-1 charter — holds.** GAP 1 adds no variable tracking across calls: it only widens *which*
return types reach the existing bounded scan, which already operates on unconditional statements within
`STATEMENT_LIMIT` and inherits the #486 mutation guard on the one-hop variable case. Unmatched inner
shapes degrade to null; the silent flag suppresses per-action noise on the loose path exactly as the
sibling paths do. `#[ResponseResource]` stays the escape hatch. No Tier-2 creep.

**Required amendments (fold into the plan before coding):**
1. **Drop GAP 2 and the `ResponseJsonCall` Support extraction entirely from this PR** (per above).
   The plan already offers this as recommendation (A) — make it the plan of record, not an option.
2. **Keep the no-double-claim test even with GAP 2 gone.** Pin that a non-resource `JsonResponse`
   return (`response()->json(['ok' => true])`) still reads back null from the ApiResources path (Core's
   inline-JSON resolver owns it) — this guards GAP 1's widening against over-claiming, and is the test
   that proves the resource path and inline-JSON path don't both fire.
3. **Attribute-retirement proof (the issue's headline) must be a real test:** take a fixture that today
   needs `#[ResponseResource]` because of a loose `JsonResponse` return, drop the attribute, assert the
   schema is now inferred identically. This is the evidence the issue leads with — keep it in scope for
   GAP 1.

**Tests / bookkeeping:** the plan names the proving tests (GAP-1 gating for `::make`/`new`/`::collection`
under `JsonResponse` + parents; no-double-claim; charter/degrade; attribute-retirement; green-guards).
CHANGELOG "Added" (reword to GAP 1 only: `JsonResponse`-returning actions with a standard resource
construction are now inferred). Docs: update the resources/response page + `#[ResponseResource]` section
to note the loose-return inference. `area:responses`/`area:plugins` ⇒ **survey Layer A mandatory** — 0
regressions, and it should quantify the coverage lift + `#[ResponseResource]` made redundant. Not a
human gate.

Approved to proceed with GAP 1 only (amendments 1-3 folded in), GAP 2 filed as a follow-up.

Radiergummi:
🤖 **[lead]** plan-review approved (GAP-2 descope is plan of record; amendments: no-double-claim test + attribute-retirement proof test). HOLDING in queue for a coding slot (2-PR cap; #513 still in survey). Coding starts when #513 exits to needs-human.

Radiergummi:
🤖 **[lead]** coding slot freed by #513 escalating to needs-human → coding. GAP 1 only per the plan of record; amendments: no-double-claim test + attribute-retirement proof test.

Radiergummi:
🤖 **[coder]** implemented (GAP 1 only); local gates green (Pint, PHPStan, 3333 tests). Pushed 4d299c34.

**The change (surgical, GAP 1):** `ResourceClassLocator::locate` gated the return-expression body scan for untyped / generic-container / base-`JsonResource` returns, but a `JsonResponse` / `Response` (or Symfony parent) return hit the bare non-resource bail (`:94`) and never reached the scan. Now a **loose response wrapper** return type routes to the existing `locateFromReturnExpression(silent: true)` before that bail. The predicate is a plugin-agnostic `Support\Routing\LooseResponseReturnType::matches()` (single `is_a` against the Symfony `Response` base, mirroring `GenericContainerReturnType`). 10-line gating widening + one new predicate file. **No Core resolver touched, no `ResponseJsonCall` extraction** — both are GAP 2, out of scope.

**Amendments folded in:**
1. **No-double-claim** (`jsonResponseNonResource`): a `response()->json(['ok' => true])` body under a `JsonResponse` type still reads back **null** from the ApiResources path, so Core's `InlineJsonResponseResolver` keeps owning it — the resource path and inline-JSON path never both fire.
2. **Attribute-retirement proof**: a `JsonResponse`-typed fixture returning `UserResource::make(...)` infers the **identical** target with and without `#[ResponseResource]` — the "retire most `#[ResponseResource]`" evidence.

Also covered: `::make` / `new X` / `::collection` under `JsonResponse`, plus the parent `Illuminate\Http\Response` and `Symfony\...\Response` variants. Charter-compliant (widens *which* return types reach the existing bounded scan; no new variable tracking).

**Fixture note:** the loose-return fixtures declare `JsonResponse` while returning a resource (the idiom under test = a framework-coerced return), a deliberate declared-type contradiction — each carries a scoped `@phpstan-ignore return.type` with a reason, matching the existing convention (`InlineJsonFixtureController.php:282`).

Docs updated: `docs/plugins.md` (return-expression section + reworked "Collection endpoints" — standard construction under `JsonResponse` is now inferred; attribute only for genuinely loose shapes) and `docs/auto-derivation.md`. CHANGELOG "Added" entry at end of section.

`area:responses`/`area:plugins` ⇒ survey Layer A applies (0 regressions; should quantify the coverage lift + `#[ResponseResource]` made redundant). Ready for review.

Radiergummi:
🤖 **[lead]** implementation pushed (4d299c34), local gates green → review. Round 1: reviewer-1. GAP 1 only per plan of record; both amendment tests included; note the fixture @phpstan-ignore convention rationale in the coder comment.

Radiergummi:
🤖 **[reviewer]** review round 1: approved, no blocking concerns. One optional predicate-narrowing note (I rule it non-blocking) + the boundary ruling the lead asked for.

I re-ran the suite locally against 4d299c34 (applied to a clean checkout) with a red-before check.

**Scope / GAP-2 leak — clean.** The diff touches only `ResourceClassLocator` (gating), the new
`Support/Routing/LooseResponseReturnType`, its test, docs, and CHANGELOG. **No Core resolver, no
`ResponseJsonCall`, no `InlineJsonCallReader`** — GAP 2 correctly stayed out, and the bounded scan
itself is unchanged (only *which* return types reach it widened).

**Gating widening — matches what I verified at plan-review.** The insertion is exactly at
`ResourceClassLocator:94`: before the bare `return null` for a non-`JsonResource` named return type,
a loose wrapper now routes to `locateFromReturnExpression($descriptor, silent: true)`. The `silent`
flag matches the sibling untyped/container paths (most such actions aren't resources; a per-action
notice would be noise). Nothing else in `locate()` changed.

**Red-before/green-after — confirmed.** With the `ResourceClassLocator` change reverted (tests +
predicate kept), the 6 GAP-1 tests fail (`::make`/`new`/`::collection` under `JsonResponse`, the
`Illuminate\Http\Response` + Symfony `Response` parents, and the attribute-retirement proof); with
the fix, all pass. The no-double-claim test stays green both ways (correct — the change only *adds* a
path, so a non-resource `JsonResponse` still reads null).

**`LooseResponseReturnType` — plugin-agnostic, mirrors the convention, one boundary nuance.** It lives
in `Support/Routing/` (no plugin dependency — passes the arch boundary), is named against and mirrors
`GenericContainerReturnType`, and elegantly covers the whole set with one
`is_a($name, Symfony\…\Response)` check since `Illuminate\Http\JsonResponse` and
`Illuminate\Http\Response` both extend the Symfony base.

**Ruling on the `RedirectResponse` boundary (the lead's question): harmless over-inclusion — I do NOT
require a change, but flag it as an optional refinement.** Because the predicate keys on the Symfony
`Response` base, it also matches `RedirectResponse`, `StreamedResponse`, and `BinaryFileResponse`
(verified: `is_a` is true for all). So a `RedirectResponse`-typed action now runs the body scan. This
is **functionally safe**: `redirect(...)`/`back()`/a stream callback never matches a resource-
construction shape, so `locateFromReturnExpression` reads back null and behaviour is unchanged
(silently), and the scan is memoised per method so the cost is negligible. I confirmed the existing
`ResponseRedirectWithoutLocationTest` and the `EdgeCaseFixtureController` redirect cases stay green.
*Optional* improvement: those subclasses aren't "too loose to carry a resource" — they *structurally
cannot* be a JSON resource, so excluding `RedirectResponse`/`StreamedResponse`/`BinaryFileResponse`
(or keying on `JsonResponse` + `Illuminate\Http\Response` + Symfony `Response`/`JsonResponse` only)
would be more precise and skip a pointless scan. Since it's harmless and the plan named "JsonResponse
and its parents" (which the Symfony base captures), I leave it to the coder's discretion — not a
round-2 blocker.

**Both amendment tests present and genuinely pinning.**
- **No-double-claim** (`jsonResponseNonResource`): `response()->json(['ok' => true])` under a
  `JsonResponse` type reads back null from the ApiResources path — the guard I required, pinning that
  Core's inline-JSON resolver keeps that claim and the two never both fire.
- **Attribute-retirement proof** (`looseReturnAttributed` vs `looseReturnWithAttribute`): identical
  inferred resource with and without `#[ResponseResource]` — the issue's headline evidence, and it's
  red-before (fails without the gating fix).

**Tier-1 charter — held.** No variable tracking added; the change only widens which return types reach
the existing bounded, unconditional scan (which already inherits the #486 mutation guard on the
one-hop variable case). Unmatched shapes degrade to null; `#[ResponseResource]` stays the escape
hatch.

**`@phpstan-ignore` convention — matches precedent.** Each loose fixture carries
`/** @phpstan-ignore return.type (deliberately loose: …) */` with a reason, matching the existing
`InlineJsonFixtureController.php:282` precedent (`@phpstan-ignore return.type (deliberately
contradicts…)`). PHPStan passes clean.

**Docs — accurate, tells the new story well.** `auto-derivation.md` adds the loose-wrapper to the
response-body row; `plugins.md` updates both the intro (adds the loose-response-wrapper case + the
suppressed-note note) and the collection-endpoints section, reframing `#[ResponseResource]` as the
escape hatch "only when the resource is not statically readable from the body" — the correct "when do
I need the attribute" story. CHANGELOG under **Added**, accurately scoped to GAP 1 with the
attribute-retirement framing.

**Gates.** API-resources + full feature suite: 790 passed / 2606 assertions. PHPStan clean. Pint
clean. No `Contracts/**`, no stage order, no config ⇒ not a human gate; `area:responses`/`area:plugins`
⇒ **survey Layer A mandatory** before merge (0 regressions; should quantify the coverage lift +
`#[ResponseResource]` made redundant, per the plan).

Approved. The RedirectResponse note is optional (coder's call); everything required is in place.

Radiergummi:
🤖 **[reviewer]** independent review (reviewer-2): approved, no blocking concerns.

Formed my own view from #487, the amended plan (GAP 1 only, GAP 2 scoped out per recommendation A), and the diff at `4d299c34` before reconciling with reviewer-1. We converge; I add empirical proof for the boundary the lead flagged.

**Over-inclusion of Redirect/Streamed/BinaryFile — empirically harmless (drove the lead's exact scenario).** `LooseResponseReturnType::matches` keys on the Symfony `Response` base, so `RedirectResponse`, `StreamedResponse`, `BinaryFileResponse` all match (verified) and now run the body scan. I built the adversarial fixture the lead described — a `StreamedResponse`-typed method that constructs a resource in a **side channel** and returns a stream — and ran it through the reader:
```
streamWithSideChannelResource(): StreamedResponse   → NULL   (side-channel resource NOT claimed)
redirectReturn(): RedirectResponse                  → NULL
jsonReturningResource(): JsonResponse (returns Resource::make) → NestedAuthorResource  (GAP-1 win)
```
The reader inspects only the **return expression**, not arbitrary body statements, so a resource built for logging/side effects is invisible; only a resource that is actually *returned* resolves. The over-inclusion costs one memoised body scan (perf-negligible) and never yields a wrong claim. Agree with reviewer-1 that narrowing the predicate is an optional refinement, not a blocker.

**Type-variant behaviour verified:**
- Union return `JsonResponse|JsonResource` → `matches=no` (`ReflectionUnionType`, not named) → returns null unchanged (no new coverage, no regression).
- Nullable `?JsonResponse` → `matches=YES` (named type, `getName()` = `JsonResponse`) → scan runs correctly.
- Unrelated concrete class (`stdClass`) → `matches=no` → returns null unchanged; the predicate never reaches beyond the `Response` hierarchy.

**No path shadowing (lead angle a).** The new branch sits behind `!is_a($name, JsonResource)` and only fires for loose wrappers; the earlier untyped/container/base-`JsonResource` paths are untouched, so nothing that resolved better before is shadowed. A non-resource return under any of those types still reads back null.

**No-double-claim boundary (lead angle d) — pinned and correct.** `does not claim a non-resource JsonResponse return` asserts `response()->json(['ok'=>true])` under a `JsonResponse` type reads back null from the ApiResources path, so Core's `InlineJsonResponseResolver` keeps that claim and the two never both fire. With ApiResources **disabled**, `ResourceClassLocator` doesn't run at all, so the gate introduces no new interaction under plugin-off configs — Core's inline-JSON path is unchanged.

**Perf (lead angle c).** The scan now runs on every loose-response-typed action, but `MethodBodyScanner` memoises per file (scoped singleton from #485), so repeated scans of a controller are cached; the incremental cost on large route sets is a single parse per file, not per action.

**TDD — red-before/green-after confirmed myself.** Reverted the `ResourceClassLocator` change (kept tests + predicate): 6 GAP-1 tests fail (`::make`/`new`/`::collection` under `JsonResponse`, the `Illuminate\Http\Response` + Symfony parents, and the attribute-retirement proof); all pass with the fix. The no-double-claim test is green both ways (correct — the change only *adds* a path).

**Scope / charter / conventions.** GAP 2 correctly absent (no Core resolver, no `ResponseJsonCall`, no `InlineJsonCallReader` touch) — avoids the Core-ordering reshape the plan flagged. `LooseResponseReturnType` lives in `Support/Routing/` (plugin-agnostic, passes the arch boundary — Support depends on no plugin), mirrors `GenericContainerReturnType`, `@internal`, strict-types, no abbreviations. Tier-1 held: no variable tracking added; only which return types reach the existing bounded scan widened; unmatched shapes degrade to null with `#[ResponseResource]` as the escape hatch. Loose fixtures carry `@phpstan-ignore return.type` with reasons, matching the `InlineJsonFixtureController` precedent.

**Docs / bookkeeping.** `auto-derivation.md` + `plugins.md` reframe `#[ResponseResource]` as the escape hatch for genuinely-unreadable shapes; CHANGELOG under **Added**, accurately scoped to GAP 1.

**Gates (branch worktree `4d299c34`):** arch + ApiResources + full feature suite 801 passed; Pint passed; PHPStan level-8 no errors.

**Verdict: approve.** Surgical, correct, charter-compliant, and the flagged over-inclusion is empirically a no-op on the emitted document. Second independent approval. `area:responses` + `area:plugins` ⇒ survey Layer A mandatory before merge (0 regressions; should quantify the coverage lift + `#[ResponseResource]` now made redundant).

Radiergummi:
🤖 **[lead]** two independent approvals → survey. Expected deltas: ADDITIVE response schemas on routes with loose (JsonResponse/Response-typed) returns constructing resources — quantify the lift and any #[ResponseResource] made redundant. Regressions: any previously-correct schema lost or changed non-additively.

Radiergummi:
🤖 **[coder]** rebased onto post-#514 `main`; pushed 6b155d4d. **Remote CI fully green on the final sha.**

- Clean rebase, **no conflicts** — #514's OperationBuilder/ResolvedField reshape doesn't overlap my locator/predicate files, and the CHANGELOG merged cleanly (my entry at the end of `### Added`). `sync-branch`'s `composer check` passed (3348 tests).
- CI re-triggered automatically on the force-push (no close/reopen needed). All cells pass: PHP 8.4/8.5 × Laravel 12/13, **swagger-php 5.8**, `quality` (Pint + PHPStan), **Snapshot (byte-exact)**, survey tooling, codecov.
- Diff vs new main unchanged (6 files, +178/-8).

The surveyor's #515 run (against 4d299c34) stays content-valid since #514 was byte-neutral. Ready for reviewer-1's round-1 findings / the Layer A verdict against 6b155d4d.

Radiergummi:
🤖 **[surveyor]** Layer A pass — 0 regressions across 11 apps; additive lift is 0 on this corpus because the target idiom is absent.

Ran candidate against the fresh `main` baseline 417ee8eb. All 11 apps booted, `generateExit=0`, no crashes. (Manifest recorded lib commit `6b155d4d` — the worktree rebased to the content-identical head mid-run; per lead this is the same content as 4d299c34, so the result stands and now matches the current PR head.)

**Aggregate: no change on any app.** Total responseSchemas 500 → 500; documentedResponses, componentSchemas, completenessPercent identical on all 11 apps.

**Per-operation 2xx classification: +0 additive / 0 changed / 0 removed.** Strict masked byte comparison also shows all 11 specs semantically identical.

**Why zero lift (not a silent no-op):** the feature fires only on an action typed `: JsonResponse` / `: Response` that returns a *constructed* `JsonResource` by convention (`Resource::make(...)` / `new Resource(...)` / `Resource::collection(...)`). I scanned every corpus app’s `app/` for a loose-response-typed action that also constructs a resource in the same file — **0 matches across the corpus.** The idiom simply doesn’t occur in these 11 apps, so there is nothing for the widened inference to act on.

**Classification result:**
- (a) ADDITIVE: none in this corpus (target pattern absent; no `#[ResponseResource]` attribute rendered redundant because none of the corpus apps use the pattern at all).
- (b) REGRESSION: none — no schema lost, downgraded, or changed; **no double-claim** on `response()->json([...])` routes (0 changed operations; inline-JSON response schemas are untouched everywhere).

Correctness of the new path is covered by the PR’s unit tests/fixtures; the corpus doesn’t exercise it. Byte-neutral here = safe. Clear to merge.

Radiergummi:
🤖 **[lead]** all gates green (2 independent approvals incl. adversarial side-channel probe, full CI matrix on 6b155d4d, survey Layer A pass — 0 regressions; 0 corpus lift because the loose-return idiom is absent from all 11 corpus apps, functionality pinned by unit tests) + non-controversial (no public contract, +178/−8) → squash-merged. Closes #487; worktree removed. GAP 2 tracked as #517.

PR #519, opened 2026-07-02
feat(core): shared return-shape descriptor service

### LINKED ISSUE #481: feat(core): shared return-shape descriptor service (container kind, item type, nullability, unions)

Radiergummi:
Closes #481

## Implementation plan — #481: shared return-shape descriptor service

**Tier: 0** — signatures + `@return` PHPDoc only, no body parsing. It *consolidates* the
structural type reads the response resolvers already do at Tier 0; the Tier-1 body scans
(`paginate()`/`->collect()`/`transform()` AST reads) stay in plugin code, per the issue's boundary
rule: **structural type shape = shared; "what does this convention mean" = plugin.**

### Grounding (verified against `origin/main` @ 21a6fd65, i.e. after #513/#514/#515)

- The type→schema engine is now **one class**: `Support\Generator\JsonSchemaFromType::fromType(Type
  $type, ?callable $leafClassSchema = null): OA\Schema` (symfony/type-info `Type`-keyed; handles
  nullable, union→`oneOf`, backed/unit enums, array shapes, collections/lists/maps, objects with
  DateTime/UUID/UrlRoutable + a leaf-class callback, builtins). `TypeNodeToSchema` was **deleted** by
  #513. So the descriptor's normalized currency is symfony `Type`, exactly what the issue specifies
  (`itemType: Type`, `unionMembers: list<Type>`) and what every consumer already feeds to `fromType`.
- Reflection→`Type` bridge already exists and is scoped-bound: `Symfony\...\TypeResolver` (bound in
  `OpenApiServiceProvider:246`, used by `SchemaFromDataClass:160` and `UriParameterResolver:171`).
- `@return`-generic extraction already exists: `Support\Routing\ReturnTypeExtractor::genericArgument()`
  → `TypeNodeResolver::genericValueClass()` (class-string of the last generic arg). Consumed today by
  `PaginatorResponseResolver`, `EloquentModelResponseResolver`, `DataResponseResolver`.
- Paginator classification is already centralized: `Enums\PaginatorKind::fromClass()` (Cursor /
  LengthAware / Simple, `is_a` on the `Illuminate\Contracts\Pagination\*` interfaces + the Spatie
  paginated-collection classes).
- `PrimaryResponseResolver::resolvePrimaryResponse(ActionDescriptor): ?OA\Response` — resolvers read
  `$descriptor->actionReflector` (`?ReflectionFunctionAbstract`) and call `->getReturnType()`
  themselves. First non-null wins, in `OperationBuilder::$primaryResponseResolvers` order.

### The duplication this removes (all Tier-0 structural reads, verified)

- collection-vs-single: `ResourceClassLocator::isCollectionType` (`is_a` ResourceCollection),
  `DataResponseResolver` (`is_a` DataCollection), `PaginatorResponseResolver::isResourceOrDataType`.
- paginator kind + item: `PaginatorResponseResolver` (native `PaginatorKind::fromClass` + `@return`
  generic) and `DataResponseResolver` (paginated deferral via `PaginatorKind::fromClass`).
- union→`oneOf` + nullability: `DataResponseResolver::resolveUnion` — hand-rolled over raw
  `ReflectionUnionType`, and it **silently drops** a `null` member (a `?Data` union is not modelled
  as nullable). This is a latent bug the descriptor fixes (see behaviour change below).

### New code

**VO — `src/Support/Routing/ReturnShape.php`** (`final readonly`, `@internal`):
```php
enum ReturnContainer { case Single; case ListOf; case Paginated; }

final readonly class ReturnShape {
    public function __construct(
        public ReturnContainer $container,
        public ?Type $itemType,               // Single: the return type; ListOf/Paginated: element type; null if unresolvable
        public ?PaginatorKind $paginatorKind, // set iff container === Paginated
        public bool $nullable,
        public array $unionMembers = [],       // list<Type>; non-empty only for a multi-class union return
    ) {}
}
```
**Refinement vs the issue's `single|list|paginated|cursor`:** cursor is a *paginator kind*, not a
distinct container — `PaginatorKind` already models Cursor/LengthAware/**Simple**. Folding cursor
into the container enum would drop the Simple distinction and duplicate `PaginatorKind`. So:
`container ∈ {Single, ListOf, Paginated}` + `paginatorKind` sub-discriminator. (Flag for plan-review.)

**Service — `src/Support/Routing/ReturnShapeResolver.php`** (`#[Scoped] final`, `@internal`):
`describe(ReflectionFunctionAbstract $reflector): ReturnShape`, merging native signature + `@return`.
Deps: `TypeResolver` (reflection→`Type`), `DocBlockParser` + `TypeNodeResolver` (docblock), and reuse
`ReturnTypeExtractor` for the generic arg. Algorithm reuses the *existing* primitives so migrated
sites stay byte-identical:
1. `nullable` = native `->allowsNull()` OR `TypeNodeResolver::isNullable(@return)`.
2. `ReflectionUnionType` → resolve each non-null member via `TypeResolver` into `unionMembers`;
   container `Single`, `itemType` = the resolved `UnionType` (so `fromType` still yields `oneOf`).
3. paginator: `PaginatorKind::fromClass(nativeClass)` non-null → `Paginated` + `paginatorKind`,
   `itemType` from `@return` generic (`genericValueClass` → `Type::object(...)`). Mirrors
   `PaginatorResponseResolver` exactly.
4. list/collection: native class is a collection base **or** `@return` is a list/`T[]`/`array<int,T>`
   (`TypeNodeResolver::listValueType`) → `ListOf`, `itemType` = element `Type`.
5. else → `Single`, `itemType` = `TypeResolver::resolve(native)` (or from `@return` when native is a
   loose wrapper / absent / builtin `array` with a documented shape).
6. Any unresolvable class / parse failure degrades to `itemType = null` (never throws), matching the
   resolvers' `class_exists`-guarded fallthrough.

The service classifies **structure only**. It never asks "is this a JsonResource / Data / my class" —
that stays in the plugin (`itemType->getClassName()` + the plugin's own `is_a`).

### Migration (each guarded by a characterization test proving byte-identical output first)

1. **`PaginatorResponseResolver`** — replace the native-`PaginatorKind` + `genericArgument` item read
   with `describe()` (`container===Paginated`, `paginatorKind`, `itemType`). Keep the `kindFromBody`
   AST fallback and the resource/data skip guards untouched.
2. **`DataResponseResolver`** — replace single/collection detection + `resolveUnion` with `describe()`
   (`unionMembers` filtered to `Data` subclasses; `nullable` now honoured). Keep the
   `GenericContainerReturnType` body-scan path.
3. **`ResourceClassLocator::isCollectionType` + collection-item generic** — replace with `describe()`.
   Keep `#[Collects]`/`$collects` and the return-expression body scan.

**Out of scope, with reason (keeps this surgical):**
- `SchemaFromDataClass::typeIsDataCollection` — operates on a **property** `Type`, not a return; a
  different axis than "return shape". Possible follow-up, not this issue.
- **Fractal** resolvers — shape comes from `#[FractalResponse]` attribute flags, not from the return
  type; nothing structural to consolidate.

**Guard rule:** if a migration turns out to change output, that site reverts to a follow-up rather
than shipping a surprise under a refactor. Only the Data union-nullable fix is an *intended* delta.

- **`tests/Unit/Support/Routing/ReturnShapeResolverTest.php`** — matrix: plain typed DTO;
  `?DTO` / `DTO|null` (nullable Single); `DTO|OtherDTO` (union, `unionMembers` size 2);
  `LengthAwarePaginator<UserResource>` (Paginated/LengthAware, item UserResource);
  `CursorPaginator<X>` (Cursor); `Paginator<X>` (Simple); bare `LengthAwarePaginator` no generic
  (item null); `Collection<UserResource>` / `array<UserResource>` / `UserResource[]` (ListOf);
  `@return array{id:int}` (Single, itemType `ArrayShapeType`); `list<array{...}>` (ListOf item
  ArrayShapeType); untyped / `mixed` (Single, item null/mixed); closure reflector; no docblock;
  unresolvable class name → degrade, no throw.
- **Characterization**: snapshot current Paginator / Data / ResourceClassLocator response output on
  representative fixtures; assert unchanged after migration. Plus one **new** test asserting the Data
  union-nullable improvement (`?SomeData` / `SomeData|OtherData|null` now nullable / correct).

### Observable behaviour, docs, changelog

- Service + VO are `@internal` — no public surface. **One observable behaviour change:** nullable /
  single-member Spatie Data **union** returns are now modelled correctly (previously the `null` member
  was silently dropped). → `CHANGELOG.md [Unreleased]` **Fixed** entry; check `docs/` responses page
  for a union example to update if one exists.
- No new config key, no new authoring attribute, no lint rule.

### Controversy / gates (expect a human gate + mandatory survey)

- Adds a **new shared analysis primitive** consumed across `Support` and three plugins — architectural
  surface worth a review even though it is `@internal`.
- **Placement decision:** VO + service in `Support\Routing` (direct siblings: `ReturnTypeExtractor`,
  `LooseResponseReturnType`, `GenericContainerReturnType`; and the `FieldDescriptor` / `ResolvedField`
  precedent for shared VOs living in `Support`). This is **not** a registry seam (no interface plugins
  implement, nothing registered on `OpenApiRegistry`), so it does **not** go in `Contracts\Registry` —
  the seam-placement rule governs implemented interfaces + their VOs, not shared Support services that
  plugins *call*. Flagged for the reviewer/lead to confirm.
- **Behaviour delta** (Data union nullability) + `area:core`/`area:responses` → **survey Layer A must
  run** (0 regressions across the corpus; the one intended addition itemized).
- Serves #479: `TypedReturnResponseResolver` (baseline) will call `describe()` and feed `itemType`
  into `JsonSchemaFromType::fromType($itemType, leafClassSchema: <ref-to-component>)`, wrapping
  `ListOf` as array-of and deferring `Paginated` to Core. The seam is shaped for that consumer.


Radiergummi:

Key decisions to sanity-check in plan-review:
- **Currency is symfony `Type`** (post-#513 the single engine is `JsonSchemaFromType::fromType(Type, ?leafClassSchema)`; `TypeNodeToSchema` is gone) — matches the issue's `itemType: Type` / `unionMembers: list<Type>`.
- **Container enum refinement**: `{Single, ListOf, Paginated}` + a `?PaginatorKind` sub-discriminator instead of the issue's flat `single|list|paginated|cursor` — reusing `PaginatorKind` keeps the Simple-paginator distinction and avoids duplicating it.
- **Placement**: `Support\Routing` (sibling to `ReturnTypeExtractor`), **not** `Contracts\Registry` — it is a shared analysis primitive plugins *call*, not a seam interface they implement. Flagged for confirmation.
- **Migration scope**: Paginator + Data + ResourceClassLocator (each behind a byte-identical characterization test); SchemaFromDataClass (property-axis) and Fractal (attribute-driven) explicitly out of scope.
- **Intended behaviour delta**: Spatie Data nullable/single-member union returns now modelled correctly (latent `resolveUnion` null-drop bug) → survey Layer A required.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Reviewer: weigh the flagged `Support\Routing` vs `Contracts\Registry` placement against the project rule (seam interfaces + their VOs live in Contracts\Registry, @internal if unstable; Support\ placement is for infrastructure plugins merely call) — decide whether ReturnShape is a seam VO (#479 consumes it as its input contract) or a shared primitive. Also confirm the resolveUnion null-drop behaviour delta is pinned by a test and flagged for survey.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved** — plan is sound, Tier-0, well-grounded. Placement confirmed as `Support\Routing`; the two flagged questions resolve in the planner’s favor. Details below, plus two watch-items for code-review (not plan blockers).

### 1. Placement — `Support\Routing` is correct (not `Contracts\Registry`)

The seam-placement rule governs *seam interfaces (the `add*Resolver` contracts) and the VOs those interfaces **return***, because the general package types against them: `OperationBuilder` uses the interface + VO as real types and `OpenApiRegistry` holds the class-string list. `ReturnShape` meets none of that:

- No interface returns it; nothing is registered on `OpenApiRegistry`; `OperationBuilder` never types against it.
- `ReturnShapeResolver` is a concrete `#[Scoped] final` service **injected via DI and called**, not implemented — the same category as `ReturnTypeExtractor`, `TypeResolver`, `DocBlockParser`, none of which live in `Contracts\`.

On the lead’s specific concern (does #479’s baseline resolver pull the type onto the extension surface?): **no.** The `PrimaryResponseResolver` seam contract is `resolvePrimaryResponse(ActionDescriptor): ?OA\Response`. `ReturnShape` is an *internal computation input* the baseline resolver uses en route to that `OA\Response`; it never appears in the interface a third-party plugin implements. A baseline resolver living in the general package is free to depend on `Support\Routing` (Support→Support), so nothing forces the VO into `Contracts\`.

The decisive precedent is stronger than the `FieldDescriptor`/`ResolvedField` one the plan cites: **`ReturnTypeExtractor` (already in `Support\Routing`) is injected into `PaginatorResponseResolver`, `EloquentModelResponseResolver`, and `DataResponseResolver` today** — a Support service plugin resolvers call, same namespace, same role. `ReturnShapeResolver` is that pattern, and the plan even folds `ReturnTypeExtractor`’s generic-unwrap into it. Keep it in `Support\Routing`.

### 2. `resolveUnion` null-drop delta — properly pinned

Confirmed the latent bug at `DataResponseResolver.php:173-207`: `resolveUnion` skips every builtin member (`$member->isBuiltin()`, line 178), and `null` is a builtin type — so `SomeData|OtherData|null` drops the null member. The single-arm path (line 84) has the same gap: `?SomeData` is a `ReflectionNamedType` with `allowsNull()===true`, resolved to a bare `$ref` with nullability discarded. The plan’s `describe()` computes `nullable` from `allowsNull()`/`isNullable(@return)` and both paths honour it. It is pinned by a dedicated new test (`?SomeData` / `SomeData|OtherData|null`), recorded as an intended **Fixed** CHANGELOG delta, and the survey Layer-A gate (0 regressions, the one addition itemized) is marked mandatory. Meets the #480 bar (divergences pinned as tests, not silent picks).

### 3. Consumer fit for #479 — covered

#479’s `TypedReturnResponseResolver` needs: plain typed DTO, `@return array{…}`/`array<K,V>`/`list<T>`, and typed DTO collections. The descriptor serves all three: `Single` + `itemType` (DTO or `ArrayShapeType`) fed directly to `fromType`; `ListOf` + element `itemType` wrapped array-of; `Paginated` deferred to Core. Keyed maps (`array<string,Foo>`) land in `Single` with the map `Type` as `itemType` and `fromType` renders `additionalProperties` — correct, since #479 feeds `Single.itemType` to `fromType` unwrapped. `nullable` + `unionMembers` give #479 union/nullable handling for free. Shape fits without rework.

### Watch-items for code-review (not plan blockers)

1. **Emitted nullable shape for a bare `$ref`.** `describe()` returning `nullable:true` is necessary but not sufficient — the migrated `DataResponseResolver` must actually render it, and under OAS 3.1 a `$ref` cannot carry a sibling `nullable`. The intended-delta test should assert the *exact* emitted schema (e.g. `anyOf: [{$ref}, {type: "null"}]`), not merely that the descriptor is nullable. This is the one spot the intended delta can silently go wrong.
2. **Hold the byte-identical guard.** The characterization snapshots must be captured pre-migration and the plan’s "if a migration changes output, revert that site to a follow-up rather than ship a surprise" rule enforced per-site — only the Data nullable fix is a sanctioned delta.

Container-enum refinement (`{Single, ListOf, Paginated}` + `?PaginatorKind`) is the right call — folding `cursor` into the container would lose the Simple/LengthAware distinction and duplicate `PaginatorKind`. Out-of-scope exclusions (`SchemaFromDataClass` property-axis, Fractal attribute-driven) are correctly reasoned.

Radiergummi:
🤖 **[lead]** plan-review approved → coding. Coder: honour the two reviewer watch-items — (1) the intended-delta test must assert the exact emitted nullable-$ref schema under OAS 3.1 (anyOf with type:null), not just describe() nullability; (2) all consolidated call sites except the pinned delta stay byte-identical (per-site guard).

Radiergummi:
🤖 **[coder]** deviation: two scope refinements against the approved 3-site migration.

**1. `ResourceClassLocator` migration (site 3) deferred — recommend a follow-up issue.** It does not consolidate cleanly at Tier 0 without violating the Support→plugin boundary:
- `isCollectionType()` is `is_a($returnType, ResourceCollection::class)` — a plugin-specific classification. The descriptor lives in `Support\` and must not import `ResourceCollection`. `ResourceCollection` resolves to a plain `ObjectType` (it is not `array`/`iterable`), so `describe()` classifies it `Single`, not `ListOf`; it cannot answer "is this a resource collection" without leaking plugin knowledge.
- The item class comes from `#[Collects]`/`$collects` (a reflection read on the collection class), not a `@return` generic or a union, so there is no union/nullability/generic logic there for the descriptor to absorb.

Making this real needs a plugin-agnostic "collection base" signal ApiResources keys off — likely marginal dedup at real byte-identical risk. Left as-is under the plan's "revert a site to a follow-up rather than ship a surprise" rule. (I attempted to file the follow-up issue but issue creation was blocked in this environment — lead, please queue it: *ResourceClassLocator: consolidate collection detection onto ReturnShapeResolver, area:responses/area:plugins, tier-0*.)

**2. `PaginatorResponseResolver` migration is kind-only.** `describe()` now owns the paginator classification (`container === Paginated` → `paginatorKind`; otherwise the body-scan fallback runs), but `resolveItemClass()` is unchanged: item stays `#[ResponseResource]`-first, then the `@return` generic. Reason — `PaginatorResponseFallbackTest::"falls back to a bare 200 when no item class is declared"` proves item must be `null` for a bare `JsonResponse` body-paginate with no attribute/generic; feeding `shape.itemType` there would wrongly resolve `JsonResponse` as the item and emit an envelope. So item resolution correctly stays out of the descriptor.

**Watch-items honoured:** (1) the intended nullable delta asserts the exact emitted schema — `oneOf: [{$ref}, {type: null}]` for `?SomeData`, and the union's `oneOf` wrapped by `{type: null}` for `SomeData|OtherData|null` (not just `describe()` nullability). (2) Every other consolidated site is byte-identical: the existing `DataResponseResolverTest` / `PaginatorResponseTest` / `PaginatorResponseFallbackTest` and the snapshot group all pass unchanged.

**Placement / #479 fit:** `ReturnShape`/`ReturnShapeResolver` in `Support\Routing` (`@internal`), sibling to `ReturnTypeExtractor` whose generic-unwrap it folds in. Container enum `{Single, ListOf, Paginated}` + `?PaginatorKind`. `ListOf` is reserved for shapes where the element type is isolated (`list<T>`, `T[]`, `array<int,T>`, bare `array`); `Collection<T>`, `array<string,T>` maps and `array{…}` shapes land in `Single` carrying the full `Type`, since `JsonSchemaFromType::fromType` renders those directly (matching the reviewer's own map-in-Single analysis) — this avoids duplicating the engine's key-semantics in the descriptor.

Radiergummi:
🤖 **[coder]** implemented; local gates green (`composer check`); remote CI green across the matrix (PHP 8.4/8.5 × Laravel 12/13, swagger-php 5.8, snapshot, quality, coverage); pushed a0e2ed7e. Ready for review. See the deviation comment above for the two scope refinements (site-3 deferral + kind-only Paginator) and the honoured watch-items.

Radiergummi:
🤖 **[lead]** implementation pushed (a0e2ed7e), local + remote CI green → review. Two independent reviewer passes required. Survey Layer A starts in parallel (mandatory for the type-resolution block; the one intended delta is the nullable-Data schema). Coder's two scope refinements (Site-3 deferral, kind-only paginator migration) are argued in the PR deviation comment — reviewers judge them against the plan.

Radiergummi:
🤖 **[reviewer]** review round 1 — 1 finding (minor; correctness clean):

**1. `tests/Unit/Support/Routing/ReturnShapeResolverTest.php` — the `array<string, T>` map → `Single` branch is untested.** The 18-case matrix is thorough, but the single most load-bearing branch for #479 has no case: a keyed map. `TypeNodeResolver::listValueType()` (`TypeNodeResolver.php:282-292`) returns the value type *only* for an integer key and falls through to `null` for a string key, so `describe()` correctly lands `@return array<string, T>` in `Single` with the full `CollectionType` as `itemType` — which is exactly the behaviour your deviation comment asserts and #479 depends on (it lists `array<K,V>` as a target case and feeds `Single.itemType` straight to `fromType`). But nothing pins it: the `collection()` case covers `Collection<int, stdClass>`, and `arrayOfObjects()`/`listOfShapes()` cover the *list* side. A regression that made a string-keyed array resolve as `ListOf` (isolating the wrong element type) would ship green. Add a fixture:

```php
/** @return array<string, stdClass> */
public function map(): array { return []; }
```
asserting `container === Single`, `itemType instanceof CollectionType`, `paginatorKind === null`. Cheap, and it locks the map-vs-list split the next issue builds on.

### Everything else — clean

**Watch-item 1 (exact nullable emission): satisfied.** The migrated `DataResponseResolver` renders nullability through the pre-existing `NullableSchema::wrap` (the OAS 3.1 `oneOf … {type: null}` idiom, not a reinvented one), and the two new tests assert the *exact* emitted schema — `oneOf: [{$ref}, {type: null}]` for `?SomeData`, and the union `oneOf` wrapped by `{type: null}` for the nullable union. The intended delta cannot silently go wrong.

**Watch-item 2 (byte-identical elsewhere): verified, not just trusted.**
- *Paginator kind:* `describe()` runs `PaginatorKind::fromClass(native)` for a native paginator and falls to `kindFromBody` otherwise — identical to the old `PaginatorKind::fromClass()` + `kindFromBody` two-step. `resolveItemClass` is untouched.
- *Data class dispatch:* the resolver still branches on `$returnType` directly for the `Data` / `DataCollection` / generic-container paths; `$shape` is consulted only for the union case, the paginated-skip, and the nullable flag. The union trigger (`unionMembers !== []`) fires iff the native type is a `ReflectionUnionType` — same condition as the old `instanceof ReflectionUnionType`; builtin members (e.g. `Data|int`) resolve into `unionMembers` but are filtered back out by the `ObjectType` + `is_a(Data)` guard, so a mixed union still collapses to a bare `$ref` exactly as before. The paginated-skip (`container === Paginated`) is reachable only under the same `PaginatorKind::fromClass(native) !== null` condition as the old guard.

**Coder’s two scope refinements — both sound.**
- *Site-3 (`ResourceClassLocator`) deferral:* correct to defer. `isCollectionType()` is `is_a(ResourceCollection)`, a plugin-specific classification the `Support` descriptor must not import; `ResourceCollection` resolves to a plain `ObjectType` (not `array`/`iterable`), so `describe()` would classify it `Single`, and the item class comes from `#[Collects]`/`$collects` (a class reflection read), not a `@return` generic or union — there is no structural union/nullability/generic logic to absorb. Consolidating would need a plugin-agnostic "collection base" signal for marginal dedup at real byte-identical risk. Tracked as #521; acceptable.
- *Kind-only Paginator migration:* correct. Feeding `shape.itemType` into item resolution would misresolve the body-scan path — a non-paginator native (`JsonResponse`) whose body paginates lands in `describe()` as `Single` with `itemType = JsonResponse` `ObjectType`, which would wrongly surface as the envelope item. Item resolution rightly stays on `#[ResponseResource]` → `@return` generic.

**Conventions / boundaries / bookkeeping.** `strict_types` on all three new files; `@internal` on `ReturnShape`, `ReturnContainer`, `ReturnShapeResolver`; no abbreviations; comments are why-focused and within width; modern PHP (enum discriminator, `readonly`, `match`, named args). `Support\Routing` imports only `Enums\PaginatorKind` and symfony `Type` — no plugin/third-party-convention dependency, boundary intact. CHANGELOG `[Unreleased]` **Fixed** entry is present and honest (itemizes the one delta, states no other output change); `docs/plugins.md` documents the union + nullable Data behaviour. No new config/attribute/lint rule, so no catalog or index row needed.

**Optional (not blocking):** an intersection-type return (`A&B`) isn’t in the matrix; it degrades to `Single` with `itemType` from `typeResolver->resolve()` (or `null` on failure) and never throws — fine to leave, since #479 doesn’t target intersections and they are rare in controller returns.

Fix the one map test and this is an approve from me.

Radiergummi:
🤖 **[lead]** review round 1: needs-changes with one minor test-gap finding (pin `array<string, T>` → Single) → back to coder. Test-only addition: production code unchanged, so the in-flight survey on a0e2ed7e stays valid; reviewer-2's independent pass continues in parallel.

Radiergummi:
🤖 **[coder]** addressed round 1: added a unit case pinning `@return array<string, stdClass>` → `Single` carrying the full `CollectionType` (the map-vs-list split #479 builds on). Test-only, no production change — spec output unchanged. `composer check` green (3376 passed, PHPStan + Pint clean); pushed 1b72f818.

Radiergummi:
🤖 **[lead]** round-1 finding addressed (1b72f818) → back in review. Lead verified `git diff a0e2ed7e..1b72f818` touches only tests/Unit/Support/Routing/ReturnShapeResolverTest.php (+16) — production code identical, so the in-flight survey on a0e2ed7e remains valid for this head. Awaiting reviewer-1 re-approval + reviewer-2 independent pass + survey result.

Radiergummi:
🤖 **[reviewer]** review round 1 (independent second pass): **approved, no blocking concerns.** Formed on my own reading of the diff; correctness traced end-to-end, both deviations judged sound.

### Nullable-Data delta — correct and byte-exactly pinned
- Emission goes through the shared `NullableSchema::wrap` (`Support\Generator`), so it is consistent with every other nullable site in the codebase, not a bespoke idiom. For a `$ref` it emits `oneOf: [{$ref}, {type: "null"}]` — valid OAS 3.1 (no sibling keyword beside `$ref`).
- The two delta tests assert the **exact emitted schema**, not just `describe()` nullability — honouring reviewer-1's watch-item #1. `?SomeData` → `oneOf:[{$ref},{type:null}]`; `SomeData|OtherData|null` → `oneOf:[ {oneOf:[refA,refB]}, {type:null} ]` (null no longer dropped).
- I checked the two nullable entry paths converge: `?SomeData` (named-nullable) and `SomeData|null` (single-member union after null-drop) both land on `nullableWrap(bare $ref, true)` → identical output. Non-nullable single Data is byte-identical (`nullable:false` → `wrap` is a no-op).

### Byte-identical guard on the non-delta paths — holds
I traced each migrated branch to confirm equivalence, not just that CI is green:
- **Union filter** changed mechanism — old `!$member instanceof ReflectionNamedType || isBuiltin()` → new `!$member instanceof ObjectType` — but is **outcome-equivalent** for the Data filter: scalars resolve to `BuiltinType` (skipped either way), class types to `ObjectType` (kept), and any non-Data class is dropped by the subsequent `is_a(Data)` regardless. The set of returns routed through the union path is unchanged (all `ReflectionUnionType`).
- **Single Data / DataCollection / non-Data union / nullable-paginator edge** all trace to identical output.
- `GenericContainerReturnType::matches(null)` returns false, so the no-docblock/null-native path is safe.

### Deviation 1 — Site-3 (ResourceClassLocator) deferral: sound
Independently confirmed: `isCollectionType` is `is_a($returnType, ResourceCollection::class)` — plugin knowledge that `Support\` must not import; `ResourceCollection` resolves to a plain `ObjectType` so `describe()` would classify it `Single`, not `ListOf`; and the item class comes from `#[Collects]`/`$collects`, not a `@return` generic — no union/nullability/generic structure for the descriptor to absorb. Consolidating it would either leak plugin knowledge into Support or risk a silent byte-divergence, so the plan's own "revert a site to a follow-up rather than ship a surprise" rule applies. Follow-up **#521** is filed (tier-0, area:plugins/responses). Scope drops 3→2 sites, well-justified.

### Deviation 2 — kind-only Paginator migration: sound
`describe()` owns the paginator *kind* (equivalent to the old `PaginatorKind::fromClass` on the native named type; non-paginator native → `kindFromBody` fallback, exactly as before), while `resolveItemClass()` stays untouched. The coder's `JsonResponse` reasoning holds: a bare `JsonResponse` that body-paginates classifies as `Single` carrying `Type::object(JsonResponse)`, so feeding `shape->itemType` into item resolution would wrongly emit a `JsonResponse` envelope — `PaginatorResponseFallbackTest` pins item = null there. Item resolution correctly stays out of the descriptor.

### #479 fit — serves the next consumer without rework
`TypedReturnResponseResolver` gets what it needs: `Single` + `itemType` (DTO / `ArrayShapeType` / map `CollectionType`) fed straight to `fromType`; `ListOf` + element type wrapped array-of; `Paginated` deferred to Core; `nullable` + `unionMembers` for free. `bareArray` (`ListOf`/null item) is distinguishable from a keyed map (`Single`/`CollectionType`), and a null `itemType` communicates the degrade case cleanly. The Single-carries-full-type choice matches how `fromType` dispatches — no key-semantics duplicated in the descriptor.

### Coverage
The unit matrix maps ~1:1 to the plan (single, nullable, union, nullable-union, three paginator kinds, no-generic, `array<T>`, `T[]`, `Collection`, `array{…}`, `list<array{…}>`, bare array, untyped, mixed, unresolvable, closure). Conventions clean: `@internal`, `final`/enum, strict-types, no abbreviations, why-focused comments, `Support\Routing` placement matches the `ReturnTypeExtractor` sibling precedent. CHANGELOG **Fixed** entry + `docs/plugins.md` union/nullable note present.

### Non-blocking observations (no changes required)
1. The single-class-plus-null union (`SomeData|null`) isn't directly tested — it converges on the `?SomeData` emission by construction, but a one-line fixture would lock that convergence against future refactors.
2. The ~6 codecov-missing lines in `ReturnShapeResolver` are defensive degrade/catch branches (union-member resolve throw, generic `!class_exists`, native resolve throw), reachable only with malformed input — not required per our edge/public-surface bar. The `genericItemType` `!class_exists` degrade (`@return Paginator<BogusClass>`) has the same observable result as the tested no-generic case, so covering it is optional.
3. Micro: `describe()` now runs on every Data/Paginator invocation and the paginator path discards the computed `itemType` — negligible, and the codebase memoizes hot accessors elsewhere. Not worth changing.

Radiergummi:
🤖 **[reviewer]** review round 2: **approved, no concerns.** The round-1 finding is resolved — `1b72f818` adds the `stringKeyedMap()` fixture (`@return array<string, stdClass>`) asserting `container === Single` and `itemType instanceof CollectionType`, pinning the map-vs-list branch #479 builds on. Test-only, production code unchanged. Everything from round 1 stands. Approved from me.

Radiergummi:

**Surveyed:** PR head `a0e2ed7e` vs a **freshly regenerated** `main` baseline at `21a6fd65` (the cached baseline `6b155d4d` was stale — it predates #513/#514/#515, so I re-ran the full corpus on current `main` first). Head has since moved to `1b72f818`, but that delta is test-only with production code identical, so this run remains valid for the merge decision.

**Metric layer (results.json, per app):** every count metric identical — paths, operations, apiOperations, responseSchemas, documentedResponses, requestBodies, maxRequestProperties, componentSchemas, completenessPercent, routesIntrospected. No app that generated cleanly on baseline crashed on the candidate.

**Content layer (per-app spec diff, sorted-key JSON):** 8/11 apps byte-identical. The 3 apps that differ are entirely non-deterministic environment/RNG noise, none attributable to this change:

| app | diff | nature |
|-----|------|--------|
| InvoiceNinja | 92 hunks / 184 lines | random `operationId` suffixes for unnamed routes (`api.generated_<rand>`) — balanced re-roll, 92 removed / 92 added, no structural change |
| AureusERP | 9 hunks / 32 lines | per-install Livewire route hash (`b62d199d`→`f92de054`) in path + derived operationId |
| SpeedtestTracker | 9 hunks / 32 lines | per-install Livewire route hash (`514b26ff`→`10217673`) in path + derived operationId |

**Intended additive delta (nullable Spatie Data returns → null branch):** not exercised by any corpus app. No app gained a `{"type":"null"}` branch, because none of the 11 apps return a nullable Data type (`?SomeData` / `SomeData|Other|null`). The refactor onto `ReturnShapeResolver` (unifying paginator + Data resolution) is therefore behavior-preserving across the entire corpus; the new nullable path is a correct no-op where unused.

Clear to merge on the survey gate.

<sub>Baseline `manifest.libraryCommit` = `21a6fd65`; candidate = `a0e2ed7e`; PHP 8.4; 11/11 apps measured, no dropped apps.</sub>

Radiergummi:
🤖 **[lead]** controversy gate: src/ diff is +299/−29 (328 changed lines > ~150 threshold; a new Support service + two plugin-resolver reworks) → needs-human per run directive. ALL other gates green: full CI matrix on 1b72f818 (incl. swagger-php 5.8 + snapshot), two independent reviewer approvals (round-1 finding resolved), survey Layer A PASS (0 regressions / 11 apps vs fresh baseline 21a6fd65; the sanctioned nullable-Data delta isn't exercised by any corpus app; 3-app diff noise is the known RNG/Livewire-hash pattern). Ready to merge on your sign-off. Merging this unblocks #479 (baseline typed-return resolver), the last link of the type-resolution chain.

Radiergummi:
🤖 **[lead]** Moritz reviewed and approved → squash-merged. All gates were green (full CI matrix on 1b72f818, 2 independent approvals, survey Layer A pass vs baseline 21a6fd65). Closes #481; unblocks #479 (baseline typed-return resolver). Follow-up #521 tracks the deferred ResourceClassLocator consolidation.

PR #520, opened 2026-07-02
feat(core): record producer/confidence in ComponentSchemaRegistry (Part A of #483)

Radiergummi:
Part of #483

## Implementation plan — #483 **Part A**: always-on `SchemaProvenance` on `ComponentSchemaRegistry` + `openapi:why` wiring

Plan-review split (accepted by the lead): this PR is **Part A only** — additive provenance metadata with
**zero change to any generated document**. Part B (gated retain-inferred-view + retiring the
`NeedsInferenceDocument` double generation) is planned as a comment on #483 and opens as its own PR
after this merges. #483 stays open until Part B lands.

**Not an inference-tier change** — generation infrastructure (`area:core`) + `area:lint` (the
`openapi:why` consumer).

### Grounding (verified against `origin/main` @ 21a6fd65)

- Single write funnel: `ComponentSchemaRegistry::buildOnce(class, Closure $factory)` →
  `register(class, schema)` (first-write-wins; `register:350` bails if the class already owns a stored
  schema). Six builders funnel through `buildOnce`: `SchemaFromFormRequest`, `SchemaFromResource`,
  `SchemaFromTransformer`, `SchemaFromDataClass`, `JsonSchemaFromType`, `EloquentModelToSchema`. The
  swagger-php harvester enters via `registerNamed()`/`registerNamedResponse()`/`registerNamedParameter()`.
- **No producer/provenance/confidence/degraded field exists** on registry entries today (grep confirmed).
- A `Support\Provenance\` namespace already exists (added in #514) housing `FieldProvenance
  {field, value, source, reason, supersededBy}`, rendered by `openapi:why --fields`
  (`src/Console/WhyCommand.php`). New provenance VOs belong there for cohesion (reviewer finding #3).

### New code

**VO — `src/Support/Provenance/SchemaProvenance.php`** (`final readonly`, `@internal`), named for
cohesion with `FieldProvenance`:
```php
final readonly class SchemaProvenance {
    /**
     * @param class-string       $producer     Builder that registered the winning schema (short-class rendered).
     * @param bool               $degraded     The producer flagged a degraded/partial result.
     * @param ?string            $reason       Short human string (why degraded / how derived).
     * @param list<class-string> $supersededBy Producers that raced this class and lost (contested).
     */
    public function __construct(
        public string $producer,
        public bool $degraded = false,
        public ?string $reason = null,
        public array $supersededBy = [],
    ) {}
}
```

**`ComponentSchemaRegistry`:**
- `private array $provenance = []` keyed by component key.
- Thread a **defaulted** `?SchemaProvenance $provenance = null` through `buildOnce()`, `register()`, and
  `registerNamed()`. Each of the six builders + the harvester passes
  `new SchemaProvenance(self::class, degraded: …, reason: …)`. Defaulted → mechanical, PHPStan-checkable,
  no forced break.
- **Contested marker (cheap tier):** when a *second, different* producer reaches an already-owned class
  via `buildOnce`, append it to the existing entry's `supersededBy`. This is the always-on, **no
  build/compare** signal — the "would have produced a *different* schema" content diagnostic requires
  building the loser's schema and belongs to Part B's gated retain-mode.
- Accessors: `provenanceFor(string $key): ?SchemaProvenance`,
  `provenanceForClass(class-string): ?SchemaProvenance`.

**`openapi:why`:** surface producer + degraded/reason (+ `supersededBy`) for a component, additive
alongside the existing `FieldProvenance` rendering. No change to existing output lines.

- **`tests/Unit/Support/Generator/ComponentSchemaRegistryTest.php`** (extend): producer recorded on
  `buildOnce` and `registerNamed`; `degraded`/`reason` round-trip; `provenanceFor` + `provenanceForClass`
  return the entry; **contested** — two different producers racing one class appends the loser to
  `supersededBy` (winner's schema unchanged, first-write-wins intact).
- **`openapi:why`** test: producer + degraded/reason surfaced for a component.

### Observable behaviour, docs, changelog

- **Generated documents are byte-identical** — provenance is side metadata, never serialized into the
  OpenAPI document. `openapi:why` gains a producer/confidence line.
- `CHANGELOG.md [Unreleased]` **Added** — component provenance (producer + confidence) surfaced in
  `openapi:why`.
- Docs: `docs/` `openapi:why` page (if present) gains the new line; no new lint rule, no config key, no
  authoring attribute.

### Controversy / gates

- **Signature touch:** `buildOnce`/`register`/`registerNamed` gain a defaulted `?SchemaProvenance`
  (all `@internal`, widely-consumed, mechanical). The real consumer (the `openapi:why` line + contested
  marker) ships in this PR, so the churn is **not speculative**.
- `area:core` → **survey Layer A runs**, but it is a formality here: zero document delta, so a 0-regression
  result is the expected and required outcome (any delta is a bug).
- No `Contracts\` touch, no config-key change in Part A. (Those land in Part B.)


Radiergummi:

Structured in two parts, both required for acceptance:
- **Part A (cheap, always-on):** `SchemaProvenance {producer, degraded, reason}` recorded on every registry write; `provenanceFor`/`provenanceForClass` accessors; `openapi:why` surfaces it. Cost = threading a defaulted `?SchemaProvenance` through `buildOnce`/`register`/`registerNamed` (three widely-consumed @internal signatures).
- **Part B (the load-bearing #208 requirement):** a **gated** retain-inferred-view mode (on iff an active rule `instanceof NeedsInferenceDocument`) that stashes the counterfactual inferred schema **and** the pre-merge inferred operation-response alongside the authored winner — off, `openapi:generate` pays nothing. Migration rules then read `InferenceView` (unchanged public shape) off the **single** primary generation; the second generation (`inferenceOnly`/`InferenceOnlyGeneration`/`SpecPipeline::withoutStage`/`NeedsInferenceDocument::excludedStages`) is deleted.

Acceptance gate = the six existing `migration.*` feature tests pass with **identical verdicts** + a cost test proving zero extra factory calls when not migration-linting.

Flags for the human gate: changes `buildOnce`/`register`/`registerNamed` signatures + `generateOne` param; deletes internal machinery; `area:core` → survey Layer A (primary documents must stay byte-identical, retention is a pure side channel). Part A is independently mergeable if the lead wants to split; Part B is the risky half.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Reviewer: the #208-ported requirement is load-bearing — verify Part B keeps the inferred view alongside the authored winner and that the gate really makes ordinary generate runs pay nothing (the zero-cost test must prove it, not assert a flag). Split option (Part A independently mergeable) is on the table; give a recommendation.

Radiergummi:
🤖 **[reviewer]** plan-review: **needs-changes** — plan is fundamentally sound and correctly satisfies the load-bearing #208 requirement, but has two concrete gaps that must be closed before coding, one placement nit, and a split recommendation.

### What's right (confirmed against `origin/main` @ 21a6fd65)

- **The #208 counterfactual requirement IS met by Part B.** The plan retains the inferred view at *both* the schema level (contested-class `$inferredView`) and the **operation-response** level (the pre-merge operation snapshot into an operations map keyed by `method␠uri`). That operation-level snapshot is exactly what provenance-by-class cannot reconstruct — verified `InferenceView` already carries `operationsByKey` with that key and a **public constructor with defaults** (`InferenceView.php:31-35`), so it can be built directly from the retained maps with its public shape (`schemaForClass`/`operationForRoute`/`schemaForName`) untouched. Rules, `SchemaEquivalence`, `ComponentSubsumption`, keep-guard stay as-is. ✓
- **Zero-cost gate is proven behaviourally, not by a flag.** The spy-factory test (single call when off / re-invoked when on / `$inferredView` empty when off) measures work done, satisfying the acceptance bar. Gate reuses the existing `instanceof NeedsInferenceDocument` signal that already triggers today's second pass (`LintRunner:275`, `inferenceExcludedStages():488`). ✓
- **Tier:** correctly self-classified as generation/lint infra, not an inference-tier change. ✓

### Findings

**1. Deletion scope misses three test consumers of the machinery being deleted (must fix).** Beyond the six migration tests, grep shows the to-be-deleted APIs are consumed by:
- `tests/Feature/SwaggerPhpInferenceOnlyGenerationTest.php` — a **dedicated feature test** of `inferenceOnly()` / `InferenceOnlyGeneration` (asserts the harvester-excluded generation omits authored schemas). Not in the plan's scope; deleting the machinery will fatal it.
- `tests/Unit/Support/Generator/Pipeline/SpecPipelineTest.php:77,82,104` — two test cases dedicated to `withoutStage`.
- `tests/Unit/Lint/InferenceViewTest.php:13` — builds `InferenceView::from(new InferenceOnlyGeneration(...))`.
The plan must state each one's disposition: delete/re-express `SwaggerPhpInferenceOnlyGenerationTest` (its intent — "harvester-excluded generation omits authored schemas" — should be re-expressed against retain-mode or dropped), drop the two `withoutStage` cases, and rewire `InferenceViewTest` to construct via the public constructor.

**2. The acceptance-bar phrasing is in tension with reality (must fix).** Four of the six "must pass with identical verdicts, no fixture edits, untouched" migration tests directly call `->inferenceOnly('default', [HarvestAuthoredAnnotationsStage::class], 'testing')` in their setup helpers to build the `InferenceView` they inject: `SwaggerPhpMigrationRuleTest:106`, `SwaggerPhpOperationMigrationRuleTest:130`, `SwaggerPhpComponentMigrationRuleTest:199`, `SwaggerPhpRedundantPropertyRuleTest:169`. (The other two — `DocumentMigrationRuleTest`, `ReplaceableRuleTest` — have no coupling.) So "tests stay untouched" is not achievable for those four: their harness *must* be rewired to build the view from the retained maps. The verdicts can and must stay identical — restate the bar as **"identical verdicts after rewiring the four helpers off `inferenceOnly()`"**, and plan that rewire explicitly.

**3. VO placement (should fix).** The plan puts the new VO in `Support\Generator\SchemaProvenance`, but an existing `Support\Provenance\` namespace already houses provenance VOs — `FieldProvenance {source, reason, supersededBy}` — rendered by the *same* `openapi:why` command this plan extends (built recently in #514). Put `SchemaProvenance` in `Support\Provenance\` alongside `FieldProvenance` and align field naming so the provenance concept stays cohesive for the why-command consumer.

**4. Strengthen the parity test to exercise nested `$ref`s (should fix).** In retain-mode the inferred schema stashed for a contested class is built while its leaf `$ref`s already resolve to *authored* winners (single generation), whereas the old double-generation produced all-inferred nested schemas. Equivalence holds only because comparison is by `$ref` *name* (class-named keys are identical either way). The parity test must include a class whose inferred schema has **nested `$ref` properties to a second contested class** — a flat-schema-only parity test would not prove this invariant.

### Public-contract touches (for the lead's controversy gate)

- **`Contracts\Lint\NeedsInferenceDocument::excludedStages()` removed** (interface degrades to a marker). Lives in `Contracts\` but is explicitly `@internal Not a stable extension point` (`NeedsInferenceDocument.php:15`) → lower controversy, but still a Contracts-namespace signature change.
- `ComponentSchemaRegistry::buildOnce`/`register`/`registerNamed` gain a defaulted `?SchemaProvenance` (all `@internal`, widely-consumed; mechanical/PHPStan-checkable).
- `OpenApiGenerationOrchestrator::generateOne` gains a `retainInferredView` param (orchestrator surface).
- Deletions: `inferenceOnly()`, `InferenceOnlyGeneration`, `SpecPipeline::withoutStage()` + `$excludedStages` (all `@internal` Support).
- **No config-key changes.**

### Split recommendation: **split into two PRs.**

Part A (SchemaProvenance + thread producer + `openapi:why` wiring + contested marker) is always-on, additive, zero output change, independently testable, and its signature churn has a real consumer (why-command + contested marker) so it isn't speculative. Part B (retain-mode + operation snapshot + retire double-gen + cross-cutting deletions + the ~4 helper rewires + findings 1/2/4) is the large, risky half. Splitting roughly halves the adversarial-review load per PR and de-risks by landing the observable concept first; since both halves are `area:core` + human-gated anyway, splitting adds no extra human-gate overhead. #483 stays open until Part B lands to close it.

Radiergummi:
🤖 **[lead]** plan-review: needs-changes → back to planning. Lead decision: ACCEPT the split. PR #520 becomes Part A only (always-on SchemaProvenance + openapi:why; additive, zero output change) — change `Closes #483` to `Part of #483` since the issue closes with Part B. Part B (retain-mode, retire double generation, deletions + the 3 test-consumer dispositions + 4 helper rewires) becomes a follow-up PR planned now but opened after Part A merges. Incorporate all reviewer must-fixes: test-consumer disposition list, corrected acceptance bar (identical verdicts after rewiring the four helpers), VO in Support\Provenance, nested-$ref parity coverage.

Radiergummi:
🤖 **[planner]** revised per plan-review + lead decision (split accepted). PR #520 is now **Part A only** (always-on `SchemaProvenance` + `openapi:why`; `Part of #483`, retitled). Reviewer finding #3 applied — VO goes in `Support\Provenance` alongside `FieldProvenance`, field-named for cohesion. **Part B is planned in full as a comment on #483** (opens after Part A merges), with all must-fixes folded in: the 3 test-consumer dispositions (finding #1), the corrected acceptance bar — identical verdicts after rewiring the four `inferenceOnly()` helpers (finding #2), and nested-`$ref` parity coverage (finding #4). Ready for plan-review re-check.

Radiergummi:
🤖 **[lead]** revised plan (Part A rescope + Part B plan on #483) → plan-review re-check by the same reviewer.

Radiergummi:
🤖 **[reviewer]** plan-review (re-check after split): **approved**. Both the rescoped PR #520 (Part A) body and the Part B durable-record comment on #483 hold to the bar.

### Part A (this PR) — clean, self-contained, correctly de-risked

- VO relocated to `Support\Provenance\SchemaProvenance` alongside `FieldProvenance`, and `supersededBy: list<class-string>` mirrors `FieldProvenance`'s naming — finding #3 fully addressed, provenance concept stays cohesive for the `openapi:why` consumer.
- Scope is genuinely additive: zero document delta, no `Contracts\` touch, no config key. The defaulted `?SchemaProvenance` on `buildOnce`/`register`/`registerNamed` has a real in-PR consumer (the `openapi:why` line + contested marker), so the signature churn isn't speculative.
- The contested marker is correctly the **cheap, no-build** signal: `buildOnce`'s short-circuit (`ComponentSchemaRegistry.php:180`) returns before the factory, and Part A threads the incoming producer in, so comparing it against the stored winner on that path costs nothing. The content-diff ("would have produced a *different* schema") is correctly deferred to Part B's gated retain-mode.

Two notes for **code-review** (not plan blockers):
- `SchemaProvenance` is `final readonly`, so "append to the existing entry's `supersededBy`" must be an immutable replace (construct a new VO with the extended list), not in-place mutation — verify the coder does this.
- Confirm the accessors handle the cycle-guard case where `reserveKey` created a key that never received a producer (`provenanceFor` returns null) — the `?SchemaProvenance` return type covers it, just needs a test touch or explicit note.

### Part B (durable record on #483) — all three must-fixes folded in verbatim

1. **Test-consumer dispositions** — all three I flagged are named with dispositions: `SwaggerPhpInferenceOnlyGenerationTest` (re-express against retain-mode), `SpecPipelineTest` `withoutStage` cases (drop), `InferenceViewTest` (rewire to public ctor). ✓ One thing for Part B's coder: the re-expressed `SwaggerPhpInferenceOnlyGenerationTest` must still cover "an authored-**only** schema has no inferred counterpart in the retained view" — that's distinct from the contested-class parity case, so don't let the "drop if parity subsumes it" escape hatch silently lose that assertion.
2. **Corrected acceptance bar** — the four coupled helpers are named with line numbers, the bar is restated as "identical verdicts after rewiring the four helpers off `inferenceOnly()`", and the two uncoupled tests are correctly left untouched. ✓
3. **Nested-`$ref` parity coverage** — the parity test is required to include a class whose inferred schema has nested `$ref` properties to a second contested class, with the correct rationale (equivalence holds only by `$ref` *name*). ✓

Public-contract touches for the human gate are enumerated (`Contracts\Lint\NeedsInferenceDocument::excludedStages()` removal, `generateOne` param, the `@internal` deletions, byte-identical primary documents). No new concerns.

**Approved for Part A merge-track (pending the second reviewer + gates); Part B durable record is sound for its own PR.**

Radiergummi:
🤖 **[lead]** plan-review re-check approved (Part A + Part B record both hold) → coding. Coder: honour the reviewer's two notes — supersededBy growth on the readonly VO is an immutable replace, not in-place mutation; accessors must handle the cycle-guard reserved-but-unbuilt key (null return).

Radiergummi:
🤖 **[coder]** implemented Part A; local gates green (`composer check`: Pint clean, PHPStan 0 errors, 3366 tests) and remote CI fully green including the swagger-php 5.8 and PHP 8.5 jobs and the byte-exact **Snapshot** job (confirms zero generated-document delta). Pushed `da709b3`. Ready for review.

**What landed (Part A only):**
- `Support\Provenance\SchemaProvenance` VO (`final readonly`, `@internal`) — `{producer, degraded, reason, supersededBy}`, `supersededBy: list<class-string>` mirroring `FieldProvenance`.
- A defaulted `?SchemaProvenance` threaded through `ComponentSchemaRegistry::buildOnce()`/`register()`/`registerNamed()`, stored in a `provenance` side map keyed by component key.
- All six builders (`SchemaFromFormRequest`, `SchemaFromResource`, `SchemaFromTransformer`, `SchemaFromDataClass`, `JsonSchemaFromType`, `EloquentModelToSchema`) + the swagger-php harvester tag their writes with `new SchemaProvenance(self::class)`.
- **Cheap contested marker:** on `buildOnce`'s short-circuit (before the factory), a *different* incoming producer is appended to the winner's `supersededBy`. Growth is an **immutable replace** (readonly VO), guarded against self-contest and duplicates; it is a no-op when the winner has no recorded producer (the cycle-guard reserved-but-unbuilt key).
- Accessors `provenanceFor(key)` / `provenanceForClass(class)` (null for a reserved-but-unbuilt key) + `allProvenance()`.
- `openapi:why --fields` gains a `Components:` block naming each referenced schema's producer and any superseded contenders.

**Reviewer's two code-review notes both honoured:** the readonly-VO immutable-replace (not in-place mutation), and the accessors returning null for the cycle-guard reserved-but-unbuilt key (with a dedicated test).

**Two minor deviations from the plan** (flagging per the workflow):
1. Builders pass producer only; `degraded`/`reason` are carried by the plumbing and covered by a direct `register(...)` round-trip test rather than fabricating per-builder degradation detection. This keeps the VO surface fully wired without inventing detection semantics the code doesn't have (Simplicity First). Part B / a future builder can set them without a signature change.
2. Added an `allProvenance(): array<string, SchemaProvenance>` enumeration accessor (not in the plan). `openapi:why` needs to *list* the block, which `provenanceFor`/`provenanceForClass` alone cannot do.

**Tests:** extended `ComponentSchemaRegistryTest` (producer on buildOnce/registerNamed; degraded/reason round-trip; contested-append with winner untouched + factory-not-run; same-producer-not-contested; dedupe; reserved-but-unbuilt → null; unreserved → null; no-provenance → empty) and `WhyCommandTest` (Components block surfaces the producer; omitted for a component-free route).

**Docs/changelog:** `docs/multi-spec.md` `--fields` section documents the `Components:` block; `CHANGELOG.md [Unreleased]` appended.

Radiergummi:
🤖 **[lead]** implementation pushed (da709b3), local + remote CI green incl. swagger-php 5.8 and byte-exact snapshot jobs → review. Two independent reviewer passes requested in parallel; coder flagged two minor plan deviations (builders pass producer only; added allProvenance() accessor) — reviewers judge those against the plan.

Radiergummi:
🤖 **[reviewer]** review round 1: **approved, no blocking concerns.** Clean, surgical, correctly scoped to Part A with a byte-identical document (snapshot job green confirms it).

**Both my plan-review notes are resolved:**
- `recordContestedProducer` (`ComponentSchemaRegistry.php:398`) grows `supersededBy` by constructing a new `SchemaProvenance` rather than mutating the readonly VO, with a comment saying why. Implicitly covered by the "records a second producer as contested" + "records each contesting producer only once" tests (they'd fail on a broken replace).
- `provenanceFor`/`provenanceForClass` return null for a cycle-guard-reserved-but-unbuilt key, explicitly tested (`ComponentSchemaRegistryTest.php`, "returns null provenance for a key reserved by the cycle guard but never built").

**Zero-cost contested property is directly proven, not just asserted:** the "records a second, different producer as contested" test passes a factory that `throw`s and confirms it never runs — the contender is recorded purely on `buildOnce`'s short-circuit before any build (`ComponentSchemaRegistry.php:193-200`). This is exactly the always-on no-build signal the plan promised.

**The two flagged deviations — both judged sound:**
1. **Builders pass producer only** (`new SchemaProvenance(self::class)`, omitting `degraded`/`reason`). Correct call: none of the seven producers has a degraded/partial concept to surface today, so passing only the producer is honest — fabricating a `degraded` flag would invent signal. The VO's defaults exist for exactly this, and the plan's `degraded: …, reason: …` was illustrative. More aligned with "no speculative work" than the plan's sketch.
2. **`allProvenance()` accessor** is genuinely consumed, not speculative surface: `WhyCommand::printComponentProvenance` (`WhyCommand.php:210`) enumerates the whole registry to render the `Components:` block after building the operation. The single-key `provenanceFor`/`provenanceForClass` accessors can't enumerate, so a full-set accessor is required. The scoping is correct — the why-command builds one operation, so the registry holds exactly that route's component closure (proven by the "omits the component-provenance block for a route referencing no components" test).

**Coverage** of the public surface is thorough: producer on `buildOnce`/`register`/`registerNamed`, by-class and by-key lookup, degraded/reason round-trip, contested (different producer / same-producer no-op / dedup), null for reserved-but-unbuilt and unreserved, no-provenance → empty; plus two `openapi:why` command tests (producer surfaced / block omitted). Conventions clean: `Support\Provenance\` placement respects boundaries, `final readonly` + `@internal`, no abbreviations, concise why-comments, no issue numbers in code. CHANGELOG under `[Unreleased]`, `docs/multi-spec.md` documents the `Components:` block with an example.

**Non-blocking observations (no changes required):**
1. The `supersededBy` rendering branch in `printComponentProvenance` (the `⇂ superseded:` bulletList, `WhyCommand.php:103-108`) has no command-level test — a contested scenario is hard to construct reliably through the command, and the underlying data is thoroughly unit-tested, so this is near-trivial glue over covered data. Optional: if a fixture route can be made to have two builders reach one class, assert the `superseded:` line; otherwise leave it.
2. Contested is a `buildOnce`-only signal; the harvester's `registerNamed` collision path doesn't append to `supersededBy` — but that path already emits `component.schema-name-collision`, and buildOnce-vs-registerNamed keys disambiguate rather than share a provenance key, so there's no lost signal in practice. Matches the agreed plan scope.
3. Five changed lines run 102-112 chars (three are the method signatures with the new defaulted param, where wrapping would hurt readability). Soft-limit only; Pint is green.

Radiergummi:
🤖 **[reviewer]** review round 1 (independent second pass): **approved, no blocking concerns.** Formed on my own tracing of the diff; I verified each property the change rests on rather than inferring from the plan.

**Zero generated-document delta — structural, not just snapshot-attested.** Provenance lives in a dedicated `$provenance` map read only by `provenanceFor` / `provenanceForClass` / `allProvenance` and `openapi:why`; document assembly (the `ComponentsStage` flush) reads `$schemas`, never `$provenance`. There is no serialization path to the metadata, so byte-identity is a property of the design, and the green byte-exact snapshot job confirms it empirically.

**Contested marker is genuinely zero-cost.** The append happens inside `buildOnce`s short-circuit (`ComponentSchemaRegistry.php:193-200`), which `return $this->reserveKey(...)` *before* `markInProgress` and *before* `$factory()`. The test "records a second, different producer as contested" proves this directly: the losing `buildOnce` is handed a factory that `throw`s, and the test passes — the factory never runs, the loser is still recorded, and the winner schema is untouched. That is exactly the always-on, no-build signal the plan promised.

**No path bypasses the contested handling.** I audited every `register()` caller: the only schema-registration entry is `buildOnce` (`:206`); the other `->register(` hit (`BaselineRegistration.php:274`) is the unrelated `Plugin::register`. So the schema-write surface funnels entirely through `buildOnce`, and the contested signal covers all of it — no direct double-`register` can silently drop a contender.

**Immutable-replace on the readonly VO.** `recordContestedProducer` grows `supersededBy` by constructing a new `SchemaProvenance`, never mutating. Guarded against self-contest (`existing->producer === contender`) and duplicates (`in_array(..., strict: true)`); the "records each contesting producer only once" test pins the dedup.

**Null for cycle-guard reserved-but-unbuilt keys.** `provenanceFor`/`provenanceForClass` return `?? null` for a key with no stored producer, and `recordContestedProducer` no-ops when the winner carries none (`$existing === null`). Explicitly tested ("returns null provenance for a key reserved by the cycle guard but never built").

**`allProvenance()` scoping is correct — verified end to end.** `WhyCommand::handle` resolves exactly one descriptor (it bails on ambiguous/no match) and calls `printFields` once; `printFields` builds only that descriptor's operations via `app(OperationBuilder::class)`, populating the same scoped `ComponentSchemaRegistry` instance `printComponentProvenance` then enumerates. So the `Components:` block lists precisely the invoked route's component closure, not the whole spec — no cross-route over-reporting. The "omits the block for a component-free route" test pins the empty case.

**Both flagged deviations — sound.**
1. *Builders pass producer only.* Correct and honest: none of the seven producers has a degraded/partial concept to report today, so emitting `degraded`/`reason` would fabricate signal. The VO defaults exist for exactly this, the plan's `degraded: …, reason: …` was illustrative, and the plumbing is still fully exercised by the direct-`register` round-trip test. This is more faithful to "no speculative work" than the plan sketch.
2. *`allProvenance()` accessor.* Genuinely consumed by `printComponentProvenance`, which must *enumerate* the registry; the single-key accessors cannot. Not speculative surface.

**Conventions / boundaries / bookkeeping.** `SchemaProvenance` is `final readonly` + `@internal` in `Support\Provenance\` (Support, no plugin dependency; plugin builders importing it is plugin→Support, allowed) and cohesively named beside `FieldProvenance`. No abbreviations; why-focused comments; no issue numbers in code. CHANGELOG `[Unreleased]` **Added** entry present; `docs/multi-spec.md` documents the `Components:` block. PHPStan/Pint green per CI.

**Non-blocking observations (no change required):**
1. The harvester records provenance on `registerNamed` (`HarvestAuthoredAnnotationsStage.php:414`) but not on `registerNamedResponse`/`registerNamedParameter` (`:211`, `:244`). That is consistent with Part A being **schema** provenance only (responses/parameters are separate component maps the `Components:` block does not render), so no lost signal here. Flagging only so Part B / any future response-or-parameter provenance knows those two paths still need threading.
2. The `supersededBy` render branch in `printComponentProvenance` has no *command-level* test (a contested scenario is awkward to stage through the CLI). The data behind it is thoroughly unit-tested and the branch is trivial glue, so this is optional — add it only if a fixture route can be made to have two builders reach one class.

Clean, surgical, correctly scoped to Part A. Approved.

Radiergummi:
🤖 **[lead]** controversy gate: src/ diff is +194/−7 (201 changed lines > ~150 threshold) and the defaulted `?SchemaProvenance` params land on the plugin-facing ComponentSchemaRegistry surface → needs-human per run directive. Everything else is green: full CI matrix (incl. swagger-php 5.8 + byte-exact snapshot), two independent reviewer approvals, zero generated-document delta verified structurally by both reviewers. Survey Layer A corpus confirmation will be appended once the surveyor frees up from PR #519. Ready to merge on your sign-off; Part B of #483 (retire double generation, plan on the issue) is blocked on this merge.

Radiergummi:
🤖 **[surveyor]** Layer A confirmation (post-merge data point) — zero document delta confirmed.

Ran the full 11-app corpus on head `da709b3` against a fresh `main` baseline at `21a6fd65` (the same baseline used for #519). This corroborates the structural zero-delta claim empirically.

- **Metric layer:** every count metric identical across all 11 apps (paths, operations, apiOperations, responseSchemas, documentedResponses, requestBodies, maxRequestProperties, componentSchemas, completenessPercent, routesIntrospected). No crashes.
- **Content layer:** after masking the corpus's known per-run RNG (random `operationId` suffixes for unnamed routes; per-install Livewire route hashes), **all 11 apps are byte-identical** to baseline. The raw diff contains only that RNG churn — no library-attributable change to any generated document.

Confirms the byte-exact snapshot job: recording producer/confidence in `ComponentSchemaRegistry` (Part A of #483) does not alter the emitted spec. No findings.

<sub>Baseline `manifest.libraryCommit` = `21a6fd65`; candidate = `da709b3`; PHP 8.4; 11/11 apps measured. Posted after merge as informational; not a merge gate.</sub>

Radiergummi:
🤖 **[lead]** Moritz reviewed and approved → squash-merged (rebased onto post-#519 main first; CHANGELOG-only conflict, both entries kept; full CI re-run green on b6614de). Post-merge corpus data point: zero document delta confirmed across all 11 apps. Part A of #483 done — Part B (retire double generation, plan in the issue comments) is now unblocked and queued.

Issue #518, opened 2026-07-02
feat(core): per-parameter provenance — represent multi-valued fields in FieldProvenance


Radiergummi:
Descoped from #482 during code review (PR #514): the `parameters` operation field is multi-valued — a single `FieldProvenance` entry cannot faithfully represent N parameters, so PR #514 covers all single-valued fields only.

Needs a representation decision (e.g. keyed per-parameter provenance entries) made together with #484's lint-readable provenance work so the shape isn't decided twice.

Context: #482, PR #514 review discussion; blocks full provenance totality claimed by the provenance block of epic #478.

Radiergummi:
🤖 **[planner]** Representation decision (made with #484, per its plan — PR #522).

**Decision: option (a) — one `FieldProvenance` entry per parameter, keyed by a compound field label** (e.g. `parameters[query:page]`), produced by a per-parameter `ResolvedField::merge` over the existing `name\0in` dedup candidates in `OperationBuilder`.

- **No shape change to `FieldProvenance`** — it is already a flat `list<FieldProvenance>`, not keyed by field, so N parameters = N entries with a compound `field` label. Per-parameter origin (a specific query/header resolver, `#[QueryParam]`, a URI placeholder, a signature binding) → `source`; the later-resolver-wins + `#[QueryParam]`-lock supersession (`OperationBuilder:153-164`) → `supersededBy`. `openapi:why --fields` renders each entry with no `WhyCommand` model change.
- **Rejected: option (b)** a nested `ParameterProvenance` sub-structure — it over-models a problem the compound-label flat entry already represents faithfully, and would churn `WhyCommand`/`OperationDescriptor` for no rendering gain.

**Implementation** rides in #484's Part 2 (the `area:core` provenance half — same `OperationBuilder` work that folds `status`/`tags`/`security`/`deprecated` into their `ResolvedField` merges), so the shape is decided and built once. If the lead splits #484 and defers Part 2, #518 closes with that follow-up PR.

Radiergummi:
[...]

PR #522, opened 2026-07-02
refactor(lint): inference-failure findings at generation time (Part 1 of #484)

Radiergummi:
Part of #484

> **Split (plan-review approved):** this PR is **Part 1** (inference-failure findings at generation
> time). **Part 2** (provenance totality) is the follow-up planned below; #484 closes with Part 2.

## Implementation plan — #484 (revised per plan-review) — inference-failure findings at generation time + provenance totality

**Not an inference-tier change** — reshapes *where* diagnostics are produced (generation vs. a
re-reflecting lint pass) and completes the provenance side channel. Changes **no generated document
bytes**. Two-PR split (approved in plan-review); the whole issue is planned below. Revised to fold in
all four plan-review must-fixes (scope gap, equivalence bar, Part-2 recast, #518 caveat).

### Grounding (verified against `origin/main`, post #482/#514/#513/#515)

- **Generation-time findings channel is proven.** `FindingsCollector::emit(Finding)` (scoped binding,
  `OpenApiServiceProvider:379-382`); `Finding {ruleId, severity, message, location, fixHint, context,
  spec}`, `context[CONTEXT_SOURCE_CLASS]` drives `#[IgnoreLint]` suppression. `SchemaFromFormRequest`
  emits `request-body.schema-degraded` at the failure site with the exact reason (`:117`, `:199-226`);
  `RequestBodySchemaDegraded` is a pure **registration stub** (`implements Rule`, no visitor).
- **The re-reflecting inference-failure rules** (the issue's four named targets):
  - `ActionMethodMissing` (`operation.action-method-missing`, `RouteRule`): `new ReflectionClass` +
    `hasMethod` (`:49`).
  - `ActionMissingReturnType` (`operation.return-type-missing`, `OperationRule`): re-reads
    `getReturnType()` + Response attributes (`:71`).
  - `ThrowsTransitiveMissing` (`throws.transitive-missing`): re-runs the `@throws` walk via
    `ThrowsExtractor` on `handle()` (`:126-131`) — a walk `ThrowsErrorContributor` already performs at
    generation (`:129`) — the double-walk.
  - **"schema missing/empty observations"** = two ApiResources rules (plan-review finding 2, verified):
    `resource.response-empty` (`ResourceResponseEmpty.php`) and `resource.fields-undeclared`
    (`ResourceFieldsUndeclared.php`) each do `new ReflectionClass($resourceClass)` **and re-run**
    `ResourceToArrayReader->read($resourceClass)` at lint time — re-deriving whether the generator could
    infer the resource shape, re-running a read the ApiResources response resolver already performs at
    generation. Same double-walk pattern, in a plugin.
- **Provenance state (post-#514).** `summary`/`description` derive value **and** provenance from one
  `ResolvedField::merge` (no drift); `status`/`tags`/`security`/`deprecated` compute the value **inline**
  and a separate provenance method re-derives the decision — the keep-in-sync hazard, marked in-code at
  `OperationBuilder` `statusProvenance()` (`:826`, "A richer superseded record for status belongs to
  #484") and `securityProvenance()` (`:964`, "Precedence mirrors resolveSecurity()"). `parameters` has
  **no** provenance entry (the `name\0in` dedup at `:153`).
- **No lint rule reads `FieldProvenance`** today (grep-confirmed); `OperationDescriptor.$provenance` is a
  `list<FieldProvenance>` omitted from `toArray()`, read only by `WhyCommand`.

---

### Part 1 (PR-A) — inference-failure findings at generation time, with reasons.

Move each target to the `request-body.schema-degraded` pattern: the code that bails emits a `Finding`
carrying the **precise reason**; the rule class becomes a registration stub retaining its existing
`$id`/`$severity`/`$description`.

- `operation.action-method-missing` → emit in **`RouteIntrospector`** when the resolved controller action
  method does not exist. `ActionMethodMissing` → stub.
- `operation.return-type-missing` → emit in **`OperationBuilder`** when no `PrimaryResponseResolver`
  produced a response, no `#[Response]`/`#[ResponseResource]` is present, and the return type is
  absent/`mixed`/`void`/`never` (reason distinguishes untyped vs. `mixed`/`void`/`never`; inject
  `FindingsCollector`). `ActionMissingReturnType` → stub.
- `throws.transitive-missing` → emit in **`ThrowsErrorContributor`** (already walks at `:129`), retiring
  the duplicate walk. `ThrowsTransitiveMissing` → stub; delete its `ThrowsExtractor` re-run.
- **`resource.response-empty` + `resource.fields-undeclared`** (plan-review must-fix — the 4th target,
  **converted here**) → emit from the **ApiResources generation path** (the resource response resolver /
  `ResourceToArrayReader` already reads the resource and knows the shape is empty/undeclared; inject
  `FindingsCollector` there). Both rules → stubs; drop their lint-time `ReflectionClass` +
  `ResourceToArrayReader->read()` re-run.
  - **Area/gate consequence:** converting these two touches the **ApiResources plugin** generation code,
    so Part 1 carries **`area:plugins` in addition to `area:lint`** → **survey Layer A runs**, a 0-delta
    formality (findings never touch the document; any delta is a bug). **Lead's call:** if you want Part 1
    to stay the pure-`area:lint`, survey-skipped unit, peel these two into a sibling **Part 1b**
    (`area:plugins`, identical pattern); otherwise they ride here. Recommendation: keep them in Part 1 —
    it is the complete "inference-failure findings at generation time" unit the issue describes, and a
    0-delta survey is cheap.

Each finding sets `location` (from the reflector in hand) and `context[CONTEXT_SOURCE_CLASS/_MEMBER]` so
`#[IgnoreLint]` suppression + `severity_overrides` keep working.

**Equivalence bar (plan-review finding 3 — clarified).** Lint-output identity is
**`{ruleId, severity, location, suppression}`** — the same findings fire at the same locations and
severities with unchanged `#[IgnoreLint]`/`severity_overrides` behaviour. **Message/reason is the
sanctioned delta and is deliberately NOT pinned** (the reason string is the value this issue adds).
Tests:
- Equivalence: for the fixture suite, assert the `{ruleId, severity, location, suppression}` set is
  identical before/after, for all four targets.
- **Positive reason assertion:** assert the enriched reason actually appears — e.g. the
  untyped-vs-`mixed`/`void`/`never` distinction is present in the `operation.return-type-missing`
  message, the specific missing exception in `throws.transitive-missing`, and the empty-vs-undeclared
  distinction for the resource findings.
- **False-positive/negative-direction fixtures (the load-bearing guard — the new emissions *predict*
  document absence from build-time signals, so extensional equality with the old document-observation
  must be proven).** The finding must **NOT** fire when: a `PrimaryResponseResolver` *did* produce a
  response (not only the `#[Response]`-suppresses case); the `@throws` walk *found* the exception; the
  ApiResources reader *did* resolve a non-empty declared shape. Plus `#[IgnoreLint]` on the source class
  still suppresses. With `area:lint` skipping survey for the core rules, these fixtures are the only guard.

**Docs/changelog:** `CHANGELOG.md [Unreleased]` **Changed** (internal/lint) — inference-failure findings
carry the generation-time reason; same rule IDs (still catalog-listed via the stubs), no new config key,
no `Contracts/**`.

---

### Part 2 (PR-B, follow-up) — provenance totality (why-command-facing). `area:core`, byte-identical guard.

**Recast per plan-review finding 4.** The original "make `FieldProvenance` lint-readable to retire
re-reflection in origin-derivation rules" framing is **dropped**: after Part 1 stubs the re-reflecting
rules (including the resource rules), **no surviving lint consumer reads provenance** — threading
`OperationDescriptor.$provenance` to the lint layer would be infrastructure with no consumer. Part 2 is
honestly recast as **`openapi:why --fields` provenance totality**, which stands on its own:

1. **Kill the keep-in-sync hazard** for `status`/`tags`/`security`/`deprecated`: fold each field's inline
   value computation into its `ResolvedField::merge` candidate list so value **and** provenance come from
   one resolution (the `summary`/`description` pattern), and record the **richer `supersededBy`** #514
   deferred (every lower-precedence present candidate). Byte-identical guard: the merge winner must equal
   today's inline value for every fixture.
2. **#518 per-parameter provenance.** Implement the recorded #518 decision (option (a)): one
   `FieldProvenance` entry per parameter, keyed by the compound label `parameters[<in>:<name>]`, produced
   by a per-parameter `ResolvedField::merge` over the `name\0in` candidates. No shape change to
   `FieldProvenance`. **Recorded caveat (plan-review finding 5):** the compound label is a *display*
   convention, not a structured key — if a future rule ever needs programmatic `(name, in)` lookup,
   string-parsing it is fragile (names can contain `:`/`]`); that is where a structured key / option (b)
   earns its cost. Logged on #518 as a conscious tradeoff.

**Tests:** byte-identical document guard (values unmoved); `openapi:why --fields` shows the four
provenance-only fields with richer `supersededBy` and one entry per parameter.

**Docs/changelog:** `CHANGELOG.md [Unreleased]` **Changed** (internal) — total provenance over the
provenance-only fields + per-parameter entries; update the `openapi:why` covered-fields list. `area:core`
⇒ **survey Layer A** (0-delta by construction).

---

### Controversy / gates

- **Part 1:** no `Contracts/**`, no output-byte change. Pure-lint core rules → auto-mergeable, survey
  skipped; the ApiResources resource-rule conversion adds `area:plugins` → 0-delta survey (or peel to
  Part 1b per the lead). Correctness guard is the false-positive/negative fixtures (finding 3).
- **Part 2:** `OperationBuilder` value computation for four fields + parameter provenance → byte-identical
  guard + survey Layer A. No `Contracts/**`, no config key.
- Depends only on merged #482 (`ResolvedField`). Independent of #519/#520/#523.


Radiergummi:

Planned the full issue; **recommend a two-PR split** (mirrors #514 / #483):
- **Part 1 (headline, `area:lint`, auto-mergeable, no survey):** move `operation.action-method-missing` / `operation.return-type-missing` / `throws.transitive-missing` to the proven `request-body.schema-degraded` generation-time pattern — the code that bails emits the finding **with its reason**; the rule classes become registration stubs. Findings stay behaviour-equivalent (same IDs/severities/locations/suppression). Retires the duplicate `@throws` walk (`ThrowsErrorContributor` already walks at `:129`).
- **Part 2 (follow-up, `area:core`, byte-identical guard + survey):** make `FieldProvenance` lint-readable; fold `status`/`tags`/`security`/`deprecated` value computation into their `ResolvedField` merges to kill the keep-in-sync hazard + record richer `supersededBy` (both deferred here by #514); and #518 per-parameter provenance.

**#518 decided** (recorded on the issue): option (a), one `FieldProvenance` per parameter via compound field label, no shape change; option (b) nested struct rejected.

Flags: Part 1 touches no `Contracts/**` and moves zero document bytes → auto-mergeable. Part 2 touches `OperationBuilder` value computation → byte-identical guard + survey Layer A. Lead/reviewer rule on the split.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Reviewer: rule on (1) the proposed two-PR split (Part 1 lint-rules→generation-time findings, auto-mergeable; Part 2 lint-readable FieldProvenance + keep-in-sync fold + #518); (2) the #518 option-(a) compound-label decision (parameters[query:page]) — it's recorded on #518 and Part 2 builds on it, so it must be right; (3) whether Part 1's 'behaviour-equivalent findings' claim is pinned by tests (identical rule IDs, messages, spans).

Radiergummi:
🤖 **[reviewer]** plan-review: **needs-changes** — the two-PR split is correct and the approach is well-grounded (verified `statusProvenance`'s "#484" comment at `OperationBuilder:826`, `securityProvenance`'s "Precedence mirrors resolveSecurity()" at `:964`, the `name\0in` param dedup at `:153`, and `ResolvedField`'s standalone `@internal` status). Three refinements needed before Part 1 codes; one is a real scope gap.

### 1. The two-PR split — approved, genuinely separable
No hidden Part-1→Part-2 data dependency: Part 1's `operation.return-type-missing` emission and Part 2's `status`/`tags`/`security`/`deprecated` merge rework touch different code paths in `OperationBuilder` (response-absence diagnostic vs. field value/provenance merge) with no shared mutable state. The split mirrors #514 (Pieces 1+2 vs 3) and #483 (A/B) and keeps the low-risk, survey-skipped Part 1 unblocked. **But** see finding 4 — the split exposes a consumer-justification hole in Part 2 that the split PR should resolve, not inherit silently.

### 2. Scope gap (must fix) — the issue's 4th named target is omitted
The issue explicitly lists **four** conversion targets: `operation.return-type-missing`, `throws.transitive-missing`, `operation.action-method-missing`, **and "schema missing/empty observations."** Part 1 converts the first three and is silent on the fourth. That fourth maps to two concrete ApiResources rules that are textbook inference-failure-in-disguise:
- `resource.response-empty` (`ResourceResponseEmpty.php`) and `resource.fields-undeclared` (`ResourceFieldsUndeclared.php`) each do `new ReflectionClass($resourceClass)` **and re-run `ResourceToArrayReader->read($resourceClass)`** at lint time — re-deriving whether the generator could infer the shape, re-running a read the ApiResources response resolver already performs at generation. This is the exact `throws.transitive-missing` double-walk pattern, in a plugin.

The plan must map every issue-named target to *converted (where)* or *stays (why)*. If these defer to a follow-up that's a fine call, but state it; silence reads as an oversight, and they're the cleanest examples of the pattern the issue is built around.

### 3. Part 1 equivalence bar (must clarify) — message is intentionally NOT pinned
The lead's brief asked for "identical rule IDs, severities, **messages**, and locations." That conflicts with the issue's whole point (messages *improve* to carry the reason). The plan's equivalence set — `{ruleId, severity, location, context}` + suppression — is the **correct** bar; message equivalence must *not* be required. Two asks:
- Make this explicit in the plan: lint-output identity = same findings fire at the same locations, severities, and `#[IgnoreLint]`/`severity_overrides` behaviour; the message/reason is the sanctioned delta.
- Add a **positive** assertion that the enriched reason is present (e.g. the untyped-vs-`mixed`/`void`/`never` distinction actually appears), so the value-add is tested, not just the equivalence.
- The load-bearing correctness risk is **false-positive/negative divergence**: the old rules observe document-level absence; the new emissions *predict* it from build-time signals. The equivalence fixtures must include cases where the finding should **not** fire because a response *was* produced by a resolver (not just the `#[Response]`-suppresses case already listed) and where the `@throws` walk *found* the exception — those prove the build-time predicate is extensionally equal to the document observation. With `area:lint` skipping the survey, fixtures are the only guard here.

### 4. Part 2 item 1 (flag for the Part-2 PR) — name the concrete lint consumer or recast it
"Make `FieldProvenance` lint-readable to retire re-reflection in origin-derivation rules" needs a **named rule that reads provenance and isn't already handled by Part 1's generation-time emission.** The plan's only candidate is `ActionMissingReturnType`'s attribute re-read — which Part 1 stubs via a different mechanism. If Part 1 (finding 2) also converts the resource rules, even fewer origin-derivation rules remain. Risk: threading `OperationDescriptor.$provenance` to the lint layer becomes infrastructure with no consumer. Part 2's PR must either name a concrete surviving consumer or honestly recast item 1 as **why-command-facing provenance totality** (the `status`/`tags`/`security`/`deprecated` + #518 work, which stands on its own) and drop the "retire re-reflection" framing. Not a split blocker — resolve it in the Part-2 plan.

### 5. #518 option (a) — sound, with one forward caveat
Option (a) survives the multi-valued case you asked about: OpenAPI identifies a parameter uniquely by **name + location**, which is exactly the `name\0in` dedup key, and the compound label `parameters[<in>:<name>]` captures both — so the same name in `query` and `header` yields two distinct entries, faithfully. No current consumer parses the label programmatically (`openapi:why` renders it as an opaque display string; no parameter-origin lint rule is in scope). **Caveat to record on #518:** the compound label is a *display* convention, not a structured key. If a future lint rule ever needs programmatic `(name, in)` lookup, string-parsing `parameters[query:page]` is fragile (and param names can contain `:`/`]`); that is the point where option (b) or a structured key earns its cost. Fine to proceed with (a) as long as this stays a conscious, recorded tradeoff.

### PR-#514 deferral accounting — complete
All #484-relevant #514 deferrals are covered: the provenance-only keep-in-sync hazard (Part 2 item 2, matches the in-code `:826` marker), the richer `supersededBy` (Part 2 item 2, "every lower-precedence present candidate"), and parameters→#518 (Part 2 item 3). #514's other follow-up (Piece 3 `PrimaryResponseResolver` widening + sentinel) is #514's own separate PR, correctly not claimed here.

### Gates
Part 1: `area:lint`, no `Contracts/**`, no output-byte change → auto-mergeable, survey skipped — agreed, contingent on the equivalence bar (finding 3). Part 2: byte-identical guard + survey Layer A, no `Contracts/**`/config — agreed.

Radiergummi:
🤖 **[lead]** plan-review: needs-changes → back to planning. Split itself approved (genuinely separable); must-fix = the fourth named conversion target (schema missing/empty observations → resource.response-empty + resource.fields-undeclared, both re-running ResourceToArrayReader + ReflectionClass at lint time) must be mapped converted-where or stays-why; equivalence tuple = {ruleId, severity, location, suppression} NOT message, with a positive new-reason assertion + false-positive-direction fixtures (area:lint skips survey, fixtures are the only guard); Part-2 item 1 needs a named surviving consumer or recast as why-command provenance totality.

Radiergummi:
🤖 **[planner]** revised per plan-review (all four findings folded in). Ready for plan-review re-check.

1. **Scope gap (finding 2) — 4th target now mapped → CONVERTED.** `resource.response-empty` + `resource.fields-undeclared` (ApiResources) join Part 1: emitted from the resource response resolver / `ResourceToArrayReader` at generation time, both rules → stubs, dropping their lint-time `ReflectionClass` + `->read()` re-run. **Area consequence:** this adds `area:plugins` to Part 1 → survey Layer A runs (0-delta formality). Flagged for the lead: keep in Part 1 (recommended — complete unit) or peel to a Part 1b to preserve Part 1 as pure-`area:lint` survey-skipped.
2. **Equivalence bar (finding 3) — clarified.** Identity = `{ruleId, severity, location, suppression}`; **message/reason is the sanctioned delta, NOT pinned.** Added a positive assertion the enriched reason appears, plus **false-positive/negative-direction fixtures**: finding must NOT fire when a resolver produced a response, when the `@throws` walk found the exception, or when the resource reader resolved a non-empty declared shape — the only guard for the pure-lint rules.
3. **Part 2 item 1 (finding 4) — recast.** Dropped the "lint-readable / retire re-reflection" framing (no surviving consumer once Part 1 stubs the rules); Part 2 is now honestly **`openapi:why --fields` provenance totality** (keep-in-sync fold + richer supersededBy + #518 per-parameter), which stands alone.
4. **#518 caveat (finding 5) — recorded on #518.** Compound label is a display convention, not a structured key; option (b) earns its cost only if a rule needs programmatic `(name, in)` lookup.

Retitled to 'Part 1 of #484'; `Closes` → `Part of` (#484 closes with Part 2).

Radiergummi:
🤖 **[lead]** revised plan → plan-review re-check. Lead ruling on the open question: keep the converted ApiResources targets IN Part 1 (single review cycle; the added area:plugins survey is a routine 0-delta confirmation) — no Part 1b.

Radiergummi:
🤖 **[reviewer]** plan-review (re-check): **approved.** All four must-fixes + the #518 caveat are folded in faithfully; the plan is ready to code.

- **Scope gap (finding 2) — closed.** The 4th target is now converted: `resource.response-empty` + `resource.fields-undeclared` emit from the ApiResources generation path (where `ResourceToArrayReader` already reads the resource), both rules → stubs, dropping the lint-time `ReflectionClass` + reader re-run. The `area:plugins` → 0-delta-survey consequence is called out with the Part-1b peel option; keeping them in Part 1 (per the lead's ruling) is the right call — it's the complete "inference-failure findings at generation time" unit the issue describes.
- **Equivalence bar (finding 3) — corrected.** Identity is explicitly `{ruleId, severity, location, suppression}`, message/reason is the sanctioned, deliberately-unpinned delta, plus a positive-reason assertion (untyped-vs-`mixed`/`void`/`never`, the specific missing exception, empty-vs-undeclared).
- **False-positive/negative fixtures — the one I specifically flagged is present.** The finding must NOT fire when a `PrimaryResponseResolver` produced a response, the `@throws` walk found the exception, and — the direction I called out — **the ApiResources reader resolved a non-empty declared shape.** That closes the extensional-equality risk (new emissions predict document absence from build-time signals) for the resource conversion specifically, which is the case with no survey backstop under `area:lint`.
- **Part 2 item 1 (finding 4) — honestly recast.** The "lint-readable to retire re-reflection" framing is dropped with the correct reasoning (after Part 1 stubs the rules, no surviving lint consumer reads provenance → threading it would be consumer-less infrastructure); Part 2 is recast as `openapi:why` provenance totality, which stands on its own (kill the keep-in-sync hazard + richer `supersededBy` + #518).
- **#518 caveat (finding 5) — recorded** as a conscious display-vs-structured-key tradeoff, logged on #518.

One light note for the coder (not a plan change): `resource.response-empty` and `resource.fields-undeclared` have **distinct emptiness gates** (JsonResource-base/abstract vs. concrete resource) — the generation-time emission must preserve which ID fires when. The `{ruleId, …}` equivalence set + the empty-vs-undeclared positive assertion already guard this; just don't collapse the two into one emission site.

Approved to code Part 1.

Radiergummi:
🤖 **[lead]** plan-review re-check approved → coding. Coder note from review: the two resource rules have distinct emptiness gates (JsonResource-base/abstract vs concrete) — the generation-time emission must preserve which rule ID fires when; the equivalence tuple + empty-vs-undeclared assertions guard it.

Radiergummi:
🤖 **[coder]** implemented Part 1; local gates green (`composer check`: Pint clean, PHPStan 0 errors, 3381 tests) — rebased onto current `main` (post #519/#520). Pushed `6ece274`. Ready for review.

**All five inference-failure findings moved to generation time, each rule → registration stub:**

| Finding | Generation-time emission site |
|---|---|
| `operation.action-method-missing` | `PathsStage` — a route whose controller method does not exist yields no operation; emitted over the **global-filtered** set so the finding's route scope matches the old `RouteRule` superset. |
| `operation.return-type-missing` | `OperationBuilder` — after response resolution, when no resolver produced a response, no `#[Response]`/`#[ResponseResource]`, and the return type is absent/`mixed`/`void`/`never`. |
| `throws.transitive-missing` | `ThrowsErrorContributor` (already walks `@throws`; injected `ThrowsExtractor`). |
| `resource.response-empty` / `resource.fields-undeclared` | `ResourceResponseResolver` — **distinct gates preserved**: base/abstract `JsonResource` → response-empty; concrete + no `#[ResourceField]` → fields-undeclared; shared empty-shape gate (`ResourceToArrayReader` + `WrappedModelLocator`). |

**Equivalence bar honoured.** Findings keep identical `{ruleId, severity, location, suppression}`: operation-scoped findings set `location` via `FindingLocation::fromDescriptor` and stamp `context[SOURCE_CLASS]` = controller (mirroring the walker's stamp), so `#[IgnoreLint]`/`severity_overrides` behave unchanged. Message/reason is the sanctioned delta. **Generated document is byte-identical** — the `snapshot` group passes.

**Tests** (rewrote the 5 rules' unit/feature tests to drive the generation path via `LintRunner`):
- Positive-reason assertions: untyped-vs-`mixed`/`void`/`never`; the specific missing exception; empty-vs-undeclared resource distinction.
- False-positive/negative fixtures (the load-bearing guard): does **not** fire when a resolver produced a response (inline `response()->json()` with a `mixed` return), when the controller redeclares the `@throws`, when the method type-hints no Action, or when the reader resolved a non-empty `toArray()` shape.
- End-to-end `#[IgnoreLint]` suppression on the controller class.

**Reviewer's coder-note honoured:** the two resource rules keep their disjoint emptiness gates (base/abstract vs concrete) at a single site without collapsing which ID fires.

**One flagged consequence (not a document delta):** because generation now emits these findings, `LoggingFindingsCollector` logs them during plain `openapi:generate` too (exactly as `request-body.schema-degraded` already does). One existing test (`UntypedReturnResolutionTest`) had an over-broad "no log mentions this method" assertion that caught the new info-level finding; I tightened it to the resolver's `NOTICE`-level refusal notice it actually tests. No other observable change.

**Area:** `area:lint` + `area:plugins` (the ApiResources conversion) → survey Layer A is a 0-delta formality.

Radiergummi:
🤖 **[lead]** implementation pushed (6ece274), remote CI green incl. byte-exact snapshot → review. Two independent reviewer passes + a 0-delta survey confirmation run in parallel. Coder flagged one consequence: generation now logs these findings via LoggingFindingsCollector, requiring one over-broad assertion tightened in UntypedReturnResolutionTest — reviewers judge that tightening.

Radiergummi:
🤖 **[surveyor]** Layer A pass — strict zero document delta across 11 apps.

**Surveyed:** PR head `6ece274` vs the hot `main` baseline at `576cefa2` (main hadn't moved since #523's baseline, so I reused it — no re-baseline needed).

**Metric layer:** every count metric identical across all 11 apps (paths, operations, apiOperations, responseSchemas, documentedResponses, requestBodies, maxRequestProperties, componentSchemas, completenessPercent, routesIntrospected). No crashes; all candidate specs are valid JSON.

**Content layer:** for **every one of the 11 apps**, the raw per-app spec diff has **0 non-RNG changed lines** — the only raw differences are the corpus's known per-run RNG (random `operationId` suffixes, Livewire route hashes). After masking that RNG, **all 11 apps are byte-identical to baseline**.

This confirms the PR's claim exactly: moving inference-failure findings to generation time (the `LoggingFindingsCollector` channel) plus the supporting `ResourceResponseResolver` / `ThrowsErrorContributor` / `OperationBuilder` / `PathsStage` wiring emit findings only — they contribute no content to any generated document. (The survey harness doesn't route the PSR logger to stdout, so the new NOTICE lines don't appear in `generate.log`; that's a harness-visibility detail, not a document concern.)

No findings. Clear to merge on the survey gate.

<sub>Baseline `manifest.libraryCommit` = `576cefa2`; candidate = `6ece274`; PHP 8.4; 11/11 apps measured, no dropped apps.</sub>

Radiergummi:
🤖 **[reviewer]** review round 1: **approved, no blocking concerns.** I traced all five conversions against their pre-change rules for the `{ruleId, severity, location, suppression}` tuple, and confirmed the reason-enrichment and every false-positive fixture.

### Equivalence per conversion
- **`operation.action-method-missing`** (→ `PathsStage`): severity `Degraded` (=old); fires on missing method, `__invoke` handled, closure/missing-class skipped (=old). No `context` — **matching the old rule, which also set none**, so `#[IgnoreLint]`-by-class was never available here and suppression parity holds (severity_overrides key on ruleId, unchanged). Covered by the **unchanged** pre-existing `tests/Feature/Lint/ActionMethodMissingTest.php` (fires / doesn't-fire / dead-route only draws this one ruleId); the deleted unit test was the now-redundant rule-class test.
- **`operation.return-type-missing`** (→ `OperationBuilder`): severity `Inconsistent` (=old); `context[CONTEXT_SOURCE_CLASS]` set (replicates the walker's operation-source stamp for the old OperationRule — end-to-end suppression proven by "is suppressed by #[IgnoreLint] on the controller class"); location.routeUri pinned; the mixed/void/never-vs-untyped reason asserted.
- **`throws.transitive-missing`** (→ `ThrowsErrorContributor`): emitted **inside the existing `@throws` walk** (retires the double-walk as planned); severity `Degraded`; context stamped; message names Action + exception + method.
- **`resource.response-empty` / `resource.fields-undeclared`** (→ `ResourceResponseResolver`, pre-existing, +123/-1 additive so the response output stays byte-identical): the **distinct emptiness gates are preserved exactly** — `isBaseOrAbstract ? response-empty : fields-undeclared`, with the concrete-declares-fields skip, the shared `toArrayReader->read()` inferable-skip, and the wrapped-model skip all mirroring the old two rules. Severity `Degraded`, context stamped.

### The three false-positive directions you named — all have dedicated fixtures
1. **resolver-produced-response** → `ReturnTypeMissingTest`: "does not fire when a resolver produced a response from the body" (`inlineJson(): mixed` body-scanned). ✓
2. **@throws-found** → `ThrowsTransitiveMissingTest`: "does not fire when the controller redeclares the transitive @throws" (+ "no Action"). ✓
3. **resource-reader-resolved-shape** → `ResourceFieldsUndeclaredTest`: "does not fire when the reader resolves a non-empty declared shape from toArray()" (+ abstract-base not flagged by fields-undeclared, concrete not flagged by response-empty — the gate boundary tested both ways). ✓

### One intended behaviour delta, correctly the sanctioned direction (flagging for the record)
`operation.return-type-missing` now emits a **strictly smaller** finding set than the old rule: the old OperationRule fired purely on return-type/attribute signals and had no knowledge of whether a resolver produced a response, so it emitted **false positives** for untyped-but-body-scanned actions. The new `responseProduced` guard removes those. This is the improvement your plan-review false-positive fixture asked for, it's byte-neutral for the document (findings ≠ document, snapshot green), and it's the only equivalence-tuple deviation — "identical minus legitimately-removed false positives," not literally identical. Worth stating plainly since the plan's bar read as strict equivalence.

### UntypedReturnResolutionTest tightening — does not mask a real signal
The old assertion ("no log record mentions `untypedScalar`") would now fail because the new info-level `operation.return-type-missing` finding is logged via `LoggingFindingsCollector` and its message contains the method name. Tightening to `level === NOTICE && contains('untypedScalar')` correctly isolates the resource reader's **NOTICE-level refusal notice** (the thing the test actually guards) from the unrelated info-level finding. It does **not** mask the real signal: a genuine resource-reader refusal for `untypedScalar` is NOTICE-level and would still trip the assertion. The tightening is precise; its one load-bearing assumption (refusal notices are NOTICE, findings are info) matches the established severity→log-level convention.

### Conventions / bookkeeping
Stubs are clean registration-only classes retaining `$id`/`$severity`/`$description`; emissions carry actionable `fixHint`s and why-focused reasons; the resource emission correctly lives in the ApiResources plugin while the core emissions sit on the Support/Core generation path; `FindingsCollector` is the proven shared channel; no `Contracts/**`, no config. CHANGELOG entry is accurate and explicit about IDs/severities/locations/suppression unchanged + document unchanged.

### Non-blocking nits
1. `PathsStage::emitMissingActionMethodFinding` instantiates `new ReflectionClass($controllerClass)` twice on the fire path (once for `hasMethod`, once for `getShortName()` in the message) — reuse one instance. Trivial, only on the finding path.
2. Patch coverage 96.4% (2 lines) — the `(unknown)` null-controller/method message fallbacks and similar defensive arms; not load-bearing.

Approved — my independent pass. Survey Layer A is a 0-delta formality (findings never touch the document).

Radiergummi:
🤖 **[reviewer]** review round 1 (independent second pass): **approved, no blocking concerns.** Formed on my own tracing of the diff and the old rules on `main`. One non-blocking divergence to confirm/document (webhook skip), plus a minor note.

**Zero document delta — structural + empirical.** All five emissions write only to the `FindingsCollector` side channel; no emission path touches `$document`. Byte-exact snapshot job green confirms it.

**{ruleId, severity, location, suppression} equivalence — verified per site; the load-bearing part is suppression, and it is correct.** I traced how the *old* rules were suppressed and confirmed the new sites replicate it exactly:
- The three operation-scoped rules (`operation.return-type-missing`, `resource.response-empty`, `resource.fields-undeclared`) were `OperationRule`s, and `SpecTreeWalker::walkOperation` (`main` :159-181) *auto-stamps* `context[SOURCE_CLASS] = controller` as a fallback so a controller-level `#[IgnoreLint]` matches. Generation-time emissions bypass the walker, so the coder **manually re-stamps** `context[SOURCE_CLASS] = controller` on each (and on `throws.transitive-missing`). Class-scope suppression is preserved, and it is directly pinned by a test ("stamps the controller class so #[IgnoreLint] can suppress it" + an end-to-end "suppressed by #[IgnoreLint] on the controller class" case).
- `operation.action-method-missing` was a **RouteRule**; the RouteRule pass (`LintRunner` :716-720) back-fills only `location` via `withLocationDefaults(fromDescriptor)`, never source-class context — so class-scope suppression never applied to it. The new `PathsStage` emission correctly **omits** context and sets `location: FindingLocation::fromDescriptor`, matching the old back-fill. (This is why the coders "operation-scoped findings stamp context" comment is accurate rather than over-general: action-method-missing is route-scoped.)
- Severities match (each stub keeps the original `$severity`; each emission hardcodes the same value: `Degraded` for the four, `Inconsistent` for return-type-missing).

**Five emission sites fire on the same conditions; false-positive direction guarded.**
- *return-type-missing:* gated on `responseProduced` (computed from the *actual* resolution result, before the `?? 200 OK` default) + `#[Response]`/`#[ResponseResource]` + a usable (declared, non-`mixed`/`void`/`never`, union/intersection-counts) return type — the old rules conditions, but read from the produced response instead of re-derived. The "does not fire when a resolver produced a response from the body (inline `response()->json()` with a `mixed` return)" fixture pins the sanctioned reduced-false-positive behavior (the whole point of emitting at generation time).
- *throws.transitive-missing:* the `resolveActionClass` + `isActionClass` (`ends-with Action` or `\Actions\` namespace) heuristic and the `ThrowsExtractor` `handle()` walk are ported from the old rule verbatim, so the flagged set is unchanged; false-negative fixtures cover redeclared `@throws` and no-Action-hint.
- *resource:* the two disjoint gates are preserved at one site — base/abstract `JsonResource` → `response-empty`; concrete-without-`#[ResourceField]` → `fields-undeclared`; a shared `ResourceToArrayReader`/`WrappedModelLocator` emptiness gate skips both. The concrete-with-fields early-return and the "does not fire when the reader resolved a non-empty declared shape" fixture close the extensional-equality risk that has no survey backstop under `area:lint`.

**UntypedReturnResolutionTest tightening — legitimate, narrow, not masking.** It restricts the "no log mentions `untypedScalar`" assertion to **`LogLevel::NOTICE`** (the resource readers refusal notice the test actually exercises), with a comment noting the new info-level `operation.return-type-missing` finding is a separate, expected channel. Emitting findings during plain `openapi:generate` via `LoggingFindingsCollector` is pre-existing behavior (`request-body.schema-degraded` already does it), so asserting the finding *not* log would be wrong. The coders flag was honest; the fix preserves the tests real intent.

**Non-blocking:**
1. **Webhook skip — confirm intended + itemize.** `emitMissingResponseSchemaFinding` early-returns when `OverrideMatcher::webhookKeyFor($action) !== null`. The old `ActionMissingReturnType` was an `OperationRule` and `SpecTreeWalker` walks webhook operations (`main` :105-106), so an untyped **webhook** action would have fired `operation.return-type-missing` before and now will not. Thats defensible (a webhook isnt response-schema-bearing the same way), but its a finding-set change beyond the sanctioned message delta and isnt covered by a fixture. Please confirm it is deliberate and note it (a one-line CHANGELOG/PR mention, or a fixture asserting no finding on an untyped webhook), the same way the inline-json false-positive was itemized.
2. **Minor:** operation findings location moved from `FindingLocation::fromOperation($operation)` (old walker) to `FindingLocation::fromDescriptor($descriptor)` (new). Both derive file/line from the same `actionReflector` and the tests pin `location->routeUri`, so theyre equivalent in practice; noting only that they are different constructors should a divergence ever surface.

Clean, careful, correctly scoped to Part 1. Approved from me; survey Layer A (0-delta) is the remaining formality.

Radiergummi:
🤖 **[coder]** addressed reviewer-1's second-pass item: the webhook boundary is now pinned. Pushed `4ab310c`.

**Confirmed intended.** The generation-time `operation.return-type-missing` emission skips webhook actions (`OverrideMatcher::webhookKeyFor($action) !== null`), because a webhook is not a response-schema-bearing operation — a missing/`mixed` return on a webhook is not the "no response schema could be inferred" defect this finding reports. The old `OperationRule` walked webhook operations too, so an untyped webhook action fired before and no longer does.

**Intended finding-set deltas (beyond the sanctioned message/reason delta), now all itemized + fixtured:**
1. **Response-aware gate** — does not fire when a `PrimaryResponseResolver` produced a response for an unusable-return action (e.g. `mixed` return with inline `response()->json()`). Fixture: *"does not fire when a resolver produced a response from the body"*.
2. **Webhook skip** (this item) — does not fire for a webhook action. Fixture added: *"does not fire for a webhook action, which is not response-schema-bearing"* (untyped `#[Webhook]` action → no finding).

Both are narrowings toward "the operation genuinely has no inferable response schema", which is the honest predicate for a generation-time emission. Test-only change; the generated document is unchanged, so the survey result stays valid. Local gates green (Pint, PHPStan 0, `ReturnTypeMissingTest` 10 passed).

Radiergummi:
🤖 **[lead]** controversy gate: src/ diff +394/−450 (844 changed lines, deletion-heavy rules→stubs refactor, far over the ~150 threshold) → needs-human per run directive. ALL gates green: full CI matrix on 4ab310c, two independent reviewer approvals, survey Layer A PASS (strict zero document delta, 11/11 apps). Both intended finding-set deltas itemized and fixtured (response-aware gate removes old false positives; webhook actions skipped — not response-schema-bearing). Ready to merge on your sign-off; #484 Part 2 (openapi:why provenance totality + #518 compound labels) is the planned follow-up.

PR #523, opened 2026-07-02
feat(requests): field-attribute overlay seam — decorate vs. declare

### LINKED ISSUE #493: feat(requests): field-attribute overlay seam — separate decorating an inferred field from declaring a missed one

Radiergummi:
Closes #493

## Implementation plan — #493: field-attribute overlay seam (decorate vs. declare)

**Tier:** infrastructure/seam, not an inference-tier change. It unifies *how* already-inferred field
metadata is decorated and creates the seam that makes the decorate-vs-declare distinction explicit
(and lintable). `area:requests` + `area:params` (+ `area:plugins` for the response-field sites).

**No hard dependency on unmerged work.** The substrate is `Support\Provenance\ResolvedField` — **merged**
(#482), explicitly documented as usable outside `OperationBuilder` for exactly this ("field-name-keyed
overlays applied on top of an already-resolved field", `ResolvedField.php:14-17`). #519 (#481,
return-shape) and #520 (#483 Part A, component registry) touch different choke points and are **not**
prerequisites — this plans against current `main`.

### Grounding (verified against `origin/main`, post #482/#514)

- **The family is dual-role today, with no seam.** `FieldAttribute` (abstract, `src/Attributes/`) is the
  shared base; `#[PathParam]`/`#[QueryParam]`/`#[RequestField]`/`#[ResponseField]` expose curated
  subsets. Only `PathParam` is near-pure human-copy (description/example/format/`pattern`, **no `type`**,
  `PathParam.php:11-16`); the other three carry the full type/constraint surface via the base. The
  "did the author speak?" primitive is `FieldDefault::Unset` (`Support/Attributes/FieldDefault.php`),
  used only inside the family's `descriptor()` merge (`FieldAttribute.php:119-168`).
- **Decoration is scattered inline, each site re-deciding precedence:**
  - Path: `UriParametersExtractor::buildParameter` — reads any `FieldAttribute` off the param
    (`:166-169`), overlays `example` (`:72-74`) and a description chain `#[PathParam] ?? @param ??
    synthetic` (`:89-93`).
  - Query: `CoreQueryParameterResolver::attributeParameters` (`:442-484`) via
    `descriptor()->toSchema()`, its own name-keyed controller-then-action merge.
  - Request fields: `RequestFieldObjectBuilder::buildProperty` (`:75`), `SchemaFromFormRequest`
    (`:150`, `:265`) via `descriptor()->applyTo()`.
  - Response fields: `SchemaFromDataClass` (`:732`, `:758`), `SchemaFromResource` (`:267`, `:273`),
    `SchemaFromTransformer` (`:164`, `:170`).
- **`QueryParam` has ~3 independent runtime readers, no shared reader:**
  `CoreQueryParameterResolver:454/460`, `OperationBuilder::explicitQueryParameterNames:385`,
  `QueryParamDuplicate:48` — each re-implements its own `getAttributes(QueryParam::class)` loop.
- **The #514 appliers do NOT cover field-level decoration** — `RequestParameterApplier` (header/cookie),
  `ResponseHeaderApplier` (response headers), `ExampleApplier` (body-level examples). There is **no**
  field-name-keyed description/example/format overlay primitive today.
- **No "attribute smell" lint rule exists** (grep-confirmed). Nearest: `field.no-effect` (empty
  attribute), `field.conflicting-type` (type/format conflict) — neither flags "attribute restates the
  reflected type."
- **`ResolvedField::merge(field, candidates)` + `FieldCandidate::present/absent` + `toProvenance()`**
  are field-name-keyed; the `OperationBuilder::resolveSummary` pattern (collect candidates in precedence
  order → `merge` → `toProvenance`) is the shape to mirror.

The issue carries a **structural seam** (byte-identical) and an **opinionated behaviour change + lint
rule** (output-affecting). They split along area/risk exactly like #514/#483/#484 — **recommend two
PRs**; lead/reviewer rule on the split.

---

### Part A (this PR) — the overlay seam + path decoration + shared reader. Byte-identical.

> **Plan amendment (plan-review §2/§5 + lead ruling).** Part A ships the overlay seam, the **path**
> decoration migration, and the `QueryParamReader` consolidation. The **query/request/response**
> decoration wiring moves to **Part B**: post-§2 those sites are sole-source (the attribute is the only
> candidate, applied through the shared `SchemaDescriptor` / `FieldDescriptor`), so they carry no
> multi-candidate precedence for the overlay to consolidate, and routing their decoration through it
> would require decoration-exclusion surgery on those shared primitives (wide byte-identity blast
> radius) to produce provenance nothing consumes until #484 Part 2. Deferred until the decorate/declare
> split actually needs the overlay's "was-inferred" signal there.

1. **Shared field-metadata overlay primitive** — `src/Support/Extraction/FieldMetadataOverlay.php`
   (`@internal`). A **pure, order-agnostic mechanism** (§2): given an already-built target
   (`OA\Schema` / `OA\Parameter` / `OA\Property`) and, per sub-field, a **caller-supplied,
   precedence-ordered `FieldCandidate` list**, it runs `ResolvedField::merge`, applies the winner, and
   records provenance. `resolve()` returns the winner for a site that post-processes before applying
   (the path optional-segment note); `apply()` sets winners directly. **The overlay hardcodes no
   precedence** — each caller owns its order, which is what preserves bytes across sites whose orders
   differ.
2. **Migrate the path decoration site** — `UriParametersExtractor`, the one site with a genuine
   multi-candidate precedence chain (`#[PathParam] ?? @param ?? synthetic`), preserving the
   optional-segment note post-processing. Output unchanged.
3. **Shared `QueryParam` reader** — `src/Support/Attributes/QueryParamReader.php`: the
   controller-then-action name-keyed merge (`mergedByName`) and the ordered raw read (`read`), read
   once. Retires the duplicated `getAttributes(QueryParam::class)` loops in `CoreQueryParameterResolver`,
   `OperationBuilder::explicitQueryParameterNames`, and `QueryParamDuplicate`. Behaviour-preserving.

**Provenance note (§5 — internal/test-only):** the overlay **produces** `FieldProvenance` records but
Part A **surfaces nothing** — no label/aggregation convention, no `openapi:why` wiring. #484 Part 2 owns
the surfaced/consumed shape (compound labels included); Part A only records the per-sub-field entries a
later change consumes, keeping the shape decided once, there.

**Tests:** byte-identical document guard via the existing path/query feature + `snapshot`-group suites
(output unchanged); a `FieldMetadataOverlay` unit test per sub-field covering attribute-present,
docblock-only, synthetic-only, `FieldDefault::Unset` vs. explicit-null, and provenance recorded
(source/reason/supersededBy); the shared reader is exercised by the existing `QueryParamDuplicate` /
`CoreQueryParameterResolver` suites (name collision + controller/action merge resolve identically).

**Docs/changelog:** `CHANGELOG.md [Unreleased]` **Changed** (internal) — shared field-metadata overlay
+ shared QueryParam reader; no observable output change, no new attribute, no config key, no lint rule.
`area:requests`/`area:params` ⇒ **survey Layer A runs**, a 0-delta formality here (any delta is a bug).

---

### Part B (recommended PR-B, follow-up) — decorate/declare split + the attribute-smell lint. Output-affecting + new lint rule; human-gated.

1. **The decorate/declare semantic split.** With the overlay in place, an attribute on an
   **already-inferred** field contributes **human-copy only** (description/example/format/deprecation);
   its type/constraint args no longer override the reflected type. The **declaring** role is unchanged
   for fields inference cannot reach (notably `RequestField`'s method-level repeatable form for bodies
   validated in an Action/service — genuinely Tier-2). The overlay already knows "was this field
   inferred?" (a candidate was present pre-overlay), which is the exact signal that routes decorate
   vs. declare.
   - **Load-bearing decision for plan-review:** what "stops carrying type/constraint args for
     already-inferred fields" means for output. **Recommendation:** for an inferred field, the
     attribute's `type`/constraints fill *gaps only* (no override of a reflected type). This changes
     output **only** where an author's attribute type *disagreed* with reflection — the case the
     philosophy calls a smell. Survey Layer A must itemize those deltas as intended.
2. **The attribute-smell lint** — `field.attribute-redundant-type` (ID; `docs/linting.md` catalog row).
   Flags a `FieldAttribute` whose `type`/`format`/constraint **restates** what reflection already
   produced. **Recommended implementation:** emit it as a **generation-time finding from the overlay**
   (per #484 Part 1's `request-body.schema-degraded` pattern — the overlay knows, at merge time, that
   the attribute value equals the inferred value), with the rule class a registration stub. This avoids
   a re-reflecting lint rule and composes with #484's direction (soft coordination, not a hard dep —
   `FindingsCollector` already exists).
3. **Wire the remaining decoration sites through the overlay** (moved from Part A): query
   (`CoreQueryParameterResolver`), request (`RequestFieldObjectBuilder` / `SchemaFromFormRequest`), and
   response (`SchemaFromDataClass` / `SchemaFromResource` / `SchemaFromTransformer`). These are
   sole-source (attribute-only, or inferred-wins gap-fill) with no precedence chain to consolidate, so
   routing their decoration through the overlay needs a decoration-exclusion mode on the shared
   `SchemaDescriptor` / `FieldDescriptor`; it earns its place here, alongside the decorate/declare split
   that consumes the overlay's "was-inferred" signal and #484 Part 2 that consumes the per-field
   provenance, rather than as speculative plumbing in Part A.

**Tests:** survey-itemized output deltas (disagreeing-type cases only); the smell finding fires when an
attribute restates a reflected type and is silent when it adds genuine copy or declares a missed field;
`#[IgnoreLint]` suppression on the source class works.

**Docs/changelog:** `docs/linting.md` catalog row for `field.attribute-redundant-type`; a
`docs/` note on the decorate-vs-declare guidance; `CHANGELOG.md` **Added** (lint rule) + **Changed**
(attribute type no longer overrides a reflected field type). `area:requests`/`params` +
new-public-lint-rule ⇒ **human gate + survey Layer A with intended deltas**.

---

### Controversy / gates

- **Part A:** touches no `Contracts/**`, no config, no output bytes → auto-mergeable after review; survey
  is a 0-delta formality. Scoped to the seam + the single path decoration site + the `QueryParam` reader
  (the query/request/response wiring moved to Part B), so the blast radius is contained; the existing
  path/query feature + `snapshot` suites are the byte-identical guard.
- **Part B:** an **opinionated output change** (attribute type stops overriding reflection) + a **new
  public lint rule** (catalog + docs) → **human gate**, survey Layer A with the intended deltas itemized.
- Splitting keeps the wide-but-safe refactor (Part A) from stalling behind the human gate the opinionated
  change (Part B) needs anyway.
- Depends only on merged #482. Coordinates with #484/#518 on the per-field provenance *shape* (already
  decided on #518) but is not blocked by them.


Radiergummi:

Planned the full issue; **recommend a two-PR split** (mirrors #514/#483/#484):
- **Part A (this PR, byte-identical):** shared `FieldMetadataOverlay` primitive applying description/example/format/deprecation via `ResolvedField` (one documented precedence, provenance recorded), migrating all four decoration sites (path/query/request/response across Core + 3 plugins) + a shared `QueryParam` reader retiring the ~3 duplicated loops. No output change, no `Contracts/**`, no lint rule → auto-mergeable behind the byte-identical guard.
- **Part B (follow-up, human-gated):** the decorate/declare semantic split (attribute type stops overriding a reflected field type — output-affecting where they disagreed) + the `field.attribute-redundant-type` smell lint, best emitted as a **generation-time finding from the overlay** (per #484 Part 1's pattern).

**Load-bearing plan-review decision:** what "stops carrying type/constraint args for already-inferred fields" means for output — I recommend type/constraints fill gaps only (no override of reflection), so Part B changes output only in the smell case.

**No hard dependency on unmerged work** — substrate is merged #482's `ResolvedField`; #519/#520 touch different choke points. Coordinates with #484/#518 on the per-field provenance *shape* (already decided on #518), produces but does not surface it. Lead/reviewer rule on the split.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Reviewer: rule on (1) the two-PR split (Part A byte-identical overlay primitive + QueryParam reader consolidation; Part B decorate/declare semantic change, human-gated); (2) the load-bearing decision the planner flags — what 'attribute stops carrying type/constraint args' means for output; planner recommends gap-fill-only; (3) the no-hard-dependency-on-#519/#520 claim (both now merged anyway — plan should rebase context onto current main); (4) coordination with #484/#518 on per-field provenance shape (produces, doesn't surface).

Radiergummi:
🤖 **[reviewer]** plan-review: **approved to proceed on Part A**; the two-PR split and the byte-identical overlay are sound. One substantive re-frame needed on the Part B load-bearing decision (**not** a Part A blocker, since Part B is deferred + human-gated) and two sharpenings. Detail below.

### 1. Two-PR split — approved

Mirrors the #514 / #483 / #520 / #484(#522) precedent exactly: the wide-but-safe refactor ships first behind a byte-identical guard, the opinionated output change waits for the human gate it needs anyway. Part A is genuinely separable — it is pure mechanism reuse on merged #482 substrate, and Part B strictly builds on it. Good call not to stall the refactor behind the semantic change.

### 2. Part A byte-identity — sound, but sharpen "one documented precedence"

The plan says both "one documented precedence per sub-field" **and** "precedence encodes todays inline order." Those only reconcile if the overlay is **pure mechanism**: `ResolvedField::merge` takes a *caller-supplied, precedence-ordered candidate list* (verified — `ResolvedField.php:41`, "the first present candidate"), so **each migrated site passes its own candidates in its own current order** and the overlay hardcodes no policy. That is what preserves bytes across four sites whose orders may differ (path is `#[PathParam] ?? @param ?? synthetic`; the others may not match). A *single hardcoded* precedence baked into the overlay would silently break byte-identity wherever two sites disagree — the one real trap here. Please state explicitly in the plan that the overlay is order-agnostic mechanism and the caller owns precedence; the four-field-kind byte-identical fixture guard then confirms it rather than being the only safety net.

### 3. The load-bearing decision (gap-fill-only) — re-frame; the philosophy split is restatement-vs-override, not agree-vs-disagree

The plan justifies gap-fill-only as: "changes output only where an authors attribute type *disagreed* with reflection — the case the philosophy calls a smell." That equivalence is **not** what the philosophy says. Per `CLAUDE.md`: an attribute is a **smell when it *restates* something a Tier-0/1 read could have inferred** (attribute value **==** reflection), and it is **healthy when it is an *intentional override*** (explicitly listed alongside "a runtime shape, a human description"). A *disagreement* (attribute **!=** reflection) is precisely the intentional-override case the philosophy calls **healthy** — not the smell.

So there are three cases, and the plan collapses two of them:
- **attribute == reflection (restatement):** redundant. This is the true smell. Flagging it is byte-neutral (winner value is identical either way) — the lint finding is the only change.
- **attribute != reflection (override):** the documented healthy case. Gap-fill-only makes reflection silently win here, which **removes the intentional-override capability** and changes output.
- **reflection absent (gap):** attribute declares. Unchanged under any model.

Gap-fill-only therefore does the opposite of what the framing claims: it is byte-neutral on the *smell* case and output-changing on the *healthy* case. Two consequences for Part B (decide at the human gate, not now):
- Decide overrides fate **explicitly**: either (a) preserve intentional override and make Part B **lint-only + byte-identical** (flag restatement, keep override winning — arguably the most philosophy-aligned outcome, and it collapses Part Bs survey to a formality), or (b) remove override deliberately and provide the correction path an author then needs (the `openapi.overrides` config escape hatch is the obvious candidate — it stays plugin-agnostic and is already the sanctioned "code cant express it" lever).
- Whichever way, the survey Layer A itemization is **disagreement (`!=`) deltas**, and the plan should call those what they are (removed overrides), not "the smell case."

The `field.attribute-redundant-type` lint rule itself is well-placed regardless: emitting it as a generation-time finding from the overlay (per #484 Part 1s `request-body.schema-degraded` pattern) is the right, non-re-reflecting design, and it keys on the `==` restatement signal the overlay has at merge time. That part I fully endorse.

### 4. Context refresh (#519/#520 merged) — no prerequisite, no missed reuse

Both are on `main` now (35392af7, 576cefa2); neither is a prerequisite (return-shape and component-registry are different choke points), so the "plans against current main" claim holds — just update the grounding header from "post #482/#514" to note #519/#520 too. **Checked whether #493 should now build on the merged provenance API instead of ad-hoc plumbing: no.** #520s `SchemaProvenance` is **component-granularity** ("which builder produced this whole component schema"); #493 needs **field-granularity** `FieldProvenance`/`ResolvedField` ("where did this fields description/example come from"). Different axis; the plan builds on the correct substrate. No missed reuse.

### 5. Coordination with #484/#518 — one granularity subtlety; keep Part A provenance internal

#518s decision (recorded on the issue and in PR #522s plan) is **one `FieldProvenance` per parameter, compound label `parameters[in:name]`, no shape change**, produced in `OperationBuilder`s `name\0in` dedup merge, surfaced by #484 Part 2. #493s overlay instead produces provenance at the **(field, sub-field)** granularity (description/example/format/deprecation each a `ResolvedField::merge`), at the *decoration sites*, not in `OperationBuilder`. These are different production points **and** different axes — so the plans phrase "uses the shape #518 settled (one entry per field via a compound label)" slightly overreaches: #518 settled the *per-parameter operation-field* shape, which is not the overlays per-sub-field record.

No hard conflict, because #493 Part A **produces but does not surface**. To keep it that way and avoid deciding the shape twice (exactly what #518 warned against): **keep Part As provenance production internal / test-asserted only, and do not bake a cross-cutting label convention** that #484 Part 2 would then have to match or migrate. Let #484 Part 2 own the surfaced/consumed shape (compound labels included). Soften the "uses the shape #518 settled" claim to "produces `FieldProvenance` records the #484 Part 2 surfacing will consume; label/aggregation convention owned there."

### Verdict

Approve Part A to build — it is a clean, well-grounded, byte-identical consolidation on merged substrate, and the split is right. Fold in the §2 mechanism-not-policy wording and the §5 internal-only provenance scoping before/while building Part A. The §3 re-frame is a **Part B** concern: record it now so the human-gate decision is made on the correct restatement-vs-override distinction, not built on the mislabeled "disagreement = smell" premise.

Radiergummi:
🤖 **[lead]** plan-review approved for Part A → coding. Binding amendments from review (treat as part of the plan): §2 the overlay is a pure order-agnostic mechanism — ResolvedField::merge takes a caller-supplied ordered candidate list and each of the four sites passes its own current precedence (a hardcoded precedence would silently break bytes); §5 Part A's per-field provenance stays internal/test-only — #484 Part 2 owns the surfaced label convention. §3 (override-vs-restatement re-frame) is recorded for Part B's human gate, not Part A scope.

Radiergummi:
🤖 **[coder]** checkpoint pushed 38360875 — the seam is in and byte-identical; flagging a scope/mechanism fork on the remaining sites before I touch shared primitives.

**Landed (byte-identical, gates green — `composer check`, snapshot group included):**
- `Support\Extraction\FieldMetadataOverlay` — the policy-free mechanism per review §2: `resolve()` / `apply()` over a caller-supplied precedence-ordered `FieldCandidate` list → `ResolvedField::merge` → apply winner → record `FieldProvenance` (internal/test-only per §5; no label convention baked). Unit-tested (resolve precedence, absent-skips, null-vs-Unset, provenance, `OA\Schema`/`OA\Parameter` targets).
- Migrated the **path** decoration site (`UriParametersExtractor`) — the one site with genuine multi-candidate precedence (`#[PathParam] ?? @param ?? synthetic`), including the optional-note post-processing preserved via `resolve()`.
- `Support\Attributes\QueryParamReader` — consolidated the three duplicated `#[QueryParam]` loops (`CoreQueryParameterResolver`, `OperationBuilder::explicitQueryParameterNames`, `QueryParamDuplicate`).

**The fork — the other three kinds are not a clean overlay migration.** Grounding on current `main`, their decoration is applied through heterogeneous, entangled paths, and none has a shared multi-candidate precedence the overlay would unify:
- **query** — `#[QueryParam]->descriptor()->toSchema()` builds a fresh `OA\Schema` with description/example/format baked in (sole source: the attribute).
- **request** — `#[RequestField]/FormRequest->descriptor()->applyTo($property)` (`SchemaDescriptor`, sole source: the attribute; these are *declaring* a field inference can't reach).
- **response** — Data uses `FieldDescriptor->applyTo($property, overwrite: false)` (a **different** descriptor primitive, inferred-wins gap-fill); Resource/Transformer use `SchemaDescriptor->applyTo()`.

To route only description/example/format through the overlay at these sites (without double-applying) I would have to modify the shared `SchemaDescriptor` **and** `FieldDescriptor` to exclude the decoration keys — both are used by many callers, so that is an invasive change with real byte-identity blast radius, and it produces only *currently-unconsumed* provenance (§5: #484 Part 2 owns surfacing; not merged).

**Recommendation:** land Part A as this seam + path migration + `QueryParamReader` (genuine, byte-safe consolidation), and wire the query/request/response decoration through the overlay in **Part B**, where the decorate-vs-declare split actually consumes the overlay's "was-this-field-inferred?" signal at those sites and #484 Part 2 consumes the provenance — rather than a wide, byte-fragile refactor now for no output or consumer benefit. This is squarely the tradeoff review §2 surfaced (sole-source sites carry no precedence to consolidate).

Lead/reviewer: happy to wire all three now via a `SchemaDescriptor`/`FieldDescriptor` decoration-exclusion mode if you'd rather establish the provenance-production seam everywhere in Part A — just confirm, since it means editing those shared primitives across all their callers under the byte-identical bar.

Radiergummi:
🤖 **[coder]** implemented; local gates green (`composer check`); remote CI green across the matrix (PHP 8.4/8.5 × Laravel 12/13, swagger-php 5.8, snapshot, quality, coverage); pushed 38360875. PR body amended per the lead ruling — Part A scope = overlay seam + path migration + `QueryParamReader`, with the plan-review §2 (pure order-agnostic mechanism, caller owns precedence) and §5 (provenance internal/test-only) wording folded in, and the query/request/response wiring moved to Part B with the sole-source rationale. Ready for review.

Radiergummi:
🤖 **[lead]** implementation pushed (38360875), remote CI green across the matrix → review. Scope as ruled: seam + path site + QueryParamReader, byte-identical; the three sole-source sites moved to Part B (deviation argued in the coder comment). Two independent reviewer passes + survey Layer A (zero-delta expectation) run in parallel.

Radiergummi:
🤖 **[reviewer]** review round 1: **approved, no concerns.** Clean, byte-identical, correctly scoped; the plan-review §2/§5 amendments and the lead-ruled scope reduction are honestly carried in both code and PR body.

**§2 pure-mechanism — compliant.** `FieldMetadataOverlay` is genuinely policy-free: `resolve()`/`apply()` take a caller-supplied precedence-ordered `FieldCandidate` list, run `ResolvedField::merge`, apply the winner, and record provenance — no hardcoded precedence. The path site passes its exact current `#[PathParam] ?? @param ?? synthetic` chain (`UriParametersExtractor::resolveDescription`) and its `example`-present-else-absent conditional, so precedence stays with the caller. This is exactly the mechanism-not-policy shape §2 required.

**§5 provenance internal/test-only — compliant.** The overlay returns `list<FieldProvenance>` but both migrated call sites discard it (`apply()` uncaptured for the example overlay; `resolve()` used only for `->value` in the description chain). No label/aggregation convention is baked; the class docblock states #484 Part 2 owns surfacing. The shape is not decided twice.

**Byte-identity — verified by reading each migrated site, not just trusting the snapshot job:**
- *Path example:* `apply($schema, [example => [present-or-absent]])` reduces to "set `$schema->example` iff the attribute example is non-null" — identical to the old `if ($fieldDescriptor?->example !== null)` guard, same `!== null` test (so falsy examples `0`/`false`/`` still apply as before).
- *Path description:* the overlay chain resolves attribute → `@param` → synthetic in the same order as the old `??` chain, and the optional-note post-processing still runs on the resolved value. The synthetic candidate is always present, so a winner always resolves (the `: ` fallback is dead-but-defensive).
- *QueryParamReader (3 loops → 1):* all three consumers preserve their prior semantics — `read()` yields controller-before-method in source order with **duplicates preserved** (so `QueryParamDuplicate`s count is unchanged), and `mergedByName()` keys by name controller-first-then-method (method shadows class), matching `CoreQueryParameterResolver`s old inline merge and `explicitQueryParameterNames`s ordered list exactly.
- Snapshot group + unchanged path/query feature suites green confirm it end to end.

**Scope reduction — honestly carried.** Part A wires the overlay only at the path site (the one with a genuine multi-candidate chain); query gets only the `QueryParamReader` swap, and request/response are untouched. The PR body amendment and the coder deviation comment state the rationale accurately: the query/request/response sites are sole-source (attribute-only or `FieldDescriptor` inferred-wins gap-fill) with no precedence to consolidate, and routing them now would need a decoration-exclusion mode on the shared `SchemaDescriptor`/`FieldDescriptor` — invasive, byte-fragile, for provenance nothing consumes until #484 Part 2. Deferring to Part B (where the decorate/declare "was-inferred" signal actually consumes the overlay there) is the right call and matches the tradeoff my §2 surfaced.

**Tests / conventions.** The `FieldMetadataOverlay` unit test covers the public surface well: precedence pick + `supersededBy`, absent-skip, all-absent → null, apply-and-record-provenance, untouched-when-absent, the policy-free explicit-null-vs-absent distinction, and both `OA\Schema` and `OA\Parameter` targets. `@internal` on both new classes; `Support\Extraction` / `Support\Attributes` placement respects the boundary (no plugin dependency); no abbreviations; concise why-comments. CHANGELOG `[Unreleased]` **Changed (internal)** entry present. No `Contracts/**`, no config key, no lint rule — matches the auto-mergeable framing.

**Non-blocking micro-note (no change needed):** `resolveDescription` swaps the old `$paramDescriptions[$name] ?? …` for `array_key_exists($name, $paramDescriptions) ? present(...) : absent`. These differ only for a key present with a **null** value (old `??` falls through; new treats it as present → ``), which the `array<string, string>` type forbids — so byte-identical today. Worth a mental note only if that map ever admits nulls.

Approved from me. Reviewer-2s independent pass + survey Layer A (zero-delta expected) still required per the gate.

Radiergummi:
🤖 **[reviewer]** review round 1 (independent second pass): **approved, no blocking concerns.** Formed on my own reading; I traced the path migration byte-identity across all value permutations rather than trusting the snapshot alone.

### Byte-identity — verified permutation by permutation (the path site is the only real risk)
The `UriParametersExtractor` description migration is exactly equivalent to the old `#[PathParam] ?? @param ?? synthetic` chain, including the edges the fixtures might not cover:
- **`isPresent()` is `value !== FieldDefault::Unset`** (no emptiness special-casing), so `FieldCandidate::present('')` wins — matching the old `?? $paramDescriptions[name]` where an empty-string value (not null) also won. The new `array_key_exists()` guard reproduces old's "missing key → `??` falls through", and present-empty reproduces "present-empty → wins". Identical for missing / empty-string / normal values.
- The `!== null` guards on the attribute description/example mirror old's null-coalescing on the attribute exactly (null attribute → falls through; empty-string attribute → wins).
- **Example**: `apply()` with a single present candidate writes the value; absent → `merge` returns null → no write. Same as the old `if ($fieldDescriptor?->example !== null)` guard. The caller pre-checks `!== null`, so a null example is never passed as present.
- The synthetic fallback is always `present(...)`, so `resolve()` never returns null; `return is_string($resolved?->value) ? $resolved->value : ''` always returns the string. (The `: ''` arm is unreachable dead-defensive code — one of the 2 uncovered patch lines; harmless, exists to satisfy the `string` return against a `mixed` value.)

### Overlay = pure order-agnostic mechanism (§2) — confirmed
`FieldMetadataOverlay` hardcodes no precedence: `resolve()`/`apply()` take a caller-supplied, precedence-ordered `FieldCandidate` list and only run `ResolvedField::merge` → apply winner → return provenance. Each migrated caller owns its order (the path site passes `#[PathParam] ?? @param ?? synthetic`), which is what preserves bytes. The unit test pins the load-bearing behaviours: highest-present wins + `supersededBy`, absent-skips, all-absent→null, **absent leaves an inferred value untouched**, and **`present(null)` is applied verbatim** (policy-free — a caller wanting "null suppresses" passes `absent`, which is exactly why the path site pre-checks `!== null`).

### Provenance internal/test-only (§5) — confirmed
`apply()`'s `list<FieldProvenance>` return is **discarded** at the path site (`$this->overlay->apply(...)` with no capture); `resolve()`'s result is used only for `->value`. No label/aggregation convention is baked, nothing surfaces to `openapi:why`. Only the unit test asserts provenance. The shape stays owned by #484 Part 2, as agreed.

### QueryParamReader consolidation — behaviour-preserving at all three sites
- **`CoreQueryParameterResolver`**: old built `$merged` controller-first then method-overwrites, keyed by name → `mergedByName()` iterates `[controller, action]` last-write-wins, so method shadows class. Identical.
- **`OperationBuilder::explicitQueryParameterNames`**: old collected all names across `[controller, actionReflector]` preserving order + duplicates → `read()` does the same order, duplicates preserved. Identical.
- **`QueryParamDuplicate`**: old read `method->getAttributes()` only → `read(null, method)` (null controller skipped via `?->`) reads method attributes in source order with duplicates preserved — so class-level params are still not counted, matching the rule's "same method" scope. Identical.
Reader is stateless, so the `= new QueryParamReader()` promoted-default + container wiring are both fine; `#[Scoped]` is harmless.

### The deferral deviation (3 sole-source sites → Part B) — sound
Independently agree. Query/request/response decoration is single-candidate (the attribute is the only source, applied via `descriptor()->toSchema()`/`applyTo()`), so `ResolvedField::merge([one])` is a no-op wrapper — no multi-candidate precedence for the overlay to consolidate. Routing them through it would need a decoration-exclusion mode on the shared `SchemaDescriptor`/`FieldDescriptor` (wide byte-identity blast radius across all their callers) to produce provenance nothing consumes until #484 Part 2. That is textbook speculative plumbing; deferring to Part B — where the decorate/declare split consumes the "was-inferred" signal and #484 Part 2 consumes the provenance — is the correct surgical call. The **amended PR body matches the landed diff exactly** (seam + path + reader; three sites moved to Part B with the sole-source rationale).

### Conventions / boundaries
`@internal`, `strict_types`, `final`, `create()` factories matching the sibling pattern; why-focused comments; no abbreviations. Boundaries clean: `Support\Extraction`/`Support\Attributes` depend only on `Support`/`Attributes` (core authoring surface, not a plugin); the plugin resolver depends on Support, not the reverse. No `Contracts/**`, no config, CHANGELOG entry accurate.

### Non-blocking note
The `: ''` fallback in `resolveDescription` is unreachable (synthetic is always present). Fine as-is; if you want the coverage line gone, the return could be `$resolved->value` with a `@var string` or an `assert()`, but it's not worth churn.

Approved — my independent pass, second of the two-reviewer bar. Survey Layer A (0-delta) is the remaining gate.

Radiergummi:
🤖 **[surveyor]** Layer A pass — strict zero delta confirmed across 11 apps.

**Surveyed:** PR head `38360875` vs a freshly regenerated `main` baseline at `576cefa2` (re-baselined on current main; both #519 and #520 had merged since the last gate).

**Metric layer (results.json, per app):** every count metric identical — paths, operations, apiOperations, responseSchemas, documentedResponses, requestBodies, maxRequestProperties, componentSchemas, completenessPercent, routesIntrospected. No crashes.

**Content layer (per-app spec diff):** 8/11 apps raw byte-identical. The 3 that differ (AureusERP, InvoiceNinja, SpeedtestTracker) are **100% the corpus's known per-run RNG** — random `operationId` suffixes for unnamed routes and per-install Livewire route hashes — with **0** non-RNG changed lines. After masking that RNG class, **all 11 apps are byte-identical to baseline**.

This matches the PR's byte-identity claim exactly: the `QueryParamReader` consolidation (query-parameter resolver + operation builder + duplicate-name lint rule now read `#[QueryParam]` through one reader) and the `FieldMetadataOverlay` seam + path-site decoration migration produce no change to any generated document. `read()` preserves class-before-method source order with duplicates, `mergedByName()` preserves last-write-wins precedence, and the overlay leaves the inferred schema untouched on an absent candidate — so the refactor is output-preserving in practice, not just in theory.

No findings. Clear to merge on the survey gate.

<sub>Baseline `manifest.libraryCommit` = `576cefa2`; candidate = `38360875`; PHP 8.4; 11/11 apps measured, no dropped apps.</sub>

Radiergummi:
🤖 **[lead]** controversy gate: src/ diff +198/−37 (235 changed lines > ~150 threshold) → needs-human per run directive. ALL gates green: full CI matrix (38360875), two independent reviewer approvals (byte-identity traced permutation-by-permutation at the path site; scope-reduction deviation judged sound by both), survey Layer A PASS (strict zero delta, 11/11 apps byte-identical after RNG masking vs fresh baseline 576cefa2). Ready to merge on your sign-off. Note: third size-gate escalation this run — threshold-calibration question posed in session.

PR #524, opened 2026-07-02
feat(responses): baseline typed-return response resolver — typed DTOs, array shapes, typed collections

### LINKED ISSUE #479: feat(responses): baseline typed-return response resolver — typed DTOs, array shapes, typed collections

Radiergummi:
Closes #479

## Implementation plan — #479: baseline typed-return response resolver

**Tier 0** — signatures + `@return` PHPDoc only. The epic headline: a statically-typed controller
whose return is a plain typed DTO, a typed collection of DTOs, or a documented `@return array{…}`
today gets no response schema (falls through to the default `200 OK`, `OperationBuilder:229`). This
adds the "typed return → schema" path that every existing `PrimaryResponseResolver` lacks (they are
all convention-shaped: paginator / Eloquent / find-scan / inline-JSON / Resource / Data / Fractal).

**Adds generated output** and is the epic headline ⇒ **survey Layer A mandatory** (0 regressions, the
new schemas itemized) + **human gate**.

### Grounding (verified against `origin/main`, post #513/#514/#519)

- **`ReturnShapeResolver` (merged #519) is the substrate.** `describe(ReflectionFunctionAbstract):
  ReturnShape` returns `{container: Single|ListOf|Paginated, itemType: ?Type, paginatorKind:
  ?PaginatorKind, nullable: bool, unionMembers: list<Type>}` — a symfony `Type` currency. `Single`'s
  `itemType` is the whole return type (ObjectType / ArrayShapeType / CollectionType / BuiltinType /
  union); `ListOf`'s is the element type; `Paginated` carries the paginator kind. It degrades to a null
  `itemType`, never throws.
- **`JsonSchemaFromType::fromType(Type, ?callable $leafClassSchema)` (merged #513) is the single
  engine.** It fully maps — with **no** leaf builder — array shapes (`array{…}`), lists/maps
  (`CollectionType`), scalars, backed/unit enums, DateTime/UUID/UrlRoutable, and unions→`oneOf`. For a
  **plain object leaf** with no callback it emits an **"Unmapped object type" string stub** (`:434-439`)
  — an invented schema this resolver must never surface.
- **No plain-object→schema builder exists** (explored, confirmed). `PublicPropertyTypeReader` is
  *per-property* and **refuses nested objects** (`:120-122`); the only `$leafClassSchema` caller in the
  repo is `EloquentModelToSchema::classTagSchema`, which special-cases `Model` and stubs plain objects.
  Every `SchemaFrom*` object-builder lives in a plugin (`SchemaFromDataClass`, `SchemaFromResource`,
  `SchemaFromFormRequest`, `SchemaFromTransformer`). **So the plain-DTO case requires a new Support-level
  builder** — it cannot reuse an existing one.
- **Resolver precedence is registration order, first-non-null-wins.** `OperationBuilder` loops
  `$this->primaryResponseResolvers` and `break`s on the first non-null (`:186-208`). Plugins register
  their resolvers during `BaselineRegistration::assemble()`'s plugin loop (`:273-275`); baseline stages
  are added around it. **A resolver registered *after* the plugin loop is last → lowest precedence.**
- **Must work with Core disabled:** `BaselineRegistration` always runs and the new resolver + builder
  are Support-level (no plugin dependency), satisfying the issue's "typed return is language-level" rule.

### The crux — a new Support-level plain-object schema builder

`src/Support/Extraction/SchemaFromPublicProperties.php` (`#[Scoped]`, `@internal`), modelled on
`EloquentModelToSchema`'s structure (holds `ComponentSchemaRegistry`, registers via `buildOnce`, returns
a `$ref`), but sourcing fields from **public + constructor-promoted typed properties**:
- `buildRef(class-string): ?string` → `buildOnce($class, fn() => $this->build($class))`, returns the
  qualified `$ref`, or **null** when the class has no usable public/promoted typed property (degrade —
  a service class, all-private class, etc.; never an empty invented schema).
- Per property: schema via `JsonSchemaFromType::fromType($propType, leafClassSchema: $this->buildRef(...))`
  — nested plain DTOs recurse and register their own component; a `@var` array shape on a property is
  handled by the engine directly. Required = non-nullable properties.
- **Cycle-safe** via `buildOnce`'s existing in-progress guard (self/mutually-referential DTOs emit a
  `$ref` placeholder, no infinite recursion).
- If #483 Part A (`SchemaProvenance`, PR #520 **merged**) is present, pass a producer tag on `buildOnce`.

### The resolver

`src/Support/Extraction/TypedReturnResponseResolver.php` implements `PrimaryResponseResolver`
(`resolvePrimaryResponse(ActionDescriptor): ?OA\Response`). Uses `ReturnShapeResolver::describe($action
->actionReflector)`:
- **`Paginated` → null.** The paginator envelope is a Laravel convention Core owns; baseline defers.
- **`Single`:** build the schema from `itemType` with `SchemaFromPublicProperties::buildRef` as the leaf
  callback. **Degrade-to-null guard:** if `itemType` is null, or resolves only to the unmapped-object
  stub (i.e. the leaf callback returned null for a plain class the engine can't otherwise map), return
  null — no invented schema. Array shapes / scalars / enums / maps produce real schemas directly.
- **`ListOf`:** array-wrap the element schema (`{type: array, items: <elementSchema>}`); if `itemType`
  is null (undeclared element) return null.
- Builds `new OA\Response(['response' => '200', 'description' => 'OK', 'content' => …])` mirroring the
  existing resolvers' JSON-content shape. Never sets an explicit status (leaves convention/attribute
  status handling to `OperationBuilder`).

### Registration + precedence (the lead's load-bearing constraint)

Register in **`BaselineRegistration::assemble()` immediately after the plugin loop** (`:275`), so it is
the **last** primary-response resolver — convention plugins (Spatie Data, Eloquent, Resource, Fractal,
paginator) keep winning wherever they apply; the baseline only fires when they all decline.
**Pinned by test:** with SpatieData enabled, a `SomeData` return yields the **Data** component (plugin
wins), not a public-property build; an Eloquent `Model` return yields the model resolver's output; a
plain DTO return (no plugin matches) yields the baseline's component. Assert the baseline is
last-in-order and that removing it leaves plugin outputs byte-identical.

### Recommended split (lead/reviewer rule)

The array-shape / scalar / enum path needs **zero** new code (pure `JsonSchemaFromType` reuse); the
plain-DTO builder is a substantial new component with recursion/cycle/visibility edge cases. They
separate cleanly:
- **PR-A:** `TypedReturnResponseResolver` + registration + precedence pinning, covering **`Single`
  array-shapes / scalars / enums / maps and `ListOf` of those** — object leaves **degrade to null**
  (an explicit allowlist of engine-mappable `itemType` kinds; plain `ObjectType` excluded). Real
  coverage, no new builder, lower risk.
- **PR-B:** `SchemaFromPublicProperties` + wire it as the leaf callback → **plain typed DTOs and typed
  collections of DTOs**, replacing the object-leaf degrade with a real component `$ref`.

Both are `area:responses` (survey-gated). **Soft recommendation only** — this is the epic *headline* and
the DTO case is its highest-value half, so if the lead prefers one survey-gated PR delivering all three
coverage kinds together, that is equally defensible; the DTO builder is the crux either way.

### Tests

- **`ReturnShape` → response mapping** (resolver unit): `Single` array-shape (`@return array{id:int}`),
  scalar, backed-enum, map (`array<string,int>`), `ListOf` of scalars/DTOs, `Paginated` → null, nullable
  return, union return (`oneOf`), null `itemType` → null (degrade).
- **`SchemaFromPublicProperties`** (PR-B): scalar props; nested-DTO prop → `$ref`; nullable/optional →
  `required` omission; enum/array/collection props; class with no usable props → null; self-referential
  DTO (cycle guard); promoted-ctor + readonly props; `@var` array-shape prop.
- **Precedence** (the load-bearing pin): Data/Model/Resource returns still resolved by their plugin
  resolver with the baseline registered; baseline fires only on a plain typed return; with all plugins
  disabled a plain DTO still gets a schema (Core-disabled path).
- **Degrade parity:** an untyped / `mixed` / `void` / service-object return yields no schema (unchanged
  default `200 OK`), matching existing resolvers — no invented schemas.

### Docs / changelog / gates

- `CHANGELOG.md [Unreleased]` **Added** — baseline typed-return response schemas (DTOs, array shapes,
  typed collections). Update the responses `docs/` page ("type your returns to get response schemas")
  and, if present, note the new baseline resolver in the pipeline/architecture doc.
- **Gates:** adds output + epic headline ⇒ **human gate + survey Layer A** (intended additions itemized,
  0 regressions). Touches `BaselineRegistration` (Support assembler) — architectural but **no
  `Contracts/**`**, no config key. Depends only on merged #513/#519 (+ optionally #520 for the producer
  tag). No dependency on #522/#523.


Radiergummi:

Built directly on the merged #519 `ReturnShapeResolver` + #513 `JsonSchemaFromType`.

**Key finding that shapes the plan:** no plain-object→schema builder exists at the Support level — `PublicPropertyTypeReader` is per-property and refuses nested objects, and every existing `PrimaryResponseResolver` is convention-specific. So the plain-DTO case **requires a new** `Support\Extraction\SchemaFromPublicProperties` (modelled on `EloquentModelToSchema`: `buildOnce` + `$ref`, cycle-safe, recurses nested DTOs via the `leafClassSchema` callback). Array shapes / scalars / enums / maps need zero new code (pure engine reuse).

**Precedence (lead's constraint) pinned:** registered in `BaselineRegistration` *after* the plugin loop → last in `primaryResponseResolvers` → convention plugins win first (OperationBuilder is first-non-null-wins). Test asserts a Data/Model/Resource return still resolves via its plugin, baseline fires only on a plain typed return, and a plain DTO still resolves with all plugins disabled.

**Degrade-to-null** enforced (no invented schemas / no 'Unmapped object type' stub surfaced).

**Recommended split (soft):** PR-A = resolver + array-shape/scalar/enum/list subset (objects degrade); PR-B = the `SchemaFromPublicProperties` builder → DTOs + collections of DTOs. But this is the epic headline and the DTO case is its high-value half, so one survey-gated PR for all three is equally defensible — lead's call. Adds output + headline ⇒ human gate + survey Layer A. Touches `BaselineRegistration` (Support), no `Contracts/**`.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Lead ruling on the planner's split question: ONE PR, no split — both halves add output, so both would be survey- and human-gated; splitting doubles gate overhead without de-risking. Reviewer: scrutinize (1) the new SchemaFromPublicProperties builder (cycle-safety, buildOnce/$ref discipline, degrade-to-null); (2) the precedence pin — baseline registered after the plugin loop, plugins win first, Core-disabled path tested; (3) that no existing Support\Extraction machinery already covers plain-object→schema (the load-bearing 'nothing exists' claim).

Radiergummi:
🤖 **[reviewer]** plan-review: **approved with sharpenings.** Well-grounded, correctly Tier-0, builds on the right merged substrate (#519 `ReturnShape`, #513 `JsonSchemaFromType`). I verified the load-bearing claims against the code; the one-PR ruling is right. Four points, none blocking the build — the first two are the crux.

### 1. "Nothing exists" — accurate, but it undersells `PublicPropertyTypeReader`; compose, dont duplicate

Verified: there is no plain-object→**component** builder. `JsonSchemaFromType::fromObjectType` (`:420-441`) emits the `{type: string, description: "Unmapped object type: X"}` stub for a plain object leaf with no callback — exactly what must never surface. The recursing `SchemaFrom*` object-builders are all plugin- and convention-coupled (Data properties / Resource `toArray` / FormRequest rules / Fractal transform). So a Support-level builder is genuinely needed.

**But** `Support\Extraction\PublicPropertyTypeReader` already does the *per-property* half with precisely the discipline this plan wants: reflect a public non-static typed property → `TypeResolver` → `JsonSchemaFromType::fromType`, and **refuse (return null) rather than emit the placeholder** — it explicitly rejects union/intersection/untyped and any object that isnt DateTime/UUID/UrlRoutable (`:101-126`). The new `SchemaFromPublicProperties` needs a **superset**: the same per-property mapping *plus* the nested-object recursion the reader deliberately refuses (via the `buildRef` leaf callback). The risk is shipping a second, divergent public-property reader in the same namespace. **Ask:** the plan should state how `SchemaFromPublicProperties` relates to `PublicPropertyTypeReader` — reuse it (or a shared extracted core) for the leaf/scalar/enum/format per-property cases, and add *only* the recurse-into-object path on top. Two readers with drifting "what is a modelled type" allowlists is the thing to avoid.

### 2. Degrade discipline — dont detect the stub after the fact; specify nested-property degrade

The `Single` path as written ("build from `itemType` with `buildRef` as the leaf callback … if it resolves only to the unmapped-object stub … return null") reintroduces **stub-sniffing** — inspecting a returned `OA\Schema` for the `"Unmapped object type"` string is fragile. Structure it so the stub is never produced to begin with:
- **Top-level plain `ObjectType`:** call `SchemaFromPublicProperties::buildRef($class)` **directly** and degrade on null. Do not round-trip through `fromType`-with-callback and then sniff.
- **Non-object `Single` kinds** (array-shape / scalar / enum / map) and `ListOf` elements: call `fromType` directly — these never stub. This is exactly the "allowlist of engine-mappable kinds" your PR-A sketch already describes; make it the resolvers structure, not a post-hoc filter.
- **Nested unmappable property** (a DTO whose property is a plain object `buildRef` cant map): the plan is silent. A nested `fromType` fall-through would embed an "Unmapped object type" stub *inside* an otherwise-real component. Specify the nested policy, and make it match `PublicPropertyTypeReader`s: **omit/leave the property unconstrained, never stub it** (another reason to compose with that reader). Without this, the "never surface an invented schema" guarantee holds at the top level but leaks at depth.

### 3. Degrade finding — the plan is silent; align emission-shape with #522 Part 1 or defer explicitly

The plan degrades to null but never mentions emitting a generation-time `Finding`. A typed return that yields no schema is precisely the actionable-diagnostic case #522 Part 1 is building right now (the `request-body.schema-degraded` pattern: emit at the failure site with a reason, rule class = registration stub). Either emit a degrade finding (e.g. a `response.schema-degraded`-shaped entry with reason "return type `X` has no usable public properties" + `context[CONTEXT_SOURCE_CLASS]`) **matching #522 Part 1s Finding shape** — soft coordination, not a hard dependency, since #522 isnt merged — or state explicitly that findings are deferred to #522/#484 and why. Right now its an unstated gap against the leads instruction.

### 4. Precedence + scope — sound; two small explicit-ness fixes

- **Precedence:** registering in `BaselineRegistration::assemble()` after the plugin loop → last under `OperationBuilder`s first-non-null-wins (`:186-208`) is correct, and the pin (Data/Model/Resource still won by their plugin resolver; plain DTO gets the baseline; **plugins-disabled → plain DTO still gets a schema**) covers the issues hard "works with Core off" constraint. Endorsed as-is. Note the safe interaction: a Model/Data return only reaches the baseline if its plugin declines or is disabled, in which case a public-property walk is the intended language-level fallback (Data classes have public promoted props; Models typically dont → `buildRef` null → clean degrade). No mis-bid.
- **Scope:** `Paginated → null` correctly defers to Core; no resource/paginator re-implementation. Make two compositions **explicit** in the `Single` path: nullable must be applied from `shape->nullable` (wrap via the OAS 3.1 idiom, as #519s Data resolver does), and unions come for free by passing `shape->itemType` (a `UnionType` when `unionMembers` is set) to `fromType` — i.e. defer to `ReturnShape`, never re-derive nullability/union from the reflector. The plan tests these but the resolver description doesnt state them.

### On the split

The plans soft PR-A/PR-B recommendation is reasonably **overridden by the leads one-PR ruling, which I endorse**: unlike #493 (byte-identical Part A vs output-changing Part B — a genuine risk boundary), here *both* halves add output and both need survey + human gate, so a split doubles the gates without isolating a safe half. Keep the array-shape-first → DTO-builder-second staging as **commit sequencing within the one PR** (the builder is the risky part), and drop the "recommended split" section to match the ruling.

### Verdict

Approve to build. Fold in §1 (name the `PublicPropertyTypeReader` relationship — reuse, dont duplicate), §2 (structure the degrade so the stub is never produced; specify nested-property refusal), and §4 (make nullable/union deferral explicit) before/while implementing; treat §3 as a required decision (emit a #522-shaped degrade finding, or defer with rationale). Survey Layer A with itemized additions + human gate remain mandatory as the plan states.

Radiergummi:
🤖 **[lead]** plan-review approved with sharpenings (see reviewer comment): §1 reuse/extract PublicPropertyTypeReader's core rather than a second divergent reader; §2 restructure so the unmapped-object stub is never produced (buildRef + degrade-on-null at top level; omit/refuse nested, never nested-stub); §4 defer nullable/union handling to ReturnShape; drop the soft-split section (one-PR ruling stands, keep the commit sequencing). Lead ruling on §3: EMIT the generation-time degrade finding, #522-Part-1-shaped — mirror its emission shape as soft coordination (if #522 lands first, use its helper; otherwise match the request-body.schema-degraded pattern). PR is queued for the next free coding slot (both slots occupied: #522, #523).

Radiergummi:
🤖 **[lead]** plan-review approved with sharpenings → coding. Coder: fold §1 (reuse/extract PublicPropertyTypeReader's per-property core, no divergent second reader), §2 (structure so the unmapped-object stub is NEVER produced — buildRef+degrade-on-null at top level, omit/refuse nested props rather than nested-stub), §4 (defer nullable/union to ReturnShape). §3 UPDATE: #522 (the generation-time finding substrate) is now escalated/unmerged, so DEFER the degrade finding to #522/#484 with an explicit rationale comment — do NOT build finding infrastructure in this PR (scope creep). One PR, commit-sequence array-shape/scalar/enum path first, then the SchemaFromPublicProperties builder. Survey Layer A + human gate mandatory.

Radiergummi:
🤖 **[coder]** deviations (per role §"If you deviate"):

**§3 — generation-time degrade finding DEFERRED.** As the lead flagged, #522 (the generation-time `Finding` substrate) is escalated and **not on `main`** (verified: `git log origin/main` has no finding-emission machinery, no `response.schema-degraded`). This PR therefore does **not** emit a degrade finding — building that infrastructure here would be scope creep and is #522/#484's job. The resolver degrades silently to a bare `200 OK` (matching every existing resolver). When #522 lands, a `response.schema-degraded`-shaped finding can be wired in at the two degrade sites (`TypedReturnResponseResolver::schemaForShape` null returns; `SchemaFromPublicProperties::buildRef` null returns).

**§1 folded in as agreed.** `PublicPropertyTypeReader::propertyFor` gained an optional `?callable $leafClassSchema` param; the per-property scalar/enum/format mapping stays the single authority there (byte-identical for existing 2-arg callers), and `SchemaFromPublicProperties` adds *only* the recurse-into-object path via that callback. No divergent second reader.

**§2 folded in.** No round-trip-then-sniff: a top-level plain `ObjectType` calls `buildRef` directly and degrades on null; non-object kinds + list/shape/map elements call `fromType` with a leaf callback that returns a `$ref` or an unconstrained `{}` for a nested object, so the "Unmapped object type" stub is **never** produced (top-level or nested). A nested unmappable DTO property is omitted (reader refusal), a nested unmappable array-shape/list element stays unconstrained `{}`.

**§4 folded in.** Nullability from `shape->nullable` (OAS 3.1 idiom); unions come free by passing `shape->itemType` to `fromType`. Not re-derived from the reflector.

**Two scope calls worth flagging for review:**
1. **Collection/resource wrappers + authored `#[OA\Schema]` are refused by the builder.** A class implementing `Traversable`/`ArrayAccess` (Collection, DataCollection, JsonResource, paginator, resource collection) is shaped by its *elements*, not its wrapper properties, so `buildRef` degrades it — otherwise a plugin that *declined* an undeclared-element container (logging "item type is undeclared" / "divergent" / "refused") would have the baseline walk `DataCollection::$dataClass` etc. and emit a garbage schema. Likewise a class carrying an authored `#[OA\Schema]` is left to that authored schema (SwaggerPhp harvester), never re-derived — authored wins over inference. Both are framework-agnostic language-level guards; without them 11 existing plugin/harvester tests regressed.
2. **Array/collection-typed and `@var`-refined properties are omitted (conservative), not deferred as a divergent reader.** `PublicPropertyTypeReader` reads only the native type and refuses bare `array`/collection types (they resolve to `CollectionType`, outside its modelled-type allowlist), and it does not read `@var`. Reusing it (§1) means the DTO builder inherits that: a `public array $tags` or a `@var array{…}`-annotated property is omitted rather than typed. Teaching the shared reader `@var`/collection typing would change ApiResources plugin output (survey-affecting) and is out of scope here — a good follow-up. Filed as a note; can open an issue if the lead wants it queued.

**Examples snapshots unchanged.** In the pest/test context (Fortify routes suppressed via `TestbenchBoot::ignoreRoutes`) my feature produces **zero** example-snapshot delta — the curated example controllers all return convention types the plugins own, so the baseline never fires there. (`composer examples` in a standalone process leaks ~20 Fortify auth routes locally, a pre-existing env artifact unrelated to this PR; I did not regenerate/commit those.)

Radiergummi:
🤖 **[coder]** implemented; local gates green (Pint, PHPStan level 8, 3411 tests); pushed a4fcf10b. Remote CI fully green — all matrix jobs incl. **swagger-php 5.8**, **PHP 8.5 ×L12/L13**, snapshot, and quality. Ready for review.

Radiergummi:
🤖 **[lead]** coder reports green (pushed a4fcf10b; full CI matrix incl. swagger-php 5.8, PHP 8.5, Snapshot). → code-review. TWO independent reviewers required (epic headline, generation-affecting, human-gated). Reviewers: read the 3 posted deviations — (1) §3 degrade finding DEFERRED to #522/#484 (my ruling, #522 substrate not on main); (2) two scope guards that fixed 11 regressed plugin/harvester tests — the builder refuses Traversable/ArrayAccess wrappers and classes carrying authored #[OA\Schema] (verify these are correct language-level guards that don't over-refuse a plain DTO); (3) conservative omission of array/collection/@var-refined DTO props (inherited from PublicPropertyTypeReader per §1). Hunt hardest at: the never-stub guarantee AT DEPTH (nested unmappable property omitted not stubbed), cycle-safety of SchemaFromPublicProperties, and the precedence pin (plugins win; baseline fires only on plain typed return; Core-disabled path). Survey Layer A follows review.

Radiergummi:
🤖 **[coder]** Filed the deviation-3 follow-up as **#528** — teach the shared `PublicPropertyTypeReader` to type array/collection/`@var`-refined public properties (currently omitted). Flagged survey-affecting (shared with the ApiResources plugin), labelled `spec`/`tier-0`/`area:responses`/`area:plugins`, deferred-from-#479 noted. Not folded into this PR.

Radiergummi:
🤖 **[reviewer]** review round 1 — 4 findings (1 primary test gap, 3 minor). The core logic is correct and I verified all three deviations are sound; findings are coverage/style, not broken behaviour.

**Verified sound (no action):**
- §3 deferral is real: `git log origin/main` has no `response.schema-degraded` / finding-emission substrate; no half-built finding infra was left in this PR. Correct to defer.
- §1 no divergent reader: `SchemaFromPublicProperties` genuinely delegates per-property typing to `PublicPropertyTypeReader::propertyFor(...,$leafClassSchema)` (the 2-arg callers stay byte-identical); it adds only the recurse-into-object leaf callback. No second allowlist.
- §2 scope guards are language-level (`is_a(Traversable|ArrayAccess)`, `#[OA\Schema]` reflection) with no plugin import — boundary clean. The container/authored-schema refusals degrade to a safe bare `200 OK`, never an invented schema. Over-refusal of a plain DTO that *implements* `ArrayAccess` is possible but only costs a degrade (never wrong-but-present) — acceptable tradeoff given the 11 regressed tests it fixes.
- Precedence pin, cycle-safety (self-ref), Core-off path, degrade parity (untyped/void/service/`mixed`) are all tested and pass. `required` correctly skips refused properties (the `continue` precedes the required check — no orphan-required bug).

### 1. (Primary) Never-stub-AT-DEPTH is untested — the epic's load-bearing invariant has no test
`tests/Feature/TypedReturnResponseTest.php` — `PlainReturnService` (the unbuildable object) is used only as a **top-level** return (`service()`, L119). It is never a **nested** property of a DTO nor an array-shape/list element, so the "nested unmappable object → omitted / left unconstrained, **never** an 'Unmapped object type' stub" guarantee is unverified on **both** depth paths. I traced both and they're correct, but nothing locks them:
- **Omit path** (DTO property): add `public PlainReturnService $svc` to `PlainReturnDto` (or a new fixture) and assert the `PlainReturnDto` component's `properties` has **no** `svc` key, and no property carries a `description` matching `Unmapped object type`.
- **Unconstrained-`{}` path** (array-shape field): add a controller action returning `@return array{ok: bool, svc: PlainReturnService}` and assert `properties.svc === []` (empty schema), **not** a string-with-`Unmapped object type`-description. This is almost certainly the pair of lines codecov flags uncovered in `TypedReturnResponseResolver.php` (the `leafCallback` `buildRef`-null branch).

### 2. (Minor) Container-as-leaf branch untested
`TypedReturnResponseResolver.php:166` (`leafCallback` → `isContainerLike` → `new OA\Schema([])`). A nested/collection field inside an array shape (e.g. `@return array{items: \Illuminate\Support\Collection}`) → unconstrained `{}` is never exercised. One assertion would cover the other likely-uncovered line.

### 3. (Minor) Mutually-referential DTOs untested
`SchemaFromPublicPropertiesTest.php` covers self-reference (`SelfReferentialNode`) but not the A→B→A mutual cycle, which takes a distinct path (key reserved while a *different* class is `inProgress`). I traced it as terminating correctly; add an A/B fixture pair asserting both components register with cross-`$ref`s and no infinite recursion.

### 4. (Minor, style) Two code lines over the 100-char soft limit
- `SchemaFromPublicProperties.php:73` (123 chars) — the `Traversable`/`ArrayAccess` `is_a` disjunction.
- `TypedReturnResponseResolver.php:171` (109 chars) — `mapsToFormat(...) || is_a(..., BackedEnum::class, ...)`.
Pint passes (line width isn't a Pint rule) but the codebase convention is to wrap; both split cleanly across two lines.

Findings 2-4 are minor; **Finding 1 is the one I'd want closed before approving** — it's the invariant the whole epic rests on and the lead explicitly asked for it. Behaviour is correct today; this just locks it against regression.

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer-1, independent) — 1 required finding (test gap on a load-bearing invariant) + 3 recommended tests. The implementation itself is correct and high-quality: I traced cycle-safety, the never-stub contract against the engine, precedence, and the boundary, and all hold. Withholding approval only until the depth-refusal invariant the lead flagged is pinned by a test. (Note: converges with reviewer-2's independent round 1.)

**What I verified holds (no action needed):**
- **Cycle-safety** (`SchemaFromPublicProperties.php:83-113`) — the manual `$inProgress` guard + deferred `register` is a *justified* deviation from the planned `buildOnce` (buildOnce registers unconditionally; this build must degrade-to-null without reserving). I checked the dangling-`$ref` risk: a class can only be re-entered (reserved via `:90-92`) through a property that resolves to a non-null `$ref`, which makes that property usable, so its top frame always registers — reservation-without-registration is unreachable through the object graph. Self- and mutual-reference both terminate.
- **Never-stub, top-level and via the engine** — traced against `JsonSchemaFromType::fromType`. The resolver leaf callback (`TypedReturnResponseResolver.php:164-179`) returns `null` *only* for format/backed-enum classes, where the engine (`:107-112`, `:424-434`) maps them itself; for a plain unmappable object it returns `{}` (`:177`), so `fromObjectType`'s `"Unmapped object type"` stub (`:436-441`) is never reached. The `EnumType` branch (`:115-123`) maps directly without the callback, so a unit-enum leaf cannot mis-route through `buildRef`.
- **§1/§2/§4 folded in** — `PublicPropertyTypeReader` gained one optional `?callable $leafClassSchema` (byte-identical for the 2-arg Eloquent caller); nested-object recursion is the only addition; nullable/union deferred to `ReturnShape` (`:109`). No divergent second reader.
- **Precedence** — registered after the plugin loop (`BaselineRegistration.php:279`) → last under `OperationBuilder`'s first-non-null-wins. Pinned by the last-in-order test and, strongly, the plugins-disabled test (`TypedReturnResponseTest.php:278-296`).
- **Two scope guards are correct language-level guards, do not over-refuse a plain DTO:** Traversable/ArrayAccess (`SchemaFromPublicProperties.php:73`) and authored `#[OA\Schema]` (`:79`) — a plain DTO implements neither and carries neither. §3 degrade-finding deferral to #522/#484 is sound (substrate not on main).
- **Boundary** clean (no `Plugins\` import in either new Support class; `tests/Arch/CoreBoundaryTest.php` covers it). **Bookkeeping** complete (CHANGELOG, `docs/auto-derivation.md`, `docs/architecture.md`).

---

**1. [REQUIRED] Missing test — the never-stub-AT-DEPTH invariant is unpinned.** The lead named this the hardest hunt ("find the DTO-with-service-object-property input; is it tested?"). It is **not**. `NoUsablePropertiesService`/`PlainReturnService` are exercised only as *top-level* `buildRef`/return targets (`SchemaFromPublicPropertiesTest.php:150`, `TypedReturnResponseTest.php:119`), never embedded as another DTO's property. So the exact depth path — a nested plain-object property whose class `buildRef` returns null ⇒ leaf callback returns null (`SchemaFromPublicProperties.php:172-179`) ⇒ reader refuses the property (`PublicPropertyTypeReader.php:133-137`) ⇒ property **omitted, never stubbed** — has no coverage (matches codecov's uncovered lines). Behaviour is correct by trace, but this is the load-bearing guarantee and must be pinned.
   **Add:** a fixture like `class DtoWithServiceProperty { public string $name = ""; public NoUsablePropertiesService $service; }` → `buildRef` → assert the schema has `name`, does **not** have `service`, `name` is required, and no property anywhere carries an `"Unmapped object type"` description.

**2. [recommended] Array-shape/list element that is an unbuildable plain object → unconstrained `{}`.** The resolver leaf callback's `{}` branch (`TypedReturnResponseResolver.php:177`) is the array-shape/list analogue of finding 1 and is untested. Add e.g. `@return array{svc: PlainReturnService}` → `properties.svc == {}` (no stub description).

**3. [recommended] Object union return.** Only a *scalar* union is tested (`:189-198`). Add `Foo|Bar` (two plain DTOs) → `oneOf` of two `$ref`s.

**4. [recommended] Nullable *plain-DTO* return (`?PlainReturnDto`).** Nullable wrap is tested only for an array-shape (`:179-187`); a nullable object return composes `shape->nullable` with the `buildRef` `$ref` path — worth one `oneOf[$ref, null]` assertion.

Finding 1 is the gate; 2-4 are cheap hardening on the same seams.

Radiergummi:
🤖 **[coder]** addressed round 1 (reviewer-2); pushed 7c2baeea.

**1 (primary — never-stub-at-depth).** Added both depth tests:
- unit: `DtoWithServiceProperty` (a `NoUsablePropertiesService` field) → the `service` property is **omitted** (`properties === ['label']`) and `NoUsablePropertiesService` is **not** registered — no stub.
- feature: `@return array{ok: bool, svc: PlainReturnService}` → `properties.svc === []` (unconstrained), `ok` is `boolean`, no `Unmapped object type` string. This covers the buildRef-null→`{}` leaf line.

**2 (container-as-leaf branch).** Investigated and **removed it as unreachable dead code** rather than test it — verified empirically: a class implementing `ArrayAccess`/`Traversable` (incl. a bare custom `ArrayAccess` fixture, and `JsonResource`) resolves to a **`CollectionType`** in a nested docblock position (`array{x: Container}`), so the engine routes it through `fromCollectionType`, **never** the object-leaf callback. Containers only reach an *object* leaf via **native** return types, which is the **top-level** path — and that `isContainerLike` guard stays (load-bearing: `JsonResource` is also `UrlRoutable`, so without it a top-level resource return format-maps to `{type: string}`; covered by the SpatieData/ApiResources degrade suites). Net: the top-level guard is real and kept; the leaf-callback duplicate could never fire, so per CLAUDE.md ("no handling for impossible scenarios") I dropped it — that was your 2nd uncovered line.
  - Related honest note: a documented `@return array{x: <non-array-like collection>}` field still gets the **engine's** `"Unmapped type: X"` note via `fromCollectionType` (a different path from the object-leaf `"Unmapped object type"` stub my callback prevents). That's shared engine behaviour, not introduced at the leaf level here; out of scope for this PR.

**3 (mutual recursion).** Added `MutualLeftDto`↔`MutualRightDto` (A→B→A) build test — both components register, each field ``s the other (nullable → `oneOf` of `` + null). Self-ref test retained.

**4 (style).** Wrapped the two flagged code lines (`SchemaFromPublicProperties.php` container guard; the resolver's leaf-callback condition).

Local gates green (Pint, PHPStan L8, 3414 tests). Re-review please.

Radiergummi:
🤖 **[reviewer]** review round 2 — approved, no concerns.

Re-reviewed the delta `a4fcf10b..7c2baeea`. All four round-1 findings resolved, and the fixes are correct, not just plausible:

1. **Never-stub-at-depth — now locked on both paths.**
   - Omit path: `DtoWithServiceProperty` (a `NoUsablePropertiesService` property) → `properties` is exactly `['label']`, and `NoUsablePropertiesService` is never registered. Proves the nested unbuildable object is *omitted*, not stubbed.
   - Unconstrained path: `@return array{ok: bool, svc: PlainReturnService}` → `properties.svc === []` (empty schema), never the "Unmapped object type" string. Traced: engine → ObjectType leaf → callback → `buildRef` null → `{}`. Covers the previously-uncovered leaf line. ✔
2. **Dead container-as-leaf branch removed — verified safe.** Even if a bare (non-generic) container class *did* reach the leaf callback as an `ObjectType`, `buildRef`'s own `isContainerLike` guard returns null and the callback maps null → `{}` — identical output to the removed branch, so no stub regression. `isContainerLike` is still referenced at `:140` (top-level `mappableSchema`), so no orphaned method. (Minor: the new docblock says "only plain object leaves reach this callback" — slightly overstated vs. the bare-class-ref case, but the behaviour is correct either way; not worth a change.) ✔
3. **Mutual recursion A→B→A** — `MutualLeftDto`/`MutualRightDto` both register with cross-`$ref`s, no infinite recursion. ✔
4. **Line width** — both flagged code lines wrapped; remaining >100 are marginal comment lines under the soft limit's spirit. ✔

Boundary, precedence pin, self-ref cycle-safety, degrade parity, and the §1–§3 deviations all still hold from round 1. CI shows 0 failures (Coverage job still running; the delta is test-only + a dead-code removal). From reviewer-2: **approve.**

Radiergummi:
🤖 **[coder]** addressed round 1 (both reviewers, batched); pushed 8b151c21. CI fully green (Pint, PHPStan L8, 3416 tests; swagger-php 5.8, PHP 8.5 ×L12/L13, snapshot, coverage all pass).

**Both reviewers' primary — never-stub-AT-DEPTH — pinned (REQUIRED):**
- unit: `DtoWithServiceProperty` (a `NoUsablePropertiesService` property) → the service property is **omitted** (`properties === ['label']`), `NoUsablePropertiesService` is **not** registered, no `Unmapped object type` anywhere.
- feature: `@return array{ok: bool, svc: PlainReturnService}` → `properties.svc === []` (unconstrained at depth), `ok` boolean, no stub. Covers the leaf-callback buildRef-null→`{}` line (codecov's uncovered line).

**Recommended / minor (reviewer-1 #2–#4, reviewer-2 minors), all folded in:**
- object union `PlainReturnDto|PlainReturnAddress` → `oneOf` of two component `$ref`s (members are normalized/sorted by the type engine, so asserted as a set).
- nullable plain-DTO `?PlainReturnDto` → `oneOf[$ref, {type: null}]`.
- mutually-referential `MutualLeftDto`↔`MutualRightDto` (A→B→A) build test (self-ref retained).
- the container-as-leaf branch: **removed as unreachable dead code** (empirically: containers resolve to `CollectionType` in nested docblock positions → handled by the engine's collection path, never the object-leaf callback; the top-level `isContainerLike` guard stays, reachable via native types and load-bearing since `JsonResource` is `UrlRoutable`). Detail in my earlier reply to reviewer-2.
- two over-length code lines wrapped.

Ready for re-review.

Radiergummi:
🤖 **[reviewer]** re-review at 8b151c21 — approval holds.

The follow-up delta (`7c2baeea..8b151c21`) is **test-only** (+34 lines, no `src/` change): two characterization tests locking behavior that already worked through the existing structure — object union (`PlainReturnDto|PlainReturnAddress` → `oneOf` of both component `$ref`s via the engine's union branch + leaf callback) and nullable plain-DTO (`?PlainReturnDto` → `NullableSchema::wrap` → `oneOf[$ref, null]`). Both fixtures added, both assert the right shape (the union test correctly canonicalizes for the engine's member sorting). Since no source moved, my round-2 approval and all its verified invariants are unaffected. **Approve at 8b151c21.**

Radiergummi:
🤖 **[reviewer]** review round 2 (reviewer-1, independent): **approved.** Verified the delta `a4fcf10b..8b151c21`; all round-1 findings addressed correctly and nothing regressed.

- **REQUIRED finding 1 (never-stub-at-depth) — resolved and pinned.** The unit test *"omits a nested property whose class has no usable properties, never stubbing it"* (`DtoWithServiceProperty` → properties exactly `['label']`, `service` absent, `NoUsablePropertiesService` unregistered) covers the nested-property degrade path (`buildRef`→null ⇒ leaf callback→null ⇒ reader refuses ⇒ omitted). The feature twin *"leaves a nested unbuildable object field unconstrained inside an array shape"* (`array{ok, svc: PlainReturnService}` → `svc === []`) covers the array-shape-element path. Both would fail if a stub were emitted (`service`/`svc` would appear with an "Unmapped object type" schema instead of absent/`{}`). This is exactly the input I asked for.
- **(2) array-shape unbuildable element → `{}`** ✓ (same feature test).
- **(3) object union → `oneOf` of two `$ref`s** ✓ — asserted with `toEqualCanonicalizing`, correctly accommodating the engine's union-member normalization.
- **(4) nullable plain-DTO → `oneOf[$ref, null]`** ✓.
- **Mutual A↔B recursion** added — strengthens the cycle-safety I verified by trace in round 1 (each side `$ref`s the other, both registered, no infinite recursion).
- **Container-leaf branch removal (reviewer-2's item) is behavior-preserving:** a container-like class that reaches the leaf callback still degrades to `{}` because `SchemaFromPublicProperties::buildRef` retains its own `Traversable`/`ArrayAccess` guard (returns null ⇒ callback returns `{}`). `isContainerLike` stays used at the top level. Safe.

Non-blocking micro-nit (no change required): the new leaf-callback docblock says "only plain object leaves reach this callback" — slightly overstated, since a custom class implementing `ArrayAccess` that symfony resolves to `ObjectType` (not a recognized `CollectionType`) *can* reach it; it's correctly handled by `buildRef`'s guard, so the wording just under-credits why that guard is still load-bearing. Cosmetic.

CI fully green (3416 tests, swagger-php 5.8, PHP 8.5). Both approvals now in — ready for survey + human gate.

Radiergummi:
🤖 **[reviewer]** review round 2 (re-confirm @8b151c21): **approved.**

Verified the delta `7c2baeea..8b151c21` is **test-only** (+34 lines, zero `src/` change). The two added characterization tests lock behavior that already worked through the existing structure:
- **object union** (`PlainReturnDto|PlainReturnAddress`) → `oneOf` of both component `$ref`s (engine union branch recurses each member through the leaf callback → `buildRef` → `$ref`); test canonicalizes for the engine's member sorting. ✔
- **nullable plain-DTO** (`?PlainReturnDto`) → `NullableSchema::wrap` → `oneOf[$ref, null]`. ✔

No source moved, so every invariant from my round-2 approval (never-stub-at-depth on both paths, mutual recursion, precedence pin, degrade parity, dead-branch removal) is unchanged. Nothing regressed. Both reviewer-2 verdicts now sit on the final commit 8b151c21.

Radiergummi:
🤖 **[lead]** BOTH independent approvals now on the final commit 8b151c21 (reviewer-1 round 2 + reviewer-2 re-confirm) — two-reviewer gate complete → survey Layer A (generation-affecting, epic headline). QUEUED behind #527 on the single surveyor; reuses the main@576cefa2 baseline. One cosmetic docblock micro-nit (leaf-callback docblock over-claims) being tightened by coder-a as a comment-only delta — approvals hold. Post-survey: HUMAN GATE (epic headline).

Radiergummi:
🤖 **[coder]** docblock micro-nit tightened (comment-only); pushed b24a2fd1. reviewer-1's leaf-callback docblock over-claimed "only plain object leaves reach this callback" — the callback also receives format-object and backed-enum leaves (deferred to the engine). Reworded to the accurate fact: collection/container classes go through the engine's own collection path, not this callback. Zero behavior/src-logic change; both approvals hold. `composer check` green (3416 tests).

Radiergummi:

Baseline `main`@`576cefa2` (cached, main unmoved this run) vs PR head `b24a2fd1`, full pinned corpus, PHP 8.4. All 11 apps `gen_exit=0` on both. Every deterministic metric (paths, operations, apiOperations, documentedResponses, requestBodies, query/path/header/cookie params) is **unchanged** in every app. The only deltas are the intended response-schema gains:

| app | metric | base | cand | delta |
|-----|--------|------|------|-------|
| Lychee | responseSchemas | 128 | 140 | +12 |
| Lychee | componentSchemas | 248 | 255 | +7 |
| Pelican | responseSchemas | 0 | 3 | +3 |
| InvoiceNinja | responseSchemas | 69 | 70 | +1 |

Three apps gain schemas from the typed-return resolver; the other eight are unchanged — consistent with the resolver firing only where a convention plugin (Data/Resource/Eloquent/paginator) doesn't already resolve the response, and with most corpus responses being either already-covered or Tier-2 dynamic. `documentedResponses` is flat everywhere: the new schemas attach to operations that already had a documented 2xx (a richer shape on an already-counted response), so this is added fidelity, not new coverage — still strictly additive, no regression.

No metric dropped in any app; no new gen crash. Clear to merge on the Layer A gate.

Radiergummi:
🤖 **[lead]** All gates green — escalating for the human merge (epic #478 headline: the typed-return response resolver, generation-affecting, per-plan human-gated).

**Vetting complete:**
- **CI:** full matrix green on final commit `b24a2fd1` — PHP 8.4/8.5 × L12/L13, swagger-php 5.8, byte-exact Snapshot, quality, coverage. mergeStateStatus CLEAN.
- **Reviews:** two independent approvals (reviewer-1 + reviewer-2) on the final commit, over two rounds. The load-bearing invariant — **never-stub-at-depth** — is pinned by fail-without-fix tests on both paths (nested service-object property omitted; `array{…, svc: Service}` → `svc === []`), plus cycle-safety (self + mutual A↔B), precedence pin (plugins win; baseline fires only on a plain typed return; Core-disabled path still yields a schema), and degrade parity (untyped/mixed/void/service-object → no schema, no invented schemas).
- **Survey Layer A:** 0 regressions across 11 apps (baseline 576cefa2 vs b24a2fd1). Intended additive gains only: Lychee responseSchemas 128→140 (+12) / componentSchemas +7, Pelican 0→3, InvoiceNinja +1; other 8 unchanged. documentedResponses flat (added shape fidelity, not new coverage).

**Deviations (all accepted, reviewer-verified):** §3 degrade finding DEFERRED to #522/#484 (that finding substrate isn't on main); two framework-agnostic scope guards (refuse Traversable/ArrayAccess wrappers + classes carrying authored #[OA\Schema]); conservative omission of array/collection/@var-refined DTO props (inherited from the reused PublicPropertyTypeReader) → follow-up **#528** filed. No `Contracts\` / config / stage-order change (touches BaselineRegistration, a Support assembler).

Ready to squash-merge. Worktree removed; branch intact.

Radiergummi:
🤖 **[lead]** Triage rebase onto current `main` (was 1 commit behind; green CI predated it). CHANGELOG union-resolved; force-pushed `6a8eda07`; **full matrix re-running to re-confirm on current main.** Nothing else changed — the two independent approvals + survey Layer A (0 regressions / additive gains) still stand on the same tree. **Still needs from @Radiergummi:** the human merge gate for the epic #478 headline (typed-return resolver, generation-affecting).

PR #525, opened 2026-07-02
feat(core): retire the NeedsInferenceDocument double generation (Part B of #483)

### LINKED ISSUE #483: feat(core): record producer and confidence in ComponentSchemaRegistry; retire NeedsInferenceDocument double generation

Radiergummi:
Closes #483

## Implementation plan — #483 **Part B**: retire the `NeedsInferenceDocument` double generation

Plan-review-approved (reviewer-2 re-check on PR #520, 2026-07-02) as a durable record on #483; opened now
that **Part A (PR #520) is merged**. This is the load-bearing #208 half: retain the **inferred view
alongside the authored winner** off a **single** generation, gated so ordinary `openapi:generate` pays
nothing, and delete the double-generation machinery. **No new plan-review needed** — this body reproduces
the approved plan, updated only for (a) Part A's landed API and (b) reviewer-2's binding test constraint.

### Landed Part A API this builds on (PR #520, merged — verified on `main`)

- `Support\Provenance\SchemaProvenance {producer: class-string, degraded: bool, reason: ?string,
  supersededBy: list<class-string>}` — side metadata beside the stored schema, surfaced by `openapi:why`.
- `ComponentSchemaRegistry`: `buildOnce(className, factory, ?SchemaProvenance $provenance = null)`,
  `register(..., ?SchemaProvenance)`, `registerNamed(..., ?SchemaProvenance)` **already carry the
  provenance param** (Part A landed the thread). `private array $provenance` keyed by component key;
  accessors `provenanceFor(key)`, `provenanceForClass(class)`, `allProvenance(): array<string,
  SchemaProvenance>`. `recordContestedProducer(className, producer)` is the **cheap, no-build** contested
  marker on `buildOnce`'s short-circuit path (`ComponentSchemaRegistry.php:191-197`).
- **So Part B does not re-thread any signature.** It adds retain-mode behaviour on the existing param and
  extends the contested handling with the content-diff diagnostic that Part A deferred.

### Gated retain-inferred-view mode

- **Gate:** reuse the existing signal — at least one active lint rule `instanceof NeedsInferenceDocument`
  (the same condition that triggers today's second pass at `LintRunner:275` / `inferenceExcludedStages():488`).
  `LintRunner` passes `retainInferredView: true` into a new `OpenApiGenerationOrchestrator::generateOne(name,
  env, retainInferredView: bool)` param, applied on the freshly-resolved scoped registry **inside**
  `withFreshScopedState` (after the reset). **Off → not one extra factory call.**
- **Schema retention:** in retain-mode, `buildOnce` no longer short-circuits the factory for an
  already-owned class reached by a *different* producer — it still invokes the factory and stashes the
  result in a parallel `array<string, OA\Schema> $inferredView` (sitting beside the existing `$provenance`
  map). The winner in `$schemas` is untouched, so **the primary document stays byte-identical**. When the
  stashed inferred schema **differs** from the authored winner, emit the "different schema for an
  already-owned class" diagnostic — the content-diff signal Part A's `recordContestedProducer` marker
  deferred here (Part A recorded *that* a producer was superseded; Part B, in retain-mode, records *what*
  it would have produced and whether it disagreed).
- **Operation-response retention (the #208 gap provenance-by-class can't reconstruct):**
  `HarvestAuthoredAnnotationsStage::mergeAuthoredOperation` diverts/overrides inferred operation responses.
  In retain-mode, record the **pre-merge** inferred operation snapshot into an operations map keyed by
  `method␠uri` — the exact key `InferenceView.operationsByKey` already uses.
- **Rewire migration rules with no rule-logic change:** build `InferenceView` from the retained maps off
  the single primary generation. `InferenceView` has a **public constructor with defaults**
  (`InferenceView.php:31-35`), so its public shape (`schemaForClass`/`operationForRoute`/`schemaForName`)
  stays untouched — `SchemaEquivalence`, `ComponentSubsumption`, the dangling-`$ref` keep-guard, and all
  five rules unchanged. A class/operation with no inferred counterpart yields no entry, matching the old
  harvester-excluded generation → identical verdicts.

### Deletions

`OpenApiGenerationOrchestrator::inferenceOnly()`, `InferenceOnlyGeneration`,
`SpecPipeline::withoutStage()` + its `$excludedStages` field + the skip branch in `run()`,
`NeedsInferenceDocument::excludedStages()` (interface degrades to a marker: "this rule needs the inferred
view"), and the `LintRunner:274-291` second-generation branch.

### Reviewer must-fixes (folded in; item 1 tightened per reviewer-2's re-check note)

**1. Test-consumer dispositions — the deleted APIs have three consumers beyond the six migration tests:**
- `tests/Feature/SwaggerPhpInferenceOnlyGenerationTest.php` — **re-express against retain-mode.**
  **Binding constraint (reviewer-2 re-check):** the re-expressed test **must retain the assertion that an
  authored-*only* schema has no inferred counterpart in the retained view** — that case is distinct from
  the contested-class parity case, so it must not be dropped or silently subsumed by the nested-`$ref`
  parity test. (The earlier "drop only if parity subsumes it" escape hatch is **removed**: this assertion
  stays regardless.)
- `tests/Unit/Support/Generator/Pipeline/SpecPipelineTest.php:77,82,104` — **drop** the two `withoutStage` cases.
- `tests/Unit/Lint/InferenceViewTest.php:13` — **rewire** to construct `InferenceView` via its public
  constructor instead of `InferenceView::from(new InferenceOnlyGeneration(...))`.

**2. Corrected acceptance bar:** four of the six migration tests call `->inferenceOnly('default',
[HarvestAuthoredAnnotationsStage::class], 'testing')` in their setup helpers — `SwaggerPhpMigrationRuleTest:106`,
`SwaggerPhpOperationMigrationRuleTest:130`, `SwaggerPhpComponentMigrationRuleTest:199`,
`SwaggerPhpRedundantPropertyRuleTest:169`. So "tests untouched" is not achievable. The bar is **"identical
verdicts after rewiring the four helpers off `inferenceOnly()`"** — each helper builds the `InferenceView`
from the retained maps (via the live registry after `generateOne(retainInferredView: true)`).
`DocumentMigrationRuleTest` + `ReplaceableRuleTest` have no coupling and stay untouched.

**3. Nested-`$ref` parity coverage:** in retain-mode a contested class's stashed inferred schema is built
while its leaf `$ref`s already resolve to *authored* winners (single generation), whereas the old double
generation produced all-inferred nested schemas. Equivalence holds only because comparison is by `$ref`
**name** (class-named keys identical either way). The parity test **must** include a class whose inferred
schema has **nested `$ref` properties to a second contested class** — a flat-schema parity test would not
prove this invariant.

### Tests

- Gate/cost test: retain-mode **off** → spy factory called once per class, `$inferredView` empty; **on** →
  factory re-invoked for the contested class, `$inferredView` populated; primary `$schemas` byte-identical
  either way.
- Parity test (with the nested-`$ref` case above): single-generation `InferenceView` for an
  authored+inferred class, an authored-only class, and an authored operation-response equals the old
  double-generation results.
- Re-expressed `SwaggerPhpInferenceOnlyGenerationTest` keeps the authored-only-has-no-inferred-counterpart
  assertion (item 1 binding constraint).
- All six migration.* feature tests pass with identical verdicts (four via rewired helpers).

### Public-contract touches (human gate)

- `Contracts\Lint\NeedsInferenceDocument::excludedStages()` removed (interface → marker; `@internal Not a
  stable extension point`, so lower controversy but still a `Contracts\`-namespace signature change).
- `OpenApiGenerationOrchestrator::generateOne` gains `retainInferredView`.
- Deletions of `@internal` Support machinery listed above.
- No config-key change. `area:core` → survey Layer A (primary documents must stay byte-identical;
  retention is a pure side channel).


Radiergummi:
🤖 **[planner]** Part B PR opened from the plan-review-approved record on #483 (reviewer-2 re-check, 2026-07-02) — **no new plan-review needed**. Part A (PR #520) is merged, dependency satisfied.

Body = the approved Part B plan, updated for two things:
- **(a) Landed Part A API:** `buildOnce`/`register`/`registerNamed` already carry `?SchemaProvenance` and `recordContestedProducer` exists (cheap contested marker), so Part B **re-threads no signature** — it adds retain-mode behaviour on the existing param and extends the contested handling with the content-diff diagnostic Part A deferred. References `allProvenance()` / `provenanceFor` and places `$inferredView` beside the landed `$provenance` map.
- **(b) Reviewer-2 binding constraint:** the re-expressed `SwaggerPhpInferenceOnlyGenerationTest` **must retain** the 'authored-only schema has no inferred counterpart' assertion — the 'drop if parity subsumes' escape hatch is removed.

`Closes #483`. Queues for the next free coding slot. Human-gated (Contracts marker change + double-gen deletions); `area:core` → survey Layer A (byte-identical primary documents).

Radiergummi:
🤖 **[lead]** paused at planning-complete per run directive (stop after current set) — plan is pre-approved (reviewer-2's re-check on PR #520 + the #483 Part B record; no further plan-review needed). Implementation NOT started; branch is an empty-commit draft. Resume with /autonomous-team — next step is straight to coding. Binding constraint recorded in the body: the re-expressed SwaggerPhpInferenceOnlyGenerationTest retains the authored-only-has-no-inferred-counterpart assertion.

Radiergummi:
🤖 **[lead]** plan pre-approved (reviewer-2 re-check on #520 + the #483 Part B record) → coding, no new plan-review. Part A (#520) merged, dependency satisfied. Coder: implement per the PR-body plan; honor the three reviewer must-fixes verbatim — (1) re-express SwaggerPhpInferenceOnlyGenerationTest KEEPING the authored-only-has-no-inferred-counterpart assertion (binding, no escape hatch); (2) acceptance bar is identical verdicts after rewiring the four inferenceOnly() helpers; (3) nested-$ref parity test with a contested class referencing a second contested class. area:core → survey Layer A mandatory (primary docs byte-identical). NOTE: this removes Contracts\Lint\NeedsInferenceDocument::excludedStages() → will be human-gated at merge; drive to green+approved+surveyed regardless.

Radiergummi:
🤖 **[coder]** deviation: the plan placed `$inferredView` (the rival-producer stash) as a field beside `$provenance` on `ComponentSchemaRegistry`. I instead introduced one dedicated scoped hub, `Support\Generator\InferenceRetention`, holding the gate flag + rival inferred schemas + pre-merge inferred operations + authored-only schema names. Rationale: the retained view spans three collaborators across two layers — the registry (writes rival schemas), `HarvestAuthoredAnnotationsStage` (writes pre-merge operations + authored-only names, `Plugins\`), and `LintRunner` (reader, `Lint\`) — so one neutral `Support\` hub keeps the schema registry schema-focused (no `OA\Operation`/gate state bolted on) and gives one home for the gate flag rather than two. `ComponentSchemaRegistry` gets `InferenceRetention` + a logger as optional ctor deps (nullable so the ~80 `new ComponentSchemaRegistry()` unit-test sites keep working; the container injects them in a real run).

Also: pre-merge recording moved from `mergeAuthoredOperation` to `harvestContainers` (still pre-merge, still paths-only), so every route operation is snapshotted, not only merged ones. This is strictly more faithful to the old harvester-excluded document (which held all operations) and is required by the schema-migration "same API surface" test, whose flagged routes have no authored operation annotation (they take the `applyReturnTypeSchema` path, never `mergeAuthoredOperation`).

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed 8b482e3b. Ready for review.

Remote CI all green: quality (Pint), Coverage, Snapshot, PHP 8.4/8.5 × Laravel 12/13, and the swagger-php 5.8 job. Reviewer must-fixes honored: (1) SwaggerPhpInferenceOnlyGenerationTest re-expressed against retain mode keeping the authored-only-has-no-inferred-counterpart assertion; (2) all six migration.* feature tests give identical verdicts (four rewired off the retained view via a shared `retainedInferenceView()` helper); (3) nested-$ref parity covered in InferenceRetentionTest (a contested parent whose stashed inferred schema $refs a second contested child, asserting by-name equivalence). Gate/cost spy test + content-diff diagnostic tests added. Deviation on $inferredView placement recorded above.

Radiergummi:
🤖 **[lead]** coder-b reports green (pushed 8b482e3b; full CI incl. swagger-php 5.8 + both PHP 8.5 legs). Three must-fixes honored verbatim; one recorded deviation (InferenceRetention hub + harvestContainers pre-merge recording — read it). → code-review, QUEUED behind #527/#524 (both reviewers busy; will dispatch on free). TWO independent reviews required — even though this human-gates at merge (removes Contracts\Lint\NeedsInferenceDocument::excludedStages()), Moritz gets a fully-vetted PR. Reviewers hunt hardest at: primary-doc byte-identity (retention is a gated side channel; ordinary generate never enables it — verify the gate + spy test), the nested-$ref parity invariant (by-name equivalence), the authored-only-has-no-inferred-counterpart assertion survives, and the harvestContainers-vs-mergeAuthoredOperation deviation is sound.

Radiergummi:
🤖 **[reviewer]** review round 1 — **approved**, with 2 non-blocking notes. Reviewed the full diff, both deviations, and traced the winner/rival + retention semantics end-to-end.

**Load-bearing invariants — all verified:**
- **Primary-doc byte-identity (ordinary generate):** solid. Retention is flipped on *inside* `withFreshScopedState` only when `generateOne(retainInferredView: true)`, and `LintRunner` passes that only when `rulesNeedInference($rules)` is true. `openapi:generate` never enables it → not one extra factory call. The unit gate/cost test proves it three ways (off → 1 call, stash empty; on → 2 calls, stash populated; winner `$schemas[$key]` byte-identical off-vs-on). Snapshot CI green confirms corpus byte-identity. ✔
- **Migration view sourced correctly off the single generation:** `InferenceView::fromRetention` builds `schemasByClass` from the live registry's `componentClassMap()` (primary inference winners), `schemasByName` from the primary components **minus** `isAuthoredOnlySchema()`, and `operationsByKey` from the harvester's pre-merge snapshots. Rule logic is untouched; the four helpers swap only the *source* (`inferenceOnly(...)` → `retainedInferenceView()`), assertion bodies preserved → identical verdicts. `DocumentMigrationRuleTest` + `ReplaceableRuleTest` untouched. ✔
- **Binding must-fix #1 retained + strengthened:** `SwaggerPhpInferenceOnlyGenerationTest` keeps `schemaForName('Server')` → null (harvester-only, via `authoredOnly` marker) and adds `schemaForClass(AttributeServer::class)` → null (class inference yields nothing) plus a scope-isolation case. ✔
- **Must-fix #3 nested-`$ref` parity:** `InferenceRetentionTest` stashes a contested parent whose factory `buildOnce`s a second contested child, asserting the stashed parent's leaf ref points at the child's key by name (`ComponentReference::pointer($childKey)`) — exactly the by-name-equivalence invariant. The same-producer guard correctly prevents double-stashing the child. ✔
- **Deletions clean:** `git grep` finds **zero** dangling refs to `inferenceOnly` / `InferenceOnlyGeneration` / `withoutStage` / `excludedStages` / `InferenceView::from`. `SpecPipelineTest` drops exactly the two `withoutStage` cases; `InferenceViewTest` rewired to the public ctor. Interface → clean marker (`interface NeedsInferenceDocument {}`, `@internal`). ✔

**Both deviations sound:**
- `InferenceRetention` hub (vs a field on `ComponentSchemaRegistry`): `@internal`, `#[Scoped]`, `Support\`, **no plugin import** — keeps the registry schema-focused and gives the gate flag one home across its three collaborators (registry / harvester / LintRunner). Nullable ctor deps preserve the ~80 `new ComponentSchemaRegistry()` unit sites; the container autowires them in a real run. Better than the planned placement.
- Pre-merge recording in `harvestContainers` (vs `mergeAuthoredOperation`): snapshots *every* route operation, not only merged ones — strictly more faithful to the old harvester-excluded document and required by the "same API surface" test whose flagged routes take the `applyReturnTypeSchema` path (no authored op annotation). Verified by the operation-migration test's `operationForRoute('get','op-redundant')`. ✔

**Non-blocking notes:**
1. **Narrow lint-time-only retention side effect (not the primary byte-identity concern, which is solid).** `retainRivalInferredSchema` runs the rival `$factory()` and guards only self-recursion (`markInProgress` on the *same* class). If a rival factory `buildOnce`s a nested class that is *not yet registered and reachable via no other path*, that nested component would be registered into the **lint-time** primary document (retention-on run) where the retention-off run's short-circuit would never have followed it. This never touches `openapi:generate` (gate off) and is benign for lint (an extra inference-only component has no authored counterpart → no spurious migration finding; the content-diff warning would fire). In practice contested top-level classes' nested graphs are already registered by the winning producer / response build, so it's likely unreachable — but the guard is asymmetric (self only, not nested), so worth a comment or a nested-registration guard if you want the lint-time document provably identical too.
2. **Style (minor):** a couple of new code lines over the 100-char soft limit — `ComponentSchemaRegistry.php:468` (125) and `:452` (119). Pint passes; the codebase otherwise wraps.

Conventions, boundary (`Support\` free of plugin deps), CHANGELOG (`Internal:` entry), and `docs/linting.md` (migration-rule descriptions reworded "second generation" → "retained inference-only view") all in order. Coverage: 2 missing lines, immaterial.

This is the highest-risk PR of the batch and it's carefully done. From reviewer-2: **approve.** (Human-gates at merge for the `Contracts\` marker change + double-gen deletions; needs reviewer-1's independent second approval + survey Layer A.)

Radiergummi:
🤖 **[reviewer]** review round 1 (reviewer-1, independent) — **1 required finding + 2 non-blocking observations.** This is a clean, well-structured retirement of the double generation; I verified every invariant the lead named and all but one hold outright. The one gap is a latent byte-identity hazard *at depth* against the explicitly-stated "winner in `$schemas` untouched" invariant, with no test covering the leak path.

**Verified holding (no action):**
- **Gate + user-facing byte-identity.** `LintRunner::rulesNeedInference` (any active rule `instanceof NeedsInferenceDocument`) → `retainInferredView` → `generateOne(..., retainInferredView:)` → `InferenceRetention::enable()` *inside* `withFreshScopedState` (after the reset). `GenerateCommand` calls `generateOne` without the flag, so **`openapi:generate` never enables retention → the user-facing document is always byte-identical.** Gate is correct.
- **Retention is a side channel.** `retainRivalInferredSchema` writes only to `InferenceRetention`, never to `$this->schemas`; the winner is untouched. The self-contained byte-identity unit test confirms `$schemas` is identical gate-on vs gate-off *for a factory with no nested refs*.
- **Nested-`$ref` parity (by-name).** The parity test's rival parent factory re-runs against an already-owned child and its `child` ref resolves to the child's key — proving the by-name equivalence the single-generation design relies on. Good.
- **authored-only-has-no-inferred-counterpart survives (binding must-fix #1).** Re-expressed `SwaggerPhpInferenceOnlyGenerationTest` keeps it in **both** forms: `schemaForName('Server')` null (name) and `schemaForClass(AttributeServer::class)` null (class), plus a scope-isolation case. `HarvestAuthoredAnnotationsStage::registerNamed` path calls `markAuthoredOnlySchema`, and `fromRetention` excludes those names. Not dropped/subsumed.
- **Six migration tests / identical verdicts (#2).** Four rewired off `retainedInferenceView()` (tests/Pest.php), which mirrors `LintRunner::fromRetention` exactly; `DocumentMigration` + `Replaceable` untouched. The "same API surface" test now compares the pre-merge retained operation ref against the harvested doc.
- **Deviation is sound.** The `InferenceRetention` hub (vs. bolting state onto `ComponentSchemaRegistry`) keeps the registry schema-focused and gives the gate flag + three retained maps one neutral `Support\` home spanning registry/harvester/LintRunner — good separation. Moving pre-merge operation recording to `harvestContainers` (all routes, incl. the `applyReturnTypeSchema` path that never hits `mergeAuthoredOperation`) is strictly more faithful to the old harvester-excluded document and is required by the schema-migration test — verified.
- **Deletions clean:** no dangling refs to `inferenceOnly` / `InferenceOnlyGeneration` / `withoutStage` / `excludedStages` (grep-clean); `NeedsInferenceDocument` → pure marker; `SpecPipeline` is a plain executor again. **Boundary:** `InferenceRetention` is `Support\` with no plugin import; `ComponentSchemaRegistry`'s new optional ctor deps autowire under `#[Scoped]`. **Bookkeeping:** CHANGELOG (#208, #483) + docs/linting.md (all three rule descriptions + the level-4 rationale) updated.

---

**1. [REQUIRED — guard or prove unreachable + test] The rival-factory invocation has no guard against registering *new nested components* into `$schemas`.** `retainRivalInferredSchema` calls `$factory()` (the losing producer's build closure) with only a self-recursion guard on the *contested class itself*. If that factory calls `buildOnce($nested)` for a nested class that is **not already registered**, the nested `buildOnce` will build and register it into `$this->schemas` (`ComponentSchemaRegistry.php` retain block, ~`:463-490`). That mutates the document `$schemas` in retain-mode only → violates the stated "gate-ON = winner in `$schemas` untouched" invariant.
   - **Not covered by tests:** the byte-identity test uses a flat factory; the nested-`$ref` parity test pre-builds the child as a winner *before* the rival runs — so neither exercises a rival whose nested ref is as-yet-unbuilt.
   - **Scope of impact (why not critical):** the gate keeps `openapi:generate` output untouched regardless, so **no user-facing spec can change**. The leak would only alter the *lint-internal* document (an extra inferred component appears in `document->components` → `fromRetention`'s `schemasByName` picks it up), which could skew a `migration.*` finding. And since `fromRetention` consumes only `inferredOperations()` + `isAuthoredOnlySchema()` — **not** `inferredSchemas()` — the whole rival-schema build currently buys only the content-diff warning, so an unguarded factory call is a poor trade.
   - **Ask:** either (a) snapshot `array_keys($this->schemas)` before `$factory()` and drop any keys the rival newly introduced (provably keeps `$schemas` untouched), **or** (b) confirm across the six `buildOnce` producers (`SchemaFromFormRequest`/`SchemaFromResource`/`SchemaFromTransformer`/`SchemaFromDataClass`/`JsonSchemaFromType`/`EloquentModelToSchema`) that a rival can never reference an unbuilt nested class, **and** add a unit test with a rival factory that `buildOnce`s an unbuilt nested class, asserting `count($registry->schemas)` / key-set is unchanged after the rival build.

**2. [non-blocking] Diagnostic may over-report when a schema transformer is configured.** The winner in `$schemas` went through `register()` → `OpenApiExtensions::applySchemaTransformers`; the rival does not, then `schemasDiffer` json-compares the two. With an active schema transformer the winner is transformed and the rival isn't, so a spurious "different schema" warning could fire even for identical inference. Diagnostic-only (no document impact); consider applying transformers to the rival before diffing, or noting the limitation. (The "stays silent when identical" test has no transformer, so it wouldn't catch this.)

**3. [non-blocking] `recordInferredOperation` shallow-`clone` coupling.** Correctly relies on `mergeAuthoredOperation` *reassigning* `$operation->responses` rather than mutating shared `OA\Response` objects in place; holds today and the comment notes it. Flagging the coupling so a future in-place merge edit doesn't silently corrupt the snapshot.

Finding 1 is the gate. CI is fully green (Pint, PHPStan L8, Snapshot, swagger-php 5.8, PHP 8.5 ×L12/L13); conventions and human-gate framing (Contracts marker change) are correct.

Radiergummi:
🤖 **[coder]** addressed round 1: pushed f54f3500.

**Finding 1 (REQUIRED) — fixed.** `retainRivalInferredSchema` now snapshots `$schemas`/`$classToKey`/`$keyToClass`/`$provenance` before the rival factory and restores them in `finally`, so any component the rival builds via a nested `buildOnce` (a nested class not already registered) is rolled back — the winner document is provably untouched, not just `$schemas` keys but the full reservation state (so a later legit build of that nested class isn't suppressed by a stray reservation). Added a unit test: a rival factory that `buildOnce`s an unbuilt `RetentionStray` leaves `componentClassMap()` unchanged, `keyFor(RetentionStray)` null, and `schemaForKey('RetentionStray')` null, while the rival's own view is still stashed.

**Non-blocking (2) — addressed.** The rival is now run through `OpenApiExtensions::applySchemaTransformers(...)` (mirroring `register()`) before `schemasDiffer`, so the diagnostic compares against the transformed winner rather than raw factory output — no spurious 'different schema' warning when a schema transformer is configured.

**Non-blocking (3) — acknowledged.** The `recordInferredOperation` shallow-clone correctness does depend on `mergeAuthoredOperation` reassigning `responses`/`parameters` (not editing in place); that holds today and the harvester is the only writer. Left as-is per your non-blocking call; happy to add a defensive deep-ish clone if you'd prefer it hardened.

Local gates green (Pint + PHPStan L8 + 3392 tests). Watching remote CI.

Radiergummi:
🤖 **[reviewer]** review round 2 (reviewer-1, independent): **approved.** Verified delta `8b482e3b..f54f3500`; the required finding is fully resolved and provably complete.

**Required finding 1 (rival factory registering new nested components) — resolved.** `retainRivalInferredSchema` now snapshots `$schemas` / `$classToKey` / `$keyToClass` / `$provenance` before `$factory()` and restores all four in `finally`, so any nested class the rival `buildOnce`s is rolled back — winner document provably untouched, and no stray reservation to suppress a later legit build.
- **Completeness check (I verified there's no un-snapshotted escape):** the only registry state a nested `buildOnce`/`register` mutates is exactly those four arrays plus `$inProgress` (self-balanced within the nested call). `$hasFileFields` is written solely by an external setter (`:547`), never on the build path, so it can't be touched inside a rival factory. The snapshot set is therefore complete.
- **Contested marker preserved:** the snapshot is taken *after* `recordContestedProducer($className)` runs at the call site, so X's own contested marker survives the restore; only the rival's *nested* provenance side effects revert. Correct.
- **Dangling-ref safety:** a nested class needed only by the losing inferred view is correctly discarded; the stashed rival schema may carry a now-dangling `$ref`, but it's inert (`fromRetention` never consumes `inferredSchemas()`; the stash feeds only the diagnostic). No document dangling ref.
- **Pinned by a fail-without-fix test:** *"rolls back a nested class a rival factory builds…"* — rival `buildOnce`s an unbuilt `RetentionStray`; asserts `componentClassMap()` unchanged, `keyFor`/`schemaForKey('RetentionStray')` null, and the rival view still stashed. Fails without the rollback (RetentionStray would leak into the class map).

**Non-blocking (2) — resolved.** The rival now runs through `applySchemaTransformers` (mirroring `register()`) before `schemasDiffer`, so the diagnostic compares transformed-vs-transformed — no spurious "different schema" warning under a configured transformer.

**Non-blocking (3)** — shallow-clone left per my explicit non-blocking call; harvester is the sole writer and the merge reassigns rather than mutates in place. No hardening needed.

PHP arrays being value-typed makes the snapshot/restore a genuine deep-enough copy for this path (nested builds add new entries / replace readonly provenance objects rather than mutating existing schema objects in place). CI green (swagger-php 5.8 + PHP 8.5). Everything from round 1 (gate, byte-identity, nested-$ref parity, authored-only assertion, six-test rewiring, deviations, deletions, boundary, bookkeeping) still holds.

That completes my review. This PR human-gates at merge (removes `NeedsInferenceDocument::excludedStages()`), but it's fully vetted from my side.

Radiergummi:
🤖 **[reviewer]** review round 2: **approved** (@ head `e1ff3cb3`).

Reviewed the delta `8b482e3b..e1ff3cb3` (the required-fix commit `f54f3500` + the cosmetic wrap `e1ff3cb3`).

**Required nested-`buildOnce` fix — correct and complete for the threat.** `retainRivalInferredSchema` now snapshots `$schemas`/`$classToKey`/`$keyToClass`/`$provenance` before the rival factory and restores all four in a `finally` (throw-safe). That is exactly the set `buildOnce` mutates, so a nested class the rival factory registers is fully rolled back — the winner document is provably untouched even at lint time, closing the note both reviewers raised. The leak test is genuinely fail-without-fix: the rival factory `buildOnce`s `RetentionStray` (never built during the winner build), and the test asserts `componentClassMap() === $before`, `keyFor(RetentionStray)` null, `schemaForKey('RetentionStray')` null, while the rival view is still stashed — without the restore, the first two assertions fail.

**Transformer-before-diff** (my round-1 note) fixed: the rival now runs through `applySchemaTransformers` before `schemasDiffer`, so no spurious content-diff warning under a configured schema transformer. `SchemaContext` import added.

**Restore-set completeness (verified, non-blocking observation):** `responses`/`parameters`/`compiledFields`/`hasFileFields` are *not* in the restore set, but they're unreachable from the threat — a rival schema factory reaches only `buildOnce` (the `schemas` path); `responses`/`parameters` are written only by `registerResponse()`/`registerParameter()`, which schema factories don't call, and `compiledFields`/`hasFileFields` are deterministic per-class memo caches (a lingering orphan entry for a rolled-back class is never serialized). So the fix is complete for what can actually leak; a one-line comment noting "the rival factory only reaches buildOnce, so only the schema maps need rollback" would document the reasoning, but it isn't required.

**`e1ff3cb3`** is cosmetic only — wraps the method signature and extracts a `$sameProducer` local; no logic change (addresses my round-1 line-width note).

Nothing from round 1 regressed — the delta touches only `retainRivalInferredSchema` + one fixture + one test. All my round-1 verified invariants (gate/byte-identity, `fromRetention` sourcing, migration rewiring with identical verdicts, binding must-fix #1, nested-`$ref` parity, clean deletions, both deviations) stand. CI shows 0 failures (matrix re-running on the new sha, expected after the push).

From reviewer-2: **approved.** This completes the two-approval gate together with reviewer-1's independent round-2 approval; #525 moves to survey Layer A + the human gate (`Contracts\` marker removal).

Radiergummi:
🤖 **[lead]** TWO-approval gate COMPLETE (reviewer-1 round 2 @f54f3500 logic + reviewer-2 round 2 @e1ff3cb3 full-delta). Final head e1ff3cb3 = f54f3500 + a whitespace-only line-wrap (non-semantic; reviewer-1's approval holds, reviewer-2 verified the delta). Required nested-buildOnce guard resolved (snapshot+restore of the 4 schema maps in throw-safe finally; fail-without-fix leak test; restore-set completeness verified — non-restored maps unreachable from a schema-building rival). → survey Layer A, QUEUED behind #527/#524 on the single surveyor (reuses main@576cefa2 baseline). CI re-running on e1ff3cb3 — must be green before the human gate. Post-survey → HUMAN GATE (removes Contracts\Lint\NeedsInferenceDocument::excludedStages()).

Radiergummi:
🤖 **[coder]** addressed round 1 (both reviewers): pushed e1ff3cb3 (on f54f3500).

**Convergent REQUIRED — nested-buildOnce leak — fixed (f54f3500).** `retainRivalInferredSchema` snapshots `$schemas`/`$classToKey`/`$keyToClass`/`$provenance` before the rival factory and restores them in `finally`, so any component the rival builds via a nested `buildOnce` is fully rolled back — the winner document is provably untouched even at lint time (not just `$schemas` keys but the full reservation state, so a later legit build of that nested class isn't suppressed). Leak test added: a rival factory that `buildOnce`s an unbuilt `RetentionStray` leaves `componentClassMap()` unchanged and the class unregistered, while the rival's own view is still stashed.

**Style — fixed (e1ff3cb3).** The two lines reviewer-2 cited (`ComponentSchemaRegistry` retainRivalInferredSchema signature @119 and the same-producer guard @125) are wrapped under 100 (guard extracted to a `$sameProducer` local).

**Non-blocking (2) — fixed (f54f3500).** Rival runs through `applySchemaTransformers` before `schemasDiffer`, so the diagnostic compares transformed-vs-transformed.

**Non-blocking (3) — rationale.** `recordInferredOperation`'s shallow clone is sound because `mergeAuthoredOperation` reassigns `responses`/`parameters` (never edits in place) and the harvester is the sole writer of those operations; a shallow clone taken pre-merge therefore preserves the inferred arrays. Left as-is per the non-blocking call.

Local gates green (Pint + PHPStan L8 + 3392 tests); remote CI all green incl. swagger-php 5.8 + both PHP 8.5 legs.

Radiergummi:
🤖 **[surveyor]** Layer A pass — clean no-op, 0 changes across 11 apps.

Baseline `main`@`576cefa2` (cached) vs PR head `e1ff3cb3`, full pinned corpus, PHP 8.4. All 11 apps `gen_exit=0` on both. A **deep full-metrics equality check** (every key — paths, operations, apiOperations, responseSchemas, documentedResponses, requestBodies, componentSchemas, maxRequestProperties, completenessPercent, lintFindings, coverage, routesIntrospected, and computed query/path/header/cookie param counts) reports **0 differing leaf values** in any app.

This is exactly the intended result: retiring the `NeedsInferenceDocument` double generation leaves ordinary `openapi:generate` output byte-identical, and the retention side channel stays gated behind the linter (never triggered by `corpus.sh`'s plain generate). No metric moved up or down; no new crash. Clear to merge on the Layer A gate.

Radiergummi:
🤖 **[lead]** All gates green — escalating for the human merge (removes a `Contracts\Lint\` signature: `NeedsInferenceDocument::excludedStages()` → marker interface; controversial-by-policy → human gate).

**Vetting complete:**
- **CI:** full matrix green on final commit `e1ff3cb3` — PHP 8.4/8.5 × L12/L13, swagger-php 5.8, byte-exact Snapshot, quality, coverage. mergeStateStatus CLEAN.
- **Reviews:** two independent approvals (reviewer-1 + reviewer-2). The convergent **required finding** — a nested-`buildOnce` leak that could touch the winner `$schemas` at lint time — is fully resolved: snapshot + throw-safe `finally` restore of `$schemas`/`$classToKey`/`$keyToClass`/`$provenance` (restore-set completeness verified), pinned by a fail-without-fix leak test. Both deviations verified sound (the InferenceRetention hub; harvestContainers pre-merge recording).
- **Survey Layer A:** a clean **byte-identical no-op** — 0 differing leaf values across all 11 apps on a deep full-metrics check. Confirms the PR's core thesis: retention is a gated side channel behind the linter; ordinary `openapi:generate` output is unchanged.

**Public-contract touch (the reason for the gate):** `Contracts\Lint\NeedsInferenceDocument::excludedStages()` removed (interface → `@internal` marker). Also deletes `inferenceOnly()` / `InferenceOnlyGeneration` / `SpecPipeline::withoutStage` / the LintRunner second-generation branch (all `@internal`). No config-key change.

One explicitly-optional non-blocking note left un-pushed by the coder (a one-line comment documenting the restore-set exclusions) to preserve the approved SHA — foldable into any future round. Ready to squash-merge. Worktree removed; branch intact.

Issue #517, opened 2026-07-02
feat(responses): read a resource wrapped in response()->json(<resource>, status)


Radiergummi:
## Problem

An action that returns a resource wrapped in an inline-JSON call —
`return response()->json(UserResource::make($x), 201);` (and the `::collection(...)` variant) —
does not get an API Resource response schema. `ReturnExpressionResourceReader::targetFromExpression`
unwraps resource-preserving *method chains* (`->additional(...)`) but does not look **inside** a
`response()->json(<data>, <status>)` call, so the resource passed as the JSON data argument is
invisible to the resource reader.

This was **GAP 2 of #487**, descoped during plan-review (see the discussion on PR #515). #487 ships
GAP 1 (routing loose `JsonResponse`-typed returns to the body scan), which is the higher-value,
non-blocked win.

## Why this is likely a no-op today (the load-bearing caveat)

Core registers **before** config plugins (`plugins: [CorePlugin::class, ...config('openapi.plugins')]`),
so `InlineJsonResponseResolver` (Core) runs **ahead of** `ResourceResponseResolver` (ApiResources)
in the first-non-null primary-response chain. For `response()->json(<resource>, 201)`, Core's
inline-JSON resolver will most likely **already emit a schema** (the resource rendered as an inline
array/object shape) and win before ApiResources is ever consulted. So teaching the resource reader
this shape, on its own, changes nothing observable.

**This issue only becomes live if a Core resolver changes** so the inline-JSON path **defers when its
`json()` data argument is a resource construction**, letting the richer resource `$ref` win over the
inline shape. That is a change to a Core resolver's emit decision and must be designed together with
the resource-reader change; measure whether the resource `$ref` is actually preferable to the inline
shape for these actions before committing.

## Proposed change (when taken up)

1. Extract a minimal `ResponseJsonCall` matcher to `Support\MethodBody\` — detect a
   `response()->json(...)` call (reuse the merged `UnqualifiedHelperCall` for the helper-shadowing
   guard) and expose its data-argument expression. Have `Plugins\Core\Support\InlineJsonCallReader`
   delegate its detection to it (no behavior change; green-guard Core's inline-JSON tests). This
   keeps the shape matcher in Support so **ApiResources does not depend on Core**
   (`tests/Arch/CoreBoundaryTest`: Support must not depend on a plugin; no plugin→plugin dep).
2. In `ReturnExpressionResourceReader::targetFromExpression`, when the expression is a
[...]

PR #467, opened 2026-06-26
feat(responses): read the OO new JsonResponse([...], status) form

Radiergummi:
## Plan — read the OO `new JsonResponse([...], status)` form (parity with `response()->json([...])`)

Closes #453.

### Tier & rationale

**Tier 1**, the *same* tier and the *same* bounded scan that already reads
`response()->json([...], $status)`. The OO construction `new JsonResponse(<data>, <status?>)` is
the identical AST argument shape (positional arg 0 = body literal, arg 1 = status), differing only
in the node kind: a `New_` instead of a `MethodCall`. No body parsing beyond the existing
first-10-statements whitelist, no variable tracking. This deliberately *reverses* the
"out of scope by design" exclusion documented at `InlineJsonResponseResolver.php:41`, backed by the
corpus evidence in the issue (Pterodactyl/Pelican use the OO form pervasively: 40 genuinely-missing
Pelican ops). No new tier ground is broken.

### Approach

`InlineJsonCallReader` already does all the work — its body reader (`readBody`) and status reader
(`resolveStatus` + `statusFromChain`) are **position-based** (`DATA_ARGUMENT_POSITION = 0`,
`STATUS_ARGUMENT_POSITION = 1`), and `JsonResponse::__construct($data = null, int $status = 200, …)`
matches those positions exactly, including the named-argument keys `data` / `status`. The chain
walker (`chainParentOf`) finds a parent via `$node->var === $current`, which is satisfied for a
`(new JsonResponse([...]))->setStatusCode(201)` chain too. So the OO form needs **only a new
class-shape matcher plus a widened parameter type** — the reading logic is reused verbatim.

**Class-name matching (precise).** `MethodBodyScanner` already runs php-parser's `NameResolver`
(with `preserveOriginalNames`) over the whole file, so a `New_` node's `->class` is a
**fully-qualified** `Name` regardless of whether the source wrote the imported `JsonResponse` or the
FQ `\Illuminate\Http\JsonResponse`. The matcher therefore resolves both identically:

```php
$node instanceof New_
    && $node->class instanceof Name                       // skip `new $class(...)` / anon classes
    && is_a($node->class->toString(), \Illuminate\Http\JsonResponse::class, true)
```

`is_a(..., true)` also accepts app subclasses of Illuminate's `JsonResponse`, and precisely rejects
an unrelated class that merely happens to be named `JsonResponse`. (See *Decision* below on the
Symfony base class.)

### Files to modify

[...]

PR #531, opened 2026-07-04
refactor(core): extract shared CallArgumentResolver to Support\MethodBody (arg-resolver dedup; matcher deferred to #517)

### LINKED ISSUE #512: feat(core): call-shape matcher registration seam for Tier-1 resolvers
Descoped from #486 during plan-review (PR #511): a small call-shape matcher utility / registration seam so a new whitelisted Tier-1 shape is closer to data (name map + arg spec) than plumbing.

Rationale for descoping: no second consumer exists yet — building the seam inside #486 would be speculative abstraction. Revisit when the next new Tier-1 shape lands and the duplication becomes concrete.

Context: #486 item 4, epic #478 (Tier-1 substrate block).

Radiergummi:
## What this does

**Refs #512.** Extracts the copy-pasted "resolve a call argument by name, else by position"
lookup into one pure, plugin-agnostic `Support\MethodBody\CallArgumentResolver` and refactors the
four on-`main` body-scan readers onto it, **byte-identically**. No spec output change; internal
refactor of the existing Tier-1 substrate (not an inference-tier change).

`#512` stays **open** as the call-shape *matcher* tracker (its titled intent). The JSON matcher
half is **deferred to #517** (its plan item 1 owns the `Support\MethodBody\` matcher extraction and
closes #512 when it lands). Hence `Refs`, not `Closes`.

## Scope (final, arg-resolver only)

**Create (`@internal`, plugin-agnostic):**
- `src/Support/MethodBody/CallArgumentResolver.php` — one static method
  `argument(array $arguments, string $name, int $position): ?Arg`, named-first then
  positional-if-unnamed. **Pure**: no spread/unpack filtering, no literal folding — callers keep
  their own such logic.
- `tests/Unit/Support/MethodBody/CallArgumentResolverTest.php` — named hit at any position,
  positional hit, differently-named-at-position → null, miss → null, negative position never
  matches a slot (the Abort param-not-in-spec case).

**Refactor onto it (byte-identical):**
- `src/Plugins/Core/Support/InlineJsonCallReader.php` — exact drop-in; private `argument()` deleted.
- `src/Plugins/Core/Support/RequestAccessorReader.php` — arg-order swap (`(position, name)` →
  `(name, position)`); private `argument()` deleted.
- `src/Plugins/QueryBuilder/Support/QueryBuilderChainReader.php` — same swap. (`spatie/laravel-query-builder`
  is `require-dev`, so this runs in the main `tests` job — in scope.)
- `src/Plugins/Core/ErrorContributors/AbortErrorContributor.php` — keeps its `HELPER_PARAMETERS`
  name→position `array_search` mapping (and the `false` guard) in the caller; delegates only the
  `(name, position)` lookup. The two-pass → single-pass collapse is equivalent for any compilable
  call (a positional arg *and* a later named arg for the same parameter is a duplicate-argument
  fatal, impossible in running code).

**Deferred (NOT in this PR):** the `ResponseJsonCallMatcher` extraction — single-consumer on `main`;
owned by #517 item 1.

## Guards

- Snapshot group (`tests/Feature/ExamplesTest.php`) + existing reader/Abort tests = the
  byte-identical proof.
- `CoreBoundaryTest` stays green: the new class lives in `Support\` and depends only on
  php-parser's `Arg` (no Core/plugin deps).
- New focused unit test for `CallArgumentResolver`.

CHANGELOG: one `[Unreleased]` internal line.

---

https://claude.ai/code/session_01YSxvAhwArZ9oxt46N5o4Js

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,6 +75,7 @@ All notable changes to this project are documented here.
 - Lint `Rule` contract now declares `id`, `severity`, and `description` as readable property requirements instead of getter methods; rule implementations declare them as plain typed properties with initializers. (#379)
 - Internal: unified the two docblock/type → schema engines on a single `symfony/type-info` `Type` representation (`TypeNodeToSchema` is removed; `JsonSchemaFromType` grew the structural array-shape/list/map cases and an optional leaf-class callback), with a `TypeNodeResolver::toType()` phpstan → symfony boundary. No spec output change except previously-unmapped array/collection shapes on the symfony path now resolve, `true`/`false` map to `boolean`, and array-shape properties serialize in sorted key order. (#480)
 - Internal: the `migration.*` lint rules now compare authored annotations against a retained inference-only view of the single generation instead of running a second, harvester-excluded generation; the `SpecPipeline` is a plain executor again (no per-run stage exclusion), and an ordinary `openapi:generate` pays nothing for the retention. (#208, #483)
+- Internal: the body-scan readers now resolve a call argument by name or position through a single shared `Support\MethodBody\CallArgumentResolver`, replacing four copies of the same logic. No spec output change. (#512)
 
 ### Fixed
 - Lint now surfaces top-level and field-level bare `oneOf` / `anyOf` schemas across components, responses, and request bodies. (#294, #318)
--- a/src/Plugins/Core/ErrorContributors/AbortErrorContributor.php
+++ b/src/Plugins/Core/ErrorContributors/AbortErrorContributor.php
@@ -16,6 +16,7 @@ use Radiergummi\OpenApi\Contracts\Registry\ErrorResponseContributor;
 use Radiergummi\OpenApi\Errors\ErrorDescriptor;
 use Radiergummi\OpenApi\Routing\ActionDescriptor;
 use Radiergummi\OpenApi\Support\MethodBody\AstLiteralEvaluator;
+use Radiergummi\OpenApi\Support\MethodBody\CallArgumentResolver;
 use Radiergummi\OpenApi\Support\MethodBody\ConditionalContextPolicy;
 use Radiergummi\OpenApi\Support\MethodBody\MethodBodyScanner;
 use Radiergummi\OpenApi\Support\MethodBody\NonLiteralValueException;
@@ -193,16 +194,14 @@ final readonly class AbortErrorContributor implements ErrorResponseContributor
      */
     private function resolveArgument(array $arguments, array $parameters, string $parameterName): ?Arg
     {
-        foreach ($arguments as $argument) {
-            if ($argument->name !== null && $argument->name->toString() === $parameterName) {
-                return $argument;
-            }
-        }
+        $position = array_search($parameterName, $parameters, true);
 
-        $index = array_search($parameterName, $parameters, true);
-        $positional = $index === false ? null : ($arguments[$index] ?? null);
-
-        return $positional !== null && $positional->name === null ? $positional : null;
+        // A parameter the helper does not declare has no positional slot; only a named argument binds it.
+        return CallArgumentResolver::argument(
+            $arguments,
+            $parameterName,
+            $position === false ? -1 : $position,
+        );
     }
 
     private function literalValueOf(Expr $expression): mixed
--- a/src/Plugins/Core/Support/InlineJsonCallReader.php
+++ b/src/Plugins/Core/Support/InlineJsonCallReader.php
@@ -17,13 +17,13 @@ use PhpParser\Node\Name;
 use PhpParser\Node\Stmt;
 use Radiergummi\OpenApi\Support\Generator\SchemaFromArrayDefinition;
 use Radiergummi\OpenApi\Support\MethodBody\AstLiteralEvaluator;
+use Radiergummi\OpenApi\Support\MethodBody\CallArgumentResolver;
 use Radiergummi\OpenApi\Support\MethodBody\ConditionalContextPolicy;
 use Radiergummi\OpenApi\Support\MethodBody\NonLiteralValueException;
 use Radiergummi\OpenApi\Support\MethodBody\SchemaDefinitionFromLiteral;
 use Radiergummi\OpenApi\Support\MethodBody\StatementNodeFinder;
 use Radiergummi\OpenApi\Support\MethodBody\UnqualifiedHelperCall;
 
-use function array_find;
 use function in_array;
 use function is_int;
 use function sprintf;
@@ -147,7 +147,7 @@ final readonly class InlineJsonCallReader
             return $chainStatus;
         }
 
-        $statusArgument = $this->argument($arguments, 'status', self::STATUS_ARGUMENT_POSITION);
+        $statusArgument = CallArgumentResolver::argument($arguments, 'status', self::STATUS_ARGUMENT_POSITION);
 
         if ($statusArgument === null) {
             return 200;
@@ -254,24 +254,12 @@ final readonly class InlineJsonCallReader
 
     /**
      * @param array<int, Arg> $arguments
-     */
-    private function argument(array $arguments, string $name, int $position): ?Arg
-    {
-        return array_find(
-            $arguments,
-            fn(Arg $argument, int $index): bool
-                => $argument->name === null ? $index === $position : $argument->name->toString() === $name,
-        );
-    }
-
-    /**
-     * @param array<int, Arg> $arguments
      *
      * @return array{0: ?Schema, 1: bool, 2: ?string}
      */
     private function readBody(array $arguments): array
     {
-        $dataArgument = $this->argument($arguments, 'data', self::DATA_ARGUMENT_POSITION);
+        $dataArgument = CallArgumentResolver::argument($arguments, 'data', self::DATA_ARGUMENT_POSITION);
 
         if ($dataArgument === null) {
             return [null, true, null];
--- a/src/Plugins/Core/Support/RequestAccessorReader.php
+++ b/src/Plugins/Core/Support/RequestAccessorReader.php
@@ -7,7 +7,6 @@ namespace Radiergummi\OpenApi\Plugins\Core\Support;
 use Illuminate\Container\Attributes\Scoped;
 use Illuminate\Http\Request;
 use PhpParser\Node;
-use PhpParser\Node\Arg;
 use PhpParser\Node\Expr\ArrowFunction;
 use PhpParser\Node\Expr\Closure as ClosureExpression;
 use PhpParser\Node\Expr\FuncCall;
@@ -16,6 +15,7 @@ use PhpParser\Node\Expr\Variable;
 use PhpParser\Node\Identifier;
 use PhpParser\Node\Name;
 use Radiergummi\OpenApi\Support\MethodBody\AstLiteralEvaluator;
+use Radiergummi\OpenApi\Support\MethodBody\CallArgumentResolver;
 use Radiergummi\OpenApi\Support\MethodBody\MethodBodyScanner;
 use Radiergummi\OpenApi\Support\MethodBody\NonLiteralValueException;
 use ReflectionMethod;
@@ -112,7 +112,7 @@ final readonly class RequestAccessorReader
 
         foreach ($calls as $call) {
             $accessor = $call->name instanceof Identifier ? $call->name->toLowerString() : '';
-            $keyArgument = $this->argument($call->getArgs(), 0, 'key');
+            $keyArgument = CallArgumentResolver::argument($call->getArgs(), 'key', 0);
 
             if ($keyArgument === null) {
                 // Zero-argument query()/cookie()/header() reads the whole bag, not a named parameter.
@@ -253,24 +253,6 @@ final readonly class RequestAccessorReader
     // region Argument resolution
 
     /**
-     * @param array<int, Arg> $arguments
-     */
-    private function argument(array $arguments, int $position, string $name): ?Arg
-    {
-        foreach ($arguments as $index => $argument) {
-            $matches = $argument->name === null
-                ? $index === $position
-                : $argument->name->toString() === $name;
-
-            if ($matches) {
-                return $argument;
-            }
-        }
-
-        return null;
-    }
-
-    /**
      * Converts a dotted key to wire notation (`filter.name` → `filter[name]`). Returns null for
      * wildcard or empty segments.
      */
@@ -309,7 +291,7 @@ final readonly class RequestAccessorReader
      */
     private function defaultValueOf(MethodCall $call, string $type): mixed
     {
-        $defaultArgument = $this->argument($call->getArgs(), 1, 'default');
+        $defaultArgument = CallArgumentResolver::argument($call->getArgs(), 'default', 1);
 
         if ($defaultArgument === null) {
             return null;
--- a/src/Plugins/QueryBuilder/Support/QueryBuilderChainReader.php
+++ b/src/Plugins/QueryBuilder/Support/QueryBuilderChainReader.php
@@ -6,7 +6,6 @@ namespace Radiergummi\OpenApi\Plugins\QueryBuilder\Support;
 
 use Illuminate\Container\Attributes\Scoped;
 use PhpParser\Node;
-use PhpParser\Node\Arg;
 use PhpParser\Node\Expr;
 use PhpParser\Node\Expr\Array_;
 use PhpParser\Node\Expr\MethodCall;
@@ -14,6 +13,7 @@ use PhpParser\Node\Expr\StaticCall;
 use PhpParser\Node\Identifier;
 use PhpParser\Node\Name;
 use Radiergummi\OpenApi\Support\MethodBody\AstLiteralEvaluator;
+use Radiergummi\OpenApi\Support\MethodBody\CallArgumentResolver;
 use Radiergummi\OpenApi\Support\MethodBody\ConditionalContextPolicy;
 use Radiergummi\OpenApi\Support\MethodBody\MethodBodyScanner;
 use Radiergummi\OpenApi\Support\MethodBody\NonLiteralValueException;
@@ -287,7 +287,7 @@ final class QueryBuilderChainReader
             return null;
         }
 
-        $nameArgument = $this->argument($call->getArgs(), 0, 'name');
+        $nameArgument = CallArgumentResolver::argument($call->getArgs(), 'name', 0);
 
         if ($nameArgument === null) {
             return $call->name->toLowerString() === 'trashed' ? 'trashed' : null;
@@ -300,24 +300,6 @@ final class QueryBuilderChainReader
         }
 
         return is_string($value) ? $this->normaliseName($value, $kind) : null;
-    }
-
-    /**
-     * @param array<int, Arg> $arguments
-     */
-    private function argument(array $arguments, int $position, string $name): ?Arg
-    {
-        foreach ($arguments as $index => $argument) {
-            $matches = $argument->name === null
-                ? $index === $position
-                : $argument->name->toString() === $name;
-
-            if ($matches) {
-                return $argument;
-            }
-        }
-
-        return null;
     }
 
     /** Strips the leading `-` direction marker from sort names. Returns null for empty strings. */
--- a/src/Support/MethodBody/CallArgumentResolver.php
+++ b/src/Support/MethodBody/CallArgumentResolver.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Radiergummi\OpenApi\Support\MethodBody;
+
+use PhpParser\Node\Arg;
+
+use function array_find;
+
+/**
+ * Resolves a call argument the way PHP binds it: by name if the caller named it, otherwise by
+ * positional index. A named argument matches at any position; an unnamed argument matches only at
+ * the given slot.
+ *
+ * Pure: it does not filter spread arguments or fold literals. Callers keep their own such logic.
+ *
+ * @internal
+ */
+final readonly class CallArgumentResolver
+{
+    /**
+     * @param array<int, Arg> $arguments
+     */
+    public static function argument(array $arguments, string $name, int $position): ?Arg
+    {
+        return array_find(
+            $arguments,
+            static fn(Arg $argument, int $index): bool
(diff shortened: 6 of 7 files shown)


Radiergummi:

Narrow build approved by Moritz on #512: extract a `JsonResponseShapeMatcher` + `CallArgumentResolver` into `Support\MethodBody\` (`@internal`), refactor `InlineJsonCallReader` + `InlineJsonResponseResolver` onto them **byte-identically**. Cross-family plugin registry explicitly out of scope.

**Dependency plan-review must verify:** builds on **#467** (which creates the duplication removed here). Working assumption: #467 merges first, #512 rebases on top — do not start coding until #467 lands on `main`.

Proof consumer = the byte-identical refactor of both JSON readers (honest without re-implementing #453, which #467 owns). Auto-merge-eligible: no `Contracts`/config/lint-rule change.

Radiergummi:
🤖 **[lead]** → plan-review. NOTE: plan is sound but #531 is **blocked on #467** (it dedupes code #467 introduces). Reviewer: assess plan quality + the #467-first sequencing; on approval it parks as plan-approved and does NOT code until Moritz merges #467 (he's away). No stacking on #467.

Radiergummi:
🤖 **[planner]** plan REVISED — decoupled from #467; ready for plan-review.

Lead re-scope (Moritz away, #467 stays human-gated/unmerged): #512 must be **independently mergeable on current `main`**, must NOT touch `InlineJsonResponseResolver` (that file is #467), and must not stack on #467.

Measured the on-main duplication:
- **Solid:** the name-or-position literal-arg resolver is copy-pasted **4×** on main — `InlineJsonCallReader`, `RequestAccessorReader`, `AbortErrorContributor`, `QueryBuilderChainReader`. Extract `Support\MethodBody\CallArgumentResolver` and refactor these onto it byte-identically. This carries #512 on its own.
- **Thin (flagged):** the JSON call-shape *detection* is already centralized in `InlineJsonCallReader` (the others delegate to it), so extracting a `ResponseJsonCallMatcher` to Support is a **single-consumer relocation** on main — its payoff is enabling the unmerged #517/#467. **Recommend deferring that half** to land with #467/#517. Kept in the plan as an option per the lead; trim at review.

No `Contracts`/config/lint-rule change; byte-identical (snapshot group is the guard). Nothing auto-merges while Moritz is away — this will stop at ready-for-merge.

Radiergummi:
🤖 **[reviewer]** plan-review: **approve with three required amendments** (plan-clarity, not rejection). The build is sound, surgical, correctly tiered (internal refactor, no new inference), and the substrate is genuinely justified *post-#467* — not speculative. But collapsing #467's two finders into one `findCanonical` introduces two precedence subtleties the byte-identical guard will **not** catch (they aren't in the snapshot corpus), and the matcher's API is under-specified where nocontent detection shares the `response()` receiver guard. Nail these in the plan before coding.

### Verdict on the framing (all hold)

- **#467-first sequencing is correct.** Pre-#467 there is only `findJsonCall` (one finder, no `MethodCall|New_` union). The duplication `findCanonical` removes — two near-identical returned-preferred two-pass finders (`findJsonCall` + `findJsonResponseConstruction`, resolver L303–350 on the #467 branch) — **only exists after #467**. Re-cutting #531 to stand alone would leave the matcher consolidating almost nothing, i.e. exactly the speculative-abstraction case #512 was descoped to avoid. So #467-first is right, and parking blocked on #467 is right.
- **Substrate genuinely justified.** Post-#467 the consolidation is concrete: `findCanonical` (2 finders → 1), `matchesHelperCall`/`matchesConstruction` (3 predicates threaded through reader + resolver), `CallArgumentResolver` (reader `argument()` + the two inline `array_find`s in the resolver's `hasExplicitStatus`/`noContentStatusArgument` = 3 sites *inside the two touched files*). #517 item 1 names the second cross-plugin consumer (ApiResources reusing the matcher from Support). This clears the #512 descope bar ("revisit when the duplication becomes concrete").
- **Scope is surgical**; keeping `RequestAccessorReader` + `AbortErrorContributor` out is the right call (see amendment 3 note).

### Required amendments

**1. `findCanonical` changes precedence in the mixed-shape case — pin it with a test and state the shift.** #467 tries `findJsonCall` (full returned-then-assigned two-pass over *helper* calls) and only on a total miss tries `findJsonResponseConstruction`. That is **helper-first globally**. A single `findCanonical` that does one returned-preferred two-pass matching *either* shape is **returned-first globally**. These differ for:
```php
return new JsonResponse([...]);   // returned construction
$x = response()->json([...]);     // assigned helper
```
#467 returns the *assigned helper* (helper-first); collapsed `findCanonical` returns the *returned construction*. The snapshot corpus won't contain this pathological mix, and no existing reader test covers it, so the "byte-identical" guard is blind to it. Add a `findCanonical` unit test pinning the chosen behavior and state explicitly in the plan that this is a (benign, pathological-only) precedence shift from #467's helper-first ordering. If strict byte-identity is required instead, `findCanonical` must preserve helper-first-then-construction (two phases), which largely defeats the collapse — so I'd take the returned-first behavior and document it.

**2. `findConditional` — the plan's own test contradicts a pure extraction.** The inline probe (#467 resolver L111–115) uses `IncludeConditionalContexts` and returns the *first* match anywhere, **including top-level**; it reads as "conditional-only" solely because it runs *after* `findCanonical` has consumed every unconditional match. Your test spec says `findConditional` returns "null when the only match is top-level" — a byte-identical extraction would return that top-level node, not null. Reconcile: either (a) keep `findConditional` a pure `IncludeConditionalContexts` finder and fix the test to expect the top-level node (correctness still rests on the resolver calling it after `findCanonical`), or (b) make it genuinely conditional-only, which is a *new* behavior beyond extraction and must be justified as such. Pick one deliberately; don't ship the contradiction.

**3. Specify where the general `response()` receiver / factory-method matcher lives.** `noContent()` detection uses `isFactoryMethodCall($node, 'nocontent')` (resolver L166), which shares `isResponseHelperCall` (the zero-arg `response()` receiver + shadow guard) with the json helper matcher. Since `matchesHelperCall` moves to Support and **Support must not depend on Core** (`CoreBoundaryTest`), the shared receiver / factory-method check must move to the matcher too, with `InlineJsonCallReader` delegating its remaining `isFactoryMethodCall('nocontent')` through it — otherwise you get either a boundary violation or a re-duplicated receiver guard. The plan lists only json/construction-specific matcher methods; add the general receiver (or factory-method) check to the matcher's stated API.

### Nice-to-have (non-blocking)

4. File the follow-up issue for `RequestAccessorReader` + `AbortErrorContributor` adopting `CallArgumentResolver`, and note it's **not purely mechanical**: `RequestAccessorReader::argument` takes args in `(position, name)` order (vs the resolver's `(name, position)`) and uses a `foreach`, and `AbortErrorContributor::resolveArgument` layers an ordered param-name→position spec on top. So the "byte-identical dupe" claim holds only for `InlineJsonCallReader::argument`; the other two need adaptation.
5. `noContentStatusArgument` carries the *same* inline `array_find` as `hasExplicitStatus`; convert both or neither for consistency (plan currently names only `hasExplicitStatus`).
6. CHANGELOG: a terse `[Unreleased]` internal line is fine; optional.

### Gate
After approval this PR **parks (plan-approved), blocked on #467**, and does **not** proceed to coding until Moritz merges #467 (currently `agent:needs-human` for reversing the deliberate `new JsonResponse` exclusion). No stacking on #467.

Radiergummi:
🤖 **[lead]** Plan APPROVED (reviewer + lead), but **parked — blocked on #467**; do not code until Moritz merges #467. Decision trail for @Radiergummi:

**Why parked, not coded:** an #467-*independent* version of #512 collapses to extracting one ~6-line name-or-position arg helper shared by just **InlineJsonCallReader + RequestAccessorReader** (QueryBuilderChainReader is a 3rd but in the optional plugin; AbortErrorContributor's param-spec shape isn't a clean target). That 2-consumer extraction is too marginal to justify on its own and would silently narrow #512 away from its titled intent (the call-shape *matcher* seam). The **matcher** — #512's actual point — only consolidates real duplication once #467 lands (findCanonical 2→1, matchesHelperCall/matchesConstruction), with #517 as the second cross-plugin consumer. So the substrate is worthwhile as a **bundle with the matcher, post-#467**, and speculative without it. Independently reached by both the planner's on-main measurement and the reviewer's plan-review.

**Reviewer's 3 required amendments** (apply when coding starts post-#467; byte-identical guard is blind to the first two):
1. findCanonical precedence: #467 is helper-first-globally; a single returned-preferred pass is returned-first-globally — differs on a returned-construction + assigned-helper mix. Pin behavior with a unit test or preserve two-phase helper-first.
2. findConditional: the inline probe returns top-level matches too; the planned 'null when only match is top-level' test contradicts a pure extraction — pick pure-finder (fix test) or conditional-only (justify as new behavior).
3. Matcher API: noContent detection shares the response() receiver guard; since Support can't depend on Core, that receiver/factory check must live in the matcher too.

**Resume path:** merge #467 → rebase #531 on main → coding with the 3 amendments. No stacking on #467 while unmerged.

Radiergummi:
🤖 **[reviewer]** plan-review (revised, #467-independent scope): **approve the re-scope with two required edits.** Ship #512 as the `CallArgumentResolver` dedup; the matcher defer is correct. My three earlier amendments (findCanonical / findConditional / nocontent-matcher API) are now **moot** — that code isn't touched (matcher deferred, `InlineJsonResponseResolver` excluded).

### (a) Byte-identical safety — the load-bearing risk. Verified GO for all four sites.

I read all four copies on `main`. A single `CallArgumentResolver::argument(list<Arg> $arguments, string $name, int $position): ?Arg` (named-first, else positional-if-unnamed) replaces every one **without behavior change**:

| Site | Adoption | Risk |
|---|---|---|
| `InlineJsonCallReader::argument` (`:258`) | exact drop-in — signature already `(name, position)` | none |
| `RequestAccessorReader::argument` (`:258`) | arg-order swap only (`(position, name)` → `(name, position)`); `foreach` vs `array_find`, identical result incl. null | none |
| `QueryBuilderChainReader::argument` (`:308`) | byte-identical to RequestAccessorReader's — same swap | none |
| `AbortErrorContributor::resolveArgument` (`:194`) | keep its `HELPER_PARAMETERS` name→position mapping in the caller (`array_search`), pass the resolved `int $position`; guard the `false` (param-not-in-spec) case | see note |

**Abort note (only real subtlety):** its resolver is two-pass (named-anywhere, then positional-at-spec-index with an unnamed guard); the shared resolver is single-pass (named-or-positional, left-to-right). These are **equivalent for any compilable call**. The one input where they'd diverge is a *positional arg at slot P* **and** a *named arg for the same parameter later* — that's a duplicate-argument fatal error, impossible in code that runs. So it's safe. Two constraints to bake into the plan so it stays byte-identical:
- `CallArgumentResolver` must stay **pure**: no `->unpack` filtering and no literal folding — every caller keeps its own unpack guard and `AstLiteralEvaluator` use (the plan says this for literal folding; add unpack explicitly).
- Abort keeps the `array_search($parameterName, HELPER_PARAMETERS[...], true)` → position mapping and its `false` guard in the caller (the resolver only takes the resolved `int`).

Add two `CallArgumentResolverTest` cases beyond the plan's list: a **named arg with a different name sitting at `position` → null** (proves no false positional match), and the **Abort spec-derived-position** case (already listed — keep it).

### (b) Defer the JSON matcher — correct, approve.

Confirmed on `main`: `InlineJsonErrorContributor` and `InlineJsonResponseResolver` both **delegate** detection to `InlineJsonCallReader` (`$this->callReader->isJsonHelperCall(...)`); the predicates are already centralized there. Extracting `ResponseJsonCallMatcher` to `Support\` is therefore a **single-consumer relocation** on `main` — its only payoff is enabling the unmerged #517 (ApiResources reuse without a Core dep) and #467 (the `new JsonResponse` variant). #517's own plan **already lists** "extract a minimal matcher to `Support\MethodBody\`" as its item 1, so deferring loses nothing and avoids exactly the speculative-abstraction the #512 descope warned against. **Drop the matcher half from #531.**

### (c) QueryBuilderChainReader — REQUIRED EDIT: pull it into scope, drop the "optional-dep CI" caveat.

The plan's "contingent on the QueryBuilder CI job / optional dep covering it" rests on a false premise. `spatie/laravel-query-builder ^7.3.0` is a plain **`require-dev`** dependency (composer.json:40), so it is installed in the standard test env and `QueryBuilderChainReader` is exercised by the **main `tests` job** (`tests/Feature/Plugins/QueryBuilder/QueryBuilderChainDetectionTest.php`, `QueryBuilderParameterResolverTest.php`). There is **no** separate optional-dep gate for it (the only special matrix job is swagger-php 5.x, unrelated). Its `argument()` is byte-identical to RequestAccessorReader's — the same trivial swap. Leaving it out would strand the new primitive with a live **4th** duplicate for no reason. Include it in #531. (Plugin→`Support` is an allowed dependency; `CoreBoundaryTest` only forbids `Support`→`Plugins`.)

### (d) `Closes #512`? — REQUIRED EDIT: no, use `Refs #512`.

#512's titled/headline intent is the *call-shape matcher / registration seam* ("a new whitelisted Tier-1 shape closer to data"), which this PR **defers**. #531 delivers only the arg-resolver ("name map") half. An unqualified `Closes #512` would mark the matcher deliverable done when it isn't. Recommend: change to **`Refs #512`**, keep #512 open as the matcher-seam tracker, and note the matcher extraction is now owned by **#517 item 1** (which closes #512 when it lands). Add a one-line follow-up note on #512 to that effect so the split is durable.

### Verdict

**Approve** with edits (c) + (d) applied and the (a) byte-identical constraints written into the plan. This is a clean, surgical, tier-neutral dedup with a real 4×-on-main justification and the snapshot group as the byte-identical proof. Nothing here is a human-gate change; it stops at ready-for-merge while Moritz is away.

Radiergummi:
🤖 **[lead]** REVERSING the earlier park — it was based on my factual misread. The reviewer's deep re-analysis (and my own verification) shows a genuine **#467-independent 4× dedup**: (1) `spatie/laravel-query-builder` is require-dev, so QueryBuilderChainReader runs in the main `tests` job — a live 4th duplicate, IN scope; (2) AbortErrorContributor::resolveArgument keeps its HELPER_PARAMETERS name→position (array_search) mapping in the caller and delegates the (name,position) lookup — cleanly extractable. Abort is NOT the un-extractable case I thought.

**Final scope → CODING (arg-resolver only; JSON matcher DEFERRED to #517):**
- New `src/Support/MethodBody/CallArgumentResolver.php` (@internal, plugin-agnostic): `argument(array $arguments, string $name, int $position): ?Arg` — name-then-position. **Pure**: no unpack filtering, no literal folding (callers keep theirs).
- Refactor all 4 byte-identically: InlineJsonCallReader (drop-in), RequestAccessorReader + QueryBuilderChainReader (arg-order swap only), AbortErrorContributor (keep array_search+false-guard+null-name guard in caller; delegate lookup).
- DROP the ResponseJsonCallMatcher half (single-consumer on main; #517 item 1 owns it).
- PR is **Refs #512, not Closes** — #512 stays open as the matcher tracker (closed by #517 when the matcher lands).
- Guards: snapshot group (ExamplesTest) + existing reader tests + CoreBoundaryTest; add a CallArgumentResolver unit test. `composer check` + rebase on main. Stop at ready — no merge (Moritz away).

Radiergummi:
🤖 **[coder]** implemented; local gates green; remote CI green (all matrix jobs incl. swagger-php 5.8, plus the byte-identical Snapshot guard); pushed d70aa0a1. Ready for review.

Scope built = arg-resolver dedup only (JSON matcher deferred to #517):
- New `src/Support/MethodBody/CallArgumentResolver.php` (`@internal`, pure, static `argument(array, string $name, int $position): ?Arg`).
- Refactored all 4 on-`main` consumers byte-identically: `InlineJsonCallReader` (drop-in), `RequestAccessorReader` + `QueryBuilderChainReader` (arg-order swap), `AbortErrorContributor` (keeps `HELPER_PARAMETERS` `array_search` name→position + false-guard in the caller; delegates the lookup; two-pass→single-pass is equivalent for any compilable call).
- `Refs #512` (matcher tracker stays open; #517 item 1 owns the matcher). CHANGELOG `[Unreleased]` internal line added.

Byte-identical proof: Snapshot group + existing reader/Abort tests all green. `CoreBoundaryTest` green (`Support\` class depends only on php-parser `Arg`).

Radiergummi:
🤖 **[lead]** coding done, full CI matrix green (incl. 8.5, swagger-php 5.8, Snapshot byte-identical guard) → code-review. Reviewer: verify the 4-site byte-identical extraction (esp. AbortErrorContributor's caller-kept array_search mapping + guards, two-pass→single-pass equivalence), CallArgumentResolver purity (no unpack/folding), Support boundary (CoreBoundaryTest), Refs-not-Closes. Drive to ready — no merge (Moritz away).

Radiergummi:
🤖 **[reviewer]** review round 1: **approved**, no blocking concerns. Clean, faithful byte-identical extraction; both my plan-review edits landed (QueryBuilder in scope, `Refs`-not-`Closes`). One optional test-strength nit below.

Verified against the diff (`d70aa0a1`):

- **(a) AbortErrorContributor equivalence holds in the actual code.** The two-pass (named-anywhere → positional-at-spec-index-if-unnamed) collapses to the single-pass `array_find` correctly. Walked the cases: param-not-in-spec (`array_search` → `false` → `-1`) disables positional matching so only a named arg binds — matches the old `$index === false ? null` branch; valid-index cases return the same arg. The one input where single-pass would differ (a positional arg at slot P *and* a named arg for the same param) is a duplicate-argument fatal error, impossible in compilable code. The `-1` sentinel is the right mechanism and the comment ("a parameter the helper does not declare has no positional slot") explains the *why*. ✓
- **(b) `CallArgumentResolver` is genuinely pure** — no `->unpack` filtering, no literal folding; docblock states it. Confirmed callers keep their own guards: `AbortErrorContributor::literalArgument` still checks `$argument->unpack`, and `InlineJsonCallReader::read` still does the top-level unpack sweep. No purity leak. ✓
- **(c) Arg-order swaps correct.** RequestAccessorReader `(args, 0, 'key')`→`(args, 'key', 0)` and `(args, 1, 'default')`→`(args, 'default', 1)`; QueryBuilderChainReader `(args, 0, 'name')`→`(args, 'name', 0)`; InlineJsonCallReader was already `(name, position)` so it's a pure drop-in. `foreach`→`array_find` is behavior-identical (first match by array key over a `getArgs()` list, else null). ✓
- **(d) Support boundary intact.** `CallArgumentResolver` imports only `PhpParser\Node\Arg` + `array_find` — no `Plugins\` dependency. Plugin→Support consumption (Core + QueryBuilder) is the allowed direction; `CoreBoundaryTest` green. ✓
- **(e) Unit test** covers named-hit, positional-hit, differently-named-at-position→null, no-match→null, and negative-position→null (the Abort `-1` sentinel). Good. ✓
- **Hygiene:** unused `use PhpParser\Node\Arg;` correctly dropped from both readers (branch files clean, no dangling `Arg` type refs). CHANGELOG has an accurate *Internal* entry ("No spec output change", #512). `Refs #512`, #512 stays **open** as the matcher tracker. ✓

**Non-blocking nit (optional):** the "resolves a named argument at any position" test puts the named `status` arg at index 1 with `position` = 1, so name-match and position-match coincide — it doesn't isolate "any position". Behavior is already locked by the byte-identical consumer tests + Snapshot, so this is cosmetic; if touched, place the named arg at a *different* index than the position (e.g. named `status` at index 0, call with position 1 → expect arg 0). Not required.

Verdict: **approve.** No spec-output change (Snapshot green), Support-only, `@internal`, no `Contracts`/config/lint/plugin surface. Ready for merge; stops here (Moritz merges on return).

Radiergummi:
🤖 **[reviewer2]** review round 1: approved, no concerns.

Independent second review (I did not author or plan this). CI matrix already green; verified the refactor is behavior-preserving on the merits.

- **AbortErrorContributor two-pass → single-pass is equivalent** for every *compilable* call. The only way the single-pass `array_find` could pick a positional arg where the old code picked a named one is an unnamed arg at `index === position` alongside a later named arg for the *same* parameter — a duplicate-argument compile fatal, impossible in running code. The `$position === false ? -1` guard preserves the original semantics: a parameter the helper doesn't declare has no positional slot, and `-1` never matches a positional index while a named arg still binds by name.
- **`CallArgumentResolver` is genuinely pure** — depends only on php-parser `Arg` + `array_find`; no spread/unpack filtering and no literal folding pulled in. Each caller retains its own such logic (Abort keeps `unpack` guard + `literalValueOf`; InlineJson keeps `readBody`, etc.).
- **Arg-order swaps correct** — RequestAccessorReader / QueryBuilderChainReader old signature was `(position, name)`; shared is `(name, position)`. `(0,'key')→('key',0)`, `(1,'default')→('default',1)`, `(0,'name')→('name',0)`. InlineJsonCallReader was already `(name, position)` (exact drop-in). `HELPER_PARAMETERS` is an ordered name list so `array_search` yields the true positional index — argument positions line up.
- **Support boundary intact** — new class lives in `Support\MethodBody` and depends only on php-parser's `Arg`; no Core/plugin deps (CoreBoundaryTest stays green). Removed `use PhpParser\Node\Arg;` / `use function array_find;` from the refactored readers are correctly orphaned by the deletions.
- **Unit test** covers named-hit-at-position, positional-hit, differently-named-at-position→null, miss→null, and the negative-position→null (Abort param-not-in-spec) case — exceeds the requested three.
- **`Refs #512`, not `Closes`** — correct; the JSON matcher half is deferred to #517.

Optional nit (non-blocking): the test suite doesn't assert that a *named* argument still binds when `position = -1` (named-hit under the abort-param-not-declared path). It's a trivial branch combination already covered indirectly, so no change required.

Approve.

Radiergummi:
🤖 **[lead]** ✅ **MERGE-READY — no decision needed, just squash-merge on return.** Two independent approvals (reviewer + reviewer2), full CI matrix green (PHP 8.4/8.5 × L12/13, swagger-php 5.8, byte-identical Snapshot, quality, Coverage), non-controversial (@internal `Support\MethodBody`, no Contracts/config/stage-order/public-signature change, ~110-line refactor). Refs #512 (matcher half deferred to #517). Held only because you're away and I committed to stop at ready rather than auto-merge in your absence — this DOES pass your hands-off gate, so if you'd rather I auto-merge fully-gated non-controversial PRs like this, say the word and I'll finish it.

PR #534, opened 2026-07-04
feat(responses): read a resource wrapped in response()->json(<resource>, status)

### LINKED ISSUE #517: feat(responses): read a resource wrapped in response()->json(<resource>, status)

Radiergummi:
Closes #517

## Implementation plan — corrected after empirical characterization

**Tier 1** (bounded AST, `area:responses`). The issue's stated premise — *"Core's inline-JSON
resolver will most likely already emit a schema (the resource rendered as an inline shape) and win
before ApiResources is consulted"* — is **empirically wrong**. A characterization probe of
`return response()->json(NestedAuthorResource::make(...), 201)` (and the `::collection(...)`
variant) on `main` produces:

| shape | status | schema | resource in `components` |
|-------|--------|--------|--------------------------|
| single | bare `200 OK` | **none** | **none** |
| collection | bare `200 OK` | **none** | **none** |

Because a resource construction (`UserResource::make($x)`) is a `StaticCall`, not an array/scalar
literal, Core's `InlineJsonCallReader::readBody` returns `bodyReadable = false` and
`InlineJsonResponseResolver` **yields** (returns null). The bare `200` is the typed-return
fallback; the authored `201` is already lost today. So:

- **Step 3 of the issue (make Core defer) is unnecessary** — Core already yields for resource-typed
  `json()` args. No Core emit-decision change, no controversial human gate on that fork.
- **Step 2 is a live improvement** — teaching the resource reader to recurse into
  `response()->json(<resource>, ...)` makes the resource `$ref` emit where today there is none.
  Status stays `200` (identical to today's fallback → **no regression**); capturing the authored
  `201` would require the resource path to read status from the body (Core's concern) — left as a
  documented follow-up, not scoped here.
- **Step 1 is still needed** — a shared `response()->json()` matcher so ApiResources can detect the
  shape without depending on Core (`tests/Arch/CoreBoundaryTest`).

### Changes

1. **`Support\MethodBody\ResponseJsonCall`** (new) — detect a `response()->json(<data>, <status>)`
   helper call and expose its data-argument `Arg`. Reuses `UnqualifiedHelperCall` for the
   helper-shadowing guard and `CallArgumentResolver` for the `data` argument. Scope: the
   `response()->json(...)` *helper* form only (not `new JsonResponse(...)`, which stays Core's).
2. **`Plugins\Core\Support\InlineJsonCallReader`** — delegate its `response()->json` helper
   detection to `ResponseJsonCall` (no behavior change; existing Core inline-JSON tests green-guard
   this).
3. **`Plugins\ApiResources\Support\ReturnExpressionResourceReader::targetFromExpression`** — when
   the current expression is a `response()->json(<data>, ...)` call, unwrap to `<data>` and continue
   the existing resource-shape resolution loop. Non-resource data degrades to null + logs (charter).

### Tests

- `ResponseJsonCall` unit test (Support/MethodBody style — build `Arg` nodes directly).
- Feature: `response()->json(Resource::make(...), 201)` → resource `$ref` (single);
  `::collection(...)` → collection shape; `response()->json($nonResource, 201)` → null + log.
- Pin no double-emit (resource path and inline-JSON path never both emit for one action).
- Survey Layer A: 0 regressions; quantify coverage change.

### Charter

Bounded-AST: unconditional statements only, no cross-call variable tracking, degrade + log on
unmatched shapes.

Refs #487, PR #515. Depends on #486 (merged).


Radiergummi:
🤖 **[lead]** Plan corrected after empirical probe (see PR body): Core already yields for resource-typed json() args — step 3 dropped, no Core emit-decision change. Implementing steps 1+2.

Radiergummi:
🤖 **[lead]** Implemented + pushed a2d71e85. Local gates green (Pint ✓, PHPStan 0 errors, 3479 tests ✓). Marked ready. Running independent review.

Radiergummi:
🤖 **[reviewer]** Independent adversarial review — 1 material finding (status fidelity), delegation no-op + boundary + conventions all verified clean.

**Verified clean:**
- **Delegation is a true no-op** — `ResponseJsonCall::matches`/`isResponseHelperReceiver` are the same predicates Core used (`isFactoryMethodCall('json',…)`/`isResponseHelperCall`): same first-class-callable/`Identifier`/`FuncCall`/zero-arg/`UnqualifiedHelperCall` guards, same data-arg position. Dropped `FuncCall`/`UnqualifiedHelperCall` imports correctly; `Name` still used.
- **Boundary** — `Support\MethodBody\ResponseJsonCall` imports only `PhpParser\*` + `Support\MethodBody\*`; `CoreBoundaryTest` holds.
- **Matcher edge cases** — no-data-arg, first-class-callable, non-`response()` receiver, same-namespace shadowing all guarded + unit-tested.
- **Conventions** — strict_types, `@internal`, verbose naming, comment style all comply.

**Finding — status fidelity (not a regression, but a known gap):**
The unwrap resolves the resource but drops the `json()` status; `ResourceResponseResolver` hardcodes `200`. So `response()->json($r, 201)` documents `200`.

**Lead analysis (baseline correction):** a `main` probe of this exact shape already emits a bare `200` with **no schema** — Core's inline-JSON resolver yields for a non-literal (resource) body and never reads the `201`, and no other resolver reads inline-json status. So the authored `201` is **already lost today**; this is not introduced here.

| | status | schema |
|---|--------|--------|
| before (main) | `200` | none |
| this PR | `200` | resource `$ref` ✓ |
| ideal | `201` | resource `$ref` |

The change is a **strict improvement** (adds the correct body under the same status). Reaching the ideal (`201`) means threading status through the resource path (`ResourceTarget` + `ResourceResponseResolver`), which the issue explicitly scoped out ("status is Core's concern"). Filed as follow-up.

Radiergummi:
🤖 **[lead]** Ready for review, **not merged** — escalating for your call (away-mode: no autonomous merge).

**State:** local gates green (Pint ✓, PHPStan 0 errors, 3479 tests ✓); one independent adversarial review done (findings above, all resolved/clean except the scoped-out status gap).

**What shipped (steps 1+2 of #517):** shared `ResponseJsonCall` matcher in Support + Core delegation (no-op refactor), and the resource reader now unwraps `response()->json(<resource>, …)`. **Step 3 of the issue was dropped as unnecessary** — an empirical probe proved Core already yields for resource-typed `json()` args (the issue's "Core wins with an inline shape" premise was wrong), so no Core emit-decision change and no controversial gate on that fork.

**Decision fork for you:**
1. **Merge as-is** — strict improvement (resource `$ref` where there was a bare 200); status fidelity (`201`) tracked as follow-up **#535**. *(recommended: small, additive, complete on its own terms.)*
2. **Expand scope** to thread the authored status now (#535 folded in) — bigger, touches shared `ResourceTarget` + `ResourceResponseResolver`, warrants its own review.

**Survey Layer A: not run** — corpus isn't bootstrapped in this session (multi-hour). Argument for low risk: the change is **structurally additive-only** — Core wins first for literal `json()` bodies (unchanged), and the reader's new unwrap only makes *more* loose-response/untyped actions resolve, never fewer, and only for the `response()->json(<resource>,…)` shape that previously refused. It cannot alter or remove any existing schema. Recommend a Layer A pass before merge if you want the corpus confirmation.

Radiergummi:
🤖 **[lead]** Remote CI **fully green** — whole matrix (PHP 8.4/8.5 × Laravel 12/13), the dedicated **swagger-php 5.8** job, and the byte-exact **snapshot** all pass. Merge-ready pending your review + the scope-fork decision above.

Issue #557, opened 2026-07-22
chore(survey): re-baseline the corpus and the private consumer-app ledger under the corrected scorer (#552)


Radiergummi:
## Why

#552 corrects three scoring defects in the survey harness (`completeness.php`,
`metrics.php`). Because those defects pulled in **opposite** directions — the
verb-based body requirement understated, while the contentless-2xx credit and the
JSON:API `{type,id}` substantiveness overstated — every historical figure the project
has published moves by an unknown amount **in an unknown direction**. None of them can
be compared against a post-#552 measurement.

This issue tracks the deliberate re-measurement, kept out of #552 so that PR stays a
reviewable scorer change rather than a corpus run.

## What needs re-baselining

- **The 11-app pinned corpus.** The cached `main` baseline in `$WS/results.json` is
  invalidated by the JSON:API unwrap in particular (it moves `responseSchemas`, which
  the surveyor diffs for regressions). Until it is regenerated, the **next PR after
  #552 will false-FAIL its survey gate** — so this is operationally urgent, not
  bookkeeping.
- **The honest coverage baseline.** Currently recorded as 34.9% (11 apps, measured
  2026-06-20, Fractal plugin off). Both the number and the default-vs-stack-enabled
  split (#443) need re-running.
- **The well-typed-bar figure** (#460) if it is derived from any affected metric.
- **The private consumer-app dogfooding ledger.** Its whole progression table — 51.3% out of the
  box → 62.0% library-only → 69.6% with the app plugin → 81.6% at `f0999f69`, plus the
  "honest ~87%" adjustment — was computed with the strict scorer. The ~87% figure was
  a *manual* correction for exactly the body-axis defect #552 now fixes properly, so it
  should be reproducible mechanically; if the corrected scorer disagrees with 87%, that
  discrepancy is itself worth understanding before either number is quoted again.

## Sequencing

Must run **after #552 merges** and **before** any subsequent survey-gated PR, or the
gate compares against a stale baseline. Practically this means it should be the first
thing done post-merge, not queued behind other work.

## Acceptance

- `$WS/results.json` regenerated on current `main` with the corrected scorer.
- The honest-coverage figure and the default/stack-enabled split re-stated with the
  measurement date and the library sha.
- The private consumer-app ledger's progression table re-measured and annotated as "corrected
  scorer", with the superseded figures kept but clearly marked so the history stays
  auditable.
- Any doc or memo quoting a superseded number is updated or marked stale.

[...]

PR #556, opened 2026-07-22
fix(survey): score completeness from evidence, not verb — three-way body + response buckets

Radiergummi:
Closes #552

## Revision log — round 2

| # | Finding | Resolution |
|---|---|---|
| 1 | **(blocking)** The acceptance invariant is false; the stated pass condition is unsatisfiable | **Acceptance rewritten.** Classified is the measurement basis; the percentage **does** move, by design. No machinery added to suppress it. Both precision errors (`202` subset, #550 attribution) corrected. |
| 2 | **(blocking)** The outcome triple cannot reproduce `documentedResponses` | `hasAnyResponse: bool` added; `maxRequestProperties` explicitly keeps its own walk. |
| 3 | The "complete" predicate still has two homes | Named once: `survey_isComplete(array $outcome, string $basis): bool`. |
| 4 | `operationsWithoutSecurity` has inverted polarity | Renamed `operationsWithSecurity`; both it and `requestBodies` added to the surveyor's watched list. |
| 5 | The playbook's `100%` gate becomes unreachable | `completeness.php` auto-detects `classify.json` (mirroring `metrics.php:434`); gate restated in reachable terms that **restore** the body axis. |
| 6 | Layer-B1 lift silently loses its request-body axis | `requestBodies` promoted to `required` in the lift schema and published as its own column. |
| 7 | `typedness.php:107` is an unnamed third consumer of `survey_substantive()` | Named; code change in scope, its published figures get the #557 staleness marker. |
| 8 | *(approval conditions)* Classify step must sit at **Phase 0**, not before the Phase-3 gate; the auto-detected `classify.json` must be **printed** | Both folded in — see *The playbook gate must stay reachable*. |

Round 1's resolutions stand unchanged except where superseded above.

---

## Tier

**Not applicable to the inference ladder** — survey-harness scoring only (`tools/survey/`), no `src/`
change and no generator plumbing. Every fix derives from the **spec plus the existing
`classify.json`**, never from new generator output.

## Scope decision: `metrics.php` is in scope (explicit, not silent widening)

All three defects live there, in `completenessPercent`:

- `metrics.php:197` — `$bodyVerbs = ['post','put','patch']` (defect 1).
- `metrics.php:262-266` — any contentless 2xx sets `$hasContentlessResponse`, credited into
  `$hasCompleteResponse` at `:289` with no status check (defect 2). The comment at `:288` claims
  *parity with `completeness.php`*.
[...]

Issue #552, opened 2026-07-22
fix(survey): completeness.php mis-scores in both directions — verb-based body requirement, synthetic empty 200 as complete, JSON:API {type,id} as substantive


Radiergummi:
## Problem

`tools/survey/completeness.php` has two scoring defects that pull the headline number
in **opposite** directions, so the net error is of unknown sign. That is bad enough
for corpus reporting, but it becomes actively harmful now that it gates autonomous
work: two of the currently-open response issues (#549, #550) would score as **zero or
negative** under the current metric despite being unambiguous fidelity improvements.

### Defect 1 — `needsBody` is verb-based (understates)

```php
$verbsBody = ['post', 'put', 'patch'];
$needsBody = in_array(strtolower($m), $verbsBody);
```

Every POST/PUT/PATCH is assumed to require a request body. Body-less writes are a
normal, correct API shape — path-only toggles (`POST .../star`, `.../verify`,
`.../apply`), create-with-defaults (`POST /apps`), and callback acknowledgements.
Documenting a body for them would be **wrong**, yet each one is scored incomplete.

In the private consumer app dogfooding run this is **26 operations** — the entire
"correctly-empty" bucket of the remaining gap, and the reason the honest figure
(~87%) diverges from the reported figure (81.3%).

### Defect 2 — a contentless 2xx always counts as a complete response (overstates)

```php
if (!is_array($content) || $content === []) {
    $hasResp = true;
    break;
} // explicit no-content 2xx (e.g., 204) is a complete response
```

The intent is right for a **204**, but the branch does not check the status code — so
the library's *synthetic fallback* `200 OK` with no content, emitted precisely when
inference found nothing, scores as a complete response. The metric rewards the
generator's give-up path.

Combined with defect 1, this produces two perverse outcomes:

- **#549** (read a controller helper's real status: `$this->empty()` → 204) moves the
  number **not at all**. Those ops already score `hasResp = true` via the synthetic
  200, and they stay incomplete on the body axis regardless. A real fidelity win
  measures as zero.
- **#550** (`never` → suppress the synthetic 2xx) moves the number **down**. Six
  fallback operations currently score complete purely because of the empty 200; doing
  the correct thing and emitting no 2xx flips all six to incomplete.

This is the same class of defect as #254 ("adding a real-but-empty schema *decreases*
responseSchemas") in a new location.

### Defect 3 — a JSON:API resource object is substantive on `{type, id}` alone (overstates)

[...]

Issue #550, opened 2026-07-22
feat(responses): read never / RedirectResponse / StreamedResponse from the declared return type (Tier-0)


Radiergummi:
## Problem

Three framework return types carry their entire response contract in the **declared
return type** — pure Tier-0 information — and the library reads none of them. All
three currently fall through to the synthetic `200 OK` with no content in
`OperationBuilder`, which documents something that never happens.

`grep -rn "RedirectResponse\|StreamedResponse" src` returns **zero hits**: this is
unbuilt, not merely incomplete.

### 1. `never` — the action cannot return

```php
public function notFound(): never
{
    throw new NotFoundHttpException(...);
}
```

A `never` return means there is **no success response at all**. The synthetic 200 is
strictly wrong. Correct behaviour: suppress the synthetic 2xx entirely. As a bonus,
when the body contains a **single unconditional `throw`** of a mappable exception,
the existing exception→response machinery should supply the real status (404 here).

### 2. `RedirectResponse` → 302 + `Location`

```php
public function landing(...): RedirectResponse
```

Should document `302 Found` with a `Location` response header (`type: string`,
`format: uri`), not an empty 200. The exact code is not knowable statically
(`redirect()` defaults to 302), so 302 is the honest default.

### 3. `StreamedResponse` (and `BinaryFileResponse`) → binary 200

```php
public function download(...): StreamedResponse
```

Should document `200` with `application/octet-stream` content and
`{type: string, format: binary}` — the OpenAPI idiom for a binary payload — rather
than an empty 200 that implies a JSON contract.

## Evidence

From the private consumer app dogfooding run (316 `/api` operations):

- `never`: 1 method, routed as the fallback → **6 operations**.
- `RedirectResponse`: **3 operations**.
- `StreamedResponse`: **1 operation** (media download).

~10 of the 54 remaining incomplete operations, and the second-largest remaining
pattern in that app after #549. Blast radius is broad — every app with OAuth/invite
landing redirects or file downloads hits this, and those are precisely the endpoints
where an empty-200 document is most misleading to a client.

## Proposal (Tier 0 — declared types only)

A new `PrimaryResponseResolver` (or an extension of `TypedReturnResponseResolver`,
which is already the "baseline typed-return" resolver and already degrades to null
rather than guessing) that matches on the declared return type:

| Declared return | Emitted |
|---|---|
| `never` | **no** primary response — suppress the synthetic 2xx |
[...]

PR #560, opened 2026-07-22
refactor(core): PrimaryResponse result type; delete the explicit-status sentinel

### LINKED ISSUE #516: refactor(core): widen PrimaryResponseResolver to a PrimaryResponse result; delete the explicit-status vendor-extension sentinel
## Problem

The body-scanned explicit success status is smuggled from `InlineJsonResponseResolver` to
`OperationBuilder` through a **transient vendor-extension sentinel**
(`OperationBuilder::EXPLICIT_STATUS_EXTENSION = 'laravel-openapi-explicit-status'`), because
`Contracts\Registry\PrimaryResponseResolver::resolvePrimaryResponse()` can only return a bare
`?OA\Response` — there is no channel to say "this response's status is explicit, don't let the
convention override it." The resolver writes `$response->x[...] = true`
(`InlineJsonResponseResolver::markStatusExplicit`) and `OperationBuilder::takeExplicitStatusMarker`
reads and strips it before serialization.

## Proposed change

Widen the contract to return a small result object:

```
PrimaryResponse { OA\Response response; bool statusIsExplicit }   // @internal, Contracts\Registry
```

`resolvePrimaryResponse(): ?PrimaryResponse` (null still = pass to the next resolver, first
non-null wins). `InlineJsonResponseResolver` sets `statusIsExplicit: true` on the result instead of
writing the vendor extension; `OperationBuilder` reads the flag off the result. Then **delete
`EXPLICIT_STATUS_EXTENSION`, `markStatusExplicit`, and `takeExplicitStatusMarker`** — the sentinel
and its widening travel together (same flow; there is no value in doing one without the other).

## Scope / blast radius

- **Contract change → human-gated.** `PrimaryResponseResolver` is in `Contracts\Registry\` (the
  documented plugin extension surface); changing its return type is a breaking change for
  third-party resolvers and must be approved by the maintainer before merge.
- **8 in-tree implementations** update to return the result object (7 trivially wrap their
  `OA\Response` with `statusIsExplicit: false`): Core (`InlineJsonResponseResolver`,
  `PaginatorResponseResolver`, `EloquentModelResponseResolver`, `FindReturnModelResponseResolver`),
  `Fortify\FortifyResponseResolver`, `ApiResources\ResourceResponseResolver`, Fractal
  (`FractalHelperResponseResolver`, `FractalResponseResolver`, `EntityTransformerResponseResolver`),
  `SpatieData\DataResponseResolver`. (Verified: `InlineJsonResponseResolver` is the **only** current
  writer of the sentinel.)

## Dependency

Depends on **#482's Pieces 1+2** (PR #514 — the `ResolvedField` merge primitive + applier
extraction). This was **Piece 3 of #482**, descoped during plan-review so #482 can merge without the
contract change gating it. See the plan-review discussion on #514.

## Tests

- All 8 implementations compile against the new return type; their existing response tests stay green.
- `InlineJsonResponseResolver` explicit status still wins over the resource convention (the behavior
  the extension protected), now via the result flag.
- No `x-laravel-openapi-explicit-status` appears in any generated document (pin it — it never did).
- Survey Layer A: 0 regressions (behavior-preserving; the sentinel was already stripped before
  serialization).

Refs #482, PR #514.

Radiergummi:
Closes #516

## Revision log — approval conditions

| # | Amendment | Resolution |
|---|---|---|
| 1 | Complete the account of the deferral cost | Property widening + **three** sites named. (The follow-up claim that PHPStan level 8 makes it safe was **retracted** at code review; the sites are enumerated by name and #550's guarantee is a test.) |
| 2 | Worked `resolvePrimaryResponse()` example | Folded in: `docs/plugin-authoring.md:269-280` fixed to implement the interface and return a `PrimaryResponse`. |
| 3 | Re-size the test sweep | 92 call sites / 4 files / **one** fake / **no** registry fake; the vendor-extension pin already exists. |
| 4 | Match `OperationConvention` form | Public promoted properties, no per-property `readonly`, qualified `@internal … seam` marker. |
| — | Lead with the layering argument | New opening section: this deletes a plugin → generator-internals coupling. |

---

## Why this is worth touching `Contracts\**`

**It deletes a plugin → generator-internals coupling.** `InlineJsonResponseResolver` imports
`Support\Generator\OperationBuilder` for one reason only: to read
`OperationBuilder::EXPLICIT_STATUS_EXTENSION` and write it into `$response->x`. A bundled plugin
reaching into a generator-internal constant to smuggle a value past a too-narrow contract is the
actual defect; the vendor-extension sentinel is its symptom. Widening the return type removes the
import, the constant, and the strip-before-serialize dance together.

That is the argument for the human gate, and it is a better one than "plumbing refactor": the
contract change buys a layering fix, not just tidier plumbing.

## Tier

**Not an inference change** — no new reading, no new evidence, no tier movement. Behaviour is
preserved exactly; the only observable difference would be a bug, not a feature.

## Deliberate public-surface change — expect a human gate

`Contracts\Registry\PrimaryResponseResolver` is the documented plugin extension surface, so this
**will not auto-merge**; the gate treats any `src/Contracts/**` change as controversial and escalates
for Moritz's sign-off. Planned as such:

- **The interface signature changes** from `resolvePrimaryResponse(ActionDescriptor): ?OA\Response`
  to `resolvePrimaryResponse(ActionDescriptor): ?PrimaryResponse`.
- **Out-of-tree plugin authors** must wrap their return value:
  `return PrimaryResponse::of($response);`. That is the entire migration — one call, no behaviour
  change, and `null` keeps meaning "pass to the next resolver". The package is pre-1.0 and
  unpublished, so no deprecation shim is warranted (and per project convention, no migration section
  is written into the docs).
- **`PrimaryResponse` itself is `@internal`** — genuinely, not defensively: it is a return-channel
  carrier with no behaviour, and its shape is expected to grow (see the third state, below). It
  lives in `Contracts\Registry\` beside the interface that returns it, per the registry-seam
  convention. Callers construct it through named constructors only.

## Correction to the issue's blast radius

The issue says **8** in-tree implementations and lists 10. The real count is **11** — it misses
`src/Support/Extraction/TypedReturnResponseResolver.php:48`, which also
`implements PrimaryResponseResolver`. That one matters beyond arithmetic: it sits in `Support\`, so
the result object must stay plugin-agnostic (it does — it carries `OA\Response` and a bool).

Full list, all of which change:

| namespace | resolvers |
|---|---|
| `Support\Extraction` | `TypedReturnResponseResolver` **(missing from the issue)** |
| `Plugins\Core\Resolvers` | `InlineJsonResponseResolver`, `PaginatorResponseResolver`, `EloquentModelResponseResolver`, `FindReturnModelResponseResolver` |
| `Plugins\ApiResources\Resolvers` | `ResourceResponseResolver` |
| `Plugins\Fortify\Resolvers` | `FortifyResponseResolver` |
| `Plugins\Fractal\Resolvers` | `FractalResponseResolver`, `FractalHelperResponseResolver`, `EntityTransformerResponseResolver` |
| `Plugins\SpatieData\Resolvers` | `DataResponseResolver` |

Ten wrap their existing return value in `PrimaryResponse::of(...)`; only
`InlineJsonResponseResolver` passes `statusIsExplicit: true` (verified: it is the only writer of the
sentinel today, at `:90`, `:386`, `:403`).

## The third state, and the explicit design call (#550)

#550 needs a `never`-returning action to say **"emit no synthetic 2xx at all"**. Today
`OperationBuilder:284-286` falls back to `new OA\Response(['response' => '200', …])` whenever nothing
resolved, and `?OA\Response` cannot express the difference — `null` already means *"I don't know,
ask the next resolver"*.

**Ruling for the design: a distinct case, not a boolean flag.**

```php
PrimaryResponse::of(OA\Response $response, bool $statusIsExplicit = false): self
PrimaryResponse::suppressed(): self   // ← #550 adds this
```

**`null` and `suppressed()` are not the same thing, and must never be conflated.** `null` means
*"I have nothing to say about this action, ask the next resolver"*: the chain continues, and if
every resolver returns `null` the synthetic `200` fallback still fires. `suppressed()` will mean
*"I claim this action, and it deliberately has no success response"*: the chain stops and **no**
fallback is emitted. Preserving that distinction is the entire reason a result wrapper beats a bare
`?OA\Response` — the old type had only one empty value and therefore could only express the first.
#550 must add the second as a new case, never by overloading `null`.

Reasoning: "here is the response" and "there is deliberately no response" are **mutually
exclusive**. A `bool $suppressesFallback` alongside a non-null `$response` would admit the
unrepresentable combination *(a response, and also suppress it)* and force every reader to decide
which wins. A named case cannot be constructed into that state. It also reads correctly at the call
site — `return PrimaryResponse::suppressed();` says what it means.

**What this PR builds: the two-state version only.** `PrimaryResponse::suppressed()` has no consumer
until #550, and speculative API is against the project's convention. Because construction goes
through named constructors from day one, #550 adds the third case **additively** — no interface
signature change, and none of the other ten resolvers are touched.

**The full cost of the deferral, stated honestly.** It is not "one branch in `OperationBuilder`":

- `PrimaryResponse::$response` must widen to `?OA\Response` — a **type change on a public property of
  the class this PR introduces**. Cheap because the class is `@internal` and constructed only through
  named constructors, but it is a change to this PR's own output, not a pure addition.
- It lands on **three** existing sites, all inside `buildOperation()` (`$autoPrimaryResponse` has six
  use sites, only these three dereference or decide on it):
  1. the `?? new OA\Response(['response' => '200', …])` fallback chain — the branch #550 wants;
  2. `$conventionSuppressedByBody`, which reads `->content`;
  3. `emitMissingResponseSchemaFinding(responseProduced:)`, which must **decide** whether
     "deliberately no response" counts as produced. That is a judgement call, not a mechanical
     unwrap, and it belongs to #550.

**Retracted: PHPStan level 8 does *not* make the deferral safe.** An earlier revision of this plan
claimed the type checker would enumerate the three sites for #550. As implemented it will not, and
the claim is withdrawn. `OperationBuilder` unwraps the result immediately at `:241-242` into
`$autoPrimaryResponse`, a local that is **already** `?OA\Response`, and all three sites below
null-guard that local. Widening `PrimaryResponse::$response` later would therefore change no
inferred type and raise **no** PHPStan error anywhere. Do not rely on a guarantee that does not
exist.

**What actually makes the deferral safe** is the two things under #550's own control: the three
affected sites are enumerated by name above, so the work is scoped rather than discovered; and
#550's guarantee must be **a test** asserting that a suppressed result emits no synthetic 2xx (and
deciding, explicitly, what `emitMissingResponseSchemaFinding(responseProduced:)` should report for
it). Without that test, #550 could ship a `never`-returning action that still emits the synthetic
`200` *and* flags a missing response schema, with nothing failing to catch it.

**Explicitly rejected: pre-emptively making `$response` nullable.** It looks free, but it would give
"result with a null response" a meaning *today* (fall through to the synthetic `200`) that is the
exact opposite of what #550 needs it to mean (emit nothing). Flipping a meaning is worse than adding
a case.

## The change, precisely

1. **New `src/Contracts/Registry/PrimaryResponse.php`** — `final readonly class`, private
   constructor, `public static function of(OA\Response $response, bool $statusIsExplicit = false):
   self`, promoted `public OA\Response $response` and `public bool $statusIsExplicit`. No behaviour,
   no logic.

   **Form follows `Contracts\Registry\OperationConvention`**, the nearest precedent in the same
   namespace: a `final readonly class` with **public promoted properties and no redundant
   per-property `readonly`**, and the qualified internal marker in that file's exact form —
   `@internal Part of the {@see PrimaryResponseResolver} seam; not a committed public contract.`
2. **`PrimaryResponseResolver::resolvePrimaryResponse()`** returns `?PrimaryResponse`; the docblock
   keeps "null to pass to the next resolver (first non-null wins)" and gains one line on what the
   result carries.
3. **Eleven implementations** wrap their return values. `InlineJsonResponseResolver` replaces its
   three `markStatusExplicit(...)` call sites with `PrimaryResponse::of($response, statusIsExplicit:
   …)` and **deletes `markStatusExplicit()`** (`:286-291`).
4. **`OperationBuilder`** — the resolver loop (`:236-246`) collects a `?PrimaryResponse`; the
   fault-boundary closure's return type follows. `$autoStatusIsExplicit` comes off the result
   (`:274`) instead of `takeExplicitStatusMarker()`. Everything downstream keeps working on
   `$autoPrimaryResponse->response`, including the `$conventionSuppressedByBody` check at `:287-291`.
   **Delete `EXPLICIT_STATUS_EXTENSION` (`:83`) and `takeExplicitStatusMarker()` (`:832-848`)** — with
   them, the `Generator::UNDEFINED` dance for emptied `->x` arrays disappears too.

The sentinel and the widening travel together, per the issue: there is no value in one without the
other.

## Files

- `src/Contracts/Registry/PrimaryResponse.php` — **new**.
- `src/Contracts/Registry/PrimaryResponseResolver.php` — return type + docblock.
- `src/Support/Generator/OperationBuilder.php` — `:83`, `:236-246`, `:274`, `:832-848`.
- `src/Support/Extraction/TypedReturnResponseResolver.php` — the one the issue missed.
- The ten plugin resolvers listed above.
- **Test sweep, measured rather than guessed:** 92 `resolvePrimaryResponse()` call sites across four
  files — `InlineJsonResponseResolverTest` (49), `FractalHelperResponseResolverTest` (26),
  `FindReturnModelResponseResolverTest` (15), `OperationBuilderFaultIsolationTest` (2). Exactly
  **one** hand-rolled fake resolver exists (in `OperationBuilderFaultIsolationTest`), and there is
  **no** registry/pipeline fake — the earlier plan claimed otherwise and was wrong.
- `tests/Unit/Core/Inference/InlineJsonResponseResolverTest.php:66` — asserts the sentinel on
  `$response->x` today; must assert `statusIsExplicit` on the result instead.
- `docs/plugin-authoring.md:40,77,126` and `docs/architecture.md:44,117` — the documented resolver
  surface.
- **`docs/plugin-authoring.md:269-280` — a worked `resolvePrimaryResponse()` example, folded into
  this PR.** There is currently **none anywhere**: that snippet declares
  `resolveResponse(): mixed` and illustrates `ResourceTargetLocator`, so it neither implements the
  interface nor shows a return value. After this PR a plugin author is told to return a
  `PrimaryResponse` with nowhere to see one constructed, and `@internal` keeps it out of generated
  API docs while being mandatory to construct. Fix the snippet to implement
  `PrimaryResponseResolver` and `return PrimaryResponse::of($response);` — same change's public
  surface, so it does not belong in a later docs phase.
- `CHANGELOG.md` `[Unreleased]` → **Changed** (public extension surface).

## Tests

Behaviour-preserving, so the bar is "nothing moves", proven positively rather than by absence:

- **All eleven implementations** compile against the new return type and their existing response
  tests stay green — the main regression net.
- **`InlineJsonResponseResolver` explicit status still wins over the convention** (`:274` →
  `:296-300`), now via the result flag: an action whose body scan yields an explicit `201` keeps
  `201` when the route convention says `200`/`204`. This is the exact behaviour the sentinel
  protected.
- **`statusIsExplicit: false` by default** — a convention status still applies over the other ten
  resolvers' responses.
- **`$conventionSuppressedByBody`** (`:287-291`, a `204` convention yielding to a content-bearing
  body-scanned response) still holds after the unwrap to `->response`.
- **No `x-laravel-openapi-explicit-status` in any generated document** — the pin **already exists**
  at `tests/.../OperationBuilderConventionStatusTest.php:132`, so this is "keep it green", not "add
  it". Nothing new to write.
- **`->x` is untouched** for a resolver that legitimately sets its own vendor extensions: previously
  `takeExplicitStatusMarker()` mutated `->x` and could reset it to `Generator::UNDEFINED`; nothing
  should now touch it at all. Guards the deletion.
- **`null` still means pass-through**: a first resolver returning `null` and a second returning a
  result yields the second's response, unchanged.
- **The fault boundary still isolates a throwing resolver** and moves to the next one (the closure's
  return type changed).
- The single hand-rolled fake (`OperationBuilderFaultIsolationTest`) updated to the new return type.

Gates: `composer test`, `vendor/bin/pint --test`, `composer lint`. PHPStan level 8 is the real
compile check for all eleven implementations — a missed one is an error, not a silent pass.

## Merge risk

- **Controversial by the standing rule** — touches `src/Contracts/**`. Expect `agent:needs-human` and
  Moritz's sign-off; do not auto-merge.
- **Survey Layer A: 0 regressions expected.** Behaviour-preserving, and the sentinel was already
  stripped before serialization, so no generated document should differ by a byte. A non-zero delta
  means the refactor changed behaviour — treat any movement as a **FAIL to investigate**, not as
  noise. Baseline provenance caveat: #552 (PR #556) changes the scorer and #557 re-baselines, so the
  surveyor must confirm which `main` SHA and which scorer version its baseline predates.
- **Dependency satisfied**: #482 Pieces 1+2 (PR #514) merged 2026-07-02.
- **Gates #550**, which consumes the third state. Land this first.


Radiergummi:

Three things for the reviewer:

- **The issue undercounts the blast radius.** It says 8 implementations; there are **11**. It misses `src/Support/Extraction/TypedReturnResponseResolver.php:48`, which also implements the interface — and that one sits in `Support\`, so it pins the requirement that the result object stay plugin-agnostic.
- **The third state (#550) is designed as a distinct case, not a boolean flag**: `PrimaryResponse::suppressed()` beside `PrimaryResponse::of(...)`. "Here is the response" and "there is deliberately no response" are mutually exclusive; a `bool $suppressesFallback` alongside a non-null `$response` would admit the unrepresentable *(a response, and also suppress it)* and force every reader to pick a winner. **This PR builds the two-state version only** — `suppressed()` has no consumer until #550, and named constructors make it additive then (no interface change, no touching the other ten resolvers, only `OperationBuilder` gaining the skip-the-fallback branch).
- **Explicitly rejected: pre-emptively making `$response` nullable.** It looks free, but it would give "result with a null response" a meaning today (fall through to the synthetic `200` at `OperationBuilder:284-286`) that is the *opposite* of what #550 needs it to mean. Flipping that later is worse than adding a case. Veto-able with the trade-off in view.

**Human gate expected:** this changes `src/Contracts/Registry/PrimaryResponseResolver`, so it will not auto-merge. Migration for out-of-tree resolvers is one call — `return PrimaryResponse::of($response);` — and `null` keeps meaning pass-to-next.

**Survey note:** behaviour-preserving, and the sentinel was already stripped before serialization, so **any** Layer A delta is a FAIL to investigate rather than noise.

Radiergummi:
🤖 **[lead]** planning → plan-review.

**I corrected the issue body during planning: 11 implementations, not 8** — see the comment on #516. The missed one, `Support/Extraction/TypedReturnResponseResolver.php:48`, is load-bearing: it sits in `Support\\`, so the project's boundary rule pins `PrimaryResponse` to being plugin-agnostic.

**Reminder that this PR will not auto-merge.** It changes `Contracts\\Registry\\PrimaryResponseResolver`; the gate treats any `src/Contracts/**` change as controversial by construction. It will be driven to green-and-approved and escalated for Moritz's sign-off.

Two design calls the planner made and explicitly invited a veto on — review them on the merits:

1. **The third state is a distinct case, not a flag** (`PrimaryResponse::suppressed()` beside `PrimaryResponse::of(...)`), because "here is the response" and "there is deliberately no response" are mutually exclusive, and a `bool` beside a non-null `$response` admits an unrepresentable combination.
2. **This PR builds only the two-state version**; `suppressed()` arrives with #550, which is its first consumer. Named constructors make that additive: no interface change, no touching the other ten resolvers, only `OperationBuilder` gaining a branch.
3. It **rejected** pre-emptively making `$response` nullable, on the grounds that "result with a null response" would mean *fall through to the synthetic 200* today (`OperationBuilder:284-286`) but *emit nothing* under #550 — and flipping an existing meaning is worse than adding a case.

I find all three sound, but they are public-surface decisions, so say so if you disagree. Also confirm the migration story for out-of-tree resolvers is genuinely one call, that `null` still means pass-to-next, and that no deprecation shim is right for pre-1.0.

Survey note from the plan: this is behaviour-preserving (the sentinel is already stripped pre-serialization), so **any** Layer A delta is a FAIL to investigate, not noise.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved with 4 required plan amendments** — no design veto. I reviewed this as a deliberate public-surface change ("is this the right shape to live with"), not just for correctness.

## The three design calls: all three hold. No veto.

**1. Distinct case, not a boolean flag — correct.** The argument is right as stated: `bool $suppressesFallback` beside a non-null `$response` admits *(a response, and also suppress it)*, and every reader would then need a precedence rule. A named constructor makes that state unconstructible. It also matches how the seam is read at the call site.

**2. Deferring `suppressed()` to #550 is genuinely additive — but the plan's account of *why* is incomplete.** I traced it: `$autoPrimaryResponse` has exactly **six** use sites, all inside `OperationBuilder::buildOperation()` — `:234`, `:237`, `:243`, `:280`, `:284`, `:290-292` — of which only `:284` and `:291-292` dereference the response. So when #550 lands: the interface is **not** re-opened, the ten other resolvers are **not** touched, and the blast radius stays inside one method. That answers your question: yes, additive.

But it costs more than "OperationBuilder gains a skip-the-fallback branch". `suppressed()` cannot be constructed unless `PrimaryResponse::$response` widens to `?OA\Response` — a type change on a public property of the class *this* PR introduces — and it lands on three existing sites, not one: the `?? ` chain (`:284`), `$conventionSuppressedByBody` (`:290-292`), and `emitMissingResponseSchemaFinding(responseProduced: …)` (`:280`), which must decide whether "deliberately no response" counts as produced (it should). *Required:* state the day-2 shape. And state the argument that actually makes the deferral **safe** rather than merely intended: under PHPStan level 8 a nullable property forces every one of those readers to be handled, so #550 cannot silently miss one. That converts "additive, trust me" into "the analyser proves it" — which is the standard this PR is being gated at.

**3. Rejecting a pre-emptively nullable `$response` — correct, and for the reason given.** It is also self-consistent with call 2: `of()` takes a non-null argument, so the property's nullability is unobservable until `suppressed()` exists. Giving "null response" a meaning today (fall through to the synthetic `200` at `:284`) that #550 inverts (emit nothing) would be a genuine trap. Confirmed.

## Confirmations you asked for

- **Migration is genuinely one call.** Verified all eleven: ten return a bare `OA\Response`-or-`null` and wrap mechanically; only `InlineJsonResponseResolver` writes the sentinel, at `:90`, `:386`, `:403`. The count of **11** is independently confirmed (`implements PrimaryResponseResolver` → 11 files).
- **`null` still means pass-to-next.** Unchanged: the loop at `:234-245` breaks on the first non-null. And the fault boundary needs **no** change — `ResolverFaultBoundary::isolate()` is `@template T` / `@return null|T` returning `mixed`, so the closure's new return type flows through. The plan's file list is right to omit it.
- **`@internal` + placement: correct, and it has an exact precedent.** `Contracts\Registry\OperationConvention` is the same construct — a `final readonly class` VO in `Contracts\Registry\`, returned by `OperationConventionResolver`, that plugin authors must construct, marked `@internal Part of the {@see OperationConventionResolver} seam; not a committed public contract.` Same tension, already resolved this way. See amendment 4 for the form.
- **`PrimaryResponse` stays plugin-agnostic**, so `Support\Extraction\TypedReturnResponseResolver` is fine: it carries an `OA\Response` and a `bool`, and `Contracts\` already imports `OpenApi\Annotations` in the interface itself. Boundary rule honoured.
- **No deprecation shim is right** — pre-1.0 and unpublished, and per project convention no migration section goes in the docs.
- **Survey framing is already right** and needs no edit: 0-delta expected, any delta is a FAIL to investigate, plus the baseline-provenance caveat against #556/#557. One more fact that supports the expectation: `InlineJsonResponseResolver` constructs a **fresh** `OA\Response` at each of its three exits and never caches, so today's read-and-strip semantics and tomorrow's carried flag cannot diverge on a reused object — which is the one way a "transient marker → result field" swap could have leaked a real behaviour change.

## Required amendments

**1. Complete the `suppressed()` deferral argument** (the property widening, the three affected sites, and the PHPStan-level-8 safety argument) — as above.

**2. The docs gap is bigger than the cited lines.** `plugin-authoring.md:40, 77, 126` are prose one-liners, and I checked the one thing that looks like a counter-example: the `JsonApiPrimaryResponseResolver` at `plugin-authoring.md:271` deliberately declares `resolveResponse(ActionDescriptor): mixed` — it illustrates `ResourceTargetLocator`, not the interface, so it needs no edit. Which means **the docs contain no worked `resolvePrimaryResponse()` example at all**. After this PR a plugin author is told to return a `PrimaryResponse` and has nowhere to see one — and because the class is `@internal`, it is excluded from generated API docs while being *mandatory* to construct. *Required:* `plugin-authoring.md` gains a minimal concrete snippet — `return PrimaryResponse::of($response);`, `null` = pass to the next resolver, and one line on what `statusIsExplicit` is for. This is the one consumer-visible gap in an otherwise behaviour-preserving change.

**3. The test sweep is mis-sized and mis-aimed.** `resolvePrimaryResponse` appears **92 times across four files**: `InlineJsonResponseResolverTest` (49), `FractalHelperResponseResolverTest` (26), `FindReturnModelResponseResolverTest` (15), `OperationBuilderFaultIsolationTest` (2). There is exactly **one** hand-rolled fake — in `OperationBuilderFaultIsolationTest`; `StreamingResponseTest`'s only match is a comment, and **no** registry or pipeline test has one. So "plugin tests and registry/pipeline tests both have them" sends the sweep after something that does not exist while under-sizing the ~90 assertion unwraps that do. *Required:* name the four files and the one fake. Also: the "no `x-laravel-openapi-explicit-status` in any generated document" pin **already exists** at `OperationBuilderConventionStatusTest.php:132` — that bullet is "keep green", not "add".

**4. Match the sibling VO's form.** `OperationConvention` uses promoted **public** properties with **no** per-property `readonly` (legal inside a `readonly` class on 8.4 — I checked — but redundant, and Pint/consistency will notice); the plan specifies "promoted readonly public". Use the same qualified `@internal Part of the {@see PrimaryResponseResolver} seam; not a committed public contract.` rather than a bare `@internal`, so it reads as *unstable shape*, not *do not use* — the whole point is that plugin authors construct it. And note the one **deliberate** divergence from that precedent — `OperationConvention` has a public constructor, `PrimaryResponse` a private one plus named constructors — in the docblock, since that divergence is exactly what makes design call 2 work; unexplained it reads as inconsistency.

## Two optional notes

- **An argument the plan under-sells, worth having in front of Moritz.** `Plugins\Core\Resolvers\InlineJsonResponseResolver` currently imports `Support\Generator\OperationBuilder` *solely* to read its constant — a plugin resolver reaching into generator internals. Deleting the sentinel removes that coupling entirely. That is a stronger case for a `Contracts\**` change than "plumbing refactor with no observable difference".
- `docs/internal/architecture-review.md:57` and `:304` list this sentinel as an open smell. A one-line "resolved by #516" keeps that record honest.

## Naming nit (no change demanded)

After the change `OperationBuilder` holds both `$autoPrimaryResponse` (a `PrimaryResponse`) and `$primaryResponse` (an `OA\Response`), so `$autoPrimaryResponse->response` sits three lines from `$primaryResponse`. It reads acceptably and the local names are pre-existing; flagging only so the coder keeps them distinct rather than "tidying" them together.

Amendments 1-4 are plan-text only — no design change, no scope change. Fold them in and this is ready to code; the merge outcome remains green-and-approved → escalate for Moritz's sign-off.

Radiergummi:
🤖 **[planner]** four amendments folded in — plan ready for `coding`.

1. **The deferral's cost is now accounted in full.** `suppressed()` requires widening `PrimaryResponse::$response` to `?OA\Response` — a type change on a public property of the class *this* PR introduces, not a pure addition — and it lands on **three** sites in `buildOperation()`, not one: the `??` fallback chain, `$conventionSuppressedByBody` (reads `->content`), and `emitMissingResponseSchemaFinding(responseProduced:)`, which has to **decide** whether "deliberately no response" counts as produced. That third one is a judgement call belonging to #550, not a mechanical unwrap. The plan now states that what makes the deferral *safe* is **PHPStan level 8**: a nullable property forces every reader to be handled, so #550 cannot silently miss one — the compiler enumerates the sites for it. That is a stronger argument than "it's additive", and it replaces the weaker one.
2. **A worked example is folded into this PR**, not deferred. `plugin-authoring.md:269-280` is not one today — it declares `resolveResponse(): mixed` and illustrates `ResourceTargetLocator`, so it neither implements the interface nor shows a return value. It gets fixed to implement `PrimaryResponseResolver` and `return PrimaryResponse::of($response);`, which both supplies the missing example and repairs a snippet that was already misleading. Agreed it belongs here: `@internal` keeps the type out of generated API docs while being mandatory to construct, so the docs page is the only place an author can see one.
3. **Test sweep re-sized to the measured numbers** — 92 call sites across `InlineJsonResponseResolverTest` (49), `FractalHelperResponseResolverTest` (26), `FindReturnModelResponseResolverTest` (15), `OperationBuilderFaultIsolationTest` (2); **one** hand-rolled fake, and **no** registry/pipeline fake. My earlier claim of registry/pipeline fakes was wrong and is corrected in place. The `x-laravel-openapi-explicit-status` bullet is now "keep green" — the pin already exists at `OperationBuilderConventionStatusTest.php:132`.
4. **Form matches `Contracts\Registry\OperationConvention`** — `final readonly class` with public promoted properties and no redundant per-property `readonly`, and the qualified marker in that file's exact form: `@internal Part of the {@see PrimaryResponseResolver} seam; not a committed public contract.`

Also taken: **the layering argument now leads the PR body.** `InlineJsonResponseResolver` imports `Support\Generator\OperationBuilder` for the sole purpose of reading its constant — a bundled plugin reaching into generator internals to smuggle a value past a too-narrow contract. That coupling is the actual defect and the sentinel is its symptom, which is a far better justification for touching `Contracts\**` than "plumbing refactor". Thanks for surfacing it.

Radiergummi:
🤖 **[lead]** plan-review → coding.

Approved with all four amendments folded in. Survey gate **applies**, and note the plan's own framing: this is behaviour-preserving (the sentinel is already stripped pre-serialization), so **any** Layer A delta is a FAIL to investigate rather than noise.

**This PR will be escalated, not auto-merged** — it changes `Contracts\Registry\PrimaryResponseResolver`, which the gate treats as controversial by construction. Drive it to green and two approvals; I will hand it to Moritz.

Coder notes:

- **11 implementations, not 8** (the issue body was short by three). `Support/Extraction/TypedReturnResponseResolver.php:48` is the one to be careful with: it sits in `Support\`, so `PrimaryResponse` must stay plugin-agnostic or the boundary rule breaks.
- **Build only the two-state version.** `suppressed()` arrives with #550, its first consumer. Do not pre-emptively make `$response` nullable — "result with a null response" would mean *fall through to the synthetic 200* today but *emit nothing* under #550, and flipping an existing meaning is worse than adding a case later.
- The `plugin-authoring.md:269-280` edit **repairs an already-misleading snippet** (it declares `resolveResponse(): mixed`), it does not merely add an example. Treat it as a correctness fix to the docs.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed `607679bb`. Ready for review.

**Local gates** (worktree, PHP 8.5 / Laravel 13):
- `composer test` — **3492 passed** (8555 assertions), including the byte-exact `snapshot` group
- `vendor/bin/pint --test` — passed
- `composer lint` (PHPStan level 8) — no errors

**What landed, against the plan:**
- `src/Contracts/Registry/PrimaryResponse.php` — new `final readonly class`, private constructor, `public static of(OA\Response $response, bool $statusIsExplicit = false)`, public promoted properties, `@internal Part of the {@see PrimaryResponseResolver} seam; not a committed public contract.` Two-state only; `suppressed()` deliberately not built.
- Interface returns `?PrimaryResponse`; `null` unchanged as "pass to the next resolver".
- All **11** implementations wrapped (including `Support\Extraction\TypedReturnResponseResolver` — `PrimaryResponse` carries only `OA\Response` + a bool, so `Support\` stays plugin-agnostic). Only `InlineJsonResponseResolver` passes `statusIsExplicit: true`; its `markStatusExplicit()` and its `use …\Support\Generator\OperationBuilder;` import are both gone — **the layering coupling is deleted**.
- `OperationBuilder`: `EXPLICIT_STATUS_EXTENSION` and `takeExplicitStatusMarker()` removed, taking the `Generator::UNDEFINED` dance and its `@phpstan-ignore` with them. The now-orphaned `OpenApi\Generator` and `is_undefined` imports were dropped.
- `ResolverFaultBoundary` untouched, as reviewed.

**Tests:** the 92-site sweep went through a shared `primaryResponseAnnotation(?PrimaryResponse): ?OA\Response` unwrap helper in `tests/Pest.php`, so the 75 pure-response assertions read exactly as before. The two explicit-status tests now assert `$result->statusIsExplicit` directly instead of poking `$response->x`. Added to `OperationBuilderFaultIsolationTest`: null pass-through to the next resolver, and a guard that a resolver’s own `->x` vendor extensions survive untouched (what the deleted `takeExplicitStatusMarker()` used to mutate). The `x-laravel-openapi-explicit-status` pin at `OperationBuilderConventionStatusTest:126` stays green (title reworded, the marker is no longer "transient").

**Docs:** `docs/plugin-authoring.md` — registry table row, the `PrimaryResponseResolver` bullet (construct via `PrimaryResponse::of()`, `statusIsExplicit` semantics), and the `ResourceTargetLocator` snippet repaired to actually `implements PrimaryResponseResolver` / `resolvePrimaryResponse(): ?PrimaryResponse` / `return PrimaryResponse::of($response);`. `docs/architecture.md` needed no change (its first-non-null-wins wording still holds).

**No deviations from the plan.** Behaviour-preserving as predicted — any survey Layer A delta is a bug, not noise.

Radiergummi:
🤖 **[lead]** coding → review. PR marked ready (`607679bb`); CI running across the full matrix including the swagger-php 5.8 job.

Coder reports local gates green (3492 tests / 8555 assertions incl. the snapshot group, Pint clean, PHPStan level 8 clean) and **no spec delta** — the byte-exact snapshot tests are unchanged, which is the load-bearing claim for a change advertised as behaviour-preserving.

**Holding it as a draft would have blocked review, so it is ready now.** The escalation for this PR happens at *merge*, not at review: it goes through both reviewers and the survey gate as normal, and then I hand it to Moritz rather than auto-merging.

Assigned to **both** reviewers for independent code review — the merge gate requires two approvals, and this one additionally has to be right about a public surface. Reviewer2 saw the plan; reviewer1 has not, which is useful here.

Radiergummi:
🤖 **[reviewer]** review round 1: **approved** — 1 should-fix and 2 nits, none blocking. I read the diff before the PR body, as asked.

The seam is the right one. The old design had a plugin reaching into an `OperationBuilder` constant to write a marker onto a swagger-php annotation so a *different* layer could read and unset it before serialization — a data channel that existed purely because the return type was too narrow, with the object mutated in flight. Replacing it with a result type deletes the channel, the import, the strip method, and the `Generator::UNDEFINED` re-clear along with its `@phpstan-ignore`. Fewer moving parts and a narrower contract.

### Q4 — behaviour preservation: confirmed independently

Ran the gates on `607679bb` myself rather than taking the report: **3492 tests pass (8555 assertions)** including the byte-exact `snapshot` group (locally it is only excluded on the swagger-php 5.x CI job), **PHPStan level 8 clean** over `src` + `tests` (1403 files — so every implementation, including the anonymous test ones, is type-checked), **Pint clean**.

Two structural checks beyond the suite:
- **No orphaned producers or readers of the sentinel.** Grepped `src/`, `tests/`, `docs/`: the only surviving mentions are the CHANGELOG entry, the retained regression test asserting `x-laravel-openapi-explicit-status` never serializes, and two lines in `docs/internal/architecture-review.md` (nit 3). The `is_undefined` import was correctly dropped — no remaining use in `OperationBuilder`.
- **The read is moved earlier and that is safe.** The flag used to be taken at the old `:269`, after the convention/additional-response block; it is now read at `:241-242`, immediately after the resolver loop. Nothing between those points touches `$autoPrimaryResponse->x`, and the only consumers of the flag (`:291`, `:301`) sit after both positions.

**One thing the commit message undersells.** `markStatusExplicit()` did `$response->x = [KEY => true]` — an unconditional **overwrite** of the whole `x` array. Any resolver that set its own vendor extensions *and* an explicit status would have lost them, and `takeExplicitStatusMarker()` then re-cleared `x` to `UNDEFINED` when it emptied. No bundled resolver hits this (only `InlineJsonResponseResolver` used the marker, and it constructs fresh responses), so "no spec output change" is true as written — but the refactor also **fixes a latent bug** for out-of-tree resolvers, and the new `leaves a resolvers own vendor extensions on the response untouched` test pins exactly that. Good instinct writing it; worth a clause in the CHANGELOG entry, since for a third-party resolver this is a behaviour change in the good direction rather than a pure no-op.

### Q1 — does #550 re-open the interface?

Mostly no, and the residue is cheap. Adding `suppressed()` leaves `PrimaryResponseResolver::resolvePrimaryResponse(): ?PrimaryResponse` **untouched** — the interface does not re-open, which is the part that matters for out-of-tree implementers. What does change is the VO: `public OA\Response $response` has to widen to `?OA\Response`. That is not additive, but it is a property-type widening on an `@internal` class with exactly **one** in-tree reader (`OperationBuilder:241`), already dereferenced as `$autoPrimaryResult?->response`, and PHPStan level 8 will point at it. Deferring is the right call, and the private constructor plus named constructor is what makes the third state a one-line addition rather than a signature change.

Worth stating plainly in the PR so #550 inherits it: **`null` and `suppressed()` must stay semantically distinct** — `null` is "not my job, ask the next resolver", `suppressed()` is "I claim this action and it has no success response". That distinction is the whole reason the wrapper beats a bare `?OA\Response`, and it is the one thing #550 could get wrong.

### Q2 — is `null`-means-pass-to-next still unambiguous?

Yes, and slightly more so than before. The loop breaks on the first non-null `PrimaryResponse` (`:236`), all-null leaves it null, and the fault boundary returning null on a throw keeps a broken resolver in the pass-to-next lane — all unchanged. The interface docblock still says it correctly. The new `falls through a resolver returning null to the next one` test pins the chain behaviour that previously had no direct coverage. The wrapper actually *removes* a latent ambiguity: a resolver that wanted to say "no body" previously had to return either a bare 200-shaped response or null, and `suppressed()` will give it a third word.

### Q3 — migration cost and discoverability (the should-fix)

The mechanical cost is genuinely one call: wrap the returned `OA\Response` in `PrimaryResponse::of(…)` and change the return type. The CHANGELOG entry spells that out including the `statusIsExplicit:` named argument, and `docs/plugin-authoring.md` is updated in all three places it needed to be — the registry table row, the interface bullet, and the worked example (which also picks up two genuine fixes in passing: it now actually `implements PrimaryResponseResolver` and uses the real method name, and `$target->isAmbiguous()` → `$target->isAmbiguous` matches the property hook at `ResourceTarget.php:14`).

**1. (should-fix, non-blocking) The `@internal` tag on `PrimaryResponse` contradicts the interface that returns it.** `PrimaryResponseResolver` is *not* `@internal`, is listed in `docs/plugin-authoring.md` as an extension point, and cannot be implemented without naming and constructing `PrimaryResponse`. Meanwhile `CLAUDE.md` defines `@internal` as "not part of the documented extension surface". Checking the whole directory, this is the **only** such pair: `OperationConvention`/`OperationConventionResolver` are *both* `@internal`, and every other non-internal interface there returns non-internal types. The practical cost is not theoretical — PHPStan's internal-usage rules and PhpStorm's inspection both flag cross-package use of `@internal`, so a third-party plugin author following the documented migration gets a warning on every `PrimaryResponse::of()` call.

The docblock's prose is exactly right ("not a committed public contract") — it is the *tag* that overstates. I'd drop `@internal` and keep the prose; alternatively mark the interface internal too, but that is wrong since it is a documented extension point. Either is a one-line change; leaving the asymmetry is the option I'd argue against.

**2. (nit) Inconsistent null handling on adjacent lines.** `:241` uses `$autoPrimaryResult?->response`, `:242` uses `$autoPrimaryResult->statusIsExplicit ?? false`. The second is correct — `??` suppresses the null-property read — but reading the same nullable two different ways one line apart invites a reader to think the second is a *different* kind of absence. `$autoPrimaryResult?->statusIsExplicit ?? false` says the same thing in the file's own idiom.

**3. (nit) `docs/internal/architecture-review.md:41,304` still describes the sentinel as live.** `:304` lists it among gaps to remove and `:41` describes `OperationBuilder` as owning `EXPLICIT_STATUS_EXTENSION` — now factually wrong. It is a dated review artifact, so annotating "(resolved in #516)" is as defensible as leaving it; what I would not do is leave it unremarked, since it is the document a future architecture pass reads.

### General sweep

Nothing else. All 11 implementations wrap at every return path (PHPStan proves it — several have multiple return sites, e.g. `DataResponseResolver` threading the type through `resolveUnion()` and the private `response()` helper). `ResolverFaultBoundary::isolate()` is `@template T` / `@return null|T`, so the boundary needed no change and the callback's narrowed `?PrimaryResponse` type flows through. The `primaryResponseAnnotation()` test helper in `tests/Pest.php` is a reasonable way to keep ~200 lines of existing assertions untouched rather than rewriting them — the diff in those three test files is mostly mechanical unwrapping, which is what makes the behaviour-preservation claim legible. Naming, strict types, docblocks, and the 100-char width all conform.

Radiergummi:
🤖 **[reviewer]** review round 1: **approved** — 1 required correction (PR body only, no code change) and 1 minor note. I reviewed the plan, so this is scoped to "did it do what the plan said, and is the public surface right", not a general sweep.

## 1. (required, PR body) The PHPStan safety claim is false as implemented — and it was my amendment that put it there.

The body says (line ~110): *"What makes the deferral safe is PHPStan level 8: widening the property to nullable forces every reader to be handled explicitly, so #550 cannot silently miss one — the compiler enumerates the three sites for it."*

The implementation makes that untrue. `OperationBuilder` unwraps immediately after the resolver loop:

```php
$autoPrimaryResponse = $autoPrimaryResult?->response;
$autoStatusIsExplicit = $autoPrimaryResult->statusIsExplicit ?? false;
```

so all three downstream sites keep operating on a `?OA\Response` local that is **already** null-guarded — `:280` `… || $autoPrimaryResponse !== null`, `:284` the `??` fallback chain, `:290` `$autoPrimaryResponse !== null && is_array(…->content)`. Widen `PrimaryResponse::$response` to `?OA\Response` and **not one of them raises a PHPStan error**: a nullable value flowing into code that already handles null is not a violation, and `?->` on the assignment absorbs the rest. Concretely, with `suppressed()` in place and nothing else changed, `emitMissingResponseSchemaFinding(responseProduced: …)` would report a missing response schema for a `never` action and the `??` chain would still emit the synthetic `200` — #550 failing silently, which is precisely what the claim promises cannot happen.

Worth being clear about the general point, because I got it wrong in plan-review too: **no** code shape restores an analyser-forced check here, because the fallback path legitimately accepts `null`. "Suppressed" and "nothing resolved" are both `null` at the type level.

*Required:* replace the claim with the true one — the deferral is safe because the three sites are **enumerated by name in this PR body**, and #550's guarantee is a **test** asserting no synthetic 2xx is emitted for a suppressed result. Do not ship a stated safety property that verification contradicts; #550 will be planned against this paragraph.

*Recommended, not required:* keep `$autoPrimaryResult` at the three decision sites rather than introducing the `$autoPrimaryResponse` alias. It buys no analyser help, but it keeps the result object visible exactly where the decisions are made, so #550's author reaches for the result rather than for a local whose `null` is ambiguous again. The alias is otherwise a good call — it is what makes the diff byte-preserving (see below).

## 2. (minor) One line short on amendment 4.

`PrimaryResponse` matches `OperationConvention` exactly on form — `final readonly class`, promoted **public** properties with no redundant per-property `readonly`, and the qualified `@internal Part of the {@see PrimaryResponseResolver} seam; not a committed public contract.` marker. What is missing is the half-sentence on the **deliberate** divergence: `OperationConvention` has a public constructor, `PrimaryResponse` a private one plus `of()`. That divergence is load-bearing (it is what makes the third state additive); unexplained beside its sibling it reads as inconsistency.

---

## Verified, each against the diff rather than the report

**Amendment 2 — real, and better than I asked.** `docs/plugin-authoring.md:270-298` now `implements PrimaryResponseResolver`, declares `resolvePrimaryResponse(ActionDescriptor): ?PrimaryResponse`, and returns `PrimaryResponse::of($response)`. It also fixes a pre-existing bug I missed: `$target->isAmbiguous()` → `$target->isAmbiguous`, correct because `ResourceTarget::$isAmbiguous` is a property hook (`src/Routing/ResourceTarget.php:14`) — the old snippet would have fatal'd if copied. The registry table row (`:40`) and the interface bullet (`:77-83`) carry the return type and the `statusIsExplicit` rationale. `docs/architecture.md:44,117` are correctly left alone: both are structural mentions that never state a return type.

**Amendment 3 — matches the diff exactly.** Four test files, the single hand-rolled fake in `OperationBuilderFaultIsolationTest`, no registry/pipeline fake, and the vendor-extension pin kept (only its title reworded) at `OperationBuilderConventionStatusTest.php:132` — which is now the **sole remaining occurrence** of the sentinel string anywhere in `src`, `tests`, or `docs`.

**Behaviour preservation — holds, and nothing was excluded to make it hold.** The diff is 22 files: `src/`, four test files, two docs, `CHANGELOG`. **No snapshot or feature test file is touched**, so the byte-exact suite is unmodified rather than adjusted, and `Snapshot (locked toolchain)` passes alongside all six matrix legs, Coverage, Tooling and quality. The mechanism corroborates it: `OperationBuilder` retains the identical `$autoPrimaryResponse` `?OA\Response` local, so every downstream decision site is unchanged line-for-line.

**Two-state discipline — clean.** `of()` only, `public OA\Response $response` non-nullable, private constructor. No `suppressed()`, no nullability, nothing pre-built "for #550".

**`null` still means pass-to-next.** Now covered by an explicit new test ("falls through a resolver returning null to the next one"). The two new anonymous fakes narrow their return to non-nullable `PrimaryResponse` — legal covariance, and a nice incidental proof that the contract tolerates it.

**`ResolverFaultBoundary` genuinely untouched** — not in the diff. Only the closure's return type moved, as its `@template T` / `@return null|T` shape predicted.

**The coupling is deleted, not relocated.** `use …Support\Generator\OperationBuilder;` is gone from `InlineJsonResponseResolver`, and `git grep 'Support\Generator\OperationBuilder' -- src/Plugins` on this branch returns **zero** hits. `EXPLICIT_STATUS_EXTENSION`, `markStatusExplicit()` and `takeExplicitStatusMarker()` are all gone, and the now-orphaned `OpenApi\Generator` and `is_undefined` imports were correctly removed with them — including the `@phpstan-ignore` that the `Generator::UNDEFINED` dance needed.

**Tests got stronger, not weaker.** The `->x`-probing helper `inlineJsonExplicit()` is replaced by direct `$result->statusIsExplicit` assertions, and two genuinely new behaviours are pinned (null pass-through; a resolver's own `->x` surviving untouched — the guard for the deletion). The new `tests/Pest.php` helper is used in both resolver test files, not dead.

**One apparent test gap, checked and dismissed.** Nothing asserts `statusIsExplicit: true` directly on the `noContent()` site — the only one that hard-codes `true`. But `OperationBuilderConventionStatusTest.php:68-71` ("keeps a `->noContent()` 204 over the store convention (201)") fails if that argument is wrong, and `:103` covers the `responseFromCall` explicit path the same way. Behaviour is pinned end-to-end at the right level; no test to file.

Fix item 1 in the body (and item 2 if you agree) and this is ready for Moritz. The code itself needs no change.

Radiergummi:
🤖 **[coder]** addressed round 1: dropped the `@internal` tag, added the clobber-bug changelog clause, stated the `null` vs `suppressed()` distinction in the PR body, and closed both nits; pushed `f9ceaec5`.

**1. `@internal` tag dropped, prose kept.** Agreed on the reasoning, and the concrete cost settles it: PHPStan's internal-usage rule and PhpStorm's inspection both fire on cross-package `@internal`, so an author following our own documented one-call migration would get a warning on every `PrimaryResponse::of()`. The docblock now reads:

```
 * Part of the {@see PrimaryResponseResolver} seam; not a committed public contract while the
 * package is pre-1.0.
```

Same caveat, no tooling signal. (Worth noting for whoever writes the next seam VO: the `OperationConvention` form is still the right template for shape — public promoted properties, no per-property `readonly` — it just carries a tag that only fits because *its* interface is `@internal` too.)

**2. Changelog clause added for the latent bug.** `markStatusExplicit()` did `$response->x = [KEY => true]`, an unconditional whole-array overwrite, so an out-of-tree resolver setting its own vendor extensions alongside an explicit status lost them silently. The entry now says the extensions are preserved rather than only "no spec output change" — the bundled resolvers never set `->x`, which is why the spec is still byte-identical while the bug was real. The `leaves a resolvers own vendor extensions on the response untouched` test in `OperationBuilderFaultIsolationTest` is the pin.

**3. PR body now states the distinction**, in the "third state" section ahead of the mutual-exclusivity argument: `null` = *ask the next resolver*, chain continues, synthetic `200` fallback still fires if everyone declines; `suppressed()` = *I claim this action and it has no success response*, chain stops, no fallback. Framed as the reason the wrapper beats `?OA\Response` at all — the old type had exactly one empty value and so could only express the first — with an explicit note that #550 must add a case rather than overload `null`.

**Nits:**
- Unified the two reads one line apart on the same nullable: both now use `?->`. The `?? false` was forced by PHPStan's `nullsafe.neverNull` (it rejects `?->` on the left of `??`); `$autoPrimaryResult?->statusIsExplicit === true` satisfies it while keeping the idiom consistent.
- `docs/internal/architecture-review.md` updated at both sites: `:41` no longer claims `OperationBuilder` owns the sentinel (the marker now rides the `PrimaryResponse` result), and the over-abstraction bullet is struck through and marked done.

⚠️ **That last edit will not appear in the diff** — `docs/internal/` is gitignored at `.gitignore:9`, so the file is untracked and, being untracked, does not even exist in the PR worktree; I applied it to the copy in the main checkout at `/Users/moritz/Projects/laravel-openapi/docs/internal/architecture-review.md`. Same class of problem as the playbook, second instance, and per the lead it belongs on #563. I did not touch `.gitignore`.

**Gates re-run after all changes:** `composer test` 3492 passed (8555 assertions, snapshot group included), `vendor/bin/pint --test` passed, `composer lint` (PHPStan level 8) no errors.

Radiergummi:
🤖 **[lead]** review → survey. `f9ceaec5`, two reviewer approvals on the prior sha, all round-1 items addressed.

Note the PR is **already marked ready** — it was marked at `607679bb` and a push does not revert that.

**This PR is advertised as behaviour-preserving**, so the survey gate is unusually meaningful here: the sentinel was always stripped before serialization, and no bundled resolver sets `->x`, so the generated document must be **byte-identical**. Per the plan, **any** Layer A delta is a FAIL to investigate, not noise to absorb. This is the cleanest possible survey signal and I want it treated that way.

Second instance of the untracked-docs problem logged: `docs/internal/architecture-review.md` edits (`:41` corrected, over-abstraction bullet struck) are real but absent from the diff, same cause as the playbook (#563). Coder correctly did not touch `.gitignore`.

Radiergummi:
🤖 **[surveyor]** Layer A PASS — generated documents byte-identical across all 11 corpus apps.

Ran a direct byte-for-byte spec diff rather than a metrics comparison, because the change is advertised as strictly behaviour-preserving and that's the sharpest possible test.

**Baseline provenance:** freshly regenerated on current `main` @ `f0999f69` (#546). The cached `$WS/results.json` baseline was stale (`44124a70`, `main`'s parent), so it was **not** used — I regenerated baseline specs against `f0999f69`. PR branch is based exactly on `f0999f69` (merge-base == `main` HEAD), so the only delta under test is the three #516 commits. Neither checkout includes PR #556 (scorer fix, still in review); this gate compares generated specs directly, not `metrics.php`/`completeness.php` scores, so #556 is irrelevant to the result either way.

**Method:** for each app, symlink vendor → library-under-test, `openapi:generate --format=json`, `shasum -a 256`. Baseline = `main` `f0999f69`; candidate = PR `f9ceaec5`.

| app | baseline vs candidate |
|-----|-----------------------|
| AdvisingApp | identical |
| AureusERP | identical |
| Bagisto | identical |
| BookStack | identical |
| Coolify | identical |
| InvoiceNinja | identical* |
| Koel | identical |
| Lychee | identical |
| Pelican | identical |
| SpeedtestTracker | identical |
| Vito | identical |

All 11 `gen_exit=0`; candidate byte-length == baseline byte-length for every app.

\* **InvoiceNinja** raw diff showed 184 changed lines — **every one** an auto-generated `operationId` suffix (`api.generated_…`, `vendor.generated_…`, `client.generated_…`) on unnamed routes. Normalizing those random suffixes makes the two specs byte-identical (matching sha256). Proven to be run-to-run RNG, not induced by #516: two generations on the **same** (`main`) library differ in the identical 184 `*.generated_` lines and **zero** other lines. This is the known operationId-RNG masking artifact; no structural delta.

Confirms the refactor is behaviour-preserving: 0 changes to paths, schemas, operations, parameters, responses, or any content across the corpus. Clears the survey gate. (This PR still routes to Moritz for the `Contracts\**` sign-off — PASS clears it *to* escalation, not to merge.)

Radiergummi:
🤖 **[lead]** survey → needs-human (escalated for sign-off, NOT a problem).

This is the **planned** escalation for a `src/Contracts/**` change, not a failure. The PR is fully ready: rebased onto current `main` (`dfd83187`), **all CI green** across the full matrix incl. swagger-php 5.8, **two independent reviewer approvals**, and a **byte-identical survey PASS** across all 11 corpus apps (behaviour-preserving as advertised). Left ready and unmerged with its worktree in place.

**Why it needs a human:** it widens `Contracts\Registry\PrimaryResponseResolver` from `?OA\Response` to `?PrimaryResponse` — the documented plugin extension surface. Out-of-tree resolvers migrate with one call (`return PrimaryResponse::of($response)`); `null` still means pass-to-next.

**The argument to weigh first** (stronger than the issue body's 'remove a plumbing sentinel'): `InlineJsonResponseResolver` currently imports `Support\Generator\OperationBuilder` solely to read a constant — a bundled plugin reaching into generator internals to smuggle a value past a too-narrow contract. This deletes that coupling. The sentinel is the symptom; the coupling is the defect.

Two decisions embedded, both documented in the PR: (1) `suppressed()` is deferred to #550, its first consumer, so this ships a two-state VO; (2) `PrimaryResponse` is **not** tagged `@internal` (unlike its sibling `OperationConvention`), because it is mandatory to construct to implement a non-internal interface, and cross-package `@internal` use trips PHPStan/PhpStorm for plugin authors. Requesting Moritz's review.

PR #568, opened 2026-07-22
fix(survey): deterministic count-DESC/key-ASC tie-break in genuinelyMissingByShape rollup

### LINKED ISSUE #533: survey: explicit count-DESC / key-ASC tie-break in synthesize.php genuinelyMissingByShape rollup
Follow-up to #428 (three-way coverage metric), non-blocking finding **F2** deferred at merge.

The `genuinelyMissingByShape` rollup in `tools/survey/synthesize.php` (was ~line 756 on the #428 branch) sorts without an explicit total tie-break, so shapes with equal counts can order non-deterministically across runs. Add an explicit **count-DESC, then key-ASC** tie-break, matching the ordering the lint rollup already uses, so the rendered survey candidate is byte-stable.

Harness-only (`area:survey`); no library code. Scope is the sort comparator only.

Radiergummi:
## What

The `genuinelyMissingByShape` corpus rollup in `tools/survey/synthesize.php` sorted with `arsort`, whose equal-count ordering is input-dependent. Shapes tied on count could therefore order non-deterministically across runs, so the rendered survey candidate was not byte-stable.

## Change

One comparator. Replaced `arsort($coverageByShape)` with an explicit total order — **count DESC, then shape ASC** — mirroring the `survey_rankByRule` lint-rollup comparator idiom already in the same file:

```php
uksort($coverageByShape, static function (int|string $a, int|string $b) use ($coverageByShape): int {
    return $coverageByShape[$b] <=> $coverageByShape[$a] ?: strcmp((string) $a, (string) $b);
});
```

Scope is the sort comparator only — no library code, no other harness changes.

## Tests

Added a dedicated case pinning the deterministic order: two shapes tied at count 1 break on key ASC, a third leads on count DESC, and the result is byte-identical regardless of the order the apps' shape maps are fed in. The existing byte-for-byte golden-snapshot test still passes (bytes unchanged).

Local gates green: `composer test` (3504 passed), `vendor/bin/pint --test`, `composer lint`.

Closes #533


Radiergummi:
🤖 **[coder]** implemented; replaced `arsort($coverageByShape)` in the `genuinelyMissingByShape` rollup with an explicit total order (count DESC, then shape ASC via `uksort`), mirroring the `survey_rankByRule` comparator idiom in the same file; added a test pinning the deterministic order (tie breaks key-ASC, byte-identical across input orders); the existing golden-snapshot test still passes. Pushed 6d143b2f; local gates green (composer test 3504 passed, pint --test, composer lint). Ready for review.

Radiergummi:
🤖 **[lead]** Fast-path → review. #533 tie-break implemented (`6d143b2f`): `arsort` → explicit count-DESC/key-ASC `uksort` in the genuinelyMissingByShape rollup, mirroring `survey_rankByRule`; tie-break test added, golden snapshot still passes. Two independent reviewer passes (reviewer1 + reviewer2) for the auto-merge gate — harness-only, non-generation-affecting, no survey.

Radiergummi:
🤖 **[coder]** removed a stray 0-byte `factory-` file that was accidentally swept in during the recovery force-push; amended the commit. Pushed 3b70689d. Diff is now exactly CHANGELOG.md + tests/Tools/SurveySynthesizeTest.php + tools/survey/synthesize.php. The three gated files are byte-identical to the previously-green commit (only `factory-` was dropped), so gates remain green.

Radiergummi:
🤖 **[reviewer]** review: **approved, no concerns.**

Fast-path confirmation of the 4 points on a harness-only, 3-file change (no `src/`):

1. **Correct, stable total order.** `uksort` with `$coverageByShape[$b] <=> $coverageByShape[$a] ?: strcmp((string) $a, (string) $b)` is count DESC then key ASC. `<=>` binds tighter than `?:`, so the count comparison decides first and the strcmp only runs on an equal-count tie — same construct as `survey_rankByRule`. Since array keys are unique, `strcmp` on two distinct keys is never 0, so the order is genuinely total: nothing is left to PHP's unstable sort. Reading counts from the by-value `use ($coverageByShape)` capture is safe (uksort reorders keys, never mutates values).
2. **Genuinely mirrors `survey_rankByRule`.** Same ordering semantics (count DESC, then string-key ASC via `strcmp`). It uses `uksort` on the map rather than `usort` on a list-of-dicts because `genuinelyMissingByShape` is consumed as a `shape => count` map (the golden snapshot depends on that shape); the comparison logic is the mirror, correctly.
3. **The test pins the tie-break and is not tautological.** `alpha`/`zebra` tie at 1 (expected `alpha` before `zebra` = key ASC), `multiple returns` leads at 3; both apps' maps are fed in both orders and asserted `toBe($expected)`. `toBe` is `assertSame` → order-sensitive `===` on the associative array, so it pins ordering, not just membership. I verified it **fails under the old `arsort`** (reverted locally and reran: the tie-break case errors at the `toBe`), confirming it distinguishes the new behaviour from both arsort and insertion order.
4. **Nothing out of scope.** Exactly `tools/survey/synthesize.php`, `tests/Tools/SurveySynthesizeTest.php` (incl. a one-line comment refresh), `CHANGELOG.md`. No `src/`.

Full `SurveySynthesizeTest` run: **18 passed / 75 assertions**, golden snapshot still byte-identical.

This is one of the two independent approvals for auto-merge.

Radiergummi:

Independent second review (not concurring with reviewer1 — verified from scratch).

1. **Total order is correct & complete.** `$coverageByShape[$b] <=> $coverageByShape[$a] ?: strcmp((string)$a,(string)$b)` — `<=>` binds tighter than `?:`, so it is `(count DESC) ?: (key ASC)`: when counts differ the spaceship is truthy and short-circuits; when equal it falls to `strcmp` on the distinct array keys. Keys are unique, so no residual ties are left to PHP's sort. `uksort` captures `$coverageByShape` by value, so count lookups stay stable while keys are reordered.
2. **Faithfully mirrors `survey_rankByRule`** (`tools/survey/synthesize.php:82`) — same `$b <=> $a ?: strcmp($a,$b)` idiom. The shape difference (`uksort`-on-map here vs `usort`-on-list there) is correct: this rollup must stay a `shape => count` map, whereas `byRule` emits a list of records.
3. **Test genuinely pins the tie-break, not tautological.** Feeding the two apps in both corpus orders and asserting byte-identical output would fail under the old `arsort`: verified empirically that `arsort` yields zebra-before-alpha for one order and alpha-before-zebra for the other (PHP 8 stable sort preserves insertion order for equal counts), while the new comparator yields `{multiple returns:3, alpha:1, zebra:1}` for both. The equal-count pair (alpha/zebra at 1) and the distinct leader (multiple returns at 3) exercise both comparator branches.
4. **Scope is exactly the 3 stated files** (`tools/survey/synthesize.php`, `tests/Tools/SurveySynthesizeTest.php`, `CHANGELOG.md`); no `src/` change, harness-only. CHANGELOG `[Unreleased]` entry present and accurate.

Comment explains the *why* (arsort input-dependence), carries no issue number in code, matches surrounding style. Remote CI green. Proportionate fix, correctly done.

Radiergummi:
🤖 **[lead]** all gates green → **squash-merged**. Two independent reviewer approvals (reviewer1 + reviewer2, both verified the tie-break test fails under the old `arsort`), remote CI green across the full matrix, non-controversial (harness-only, 0 src/ lines, no Contracts/config/signature). Closes #533. Branch + worktree removed.

PR #558, opened 2026-07-22
fix(responses): scan the whole method for the return, not the first 10 statements

Radiergummi:
Closes #555

## Revision log — round 2 (guard mechanism)

| # | Finding | Resolution |
|---|---|---|
| 1 | **(blocking)** Guard 2's predicate must be wider than `resolveVariable():74-77` | **Predicate spelled out node-kind by node-kind**, with a test per shape; it is the *correct* predicate, and the narrow one is the bug (**#561**). |
| 2 | **(blocking)** `$resolutionReturnIndex` under-specified for nested returns; unspecified default is the unsafe one | Defined as the index of the **containing top-level statement**, with an explicit **fail-closed** default: no index ⇒ **empty** dominating region. |
| 3 | State fields must be written and reset together, before the docblock path | Both initialised at the top of `resolve()` (ahead of `targetFromReturnTag()` at `:195-199`) and assigned in the same statement at every write site. |
| 4 | The new ambiguity note breaks the silent-probe contract | The ambiguity refusal is **silent**; the existing outer note still fires once per action on the non-silent path. |
| 5 | *(round 3, lead)* The rebinding table must be consumable by #561, not re-derived | Extracted to a shared `Support\MethodBody\VariableRebinding` helper with its own test, rather than an inline predicate. Deliberately a slightly wider diff. |
| 6 | *(approval conditions)* Off-by-one in guard 2's region; helper naming/scope/citation | Region pinned to **`[assertIndex+1 … containingReturnIndex]` inclusive** with a boundary test that discriminates. Name stays `VariableRebinding`; `resolveVariable()` behaviour untouched (**#561** owns it); the narrow predicate reframed as the bug, not the baseline; #559 → #561 citation fixed. |

Round 1's resolutions stand. The reviewer declined to veto the rejection of the
single-unconditional-assignment guard and strengthened it: that guard uses the *same*
`Assign`-to-`Variable` predicate, so `foreach ($others as $subject)` defeats it identically — it is
not merely costly but **ineffective** against the residual class, while breaking a multi-branch idiom
that resolves correctly today.

---

## Tier

**Tier 0, blocked by a Tier-1 bound.** The resolution itself (a `Category`-typed parameter as the
`->toResource()` receiver) is pure reflection over a parameter's declared type; the only body work is
*locating* the method's single return. Nothing here adds dataflow.

## Premise correction (retained; #555 amended to match)

All three sites are the same scan — return discovery: `:201` `resolve()`, `:377`
[...]

Issue #529, opened 2026-07-02
Header/cookie parameter dedup is case-sensitive


Radiergummi:
## Summary

Request header and cookie parameters are deduplicated **by exact name string**, so two spellings of the same header/cookie that differ only in case emit as **two separate parameters** in the operation, e.g. an inferred `X-Api-Key` read plus a `#[Header('x-api-key')]` attribute:

```php
#[Header('x-api-key')]
public function show(Request $request): JsonResponse
{
    $key = $request->header('X-Api-Key'); // inferred header: X-Api-Key
    // …
}
// → emits BOTH `X-Api-Key` (inferred) and `x-api-key` (attribute) as header params
```

HTTP header names are case-insensitive (RFC 9110 §5.1), so these are the same header and should collapse to one parameter (the attribute winning, per the `#[Header]`/`#[CookieParam]` precedence lock). Cookie names are case-*sensitive* per RFC 6265, so this specifically concerns **headers**; the cookie path is included only for symmetry review.

## Scope / status

- **Pre-existing, not a regression.** The attribute-only header path already deduped by exact name (`RequestParameterApplier::headerParameters()` keys `$byName[$instance->name]`); #527 (cookie/header accessor inference) only made it *reachable* by a second source (an inferred read + an attribute), which is how it surfaced during review.
- **Latent / author-triggered:** it only manifests when the author writes a header attribute whose case differs from the code's `->header('…')` literal (or two attributes differing only in case). No corpus app is known to hit it.
- Surfaced by reviewer-2 during the #527 review; explicitly **not** folded into #527 (kept surgical).

## Likely fix

Normalize the dedup key for header parameters to a case-fold (e.g. lowercased name) in the `(name, in)` fold in `OperationBuilder::build()` and/or `RequestParameterApplier`, while preserving the author's original casing on the emitted `name`. Decide whether cookies stay case-sensitive (per RFC 6265) — probably yes. Add tests: inferred `X-Api-Key` + `#[Header('x-api-key')]` → one header param (attribute wins, attribute's casing).

Radiergummi:
🤖 **[lead]** admitted → planning. Independent bug (header/cookie dedup case-sensitivity) on already-merged param code; safe to drive to ready while Moritz is away.

Issue #488, opened 2026-07-01
feat(params): generalize the request-accessor reader to cookie() and header()


Radiergummi:
## Problem

`RequestQueryAccessorReader` proves the `$request->accessor('x')` Tier-1 whitelist shape works — but only for query parameters. `$request->cookie('x')` and `$request->header('x')` are the **identical call shape at the identical location**, yet cookies and headers are 100% attribute-driven (`#[CookieParam]`, `#[Header]`) purely because the reader was never parameterized over accessor kind. The asymmetry: query params are inference-first with the attribute as precedence tier 4; cookies/headers have no inference at all.

## Proposed change

Generalize the accessor reader over `(method name → parameter location)`: `query`/`input` → query (as today), `cookie` → cookie, `header` → header. Same bounded scan, same literal-name-only whitelist, same graceful degradation. Inferred cookie/header params slot into the existing `(name, in)` dedup in `OperationBuilder::build`, with the explicit attributes keeping authoring precedence exactly as `#[QueryParam]` does.

## Why

Proven inference generalized across sibling accessor kinds — minimal new surface, kills the most mechanical uses of two attributes. Tier 1, in-charter, cheap.

Radiergummi:
🤖 **[lead]** admitted → planning. tier-1 area:params, generation-affecting (survey-gated). The proven substrate is `RequestQueryAccessorReader` (the `$request->accessor('x')` Tier-1 whitelist). Generalize it over (method name → parameter location): query/input→query (as today), cookie→cookie, header→header — same bounded scan, same literal-name-only whitelist, same graceful degradation. Inferred cookie/header params slot into the existing (name,in) dedup in OperationBuilder::build; explicit #[CookieParam]/#[Header] keep authoring precedence exactly as #[QueryParam] does. Ground against current main; name files/tests; surface any Contracts/config touch (none expected) for the human-gate call.

Issue #543, opened 2026-07-22
feat(responses): resolve ->toResource() receiver via an identity @return T generic on the passthrough callee


Radiergummi:
## Summary

Resolve a `->toResource()` receiver whose declared **generic** return type is an identity template parameter bound to an argument — e.g. a method `resolve<T of Model>(T $x): T`. Today `ReturnExpressionResourceReader` only recognises a passthrough when the callee's body has a single identity `return $x;`; a method with the same *declared* identity generic but a multi-branch body is refused, even though the declared return type alone fully specifies the result.

## Motivation (dogfooding: a private consumer app, run 2)

The single dominant response gap in the app is this one method, hit **30 times**:

```php
// controller
public function single(GetClaimedDomainRequest $request, ClaimedDomain $claimedDomain): JsonResource
{
    return $request->resolve($claimedDomain)->toResource();
}

// App\Http\Requests\Support\ResourceRequest
public function resolve(Model $instance): Model
{
    if (!$this->hasAnyInclude() && $this->alwaysIncludesFromAttributes() === []) {
        return $instance;                                   // identity path
    }
    return $this->getQueryBuilder($instance)->firstOrFail(); // query-builder path
}
```

`calleeReturnedParameterIndex()` requires exactly one identity return statement, so the second (query-builder) branch makes it — correctly, for a plain `: Model` signature — refuse. Borrowing the argument's type off a bare `: Model` signature would be unsound (that was the #541 review-fix).

But the honest, sound way to express "this returns exactly what you pass in" is a generic identity annotation:

```php
/**
 * @template T of \Illuminate\Database\Eloquent\Model
 * @param T $instance
 * @return T
 */
public function resolve(Model $instance): Model
```

With that, the **declared return type is `T` = the argument's type**, regardless of how many branches the body has. No body proof is needed or wanted — the generic is a Tier-0 declared-type signal.

## Proposed behaviour

In `ReturnExpressionResourceReader::methodCallModelClass()`, before the single-return body proof (`calleeReturnedParameterIndex`):

1. Read the callee's `@template`/`@param T`/`@return T` via the existing `DocBlockParser` + `TypeNodeResolver` stack.
2. If the declared return is a bare template parameter `T` that is bound to exactly one parameter's type (identity generic), resolve the corresponding call-site **argument**'s model class (reusing `receiverModelClass()` on the argument, exactly as the current passthrough does) and return that.
[...]

PR #570, opened 2026-07-22
feat(params): reserved-header denylist for the request-accessor reader

### LINKED ISSUE #530: feat(params): reserved-header denylist for the request-accessor reader
Surfaced by the #527 (#488) survey Layer A run on the consumer corpus.

## Observation
Now that `$request->header('x')` is inferred as an `in:header` parameter (#488, PR #527), the reader surfaces some **reserved / protocol headers** as documented request params where that is arguably noise rather than an API contract:

- **`Content-Type` / `Content-Length`** — reading these in a controller is typically a control-flow check, not a client-supplied API parameter (seen on AureusERP / Pelican / SpeedtestTracker).
- **`Authorization`** — this is **security-scheme territory** (documented via the security requirement, not as a free-form header param). Seen on Coolify.

App-legitimate custom headers (`X-Hub-Signature-256`, `X-API-TOKEN`, `X-API-COMPANY-KEY`, `Stripe-Signature`, etc.) are correctly surfaced and should stay.

## Proposed change
A small, bounded **reserved-header denylist** applied to inferred (not attribute-authored) header params: skip inferring a header param when the literal name matches a reserved set (`Authorization`, `Content-Type`, `Content-Length`, and likely the other RFC 9110 representation/negotiation/control headers — `Accept`, `Content-*`, `Host`, etc.). Case-insensitive match (see also #529). Attribute-authored `#[Header]` stays authoritative and is never filtered (author's explicit intent).

Tier-1: a static name-list check on the already-inferred name; no new scanning. `area:params`. Deferred from #527 — the base generalization is correct as-is; this is a quality refinement to reduce protocol-header noise.

Radiergummi:
Closes #530

## Problem

Since #488/#527, `$request->header('X')` is inferred as an `in: header` request parameter. That
correctly surfaces app-legitimate custom headers (`X-Hub-Signature-256`, `X-API-TOKEN`,
`Stripe-Signature`, …) but also surfaces **reserved / protocol headers** as if they were free-form
API parameters, which is noise, not an API contract:

- `Authorization` — security-scheme territory, documented via the security requirement (Coolify).
- `Content-Type` / `Content-Length` — control-flow checks, not client parameters (AureusERP, Pelican, SpeedtestTracker).
- The rest of the RFC 9110 negotiation / representation / conditional / control headers.

## The read (bounded, Tier-1)

A static, case-insensitive name check against a fixed reserved set, applied to an **already-inferred**
header name. No new scanning — it filters a read the reader already produced. Reserved header reads
are simply not emitted.

**Inferred-only by construction.** `#[Header]` / `#[CookieParam]` attributes never pass through
`RequestAccessorReader` — they are applied separately by `RequestParameterApplier`. Filtering inside
the reader therefore touches only inferred reads and can never drop an author's explicit `#[Header]`,
satisfying the "attribute-authored stays authoritative, never filtered" requirement automatically.

**Case handling** mirrors the #529 fix (closed): HTTP header names are case-insensitive (RFC 9110
§5.1). The reserved set is stored lowercased; the read name is compared via `strtolower()`. The
surfaced parameter keeps its original literal casing (only the comparison is folded). Cookies stay
case-sensitive (RFC 6265) and are **not** affected — the denylist is header-location only.

## Home

`src/Plugins/Core/Support/RequestAccessorReader.php` (the reader #488/#527 added). In the `read()`
loop, after `$location` and `$name` are computed and before the `AccessorRead` is built:

```php
if ($location === 'header' && $this->isReservedHeader($name)) {
    continue; // reserved/protocol header: not an API parameter
}
```

Reserved reads are skipped silently — this is intentional filtering, not an unreadable-accessor
failure, so they are **not** added to `$unreadableAccessors` and produce no note.

### Constant, not config

Per the flat-config preference and the issue's "bounded static name-list", the reserved set is a
**class constant** (`private const array RESERVED_HEADERS`), not a config key. Reserved headers are
protocol-defined and stable; there is no per-app variation worth a config surface. The escape hatch
for the rare app that genuinely wants to document a reserved header as a parameter already exists and
is authoritative: an explicit `#[Header('Authorization')]` is never filtered.

### The reserved set (design decision — reviewer confirm)

Guiding invariant, proposed: **the denylist contains no `X-` header.** Every `X-*` header therefore
always surfaces, which keeps every app-custom header (the legitimate case the issue calls out) safe
by construction; a reviewer never has to reason about whether a given `X-*` was accidentally caught.

Reserved set (all lowercased), RFC 9110 + auth + framework transport:

- **Auth:** `authorization`, `proxy-authorization`
- **Representation (request bodies):** `content-type`, `content-length`, `content-encoding`,
  `content-language`, `content-location`, `content-md5`, `content-range`
- **Negotiation:** `accept`, `accept-charset`, `accept-encoding`, `accept-language`
- **Conditionals:** `if-match`, `if-none-match`, `if-modified-since`, `if-unmodified-since`, `if-range`
- **Controls / connection / transport:** `host`, `connection`, `cache-control`, `pragma`, `expect`,
  `range`, `te`, `trailer`, `transfer-encoding`, `upgrade`, `via`, `max-forwards`, `forwarded`,
  `date`, `cookie`
- **Client context (protocol, not API params):** `user-agent`, `referer`, `origin`, `from`

Deliberately **excluded** (kept surfaced) so the invariant holds and app headers survive:
`x-requested-with`, `x-forwarded-for`, `x-forwarded-proto`, `x-forwarded-host` and all other `X-*`.
If the reviewer prefers to catch the well-known `X-Forwarded-*` / `X-Requested-With` proxy headers,
that is a small, explicit addition — flagged, not assumed.

## Files touched

- `src/Plugins/Core/Support/RequestAccessorReader.php` — add `RESERVED_HEADERS` constant + `isReservedHeader()`; the one guard in the `read()` loop.
- `tests/Fixtures/QueryAccessorFixtureController.php` — fixture actions reading reserved + custom headers, and a `#[Header]`-authored reserved name.
- `tests/Feature/QueryAccessorParametersTest.php` — the end-to-end + attribute-interaction assertions.
- (Optionally `tests/Unit/Core/Resolvers/CoreQueryParameterResolverTest.php` if a unit-level assertion reads cleaner for the pure filter cases.)
- `docs/auto-derivation.md` — one sentence in the "Request parameters from the method body" section.
- `CHANGELOG.md` — `[Unreleased] → Changed`.

## TDD test list (failing test first)

Positive filtering:
1. `$request->header('Authorization')` → **no** header parameter emitted.
2. `$request->header('Content-Type')` and `$request->header('Content-Length')` → not emitted.
3. A negotiation/conditional/control representative (`Accept`, `If-None-Match`, `Host`) → not emitted.

Case-insensitivity (the #529 precedent):
4. `$request->header('content-type')`, `'CONTENT-TYPE'`, `'Content-Type'` → all filtered (folded match).

Attribute stays authoritative (never filtered):
5. `#[Header('Authorization')]` on an action → the `Authorization` header parameter **is** emitted
   (attribute wins; the denylist does not touch the attribute path). Include a variant where the
   action also has `$request->header('Authorization')` — one `Authorization` param, from the attribute.

Custom headers still surface (no over-filtering):
6. `$request->header('X-Api-Key')`, `$request->header('Stripe-Signature')`,
   `$request->header('X-Forwarded-For')` → each still emitted (the `X-` invariant + non-reserved names).

Non-regression:
7. Query / cookie accessor inference is unchanged (the guard is header-location-only): a
   `$request->cookie('Content-Type')` style read on the cookie location is **not** filtered.

## Docs & CHANGELOG

- `docs/auto-derivation.md`, "Request parameters from the method body": add a sentence — reserved /
  protocol header names (`Authorization`, `Content-*`, `Accept*`, `Host`, …) are not inferred as
  parameters; document one explicitly with `#[Header]` if it is genuinely part of the API contract.
- `CHANGELOG.md` `[Unreleased] → Changed` (own line, end of the Changed list):
  > Inferred `in: header` parameters now skip reserved / protocol header names (`Authorization`,
  > `Content-Type`, `Content-Length`, the RFC 9110 negotiation / representation / conditional /
  > control headers, `Host`, `Cookie`, `User-Agent`, …), matched case-insensitively, to cut
  > protocol-header noise from `$request->header(...)` reads. Custom (`X-*`) headers are unaffected,
  > and an explicit `#[Header]` is never filtered. (#530)

## Independence & merge-risk

- **Independent.** Sits in `RequestAccessorReader` (request-param inference), untouched by the #558
  (response-scan) and #560 (`PrimaryResponse`) hot zones. No gate found.
- **Low-risk, additive filter.** Generation-affecting only in the narrow sense of *removing* reserved
  header params from ~a handful of corpus operations (AureusERP / Pelican / SpeedtestTracker /
  Coolify). No public contract change; a class constant, not config.
- **One controversy to surface for review:** the exact membership of the reserved set (the `X-`
  invariant and the excluded proxy headers). Everything else is mechanical.
- Snapshot group: refresh any snapshot app whose spec legitimately loses a reserved header param.


Radiergummi:

Radiergummi:
🤖 **[lead]** planning → plan-review. #530 reserved-header denylist, home `RequestAccessorReader::read()` (inferred-only by construction — `#[Header]` never routes through this reader, so authored headers can't be filtered). Constant not config; case-insensitive (mirrors #529); header-location-only guard. → reviewer for plan-review. **Key decision to settle:** the exact reserved set — planner proposes the invariant *no `X-` header is ever denylisted* (so `X-API-TOKEN`/`Stripe-Signature`/`X-Forwarded-For` always surface); open sub-question is whether `X-Forwarded-*`/`X-Requested-With` should be exceptions. Also sanity-check the RFC 9110 set isn't over-broad (must not drop app-legit custom headers).

Radiergummi:
🤖 **[reviewer]** plan-review: **approved** — sound, correctly scoped, Tier-appropriate. Ruling on the one open decision below.

### 1. Inferred-only guarantee — airtight, confirmed
Verified in the source: `#[Header]` params are built by `RequestParameterApplier::headerParameters()` from `descriptor->controllerAttributes()/actionAttributes()` (pure reflection over attributes), while `RequestAccessorReader::read()` only scans method-body `$request->header(...)` calls. They are structurally distinct sources — an attribute is never a method-body call, so it can never enter the reader. A denylist inside `read()` therefore cannot drop an authored `#[Header]` **by construction**; no extra inferred-only guard is needed. The plan's claim holds as written. The guard placement is right (after `$location`/`$name`, after the `$name === null` continue, before `new AccessorRead`).

### 2. Reserved set — RULING
- **Adopt the `X-` invariant as proposed.** "No `X-` header is ever denylisted" is the right bright line. The entire motivating corpus case is custom headers (`X-Hub-Signature-256`, `X-API-TOKEN`, `X-API-COMPANY-KEY`, `Stripe-Signature`), and the risk asymmetry favours it: a reserved header leaking through as one noisy param is cheap; silently dropping a real API-contract header is expensive and confusing. The invariant makes the filter trivially safe to reason about.
- **Do NOT add `X-Forwarded-*` / `X-Requested-With` exceptions.** Keep the bright line pure. The value of the invariant is that it requires zero case-by-case reasoning; punching a hole for proxy headers reintroduces exactly the "is my `X-` header caught?" question it eliminates. There is **no corpus evidence** these recur as noise (the issue's offenders are all `Content-*`/`Authorization`), and `X-Forwarded-For` is at least arguably documentable. Per the repo's measure-first ethos: leave them surfaced; revisit only if a future survey shows them recurring as spurious params (a one-line addition, as the plan notes).
- **RFC 9110 set is not over-broad.** Walked every entry: auth (`authorization`, `proxy-authorization`), representation (`content-*`), negotiation (`accept*`), conditionals (`if-*`), controls/transport (`host`, `connection`, `cache-control`, `pragma`, `expect`, `range`, `te`, `trailer`, `transfer-encoding`, `upgrade`, `via`, `max-forwards`, `forwarded`, `date`, `cookie`), client context (`user-agent`, `referer`, `origin`, `from`). Every one is a genuine protocol/negotiation/auth/transport header; none is a plausible API-contract parameter. Borderline names (`origin`, `from`, `date`) are browser/protocol-set, correctly filtered. `cookie` is correctly caught here (cookie params are their own location). No entry needs removing.

### 3. Case-insensitivity & location scope — correct
Storing the set lowercased and comparing via `strtolower($name)` while emitting the original casing mirrors #529 and matches the existing `RequestParameterApplier` header dedup (`strtolower($instance->name)`). The `$location === 'header'` guard keeps cookies (case-sensitive, RFC 6265) and query untouched; test 7 pins that. Good.

### 4. Tier — appropriate
This is a static name-list check on an already-inferred name, no new AST scanning — at or below the Tier-1 bar. Fine.

### Test list
Covers the surface well: positive filtering (auth/representation/negotiation/conditional/control reps), case-folding, attribute-stays-authoritative (incl. the both-present variant → one param from the attribute), custom-`X-` survival, and the cookie-location non-regression. No gaps to add. One note for the coder: the both-present variant in test 5 passes trivially because the inferred `Authorization` is filtered out entirely (only the attribute param remains) — that is the correct outcome, just be aware it does not exercise reader/attribute *precedence* (out of scope here).

Approved to move to coding with the ruling above (adopt the `X-` invariant; no proxy-header exceptions).

Radiergummi:
🤖 **[lead]** plan-review → coding. reviewer1 approved with rulings: adopt the **`X-` invariant** (no `X-` header ever denylisted); **no `X-Forwarded-*`/`X-Requested-With` exceptions** (keep the line pure; revisit only on future survey evidence); RFC 9110 set verified not over-broad; inferred-only guarantee airtight by construction. → coder1.

Radiergummi:
🤖 **[coder]** implemented; local gates green (`composer test` 3513 passed, `vendor/bin/pint --test` clean, `composer lint` 0 errors); rebased onto `origin/main` (ab4b19e8, #568) and pushed `092b42f7`.

**Implementation (per the approved plan + reviewer1's rulings):**
- `RequestAccessorReader`: added a `private const array RESERVED_HEADERS` (lowercased) + `isReservedHeader()`, and one guard in the `read()` loop — `if ($location === 'header' && $this->isReservedHeader($name)) { continue; }`. Reserved reads are skipped **silently** (not recorded as unreadable, no note). Case folded via `strtolower()`; the emitted param keeps its original casing.
- **`X-` invariant honoured:** the set contains **no** `X-*` header — no `X-Forwarded-*`/`X-Requested-With` exceptions added, so every app-custom header still surfaces by construction.
- **Header-location only:** the guard is gated on `$location === 'header'`; cookies/query untouched.
- Constant, not config (protocol-defined, stable). The escape hatch is the existing `#[Header]`, which never routes through this reader.

**Reserved set** (RFC 9110 + auth + transport, all lowercased): `authorization`, `proxy-authorization`; `content-type/-length/-encoding/-language/-location/-md5/-range`; `accept`, `accept-charset/-encoding/-language`; `if-match`, `if-none-match`, `if-modified-since`, `if-unmodified-since`, `if-range`; `host`, `connection`, `cache-control`, `pragma`, `expect`, `range`, `te`, `trailer`, `transfer-encoding`, `upgrade`, `via`, `max-forwards`, `forwarded`, `date`, `cookie`; `user-agent`, `referer`, `origin`, `from`.

**Tests (TDD, unit + feature):** reserved names filtered when inferred; case-insensitivity (`content-type`/`CONTENT-TYPE`/`Content-Type` all filtered); `#[Header('Authorization')]` + a same-name `$request->header('Authorization')` → exactly **one** `Authorization` header param (from the attribute); `X-Api-Key`/`Stripe-Signature`/`X-Forwarded-For` still surface; a reserved name on the **cookie** location is not filtered. Full suite green (no snapshot churn).

Docs (`docs/auto-derivation.md` boundaries list) + CHANGELOG (`[Unreleased] → Changed`, end of section) updated. `src/` diff is 38 lines (constant + guard + one method) — no Contracts/config-key/signature change. Ready for review.

Radiergummi:
🤖 **[lead]** coding → review. #530 implemented `092b42f7` (rebased onto post-#568 main): `RESERVED_HEADERS` constant + `isReservedHeader()` + one guard in `RequestAccessorReader::read()`; X- invariant honoured, case-insensitive, header-location-only, `#[Header]` never filtered. Local gates green (3513 passed). src/ +38, no Contracts/config/signature → auto-merge candidate if survey passes. Two independent reviewer passes (reviewer1 + reviewer2) → then **survey gate** (generation-affecting).

Radiergummi:

Independent second review (I did not plan-review this — reviewed from scratch).

1. **Guard correctness.** The skip sits in `read()` after the `$name` literal-extraction + null-check and before `AccessorRead` construction (`RequestAccessorReader.php`, `+166`), gated on `$location === 'header' && isReservedHeader($name)`. So cookies/query/input are untouched (query reserved names still surface — proven indirectly, and the cookie-location test pins the gate). `isReservedHeader()` lowercases only for the `in_array(..., true)` compare; the `$name` carried into `AccessorRead` keeps original casing — the `X-Api-Key`/`Stripe-Signature`/`X-Forwarded-For` test asserts the surfaced keys retain source casing. The skip is a bare `continue` that does **not** push onto `$unreadableAccessors`, so it correctly earns no generation-log note (intentional filtering, not an unreadable shape).
2. **`X-` invariant.** No `X-*` entry in `RESERVED_HEADERS`; the RFC-standardized `Forwarded` is filtered while the de-facto `X-Forwarded-For` is not — a consistent standardized-vs-custom split, and the test pins `X-Forwarded-For` (plus non-`X-` `Stripe-Signature`) surfacing. No `X-Requested-With`/`X-Forwarded-*` special-casing.
3. **`#[Header]` authoritative.** The reader only sees method-body `$request->header()` reads; attributes never route through it. The `authorizationHeaderAttribute` test asserts exactly one `header:Authorization` param, sourced from the attribute (`description: 'Bearer token.'`) — the inferred read of the same name is filtered, so no duplicate. Correct.
4. **Set not over-broad.** RFC 9110 auth/representation/negotiation/conditional/control/transport + client-context headers only — all protocol plumbing. Borderline entries (`Accept` for media-type versioning, `Range` for range-pagination) are defensible to filter because the `#[Header]` escape hatch always re-surfaces them, consistent with the tier philosophy (infer the common case, attribute for the exception). Nothing in the set is plausibly a free-form API-contract header.
5. **Scope.** Exactly 6 files: `src/Plugins/Core/Support/RequestAccessorReader.php` + 2 tests + 1 fixture + `docs/auto-derivation.md` + `CHANGELOG.md`. No `src/Contracts`, no config key, no public-signature change; `RESERVED_HEADERS`/`isReservedHeader()` both private, constant not config.

Conventions clean: strict-types intact, `use function strtolower;` matches the file's function-import style, docblocks concise and on the *why*, no abbreviations, no issue number in code. Tests are dual-layer (unit resolver + feature generate path) and would fail without the guard. CHANGELOG + docs both updated. CI green across the matrix incl. the 5.8 job and snapshot. Second approval — clear to merge-gate.

Radiergummi:
🤖 **[reviewer]** review round 1: **approved, no concerns.** Implementation honours every plan-review ruling; verified the diff and ran the suite (49 passed / 131 assertions in `QueryAccessorParametersTest` + `CoreQueryParameterResolverTest`).

1. **Inferred-only — confirmed.** The guard `if ($location === 'header' && $this->isReservedHeader($name)) continue;` sits in `RequestAccessorReader::read()` (the body-scan path); `#[Header]` params are built independently by `RequestParameterApplier` from descriptor attributes and never enter this reader. The `authorizationHeaderAttribute` fixture carries **both** `#[Header('Authorization', description: 'Bearer token.')]` and `$request->header('Authorization')`, and the test asserts exactly **one** `Authorization` header param whose `description` is `Bearer token.` — proving the authored header survives and the inferred read is filtered.
2. **`X-` invariant — confirmed.** `RESERVED_HEADERS` contains no `X-*` entry; `X-Api-Key`, `Stripe-Signature`, and `X-Forwarded-For` all still surface (test passes). No `X-Forwarded-*` / `X-Requested-With` exception was added, per the ruling.
3. **Reserved set — matches what I approved, item for item** (auth, `content-*` representation, `accept*` negotiation, `if-*` conditionals, the control/transport block incl. `host`/`cookie`/`date`, client-context `user-agent`/`referer`/`origin`/`from`). Nothing app-legitimate.
4. **Case-insensitive & header-location-only — confirmed.** `isReservedHeader()` compares `strtolower($name)` against the lowercased set and the emitted parameter keeps the original casing (`$name` is unchanged). The `content-type`/`CONTENT-TYPE`/`Content-Type` fixture is fully filtered, and `$request->cookie('Content-Type')` still surfaces as a `cookie:Content-Type` param — the guard's `$location === 'header'` clause holds.
5. **Silent skip / const / no surface change — confirmed.** The reserved read is `continue`-d before any `$unreadableAccessors[]` append, so it earns no note. `private const array RESERVED_HEADERS` — a class constant, not config; no Contracts, config-key, or signature change (only a `strtolower` import added).

Bookkeeping: `CHANGELOG.md [Unreleased] → Changed` entry present and accurate; `docs/auto-derivation.md` gains an honest sentence framing the tier/boundary and the `X-` invariant. Conventions clean (strict-types, concise why-comments, no issue numbers in code, 100-col).

This is one of the two approvals for the auto-merge gate. Deferring to your separate remote-CI and survey gates.

Radiergummi:
🤖 **[lead]** review → survey. Both independent code approvals in (reviewer1 + reviewer2), CI green across the matrix, non-controversial (38 src lines). surveyor-2 running Layer A — verifying the only spec delta is the intended reserved-header param removals (AureusERP/Pelican/SpeedtestTracker/Coolify), no other regression.

Radiergummi:
🤖 **[surveyor]** Layer A **pass** — only the intended reserved-header param removals (AureusERP, Coolify, Pelican, SpeedtestTracker); 0 real regressions across 11 apps.

Baseline regenerated on current `main` (`3c9c283a`) — the cached baseline was stale (`44124a70`, pre-#543). Candidate: PR head `092b42f7`. Both runs: all 11 apps `bootOutcome=booted`, `generateExit=0`, no new crash.

**Non-`parameters` metrics** (paths, operations, responseSchemas, documentedResponses, requestBodies, componentSchemas, operationsWithSecurity, completenessPercent): equal-or-better on **every** app — no regression anywhere.

**Reserved headers removed** (every net parameter drop, all `in: header`):

| app | removed header(s) | route |
|-----|-------------------|-------|
| AureusERP | `Content-Length` | `POST /livewire-*/update` |
| Coolify | `Authorization` | `POST /api/v1/sentinel/push` |
| Coolify | `Content-Type` ×2 | `POST /webhooks/source/{gitea,github}/events/manual` |
| Pelican | `Content-Length` | `POST /livewire-*/update` |
| SpeedtestTracker | `Content-Length` | `POST /livewire-*/update` |

6 removals, all in the RFC 9110 reserved set at `in: header` — exactly the four apps #530 predicted. No app outside that set lost a parameter.

**Denylist is not over-broad** — `X-*` / app-contract headers still surface in the candidate specs:
- Coolify: `Stripe-Signature`, `X-GitHub-Delivery`, `X-GitHub-Event`, `X-GitHub-Hook-Installation-Target-Id`, `X-Gitea-Delivery`, `X-Gitea-Event`, `X-Hub-Signature-256`
- InvoiceNinja: `X-API-TOKEN`, `X-API-SECURITY`, `X-API-COMPANY-KEY`
- Vito: `secret`

(The apparent `livewire-<hash>` path-param churn is Livewire's per-boot random route hash — a balanced -4/+4 of identical `component`/`filename` params, not a spec change.)

Clear to merge.

Radiergummi:
🤖 **[lead]** all gates green → **squash-merged**. Two independent code-review approvals (reviewer1 + reviewer2), remote CI green across the full matrix (incl. swagger-php 5.8 + snapshot), and **survey Layer A PASS** — surveyor-2 confirmed the only spec delta is 6 intended reserved-header `in:header` param removals (Content-Length/Authorization/Content-Type) on the 4 predicted apps (AureusERP/Coolify/Pelican/SpeedtestTracker), 0 regressions on every other metric, and `X-*`/app-contract headers (Stripe-Signature/X-GitHub-*/X-API-TOKEN) all still surface. Non-controversial (38 src lines, private constant+method, no Contracts/config-key/signature). Closes #530. Branch + worktree removed.

PR #571, opened 2026-07-22
feat(requests): let RuleDocumentation express an object property allowlist (additionalProperties/propertyNames)

### LINKED ISSUE #548: feat(requests): let RuleDocumentation express an object property allowlist (additionalProperties/propertyNames)
## Summary

`RuleDocumentation` cannot express an **object property allowlist** — a fixed set of permitted keys with `additionalProperties: false`. A `SelfDocumentingRule` for a rule that constrains an object's shape (allowed keys, an enum of keys) can therefore only degrade to a description-only `type: object`.

## Motivation (dogfooding: a private consumer app)

Implementing `SelfDocumentingRule` on the app's custom rules (rule.unknown 128 → 9) surfaced this. `App\Rules\KeysIn` (used **26 times**) validates that an object contains only a fixed set of keys:

```php
readonly class KeysIn implements ValidationRule
{
    public function __construct(protected array $values, ...) {}
    // passes when array_diff_key($value, array_flip($this->values)) === []
}
```

The faithful schema is `{ type: object, additionalProperties: false, properties: { <key>: {}, … } }` (or at least the allowed key set), but `RuleDocumentation` has no field for it, so the best achievable is:

```php
new RuleDocumentation(
    description: 'An object limited to these keys: a, b, c.',
    type: 'object',
);
```

The constraint survives only as prose.

## Proposed enhancement

Add optional fields to `RuleDocumentation` for object-shape constraints, applied by the same "write only when the descriptor has no value yet" merge policy as the existing fields:

- `additionalProperties: bool|null` (emit `additionalProperties: false`).
- `propertyNames: list<string>|null` — the allowed key set. Either materialise `properties: { key: {} }` for each, or emit a JSON-Schema `propertyNames: { enum: [...] }`.

(Naming/shape open to design — the point is a first-class way to express "these keys, nothing else".)

## Scope

Small, additive VO change plus the write path in `ValidationRulesToSchema` (or wherever `RuleDocumentation` is applied to a field descriptor). No behavioural change for rules that don't set the new fields.

## Notes

Found while implementing `SelfDocumentingRule` across 8 real custom rules; `KeysIn` was the only one the VO could not represent faithfully.

Radiergummi:
Closes #548

> **This is a design *proposal*, not a settled contract.** `RuleDocumentation` is the return type of
> the public `Contracts\Extraction\SelfDocumentingRule` extension interface (documented, with a
> worked example in the interface docblock), so it is effectively **public plugin-facing surface**.
> Field naming and shape are **Moritz's call** — the sections below give a recommendation with
> rationale, and flag the alternatives explicitly. **This PR will escalate to human review.**

## Problem

`RuleDocumentation` cannot express an **object property allowlist** — a fixed key set with
`additionalProperties: false`. A `SelfDocumentingRule` for a shape-constraining rule can only
degrade to a prose-only `type: object`. Motivating case: the app's `KeysIn` rule (used 26×;
`array_diff_key($value, array_flip($this->values)) === []`), whose faithful schema is
`{type: object, additionalProperties: false, propertyNames: {enum: [...]}}` — the constraint
currently survives only in the description string.

## Independence pre-check

**Fully independent.** Validation-rule→schema surface; no overlap with the #558 (response-scan) or
#560 (`PrimaryResponse`) hot zones. Scanned all open PRs — **none** touch `RuleDocumentation`,
`ValidationRulesToSchema`, `FieldDescriptor`, or `SelfDocumentingRule`. No gate, no collision.

## How public is `RuleDocumentation`, and the merge policy it must follow

- **Location:** `src/Support/Extraction/RuleDocumentation.php` (a `Support\` VO), but it is the
  return type of `Contracts\Extraction\SelfDocumentingRule::documentation()` — the public extension
  point a host app implements on its custom rules. So new fields are **public API additions**.
- **Merge policy (must be preserved):** its docblock states "Schema constraints are written only
  when the target descriptor has no value yet; `$description` is appended rather than overwritten."
  Confirmed in `ValidationRulesToSchema::applySelfDocumentingRule()` (`src/Support/Extraction/ValidationRulesToSchema.php:901`):
  every field writes under a `$doc->x !== null && $field->x === null` guard onto a
  `FieldDescriptor`. The two new fields follow the same guard.
- **Consumption chain:** `SelfDocumentingRule::documentation()` → `applySelfDocumentingRule()` writes
  onto `FieldDescriptor` (`src/Support/Extraction/FieldDescriptor.php`) → `FieldDescriptor::applyTo(OA\Schema)`
  serialises to the swagger-php schema (its `$overwrite` / `is_undefined($target->…)` guard is the
  second half of the same merge policy). `FieldDescriptor` already carries `?array $properties` and
  `applyTo()` already reads/writes `$target->additionalProperties`, so most of the emit machinery
  exists.

## swagger-php support (verified across the supported range)

The supported range is `^5.8 || ^6.1.2`. `OA\Schema` models **all three** relevant keywords on both
ends:

- `additionalProperties` — long-standing; present in 5.8.0 and 6.2.0.
- `propertyNames` — **verified present in 5.8.0** (`Annotations/Schema.php`, `public $propertyNames`)
  and in 6.2.0 (`Annotations/JsonSchemaTrait.php:156`). So `propertyNames` is **not** blocked by the
  5.8 floor. (This is distinct from the known unmodeled-keywords gap #140, which is `if`/`then`/`else`/
  `not`/`dependentRequired` — `propertyNames` is modelled.)
- `properties` — universal.

**Caveat that shapes the emit:** per the known 5.8-vs-6.x validation difference, a *raw nested array*
schema value can be rejected by 5.8's validator. So the `propertyNames` value must be built as an
`OA\Schema` object (`new OA\Schema(['enum' => …])`), never a raw array — the same way `enum` is
already set elsewhere. With that, the byte-exact `snapshot` group (excluded from the 5.x job anyway)
and the 5.8 CI job both stay green.

## Proposed field shape (recommendation + alternatives)

Add two optional fields to `RuleDocumentation`, JSON-Schema-keyword-aligned to match the existing
fields (`enum`, `pattern`, `minLength`, … already mirror keyword names):

```php
public ?bool $additionalProperties = null,        // false → emit additionalProperties: false
/** @param null|list<string> $propertyNames */
public ?array $propertyNames = null,               // the allowed key set
```

**Recommended emit:** `propertyNames` → `propertyNames: { enum: [...] }`, and `additionalProperties:
false` when set. Rationale:
- `propertyNames: {enum}` is the **most faithful and compact** expression of "every key must be one
  of this set" — exactly `KeysIn`'s semantics — and is verified available on 5.8+6.x.
- `additionalProperties: false` alongside it makes the closed-object intent explicit and belt-and-braces.

**Alternative emit (materialise `properties`):** `properties: { key: {} }` per allowed key +
`additionalProperties: false`. Reuses `FieldDescriptor::$properties` wholesale. Rejected as the
default because it is verbose (an empty `{}` schema per key) and `properties` means "these keys *may*
appear" — only `additionalProperties: false` makes it closed — whereas `propertyNames` states the
closed key constraint directly. Since `propertyNames` is confirmed on 5.8, compatibility does **not**
force this fallback. Presented so Moritz can choose it if preferred.

**Naming alternatives for Moritz** (I recommend the keyword-aligned names above): `allowedKeys` /
`closedObject` read more domain-y but break the "mirror the JSON-Schema keyword" convention the other
fields follow. Scope of `additionalProperties`: kept `?bool` (only `false` is useful for this case);
a full `bool|OA\Schema` typed-additionalProperties is a separate, larger feature — out of scope.

## Files touched

- `src/Support/Extraction/RuleDocumentation.php` — add `?bool $additionalProperties`, `?array $propertyNames` (+ `@param list<string>`); extend the class docblock.
- `src/Support/Extraction/FieldDescriptor.php` — add matching fields; in `applyTo()`, emit
  `$target->additionalProperties` and `$target->propertyNames = new OA\Schema(['enum' => …])` under
  the existing `$overwrite`/`is_undefined` guard.
- `src/Support/Extraction/ValidationRulesToSchema.php` — in `applySelfDocumentingRule()`, two guarded
  writes (`$doc->x !== null && $field->x === null`); sanitise `propertyNames` to strings and emit a
  `rule.invalid-*` finding for a non-string entry, mirroring the existing `$doc->enum` sanitisation
  (id `rule.invalid-enum-value`) — reuse or add a sibling id (Moritz/reviewer call).
- `docs/extending/custom-validation-rules.md` — document the two new fields + the `KeysIn` worked example.
- `CHANGELOG.md` — `[Unreleased] → Added`.

## TDD test list (failing test first — `tests/Unit/Support/Extraction/SelfDocumentingRuleTest.php`)

1. A rule returning `additionalProperties: false` on an object field → schema emits `additionalProperties: false`.
2. A rule returning `propertyNames: ['a','b','c']` → schema emits `propertyNames: { enum: ['a','b','c'] }`.
3. `KeysIn`-shaped rule (both fields + `type: 'object'`) → `{type: object, additionalProperties: false, propertyNames: {enum: [...]}}`.
4. **Merge policy — no overwrite:** a field that already has `additionalProperties` / `propertyNames`
   set (from a prior rule or composed attribute) is **not** overwritten by the rule.
5. **No-op:** a rule that leaves both fields null emits neither keyword (byte-identical to today).
6. **Sanitisation:** a non-string `propertyNames` entry is dropped and a `rule.invalid-*` finding is
   emitted (mirror the enum case); an all-invalid list emits nothing.
7. **Serialisation safety:** the generated document serialises cleanly (guards the "value is an
   `OA\Schema`, not a raw array" requirement) — assert via the normal generate path, and rely on the
   5.8 CI job for the cross-version guarantee.
8. **End-to-end** (feature test, likely alongside the request-body/validation tests): a FormRequest /
   request DTO field whose rules include a `KeysIn`-style `SelfDocumentingRule` → the request body
   schema shows the allowlist.

## Docs & CHANGELOG

- `docs/extending/custom-validation-rules.md`: add the two fields to the `RuleDocumentation` reference
  and a `KeysIn` example emitting a closed-object allowlist.
- `CHANGELOG.md` `[Unreleased] → Added` (own line, end of the list):
  > `RuleDocumentation` can now express an object property allowlist: `additionalProperties` (emit
  > `additionalProperties: false`) and `propertyNames` (the allowed key set, emitted as
  > `propertyNames: { enum: … }`). A `SelfDocumentingRule` for a shape-constraining rule (e.g. a
  > "keys in" rule) now produces a faithful closed-object schema instead of a prose-only
  > `type: object`. (#548)

## Merge-risk

- **Public-surface change → human gate (expected).** Two new fields on a documented extension VO.
  Additive and back-compatible (all optional, default null); the risk is purely the naming/shape
  contract, which is why this is framed as a proposal for Moritz.
- **Generation-affecting** only for apps whose custom rules opt into the new fields — zero change for
  every existing rule (both default null, guarded writes).
- Refresh any snapshot app that adopts the fields (none in-tree yet).


Radiergummi:
🤖 **[planner]** plan posted — see PR body. Ready for plan-review. Note: this is a design *proposal* on public surface (RuleDocumentation) and will escalate to Moritz for the final field naming/shape call.

Radiergummi:
🤖 **[lead]** planning → plan-review (technical validation only; the field naming/shape is Moritz's call → escalates after). Proposal: add `?bool $additionalProperties` + `?array $propertyNames` to `RuleDocumentation` (return type of the public `Contracts\Extraction\SelfDocumentingRule`), emit `propertyNames: {enum: […]}` + `additionalProperties: false`. planner-2 verified `propertyNames` is modelled in swagger-php **5.8.0 and 6.2.0** (not the #140 unmodeled-keywords gap). → reviewer2.

Radiergummi:
🤖 **[reviewer]** plan-review: technically sound — all four premises verified against source. Two concrete edge-case test-gaps to add, and a recommendation for Moritz on emit shape. Naming/shape remains Moritz's call.

**Premise checks (all confirmed):**

1. **swagger-php modelling — verified on the real supported range.** `propertyNames` is `public $propertyNames = Generator::UNDEFINED;` in **5.8.0** `src/Annotations/Schema.php:430` (and `additionalProperties` at `:398`), and in **6.2.0** it lives in `Annotations/JsonSchemaTrait.php:156` (used by `Schema`). So it is modelled on both ends — **not** the #140 unmodeled-keyword gap (that gap is `if`/`then`/`else`/`not`/`dependentRequired`). Caveat is correct and material: the `$propertyNames` property is **untyped** (bare `= Generator::UNDEFINED`), so per our known 5.8-vs-6.x validation divergence the value must be built as `new OA\Schema(['enum' => …])`, never a raw nested array. The plan specifies exactly that.
2. **Emit machinery largely exists.** `FieldDescriptor` carries `?array $properties` (`FieldDescriptor.php:91`) and `applyTo()` reads/writes `$target->additionalProperties` (`:212`, `:243`). One clarification: today `additionalProperties` is only *read* for map-detection and *migrated* in the nullable block — there is no first-class descriptor field that **writes** it, so the `additionalProperties` emit is a genuinely new write (the plan says "add matching fields", so this is understood, not a contradiction).
3. **Merge policy preserved.** `applySelfDocumentingRule()` writes every field under `$doc->x !== null && $field->x === null` (`ValidationRulesToSchema.php`, confirmed for type/format/pattern/enum/minLength/maxLength), and `applyTo()`'s `is_undefined($target->…)` guard is the second half. The plan's two new writes follow both. Confirmed.
4. **Publicness framing correct.** `RuleDocumentation` is the return type of `Contracts\Extraction\SelfDocumentingRule::documentation()` (`SelfDocumentingRule.php:37`), a documented extension interface with a worked example in its docblock. New constructor params are a public-surface addition → escalation is right.

**Two concrete test-gaps the plan should add (both real, both in the emit path):**

- **A — map-detection collision.** `applyTo()` computes `$targetIsMap = $target->type === 'object' && is_defined($target->additionalProperties)` (`:212`). A `KeysIn` field sets **both** `type: object` and `additionalProperties: false`, which flips `$targetIsMap` true — a flag designed to detect `array<string,T>` **maps**, not closed objects. It currently only gates the `items` branch (benign when `items` is null, as here), but the coder must add an assertion that a closed-object allowlist is **not** mis-serialised as a map (no stray `items`/type coercion). This is test #3/#7 territory but should be called out explicitly.
- **B — nullable migration must carry `propertyNames`.** The nullable-object wrapper (`:228`–`:246`) migrates `items`/`properties`/`additionalProperties` into the inner `oneOf` schema but (naturally) not `propertyNames`. Once `propertyNames` is emitted onto `$target`, a **nullable** closed-object would leave `propertyNames` on the outer wrapper while `additionalProperties` moves inner — an inconsistent/likely-invalid split. The plan must (1) add `propertyNames` to that migration block and (2) add a test: a nullable `KeysIn`-style field emits a well-formed nullable closed object. The current TDD list has no nullable case.

**Recommendation for Moritz (emit shape — not a blocker):** I endorse `propertyNames: {enum: […]}` over materialising `properties: {key: {}}`. It is the faithful expression of `KeysIn`'s "every key must be one of this set" (JSON Schema `propertyNames` constrains *all* property names), whereas `properties` means "these keys *may* appear" and only closes via `additionalProperties: false`. It is also more compact and is confirmed available on 5.8, so compatibility does not force the `properties` fallback. Note that with `propertyNames: {enum}` present, `additionalProperties: false` is strictly **redundant** (no property can have a name outside the enum) — harmless belt-and-braces, worth keeping as explicit intent, but Moritz may prefer to drop it. Keyword-aligned field names (`propertyNames`/`additionalProperties`) are consistent with the existing VO fields; the domain-y alternatives (`allowedKeys`/`closedObject`) break that convention — I lean keyword-aligned, but that is Moritz's decision.

No scope, boundary, or convention concerns: `Support\` VO, no plugin/third-party dependency, additive and default-null (zero change for existing rules), CHANGELOG + `docs/extending/custom-validation-rules.md` both accounted for.

Radiergummi:
🤖 **[lead]** plan-review → needs-human — **DESIGN DECISION REQUESTED (your call on the public API shape).** The proposal is technically vetted end-to-end (reviewer2 confirmed all 4 premises in source, incl. fetching the 5.8.0 tag: `propertyNames` present at `Schema.php:430` and 6.2.0 `JsonSchemaTrait:156` — **not** the #140 gap). No code written yet — this is plan-only, awaiting your shape decision before implementation.

**Decision needed:** the field naming/shape on `RuleDocumentation` (public — return type of `Contracts\Extraction\SelfDocumentingRule`).

**Team recommendation:** add `?bool $additionalProperties` + `?array $propertyNames` (JSON-Schema-keyword-aligned, matching existing `enum`/`pattern`); emit `propertyNames: {enum: […]}` (as `new OA\Schema(['enum'=>…])`, never a raw array — 5.8 validator) + `additionalProperties: false`. Endorsed over materialising `properties: {key:{}}` (faithful + compact; 5.8 support confirmed so no compat forcing the fallback). `additionalProperties:false` is strictly redundant beside `propertyNames:{enum}` but harmless. Alternatives presented in the PR body: `allowedKeys`/`closedObject` naming; `properties`-materialisation emit.

**Two emit-path test-gaps to fix when coded** (reviewer2 found; NOT yet in the TDD list): (A) `additionalProperties:false`+`type:object` flips `applyTo()`'s `$targetIsMap` flag (FieldDescriptor:212) — needs a 'closed object ≠ map' assertion; (B) the nullable-object migration block (:228–246) won't carry the new `propertyNames` into the inner oneOf — must be extended + a nullable test added.

Once you pick the shape, this can be coded (independent of #558/#560), reviewed, surveyed, and merged. Worktree/branch left in place. Ready, not merged.

Radiergummi:
🧑 **[Moritz]** Design decision: go with the team's recommended shape — keyword-aligned `?bool $additionalProperties` + `?array $propertyNames` on `RuleDocumentation`, emitting `propertyNames: {enum: […]}` (built as `new OA\Schema(['enum' => …])`, never a raw array) + `additionalProperties: false`. Rejecting the domain-y `allowedKeys`/`closedObject` alternative. Unblocked for coding — fold in the two emit-path test-gaps (map-detection collision + nullable `propertyNames` migration) the reviewer flagged.

Radiergummi:
Yeah, that shape tracks. Let's do it this way.

Issue #555, opened 2026-07-22
fix(responses): STATEMENT_LIMIT=10 hides a trailing return, refusing canonical Model-typed toResource() receivers


Radiergummi:
## Problem

`ReturnExpressionResourceReader` scans only the **first 10 top-level statements** of a
method (`STATEMENT_LIMIT = 10`, `ReturnExpressionResourceReader.php:83`, applied at
`:201`, `:377`, `:962`). A method whose `return` sits past that window is never seen,
so the receiver resolution introduced by #540/#543/#545/#547 never runs — even when
the receiver is the **canonical, simplest case**: a route-model-bound, Model-typed
method parameter.

```php
public function update(UpdateCategoryRequest $request, Category $category): JsonResource
{
    if ($request->hasName())        { $category->name = $request->getName(); }
    if ($request->hasDescription()) { $category->description = $request->getDescription(); }
    // … six more single-field guards …

    $category->save();
    $category->load(Category::RELATION_CREATED_BY);

    return $category->toResource();   // ← statement 13; never reached by the scanner
}
```

`$category` is a `Category`-typed parameter. Resolution needs no dataflow, no body
proof, no generics — it is the textbook Tier-0 read the receiver family was built for.
It refuses purely because of an arbitrary window.

The guard that actually matters for soundness is "**exactly one unconditional
return**", which the reader already enforces. The statement window adds nothing to
that guarantee — it only decides whether the return is *found*.

## Why the limit exists, and why it is miscalibrated here

The Tier-1 rationale ("match a small whitelist of call shapes in the first N
statements") is about bounding the cost and blast radius of *pattern matching* over a
body. Locating a method's single trailing `return` is not that: it is cheap, and the
shape being matched is on the return expression itself, not on the statements
preceding it. Applying a scan cap designed for inline `validate()`/`abort()` matching
to return-statement discovery conflates two different scans.

Methods with a run of guard clauses followed by a trailing return are a mainstream
shape — conditional field assignment, validation guards, authorization checks. Ten
statements is easily exceeded by ordinary code.

## Evidence

Measured on a private consumer app (316 `/api` operations), against library `main` @
`f0999f69` with the full receiver family merged:

- **18** operations still `$ref` an abstract resource base (down from 78 pre-#540).
- **102** controller methods declare an abstract return type and return via
  `->toResource()`.
- **6** of those exceed the 10-statement window and are therefore refused outright:

  | Statements | Method |
  |---|---|
[...]

Issue #540, opened 2026-07-21
Resolve ->toResource() receiver model from any declared type, not just Model-typed parameters


Radiergummi:
## Problem

`ReturnExpressionResourceReader` resolves a bare `$model->toResource()` only when
the receiver is a **Model-typed method parameter**. Every other receiver refuses:

```
calls ->toResource() on a receiver whose model type is not statically knowable;
the concrete resource stays unresolved. Annotate the action with #[ResponseResource].
```

But the common refused receivers carry a **statically declared type** — this is
Tier-0 information, not dataflow:

```php
// declared return type on the action class
return $action->handle(request()->user())->toResource();

// declared return type on the form request
return $request->resolve($contact)->toResource();

// typed relation / property
return $document->mediaItem->toResource();

// local assigned from a typed expression, often with an assert() narrowing
$customer = $request->user()->customer;
assert($customer instanceof Customer);
return $customer->toResource();
```

Once the receiver's model class is known, the rest of the chain already works:
`conventionalResourceFor()` reads Laravel's `#[UseResource]` attribute and
`guessResourceName()` correctly.

## Measurement (a private consumer app, Laravel 13, 316 `/api` ops)

After #536/#538 landed, resource schemas are populated and empty component
schemas dropped 58 → 8. What remains:

| | count |
|---|---|
| `/api` ops still resolving to the abstract `JsonResource` | **47** |
| generator notes: `->toResource()` receiver not statically knowable | **43** |
| generator notes: `->toResourceCollection()` without a named class | 2 |

68 of the app's 72 abstract-typed actions return a bare `toResource()`; ~31 already
resolve because the receiver is a route-model-bound parameter. The 43 that fail
differ only in *where* the type is declared.

## Why not just annotate

`#[ResponseResource]` is the documented escape hatch and would close all 43 — but
per the inference philosophy an attribute is a **smell** when it restates what a
Tier-0 read could infer. A declared return type on `Action::handle()` is exactly
that. 43 attributes across one app would be papering over a reflection gap.

## Proposed change

Extend receiver-type resolution to any expression whose type is **declared**:

1. Method call → the method's declared return type (`ReflectionMethod::getReturnType()`),
   resolving the callee from a typed variable/parameter/property.
2. Property fetch → the declared property type or `@property`/`@property-read` /
   relation generic (`WrappedModelLocator` and `TypeNodeResolver` already do this).
3. Local variable → the declared type of the expression assigned to it, and/or an
[...]

Issue #536, opened 2026-07-21
Read Laravel 13's first-party JsonApiResource (toAttributes/toRelationships/toLinks/toMeta)


Radiergummi:
## Problem

Laravel 13 ships first-party JSON:API resources at
`Illuminate\Http\Resources\JsonApi\JsonApiResource`. Their public surface is **not**
`toArray()`; it is:

```php
public function toType(Request $request): string
public function toId(Request $request): string
public function toAttributes(Request $request): array     // ← the field bag
public function toRelationships(Request $request): array
public function toLinks(Request $request): array
public function toMeta(Request $request): array
```

All resource reading in `Plugins\ApiResources` is built around `toArray()`
(`ResourceToArrayReader::readFields()` hardcodes the method name). Against a
`JsonApiResource` subclass the reader finds no override, reads nothing, and
`SchemaFromResource` emits:

```yaml
UserResource:
  type: object
  properties: {}
```

…for every such resource.

This is a **fidelity** bug as well as a coverage one. A native JSON:API resource
responds with `{data:{type,id,attributes:{…},relationships:{…}}}`, but the generic
`JsonResource` path models a flat `{data:{…fields}}` envelope — so any fields we
did emit would sit at the wrong nesting depth.

`JsonApiResource` is now vanilla Laravel, which puts it squarely in the Core /
ApiResources remit.

## Evidence (dogfooding a private consumer app)

Real Laravel 13 app, 316 operations under `/api`, package installed with **no**
configuration:

| Metric | Value |
|--------|-------|
| Complete operations (substantive 2xx + body) | 162 (**51.3%**) |
| Ops whose response `$ref`s an *empty* resource schema | **117** (37%) |
| Empty component schemas that are `*Resource` | 54 of 58 |
| App resources extending `JsonApiResource` | 66 |
| …whose `toAttributes()` is a single `return [...]` literal | **56 (88%)** |

The 88% figure matters: these bodies are plain array literals with
`self::CONST => expr` entries — exactly as statically readable as the `toArray()`
bodies already handled, and `AstLiteralEvaluator` already resolves class
constants. This is a bounded extension of the existing reader, not new machinery.

## Proposed change

1. **Generalize `ResourceToArrayReader`** to read a named method. Everything below
   `readFields()` (literal walking, `when()`/`whenLoaded()` handling, model-backed
   type inference) is already method-name-agnostic; only the `'toArray'` string and
   `overridesToArray()` need to become parameters.

[...]

PR #573, opened 2026-07-23
feat(lint): operation.response-type-abstract — flag a response typed as an abstract resource base

### LINKED ISSUE #553: feat(lint): operation.response-type-abstract — flag a response typed as an abstract resource base
## Problem

When a controller declares an abstract API-Resource base as its return type, the
library resolves it honestly and emits a `$ref` to a component schema with no
properties:

```php
public function updateSelf(UpdateProfileRequest $request): JsonResource
```

```yaml
$ref: '#/components/schemas/JsonResource'   # → properties: {}
```

Nothing is wrong library-side — this is the documented "type your returns" contract
working exactly as specified, and the fix is entirely app-side (narrow the return
type to the concrete subclass). But today the *only* symptom the consumer sees is an
unhelpfully empty schema in the output, with no indication that (a) this is why, and
(b) it is a one-word fix in their own code.

In the private consumer app dogfooding run this was **78 of 316 `/api` operations** —
the single largest self-serviceable gap in the app, and invisible without a manual
schema walk.

## Proposal

A lint rule that fires when an operation's primary response `$ref`s a schema derived
from an **abstract** resource base, or from a known framework base class
(`Illuminate\Http\Resources\Json\JsonResource`,
`Illuminate\Http\Resources\Json\ResourceCollection`,
`Illuminate\Http\Resources\Json\AnonymousResourceCollection`,
`Illuminate\Http\Resources\JsonApi\JsonApiResource`), rather than a concrete subclass.

Message should name the fix, not just the symptom — something like:

> `operation.response-type-abstract`: the response type is an abstract resource base
> (`JsonResource`); narrow the return type to the concrete resource subclass to get a
> response schema.

Suggested ID: `operation.response-type-abstract` (`area:lint`, default severity
`warning`).

## Design notes

- Should fire on the **declared return type**, so it needs either the provenance to
  reach back to the source (see #208 / #484, which make inference-failure reasons
  lint-readable) or a resolution at generation time that the rule can read. If #484
  lands first, this may be better expressed as an inference-failure finding emitted at
  generation time rather than a spec-tree rule — worth deciding during planning.
- **Do not fire** when the app has legitimately authored a schema for that operation
  via `#[Response]` / `#[ResponseResource]` — the escape hatch was used correctly.
- **Do not fire** for `AnonymousResourceCollection` when the element type *was*
  resolved (the collection base is expected there; the element is what matters).

## Caveat worth documenting alongside the rule

Narrowing the return type is **not** always available to the consumer. Laravel
declares `Model::toResource(?string $resourceClass = null): JsonResource`, so an app
using the `->toResource()` idiom cannot narrow its controller return type without
PHPStan reporting an error on every such method (observed: 32 errors across 26
controllers in the dogfooding app). The rule's message and the docs page should say
so, and point at the receiver-resolution family (#540/#543/#545/#547) as the reason
the library resolves those cases itself rather than asking the app to annotate.

Filing this without that caveat would generate findings the consumer cannot act on.

## Acceptance

- The rule fires on an operation whose response refs an abstract/framework resource
  base, with a message naming the concrete-subclass fix.
- It stays silent when a schema was authored explicitly, and when a collection's
  element type resolved.
- A row is added to the hand-maintained catalog in `docs/linting.md`.

Radiergummi:
Closes #553

## Implementation plan — `operation.response-type-abstract`

A new lint rule that flags an operation whose primary (2xx) response `$ref`s a schema
derived from an **abstract** resource base or a **known framework** resource base, rather
than a concrete subclass — the honest symptom of the "type your returns" contract, made
self-serviceable.

### Tier & path

**Tier 0.** The rule reads the *generated* document (spec tree) plus reflection on the
component's already-recorded source class. No method-body parsing. `area:lint`,
non-generation-affecting → **Fast-path** (coder → reviewer ×1 → merge; no survey).

### Preconditions — verified during planning

- **#555 has landed** (merged as PR #558, `d2c45fa`, whole-method return scan). The lead's
  gating comment ("emitting a narrow-your-return finding for the `STATEMENT_LIMIT=10` cases
  would be wrong advice") is therefore satisfied at the source: those receivers now resolve
  to a concrete resource and won't present as an abstract base. The residual findings this
  rule emits are genuinely consumer-fixable.
- The `->toResource()` receiver family (#540/#543/#545/#547, all merged) already resolves
  most `->toResource()` cases library-side, so they ref the concrete resolved resource, not
  `JsonResource`. What remains firing is the case the caveat documents: an app that *declares*
  `: JsonResource` (or an app-defined abstract base) as the return type.
- **#484 / #208 are OPEN** — the generation-time inference-failure / source-map path the issue
  floated is unavailable. So this ships as a **spec-tree rule**, as the issue's fallback design
  specifies. No dependency on either.

### The mechanism the tree already provides

- `Lint\Tree\ComponentSchemaNode` carries `?string $sourceClass` — the FQCN that produced the
  component schema. This is the whole enabler: the rule reaches from a response `$ref` back to
  the PHP class behind it without any new plumbing.
- `Lint\TreeIndex::$componentsByName` (`schema name → ComponentSchemaNode`), exposed via
  `LintContext::$index`, resolves `ResponseNode::$schemaRef` (a component name) to that node.
- `ResponseNode` already exposes `isSuccess` (2xx), `schemaRef`, and `fields`.

### What fires (predicate)

On each **2xx** response with a non-null `schemaRef`, resolve the component node. Fire when
**both**:

1. Its `sourceClass` is a resource base — either
   - one of the framework bases (exact FQCN match):
     `Illuminate\Http\Resources\Json\JsonResource`,
     `…\Json\ResourceCollection`,
     `…\Json\AnonymousResourceCollection`,
     `…\JsonApi\JsonApiResource` (Laravel 13, first-party — see #536/#537); **or**
   - an **abstract** class that is `is_a(..., JsonResource::class, true)`
     (`class_exists` guard + `ReflectionClass::isAbstract()`), covering app-defined abstract
     resource bases; **and**
2. The referenced component schema is **empty** (`$component->fields === []`).

Gate (2) is how the two "do not fire" exemptions from the issue are satisfied *structurally*,
without special-casing:

- **`AnonymousResourceCollection` with a resolved element** produces a schema with items /
  fields, so it is not empty → no fire. An unresolved one is empty → fires. (Fixture covers
  both; the coder verifies the actual rendering and, if a resolved collection turns out to
  carry its element via a nested `items` ref rather than `fields`, tightens the emptiness test
  accordingly.)
- **Explicitly authored response** (`#[Response]` / `#[ResponseResource(Concrete::class)]`)
  resolves to a concrete class or inline schema, whose `sourceClass` is not an abstract/
  framework base → no fire. A fixture asserts this; **if** the assertion fails (authoring
  somehow still refs an abstract base), add an explicit guard reading the operation's action
  attributes via `LintContext` — but the expectation is it is already structural.

Scope note: limited to the **JsonResource family + JsonApiResource**, per the issue
("abstract *resource* base"). Abstract Spatie Data bases are out of scope.

### Files

- **New** `src/Lint/Rules/OperationResponseTypeAbstract.php`
  - `implements Rule, ResponseRule` (the `ResponseRule` visitor).
  - `public string $id = 'operation.response-type-abstract';` (ID verbatim from the issue).
  - `public Severity $severity = Severity::Underspecified;` — the schema is emitted but empty;
    same tier as its sibling `response.success-empty-body`. (The enum has no "warning"; the
    issue's informal "warning" maps to the `Underspecified` tier, catalog severity `2`.)
  - Message names the **fix**, not the symptom, e.g.:
    "Response {status} on {METHOD /path} refs an abstract resource base ({ShortName}); narrow
    the action's return type to the concrete resource subclass to get a response schema."
  - `fixHint`: narrow the return type; **caveat** — if the value comes from `->toResource()`,
    narrowing may be impossible (Laravel types `Model::toResource(): JsonResource`, so PHPStan
    rejects the narrower type); the library resolves those cases itself via the receiver family.
- **Modify** `src/Support/Generator/BaselineRegistration.php` — import + add
  `OperationResponseTypeAbstract::class` to the baseline lint-rule list (append near
  `ResponseSuccessEmptyBody::class`, ~L168). No config entry needed — baseline rules default
  their severity from the class property; `config/openapi.lint.rules` holds only overrides.
- **Modify** `docs/linting.md` — add a catalog row inside the `lint-rule-catalog` markers,
  severity-2 group:
  `| operation.response-type-abstract | 2 | The primary response refs an abstract or framework resource base (e.g. JsonResource); narrow the return type to the concrete resource subclass to get a schema. |`
  Add a short caveat paragraph near the rule mentioning the `->toResource()` un-narrowable case
  and pointing at the receiver-resolution family as the reason the library handles those.
- **Modify** `CHANGELOG.md` — `[Unreleased] → Added`: one line for the new rule (#553).

### Tests — `tests/Unit/Lint/Rules/OperationResponseTypeAbstractTest.php`

Model on `ResponseSuccessEmptyBodyTest`. TDD; the failing "fires" test comes first. Cases:

- **fires** — 2xx response refs a component whose `sourceClass` is `JsonResource` and the
  schema is empty.
- **fires** — response refs an **app-defined abstract** resource base (`abstract class Foo
  extends JsonResource`), empty schema.
- **fires** — `ResourceCollection` / `JsonApiResource` bases (parameterised).
- **silent** — response refs a **concrete** resource subclass with fields.
- **silent** — `AnonymousResourceCollection` whose **element resolved** (non-empty schema).
- **fires** — `AnonymousResourceCollection` whose element did **not** resolve (empty).
- **silent** — response authored via `#[ResponseResource(Concrete::class)]` (concrete ref).
- **silent** — non-2xx (4xx/5xx) response refs an abstract base (only the primary success
  response is in scope).
- **silent** — no `schemaRef` (inline body or bodiless) — no `componentsByName` lookup.
- **silent** — `schemaRef` resolves to no component (broken ref is `ref.broken`'s job).

Fixtures: small resource/controller classes under `tests/Fixtures` (an abstract base, a
concrete subclass, an anonymous collection) — reuse existing resource fixtures where present.

### Observable-change surface

- New public lint rule → **`docs/linting.md`** catalog row + caveat (above) and **CHANGELOG**.
- `openapi:lint --list` is the live catalog; the doc row is the hand-maintained mirror.

### Controversy

Low. New isolated lint rule; no `Contracts` change, no stage-order change, no config key, no
public-signature change. Default severity is `Underspecified` (opt-out via the standard
`config/openapi.lint.rules` override, no new config surface). Expected to pass Fast-path.


Radiergummi:

Radiergummi:
🤖 **[lead]** plan posted → plan-review.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved.**

Verified every claimed mechanism against `main` — all present and used exactly as the plan describes:
- `ComponentSchemaNode::$sourceClass` (`?class-string`) — the reach-back enabler.
- `TreeIndex::$componentsByName` (schema name → node), exposed via `LintContext::$index` — already used by `SchemaRequiredWithoutProperty`/`ref.broken`.
- `ResponseNode::$isSuccess` / `$schemaRef` / `$fields`.
- Sibling `ResponseSuccessEmptyBody` is the right template (same `Severity::Underspecified`, `ResponseRule` visitor, baseline-registered with severity defaulted from the class property, no config entry).

Tier/scope/bookkeeping all correct: Tier 0 (spec-tree read + reflection, no body parsing), #484/#208 correctly noted OPEN so the fallback spec-tree rule ships (not the generation-time variant), #555 precondition satisfied via #558. Proving tests, docs catalog row + caveat, and CHANGELOG entry are all named. The two "do not fire" exemptions are handled structurally by the empty-schema gate rather than special-casing — good.

Two minor notes for the coder (not blockers, no re-plan):
1. The `Rule` contract requires a third property `public string $description` (shown by `openapi:lint --list`), which the plan omits from the class sketch. Add it and keep its text aligned with the `docs/linting.md` catalog row so the live catalog and the hand-maintained mirror match.
2. The visitor interface is `Radiergummi\OpenApi\Lint\Visitors\ResponseRule` (namespace `Lint\Visitors`, not `Rules\Visitors\`) — the plan just says "ResponseRule visitor", which is fine.

The one real correctness risk is the emptiness gate (`$component->fields === []`): a resolved `AnonymousResourceCollection` may carry its element via `raw`/nested `items` rather than `fields`. The plan already flags this and instructs the coder to verify the actual rendering and tighten the test accordingly — that is the right call; the anonymous-collection resolved/unresolved fixtures must confirm it empirically before merge.

Radiergummi:
🤖 **[lead]** plan-review approved → coding.

Radiergummi:
🤖 **[coder]** implemented; local gates green (`composer check`: Pint, PHPStan L8, 3622 tests); pushed `d72ffefe`. Ready for review.

**Deviation from the plan — the mechanism changed (with empirical proof).** The plan's predicate was "a 2xx response whose top-level `$ref` points at an abstract-base component." Probing the real pipeline showed that premise does not hold for API Resource responses:

- An unresolvable `: JsonResource` / app-abstract-base return renders as an **inline envelope** — `{properties: {data: {$ref: JsonResource}}, type: object}` — so the response's top-level `schemaRef` is **null**; the offending component is reached through the `data` field's ref. The empty component *is* emitted (`JsonResource: {properties: [], type: object}`) and **does** carry its `sourceClass` in `componentClassMap` (`JsonResource → Illuminate\…\JsonResource`, `AppAbstractResourceBaseFixture → …`), which is the enabler.
- So the rule now collects candidate component names from **the top-level `schemaRef` *and* every response field ref** (recursively), resolves each via `TreeIndex::$componentsByName`, and fires when the component is empty **and** its `sourceClass` is a framework resource base (exact FQCN) or an app-defined abstract subclass of `JsonResource`/`JsonApiResource`.

**reviewer-1's correctness risk (collections) — discharged empirically.** Verified rendering of both collection shapes through the real generator:
- *Resolved* `AnonymousResourceCollection` renders inline as `{data: {type: array, items: {$ref: ConcreteResource}}}` — the element ref is concrete and sits under `items` (not walked into response fields), so no abstract-base component exists → **silent**. A Feature-test case asserts this.
- *Fully unresolved* collection emits a **bodiless** 200 (no component at all) → covered by `response.success-empty-body`, not this rule → **silent**. (The plan's "unresolved collection fires" case was empirically wrong.)
- The emptiness gate was still tightened per the note: it rejects a component carrying `items`/`allOf`/`oneOf`/`anyOf`, not just `fields === []` (unit-tested).

reviewer-1's other two notes are in: `public string $description` added (aligned with the docs row); visitor namespace is `Lint\Visitors\ResponseRule`.

Tests: `tests/Feature/Lint/ResponseTypeAbstractTest.php` (real pipeline: framework-base fires, app-abstract fires, concrete + resolved-collection silent) and `tests/Unit/Lint/Rules/OperationResponseTypeAbstractTest.php` (predicate edges: non-2xx, broken ref, no-ref, items-bearing emptiness gate, top-level-schemaRef path, id/severity). Docs catalog row + `->toResource()` caveat and CHANGELOG entry added.

Radiergummi:
🤖 **[coder]** remote CI green across the full matrix (PHP 8.4/8.5 × Laravel 12/13, swagger-php 5.8, quality, snapshot, coverage) at `d72ffefe`. Ready for review.

Radiergummi:
🤖 **[lead]** implemented (d72ffefe) → code-review. Queued for two independent approvals; reviewers finishing #574 first.

Radiergummi:
🤖 **[reviewer]** review round 1 — 1 finding (a missing acceptance test), plus 1 optional. Implementation and deviation are otherwise verified correct; this is reviewer-2, independent of reviewer-1.

**Deviation accepted.** I independently confirmed the envelope-vs-top-level mechanism against the tree: resource payloads render as an inline `{data: {$ref: …}}` envelope, so the abstract-base component is reached through response field refs, not `schemaRef`. `collectFieldRefs` walks the finite inline `FieldNode` tree (`children`), collecting only ref *strings* — it never follows a ref into a component, so there is no cycle/recursion risk. The resolved-`AnonymousResourceCollection` case is silent because its element sits under `items` (not a walked `FieldNode.ref`), and the fully-unresolved case is bodiless (owned by `response.success-empty-body`) — both empirically correct.

**Verified correct:**
- 2xx gate; dedup-by-base via `$seen` (only set after a fire, so non-matching bases are harmless).
- `isAbstractResourceBase`: framework FQCN exact-match OR `class_exists` + `is_a(JsonResource/JsonApiResource, strict)` + `ReflectionClass::isAbstract()`. The `class_exists` guards on `RESOURCE_ROOTS` correctly no-op `JsonApiResource` on Laravel 12 where it is absent; the framework-base string list needs no `class_exists` for exact match. Sound.
- `isEmptySchema`: rejects `fields !== []`, `items`, `allOf`, and (via `SchemaAccessor::classifyComposition`) a single-branch or 2+-branch `oneOf`/`anyOf` — I read `classifyComposition`; `branch !== null || uninspectedComposite` both short-circuit to non-empty, so the coder's "rejects items/allOf/oneOf/anyOf" claim holds.
- `Severity::Underspecified = 2` matches the docs catalog row.
- Boundary: rule lives in `src/Lint/Rules/`, reads the generated tree + reflection on the recorded `sourceClass`; plugin-agnostic, registered in `BaselineRegistration` baseline list, never gated behind the ApiResources plugin (with Core/ApiResources off it simply never fires). Correct per the package's "Core-disabled must still work" rule.
- Nice edge already covered: `ConcreteUnitResource` with **empty** fields → silent proves gate 1 excludes concrete-but-empty resources (distinguishes "empty because abstract" from "empty because unresolved concrete `toArray`").
- Bookkeeping: CHANGELOG `[Unreleased] → Added`, docs catalog row + `->toResource()` caveat paragraph, no new config key. TDD followed; local + full remote matrix green.

**1. (finding — please add) The "authored response stays silent" acceptance criterion has no direct test.** Issue #553 acceptance names it explicitly ("It stays silent when a schema was authored explicitly"). It is *mechanistically* covered (an authored `#[ResponseResource(Concrete::class)]` resolves to a concrete `sourceClass`, and `#[Response]` inline → `sourceClass` null → the `base === null` guard), and the concrete/null paths are unit-tested — but there is no assertion tying the authoring escape hatch to silence. Add one Feature case: an action carrying `#[ResponseResource(ConcreteThingResourceFixture::class)]` (or `#[Response]`) whose declared return type is `JsonResource`, asserting `responseTypeAbstractFindings(...) === []`. This proves the escape hatch suppresses the rule even when the *declared* type is the abstract base — the exact scenario a consumer would hit.

**2. (optional) `oneOf`/`anyOf` emptiness-gate rejection is unit-tested only for `items`.** `classifyComposition` handling is correct by reading, but only the `items` branch has a unit test (`stays silent when the component schema carries array items`). A parameterised sibling asserting silence for a component whose `raw` carries a single-branch `oneOf` (and a 2-branch `anyOf`) would pin the composition gate directly. Nice-to-have, not required.

No correctness defects found. Once finding 1 lands, I approve.

Radiergummi:
🤖 **[coder]** addressed round 1; pushed `c5901136`. `composer check` green (Pint, PHPStan L8, 3624 tests).

- **REQUIRED (finding 1):** added a Feature case `authoredEscapeHatch` — an action with `#[ResponseResource(ConcreteThingResourceFixture::class)]` whose **declared return type is `JsonResource`** (the abstract base); asserts the rule emits **no** findings, proving the authoring escape hatch suppresses it even against an abstract declared type. (Structurally: the attribute resolves the response to the concrete `ConcreteThingResourceFixture` component, whose source class is not an abstract/framework base.)
- **OPTIONAL (finding 2):** added a unit case asserting the emptiness gate stays silent when the component's raw schema carries a `oneOf` composition branch (previously only `items` was covered).

Radiergummi:
🤖 **[reviewer]** review round 1: approved, no concerns. (This is reviewer-1, the independent second approval — I did the plan-review; this verdict is from reading the pushed code at `c5901136`, not from reviewer-2's.)

I did not rubber-stamp — I re-derived the load-bearing claims against a fresh checkout and ran the gates myself.

**Core-mechanism deviation — confirmed correct.** Resource payloads render as an inline `{data: {$ref}}` envelope, so the offending component is reached through response **field refs**, not top-level `schemaRef`. `collectFieldRefs` walks only the finite inline `FieldNode` tree (`children`) collecting ref *strings* — it never follows a ref into a component, so no cycle. It deliberately does **not** descend into array `items` or into a referenced component's own fields, which is what keeps the resolved `AnonymousResourceCollection` (concrete element under `items`) and nested-relation cases out of scope — correct per the issue's "top-level return type" framing.

**Emptiness gate — independently verified sound against false positives.** `isEmptySchema` never inspects `raw->properties` directly, which looked risky, so I traced `SpecTreeBuilder::buildComponents` → `buildFields` → `collectComposedProperties`: `fields` is *derived* from the schema's (composed) properties, so `fields === []` genuinely implies no properties. Combined with the `items`/`allOf` `is_defined` checks and `SchemaAccessor::classifyComposition` (which I read: single non-null branch → `branch !== null`; 2+ → `uninspectedComposite`), the gate rejects fields/items/allOf/oneOf/anyOf. No path emits a false positive on a populated component.

**Adversarial probes, all clean:**
- Gate 1 (`isAbstractResourceBase`): framework FQCN exact-match, else `class_exists` + `is_a(JsonResource/JsonApiResource, strict)` + `ReflectionClass::isAbstract()`. A concrete app resource that renders empty (unresolved `toArray`) is correctly **not** flagged (not abstract) — that's `response.success-empty-body`/inference territory. `JsonApiResource` FQCN is inert on Laravel 12 (never a `sourceClass` there).
- Dedup `$seen` is per-response and set only after a fire, so a base appearing in both `schemaRef` and a field ref fires exactly once; non-matching bases are harmless. No double-fire.
- 2xx-only gate; broken-ref / no-ref / non-2xx all short-circuit (unit-tested).
- reviewer-2's required `authoredEscapeHatch` Feature case is present and green: `#[ResponseResource(Concrete)]` on an action *declared* `: JsonResource` emits **no** finding — the exact consumer scenario.

**Gates (ran locally in a clean worktree at `c5901136`):** target suite **17/17 pass**, PHPStan L8 **no errors**, Pint **passed**. Bookkeeping complete: baseline registration (plugin-agnostic list, never gated behind ApiResources), `docs/linting.md` catalog row **matching** the `public string $description`, `->toResource()` caveat paragraph, CHANGELOG `[Unreleased] → Added`.

The single codecov-missing line is the defensive `'<unknown operation>'` fallback (`operation() === null`), which is unreachable in normal tree-walk — not worth a test.

**Optional nit (non-blocking):** the finding message says "refs an abstract resource base (JsonResource)" though `JsonResource` is technically a *concrete* framework class used as a base. The `$description`/docs say "abstract **or framework** resource base" (accurate); the message parenthesises the real class name so the consumer sees exactly what to fix. Cosmetic only — fine to leave.

Non-generation-affecting lint rule → Fast-path, no survey needed. From the code side this is ready; with both reviewer approvals in, it can merge.

Radiergummi:
🤖 **[reviewer]** approved.

Re-verified the test-only delta d72ffefe..c5901136 (two test files, +24 lines, no src changes) — my round-1 finding is satisfied:

- **Finding 1 (required) — closed.** `authoredEscapeHatch`: a controller action carrying `#[ResponseResource(ConcreteThingResourceFixture::class)]` whose **declared** return type is the abstract `JsonResource`, added to the "does not fire" dataset asserting no findings. This is exactly the escape-hatch scenario acceptance names — and its passing proves authoring wins over the abstract return-type path (concrete `sourceClass` → not an abstract base → silent), not merely that a concrete return type is silent.
- **Finding 2 (optional) — added.** `stays silent when the component schema carries a composition branch`: a 2-branch `oneOf` → `classifyComposition` yields `uninspectedComposite: true` → `isEmptySchema` returns false → silent. Pins the composition arm of the emptiness gate directly.

Combined with round 1 (mechanism/deviation verified, boundaries, docs catalog row + `->toResource()` caveat, CHANGELOG, severity, real-generator feature coverage), this is a clean approval. No concerns.

Radiergummi:
🤖 **[lead]** two independent approvals (reviewer-1 + reviewer-2) + full matrix CI green (incl. swagger-php 5.8) + non-controversial (one new isolated lint-rule file, 2-line registration, no Contracts/config/severity-default changes) → squash-merged. Closes #553.

Issue #559, opened 2026-07-22
fix(core): SingleReturnArrayLiteralFinder's STATEMENT_LIMIT=10 hides trailing returns in toArray()/transform()/rules()


Radiergummi:
## Problem

`SingleReturnArrayLiteralFinder` (`src/Support/MethodBody/SingleReturnArrayLiteralFinder.php:25`)
caps its scan at the first 10 top-level statements:

```php
public const int STATEMENT_LIMIT = 10;
…
$statements = $this->scanner->firstStatements($method, self::STATEMENT_LIMIT);
```

It then does nothing but **locate a single return** — `methodLevelReturns($statements)`, refuse
unless `count === 1`, and resolve a `return $variable;` through
`ReturnExpressionResolver::resolveVariable()`. That is the identical conflation #555 fixes in
`ReturnExpressionResourceReader`: a scan window designed to bound *Tier-1 pattern matching over
preceding statements* is applied to *return discovery*, where it bounds nothing and only decides
whether the return is found.

The soundness guard that matters — "exactly one unconditional return", plus `resolveVariable()`'s
≠1-assignment and mutated-after-assignment refusals — is independent of the window.

## Blast radius

This finder is shared by three consumers, so the effect is on schema **content** rather than on
which class is resolved:

- API Resource `toArray()` — a resource whose `toArray()` runs more than 10 statements before its
  `return [...]` yields no field shape.
- Fractal `transform()` — same, via the Fractal plugin.
- `FormRequestStaticRulesReader` (`src/Plugins/Core/Support/FormRequestStaticRulesReader.php:38`),
  which aliases the constant — a `rules()` method with more than 10 leading statements
  (conditional rule building, `$rules = [...]` after setup) yields no request-body schema.

`toArray()` and `rules()` methods that build up state before returning are ordinary code; ten
statements is easily exceeded.

## Borderline third case

`FindReturnModelResponseResolver` (`src/Plugins/Core/Resolvers/FindReturnModelResponseResolver.php:47`)
also declares `STATEMENT_LIMIT = 10`, but it sits between the two categories: it scans for a
directly-returned lookup call, so the window is doing *some* pattern-matching work. Decide
explicitly during planning whether it belongs with the return-discovery fix or with the eight
readers whose bound is correct — do not fold it in silently.

**Not affected, and correct as they are** — these genuinely match whitelisted call shapes across
preceding statements, which is exactly what the Tier-1 bound is for: `InlineValidatorRulesReader`,
[...]

PR #574, opened 2026-07-23
fix(core): ReturnExpressionResolver refuses foreach/destructuring/ref rebindings instead of resolving to the wrong expression

### LINKED ISSUE #561: fix(core): ReturnExpressionResolver misses foreach/destructuring/ref rebinding, resolving locals to the wrong expression
## Problem

`ReturnExpressionResolver::resolveVariable()` decides whether a local is safe to resolve
by counting assignments to it. Its predicate recognises **only** `Assign` with a plain
`Variable` target (`:74-77`), and its companion `mutatesVariable()` (`:115-128`)
recognises **only** `AssignOp` and `Assign` into an `ArrayDimFetch`.

Every other way PHP rebinds a name is invisible to both. So a variable that is rebound
before the return still counts as "assigned exactly once, never mutated", and the
resolver returns the **first** assignment's expression — resolving to something the
method does not actually return.

```php
$subject = $repository->find($id);       // Assign → counted
foreach ($others as $subject) {          // Foreach_::$valueVar → INVISIBLE
    // …
}

return $subject->toResource();           // resolves against $repository->find($id)
```

`count($allAssignments) !== 1` passes (one), `count($unconditionalAssignments) !== 1`
passes (one), `$unconditionalMutations` is empty. The guard is satisfied and the wrong
type is emitted.

Rebinding forms missed by both predicates:

| Form | Seen by `resolveVariable`? |
|---|---|
| `foreach ($xs as $subject)` / `as $subject => $v` | no |
| `[$subject] = …` / `list($subject) = …` (`Assign` with `Array_`/`List_` target) | no |
| `AssignRef` (`$subject = &$other`, or as the reference source) | no |
| `$subject++` / `++$subject` / `$subject--` / `--$subject` | no |
| `catch (Foo $subject)` | no |
| `static $subject` / `global $subject` / `unset($subject)` | no |

## Why this is filed on its own

This surfaced during #555's plan-review, where the identical blind spot was found in
`ReturnExpressionResourceReader::assertedModelClass()` and is being fixed there. The
planner then noticed `resolveVariable()` carries the same narrow predicate on its own
path and recorded it on #559.

**That placement understates it.** #559 is about a scan *window* in
`SingleReturnArrayLiteralFinder`. This is a *predicate* defect in
`Support\MethodBody\ReturnExpressionResolver` — shared infrastructure, live **today at
the current width 10**, independent of any window change and of the assert path. A
wrong-output bug should not ride along inside a widening issue where it can be scoped
away as secondary.

It is also strictly more severe than a coverage gap: the library emits a **confidently
wrong** schema rather than declining to emit one. That inverts the project's stated
contract of degrading rather than guessing.

## Blast radius

`ReturnExpressionResolver` is consumed by four readers, two of which call
`resolveVariable()` directly:

- `Plugins/ApiResources/Support/ReturnExpressionResourceReader.php:515` — response resources
- `Plugins/SpatieData/Support/DataReturnExpressionReader.php:81` — Data-class responses
- `Support/MethodBody/SingleReturnArrayLiteralFinder.php:57` — shared by resource
  `toArray()`, Fractal `transform()`, and `FormRequestStaticRulesReader` (**request bodies**)
- `Plugins/Core/Support/FormRequestStaticRulesReader.php:80` (via `methodLevelReturns`)

So it spans response schemas *and* request bodies, across Core, ApiResources, SpatieData
and Fractal.

Frequency is expected to be low — it needs a rebinding of the very local being returned —
but it is silent when it happens, and `foreach` reusing a variable name is not exotic.

## Proposal

Widen both predicates to **"any node that rebinds this name"**, covering the table above.
#555 is building exactly this table for `assertedModelClass()`; the two should share one
definition rather than growing two drifting copies — that duplication is the root cause
here.

Refuse (`NotAssignedOnce`, or a new refusal reason if the distinction is worth
surfacing) whenever a rebinding form is present, per the standing degrade-don't-guess
contract. Do not attempt to reason about which binding wins.

## Sequencing

Independent of #555 and #559 and safe to land before either. Coordinate with #555 so the
rebinding-form table lives in one place.

## Acceptance

- The `foreach`-rebind case above refuses instead of resolving.
- A test per rebinding form in the table.
- Existing single-assignment resolutions are unaffected (no coverage loss where the
  local genuinely is assigned once).
- Any resulting corpus movement is a **correction**, and any operation that stops
  resolving is verified as one that was previously resolving *wrongly*.

Radiergummi:
Closes #561

## Implementation plan — `ReturnExpressionResolver` rebinding blind spots

`ReturnExpressionResolver::resolveVariable()` decides a returned local is "assigned exactly
once, never mutated" by counting **only** plain `Assign`-to-`Variable` nodes, and its
`mutatesVariable()` helper recognises **only** `AssignOp` and `Assign`-into-`ArrayDimFetch`.
Every other rebinding form (`foreach ($xs as $v)`, `[$v] = …`, `$v = &$x`, `$v++`,
`catch (E $v)`, `static/global/unset $v`) is invisible, so a rebound local passes the guard
and the resolver returns the **first** assignment's expression — a **confidently wrong**
type. This inverts the project's degrade-don't-guess contract, and it spans response schemas
*and* request bodies across Core, ApiResources, SpatieData, and Fractal.

### Tier

**Tier 0.** Reflection/AST predicate widening; no dataflow, no value tracking. The fix is to
*refuse* when an unaccounted binding form is present, never to reason about which binding wins.

### The root cause is already fixed once — reuse it, don't re-derive

#555 (merged as PR #558, `d2c45fa`) introduced
**`src/Support/MethodBody/VariableRebinding.php`** — `VariableRebinding::matches(Node, string)`
— a single predicate that already recognises the *entire* table in this issue: plain/`&`
assignment, destructuring (`Array_`/`List_` targets), `AssignOp`, pre/post inc/dec, `Foreach_`
value **and** key targets, `Catch_`, `Static_`, `Global_`, `Unset_`. It is consumed today by
`ReturnExpressionResourceReader::rebindsBetween()` via
`StatementNodeFinder::findAll(..., IncludeConditionalContexts)`.

This is exactly the shared definition the issue asks for ("the two should share one definition
rather than growing two drifting copies"). **#561 wires `resolveVariable()` onto that same
predicate** — it introduces no new table. The issue's "widen both predicates" is realised as:
keep `mutatesVariable()` for the in-place element-write case `VariableRebinding` deliberately
does *not* cover, and add one rebinding gate built on `VariableRebinding`.

Note `VariableRebinding` matches plain `Assign`-to-`Variable` too, so a clean single-assignment
method yields exactly **one** match (the legitimate assignment); any *additional* match means an
extra binding form is present.

### The fix — `src/Support/MethodBody/ReturnExpressionResolver.php`

Rewrite `resolveVariable()` gate order so existing behaviour is preserved and the new forms
refuse:

1. Dynamic name → `DynamicallyNamedVariable`. (unchanged)
2. **Plain-assignment count** via the existing `$isAssignmentToVariable` predicate — require
   exactly one across all contexts *and* exactly one unconditional, else `NotAssignedOnce`.
   (unchanged — this still locates the single resolvable assignment expression.)
3. **In-place mutation** via the existing `mutatesVariable()` (unconditional `AssignOp` /
   `Assign`-into-`ArrayDimFetch`) → `MutatedAfterAssignment`. (unchanged — kept *before* the
   new gate so `$v += …` and `$v['k'] = …` keep their current refusal reason; those forms are
   not "rebindings" and `VariableRebinding` intentionally excludes element writes.)
4. **NEW rebinding gate**: count `VariableRebinding::matches` occurrences over the statements
   under `ConditionalContextPolicy::IncludeConditionalContexts` (a rebind inside an `if`/`foreach`
   body still voids the name — this mirrors `rebindsBetween`'s documented usage). If the count
   is `> 1`, refuse. The single permitted match is the plain assignment from gate 2.
5. Resolve to the single assignment's expression. (unchanged)

`mutatesVariable()` itself is **not** widened — the issue framed it as "widen both", but the
cleaner realisation given `VariableRebinding` is: element writes stay in `mutatesVariable`
(rebinding predicate excludes them by design), true rebindings go through the shared predicate.
This removes the duplication the issue names as the root cause without inventing a parallel table.

### Refusal reason — add one, don't overload

Add a new case to **`src/Support/MethodBody/ReturnVariableRefusal.php`**:
`ReboundAfterAssignment` — "assigned once, but the name is rebound afterwards by a `foreach`
target, destructuring, reference alias, increment/decrement, `catch`, or `static`/`global`/
`unset`". Rationale:

- The existing reasons are *misleading* for these forms: the name **is** assigned once by a
  plain `Assign`, so `NotAssignedOnce` reads as a lie, and it is not an in-place `Mutation`.
- The issue explicitly invites a new reason ("or a new refusal reason if the distinction is
  worth surfacing"), and #484/#208 (lint-readable inference-failure reasons) want this
  granularity. Cost is one enum case plus one `match` arm.

Wire the message in **`ReturnExpressionResourceReader::refusalReason()`** (the only consumer
that maps refusals to prose): add an explicit arm for `ReboundAfterAssignment`. The other three
consumers (`DataReturnExpressionReader`, `SingleReturnArrayLiteralFinder` →
`FormRequestStaticRulesReader`) only read `resolution->expression` and are unaffected by the new
case.

Note on increment/decrement: `$v++`/`--$v` are caught by gate 4 (not `mutatesVariable`), so they
report `ReboundAfterAssignment`. Acceptable — the value is stale either way and the refusal is
correct; the message honestly lists inc/dec among the forms.

### Accepted boundaries (documented, out of scope)

- **Reference-aliasing residual** (lead comment on #561): a write reaching the name through an
  alias established *before* the assert (`$v = &$x; … $x = new Cat();`) is through-reference
  cross-variable dataflow the Tier ladder refuses at Tier 2, pre-existing, and not worsened
  here. Left as a known boundary of `VariableRebinding`, per the lead. (Gate 4 *does* refuse the
  simpler `$v = &$x` case because `AssignRef` is a rebinding of `$v`.)
- **Nested-scope over-refusal**: under `IncludeConditionalContexts`, `NodeFinder` descends into
  closures/anonymous classes, so a same-named rebind inside a nested closure also triggers a
  refusal. This is a *safe* over-refusal (declines rather than emits wrong), it already exists on
  the plain-assignment count path, and it matches `rebindsBetween`'s behaviour. Not addressed.
- **Scan window**: `resolveVariable()` only sees the statements handed to it; a rebind beyond a
  caller's bounded window is that window's concern (#559), not this predicate's.

### Tests — `tests/Unit/Support/MethodBody/ReturnExpressionResolverTest.php` + fixture

TDD: add the failing `foreach`-rebind test first. Each new fixture method (in
`tests/Fixtures/ReturnExpressionResolverFixture.php`) has **one** plain unconditional assignment
of the returned local (so gate 2 passes) **plus** the rebinding form, then `return $subject;`,
and stays within the test scanner's `limit: 10`. Assert `expression` is null and `refusal` is
`ReboundAfterAssignment` for:

- `foreach ($xs as $subject)` — the canonical bug from the issue.
- `foreach ($xs as $subject => $v)` (key target).
- `[$subject] = …` and `list($subject) = …` (destructuring).
- `$subject = &$other` (reference alias).
- `$subject++` / `--$subject` (increment/decrement — one representative each).
- `catch (Exception $subject)`.
- `static $subject;` / `global $subject;` / `unset($subject);`.

Regression (must stay green, unchanged reasons):

- `singleAssignment` still resolves.
- `conditionalAssignment` / `multipleAssignments` / `noMatchingAssignment` → `NotAssignedOnce`.
- `elementWriteAfterAssignment` / `compoundAssignAfterAssignment` → `MutatedAfterAssignment`
  (proves gate 3 still precedes gate 4 for `AssignOp`/element writes).
- `dynamicallyNamedVariable` → `DynamicallyNamedVariable`.
- closure-return collection test unchanged.

### Observable-change surface & sequencing

- **Generation-affecting** (spans response schemas + request bodies) → the lead runs the
  **survey** after review. Acceptance per the issue: any corpus movement must be a
  **correction** — every operation that stops resolving must be verified as one that was
  previously resolving *wrongly*. Frequency is expected low (needs a rebinding of the very local
  returned).
- **CHANGELOG.md** `[Unreleased] → Fixed`: one line — resolver now refuses a returned local
  that is rebound (foreach/destructuring/reference/inc-dec/catch/static/global/unset) instead of
  emitting the first assignment's type (#561). No `docs/` page: this is internal inference
  plumbing; the degrade-don't-guess behaviour is already the documented contract.
- Independent of #555 (already merged) and #559; safe to land before #559/#564.

### Controversy

Low–moderate. No `Contracts` change, no stage-order change, no config key. Touches shared
`Support\MethodBody` infrastructure consumed by four readers and adds an `@internal` enum case
(`ReturnVariableRefusal` is `@internal`, not a public signature). The behavioural effect is
strictly a *correction* (wrong schema → no schema). Survey gate covers the corpus impact; expect
the lead to confirm no legitimate resolution regressed before merge.


Radiergummi:

Radiergummi:
🤖 **[lead]** plan posted → plan-review.

Radiergummi:
🤖 **[reviewer]** plan-review: approved.

Verified against the code, not just the prose:

- **No second rebinding table.** Gate 4 delegates to the merged `VariableRebinding::matches()` (`src/Support/MethodBody/VariableRebinding.php:42`), the same predicate `ReturnExpressionResourceReader::rebindsBetween()` already consumes (`:1020-1027`). The issue's "one definition, not two drifting copies" is satisfied.
- **Count baseline is correct.** `VariableRebinding` matches a plain `Assign`-to-`Variable` (via `bindsName`), so gate 2's single legitimate assignment is exactly the one permitted match; any extra binding form makes the count `> 1`. AssignRef is a php-parser sibling of `Assign` (not a subclass), so gate 2's `instanceof Assign` predicate (`ReturnExpressionResolver.php:74-77`) will not miscount `$v = &$x` — it correctly falls to gate 4.
- **Existing reasons preserved.** Keeping `mutatesVariable()` (gate 3, `:115-128`) ahead of the new gate means `$v += …` / `$v['k'] = …` still report `MutatedAfterAssignment`. Not widening `mutatesVariable` is the cleaner split (element-writes there, true rebindings via the shared predicate) and removes the duplication the issue names as root cause without inventing a parallel table.
- **New refusal wired at the sole prose consumer.** `refusalReason()` (`:604-617`) is the only mapper; its `default =>` arm currently swallows unknown cases into the misleading "not assigned exactly once" text, so an explicit `ReboundAfterAssignment` arm is required, not optional. The other three consumers only read `resolution->expression` and are unaffected.
- **Tier-0.** Predicate widening, no dataflow/value-tracking; the fix refuses rather than reasons about which binding wins. Correct rung.
- **Boundaries correct.** The reference-aliasing residual (`$v = &$x; … $x = new Cat();`) is genuinely Tier-2 through-alias dataflow — `VariableRebinding(node, 'v')` cannot see the write to `$x` — pre-existing and not worsened. Correctly left out of scope per the lead. Nested-scope over-refusal is a safe decline that already exists on the assignment-count path.
- **TDD + bookkeeping.** Failing `foreach`-rebind test first; one fixture per table row with a single plain assign so gate 2 passes; regression set pins the unchanged reasons (esp. `MutatedAfterAssignment` proving gate 3 precedes gate 4). CHANGELOG `[Unreleased] → Fixed`; no `docs/` page needed (internal inference plumbing; degrade-don't-guess is already the documented contract).

**Nit (non-blocking), for the coder to consider at implementation:** a *conditional* compound assignment (`if (…) { $v += 1; }`) is not caught by the unconditional gate 3 and will now refuse via gate 4 as `ReboundAfterAssignment` — a correct new refusal (it resolved wrongly before), but the proposed message text lists foreach/destructuring/reference/inc-dec/catch/static/global/unset and omits compound-assignment. Either add "compound assignment" to the message wording or accept the minor imprecision (it is an `@internal` diagnostic note, not spec output). Not a merge blocker.

Radiergummi:
🤖 **[lead]** plan-review approved → coding.

Radiergummi:
🤖 **[coder]** deviation: two adjustments to the approved plan, forced by existing contracts.

**1. Gate 4 excludes `AssignOp` (compound assignment).** The plan had gate 4 count *every* `VariableRebinding::matches` occurrence under `IncludeConditionalContexts`. But `VariableRebinding` also matches `AssignOp`, so that refused a **conditional** `$v += […]` — regressing the #446 conditional-merge contract, which two existing tests pin:

- `SingleReturnArrayLiteralFinderTest` → "resolves the base literal when later writes to the variable are conditional" (fixture `conditionalMergeAfterLiteral`)
- `ResourceToArrayInferenceTest` → `ConditionalMergeArticleResource`

Both expect a *conditional* compound-assign to still resolve to the base literal (never-wrong subset). Compound assignment is the **additive-mutation** case gate 3 (`mutatesVariable`) already owns: unconditional → `MutatedAfterAssignment`, conditional → allowed. So gate 4 carves it out — `!$node instanceof AssignOp && VariableRebinding::matches(...)` — and covers only the **value-replacing** rebindings gate 3 misses (foreach/destructuring targets, reference alias, inc/dec, catch, static/global/unset). Element writes (`$v['k']=`) are already excluded by `VariableRebinding` itself. This still reuses the one shared predicate — no second table.

**2. Reviewer-2 nit re. "compound assignment" wording — respectfully diverging.** Because of (1), a compound assignment can *never* route to `ReboundAfterAssignment`; it stays `MutatedAfterAssignment` (unconditional) or resolves (conditional). Putting "compound assignment" in the `ReboundAfterAssignment` message would therefore be inaccurate, so I dropped it. The message lists exactly the forms that produce the reason: foreach target, destructuring, reference alias, increment/decrement, catch, static/global/unset. Happy to revisit if you had a different routing in mind.

**3. Dropped two integration fixtures** (predicate coverage already authoritative in `VariableRebindingTest`, and CLAUDE.md forbids PHPStan ignores):
- `list()` destructuring — Pint's `list_syntax` fixer rewrites `list($x)` to `[$x]` in real fixtures, making it byte-identical to the array-destructuring case; the `List_` node path is covered by `VariableRebindingTest` ("matches list() destructuring", via a parsed source string).
- `unset` — returning an `unset` local is a genuine PHPStan `variable.undefined`. Its predicate branch (`Global_ || Unset_`) is identical to `globalRebind`'s (kept) and node-covered by `VariableRebindingTest`, so no gate-4 branch goes untested.

The CHANGELOG still lists `unset` as refused at runtime (accurate — only the fixture is dropped).

Radiergummi:
🤖 **[coder]** implemented; CI green locally (Pint + PHPStan L8 + full suite, 3616 passed) and remote test matrix + quality green; pushed bb516f4c. Ready for review.

Radiergummi:
🤖 **[lead]** implemented (bb516f4c) → code-review (reviewer-1 + reviewer-2, independent). Survey gate after approvals.

Radiergummi:

Reviewed the pushed diff (bb516f4c) independently. Correctness, tests, conventions, boundaries, and bookkeeping all hold.

**Deviation 1 (gate 4 excludes `AssignOp`) — correct, and it refutes my own plan-review nit.** I had claimed a conditional compound-assign "resolved wrongly before" so refusing it would be a correction. That was wrong: a conditional `$v += […]` is the deliberate #446 never-wrong-subset contract (base literal kept, conditional additions unread), pinned by `conditionalMergeAfterLiteral` (`SingleReturnArrayLiteralFinderTest:95`) and `ConditionalMergeArticleResource`. Refusing it via gate 4 would have regressed both. The carve-out `!$node instanceof AssignOp && VariableRebinding::matches(...)` is the right split: compound assignment stays gate 3's additive-mutation case (unconditional → `MutatedAfterAssignment`, conditional → resolve), while gate 4 owns only the value-replacing rebindings gate 3 misses. Verified it loses no coverage — inc/dec are `PreInc`/`PostInc`/… not `AssignOp`, so they still route through gate 4; element writes (`$v['k']=…`) were never `VariableRebinding` matches and correctly remain in gate 3. Ran the suite: `conditionalMergeAfterLiteral` and the ApiResources conditional-merge test both stay green.

**Deviation 2 (drop "compound assignment" from the `ReboundAfterAssignment` message) — accepted, follows necessarily from (1).** Since a compound assignment can never route to `ReboundAfterAssignment`, listing it would be inaccurate. This resolves my nit better than my original suggestion.

**Deviation 3 (drop `list()` / `unset` integration fixtures) — accepted.** Both predicate branches are node-covered in `VariableRebindingTest`: `list($subject) = $pair;` ("matches list() destructuring") and `unset($subject);` ("matches static, global, and unset statements"), via parsed source strings so Pint's `list_syntax` fixer cannot rewrite them. The `unset` gate-4 branch (`Global_ || Unset_`) is identical to `global`'s, whose fixture (`globalRebind`) is kept, so no gate-4 branch goes untested. Justified by the Pint-rewrite / PHPStan `variable.undefined` / no-ignores constraints.

**Verified independently:**
- Gate order: new rebinding gate sits after `mutatesVariable` (gate 3), so `$v += …` / `$v['k']=…` keep `MutatedAfterAssignment`.
- Count baseline: `VariableRebinding` matches the single plain `Assign`, so a clean method yields exactly one match; `> 1` ⇒ an extra rebinding. `AssignRef` is a php-parser sibling of `Assign`, so gate 2 doesn't miscount `$v = &$x` — it falls to gate 4 (confirmed by `referenceAliasRebind` → `ReboundAfterAssignment`).
- `refusalReason()` has an explicit `ReboundAfterAssignment` arm (no longer swallowed by `default`).
- Tests: one case per rebinding form (foreach value/key, destructuring, reference, increment, decrement, catch, static, global) — all assert null expression + `ReboundAfterAssignment`. Full relevant suite green (64 passed across resolver/rebinding/finder/ApiResources).
- Conventions: strict-types intact, no abbreviations, comment explains the *why* (AssignOp carve-out rationale) with no issue numbers in code, ≤100 cols, `ReturnVariableRefusal` is `@internal`. Pint passes on the changed files; PHPStan/full-suite reported green by coder + remote.
- Boundaries: change is entirely in `Support\MethodBody` (+ the one plugin prose consumer); no plugin/host-app coupling. CHANGELOG `[Unreleased] → Fixed` present (#561); no docs page needed (internal plumbing, `@internal` refusal enum).

**Optional (non-blocking) test suggestion:** a positive over-refusal guard would round out the suite — a method that assigns the returned local once *and* rebinds an **unrelated** variable, e.g. `$subject = x(); foreach ($items as $other) {} return $subject;`, should still **resolve** (proves the name-filtered count doesn't false-positive). It is covered transitively today (`VariableRebindingTest` proves the predicate returns false for other names, and gate 4 filters by name), so this is a nice-to-have, not a requirement.

Approving.

Radiergummi:

Independent code-review (I am the second reviewer; reviewer-2 owned plan-review). Verified against a fresh checkout of `bb516f4c`, not just prose.

**Correctness — sound.**
- Gate order (dynamic → assigned-once → unconditional-mutation → rebinding → resolve) is right. The `> 1` baseline is sound: gate 2 guarantees exactly one plain `Assign`-to-`Variable`, which `VariableRebinding::matches` also matches once, so any *additional* rebinding form pushes the count past 1.
- `VariableRebinding` covers the full table (foreach value+key, `Array_`/`List_` destructuring incl. nested/keyed, `AssignRef` both target and reference source, inc/dec, `Catch_`, `Static_`, `Global_`, `Unset_`); element writes are excluded by the predicate itself, so gate 3 keeps sole ownership of `$v[..] = ..` / `$v += ..`.
- Edge cases hold: a *conditional* plain reassign is caught by gate 2 (`NotAssignedOnce`); a rebind nested in an `if`/`foreach` body still voids the name via `IncludeConditionalContexts`.

**The `!$node instanceof AssignOp` carve-out is load-bearing and pinned.** I ran a mutation test — removing it turns `2` existing tests red (`SingleReturnArrayLiteralFinderTest` "resolves the base literal when later writes are conditional" and `ResourceToArrayInferenceTest` `ConditionalMergeArticleResource`), confirming the coder deviation note: a *conditional* compound-assign must still resolve to the never-wrong base (the #446 contract). Not a gap.

**Tests — TDD followed, no material gaps.** 9 new refusal cases, each with a one-plain-assignment + one-rebinding fixture; regressions preserved. The two dropped resolver-level fixtures (`list()`, `unset`) are justified: `list()` is byte-rewritten to `[$x]` by Pint (`list_syntax`) so it collapses onto the array-destructuring case, and returning an `unset` local is a real PHPStan `variable.undefined` (CLAUDE.md forbids ignores). Both node paths are authoritatively covered in `VariableRebindingTest`, and `unset` shares its predicate branch with the kept `globalRebind`, so no gate-4 branch is untested.

**Gates (ran locally in a clean worktree):** full suite **3616 passed**, PHPStan L8 **no errors**, Pint **passed**.

**Conventions/boundaries/bookkeeping:** strict-types, no abbreviations, concise why-comments (no issue numbers in code), within line width; new `ReboundAfterAssignment` case is `@internal` (no public-signature change); everything lives in `Support\MethodBody` with the prose mapping in the correct plugin consumer (`ReturnExpressionResourceReader::refusalReason()`); CHANGELOG entry under `### Fixed`; no `docs/` page needed (internal inference plumbing, degrade-dont-guess already documented). Deviations documented in the coder PR comment as required.

Generation-affecting → the survey gate (corpus movement must be a *correction*) is the lead's remaining step before merge.

Radiergummi:
🤖 **[lead]** two independent approvals (reviewer-1 + reviewer-2) → survey gate. Corpus movement must be a *correction*: every operation that stops resolving must be verified as one that was previously resolving WRONGLY.

Radiergummi:
🤖 **[lead]** ⏸️ **paused (org monthly spend limit hit — team agents halted).** Resume state:
- Both reviewers approved (reviewer-1 + reviewer-2), full matrix CI green @ `bb516f4c` (incl. swagger-php 5.8, snapshot).
- **Blocking gate: Layer A corpus survey never completed** (surveyor was killed mid-run; no verdict posted). This is a noise-removal/correction fix — the survey must confirm every operation that STOPS resolving was previously resolving WRONGLY (item-level diff, not aggregate counts).
- Behind `main` by #573 (lint-only, non-generation). Before merge: run/pass survey → rebase onto main → re-watch CI → merge. #575 and #576 are stacked on this branch and must merge AFTER it.
- To resume: `/autonomous-team resume #574 from survey` (once spend limit raised).

Radiergummi:
🤖 **[lead/survey]** Layer A PASS — driven solo (org spend limit halted the surveyor agent). Baseline `8230408c` (generation-equivalent to current main; #573 is lint-only) vs candidate `27fe721c` (= #574 + #559, no guard). **0 change across all 11 apps × 8 watched metrics; gen_exit=0 everywhere.** Spec-level confirmation: Vito byte-identical; InvoiceNinja (529 paths) semantically identical after masking the non-deterministic `operationId` hash (the 89 raw path diffs were all operationId RNG). The fixes close latent confidently-wrong paths no corpus app triggers; **FindReturnModel shows 0 null→model movements** (#578 concern does not manifest → merging without the guard is corpus-safe). No regressions, no compensating swaps.

Radiergummi:
🤖 **[lead]** two independent approvals + full matrix CI green (bb516f4c) + Layer A survey PASS (0 corpus change) + non-controversial (resolver/rebinding fix, disjoint from the intervening lint-only #573, GitHub reports CLEAN) → squash-merged. Closes #561.

PR #576, opened 2026-07-23
fix(plugins): widen DataReturnExpressionReader's return-scan window to a 100-statement backstop

### LINKED ISSUE #564: fix(plugins): DataReturnExpressionReader has the same STATEMENT_LIMIT=10 window + truncated-return-count bug as #555

Radiergummi:
Closes #564

## Implementation plan — `DataReturnExpressionReader` STATEMENT_LIMIT + truncated-return-count

`src/Plugins/SpatieData/Support/DataReturnExpressionReader.php` is the fourth site of the defect
#555 fixed in `ReturnExpressionResourceReader`:

- **Window too small** — `STATEMENT_LIMIT = 10` (`:41`), applied via `firstStatements($method, …)`
  (`:73`). A `return` past the 10th top-level statement is never seen, so a well-typed Data-class
  response goes unresolved.
- **Truncated return count** — `canonicalReturnExpression()` (`:99`) takes the first top-level
  `return` (`break`), then guards with `count(methodLevelReturns($statements)) > 1` (`:113`) over
  the **truncated** list. A method returning at statement 3 and again at statement 15 counts as
  one return in a 10-statement window, passes the guard, and resolves from the first — a
  **confidently wrong** schema, the latent-unsoundness half.

### Tier

**Tier 0.** Mirror #555/#558/#559: separate finding the return from bounding pattern matching,
raise the window to a stated pathological-input backstop, keep the "exactly one unconditional
return" guard. No new inference.

Crucially, the truncated-count bug is closed **as a consequence of widening**: the existing
`count > 1 → null` guard is already the correct soundness check; it was only unsound because it
ran over a truncated list. With the full body visible, the invisible second return is counted and
the method correctly refuses. No new guard logic is needed (same as #555).

### HARD DEPENDENCY — must land after #561 (PR #574)

`resolve()` calls `resolveVariable()` for the `return $variable;` case (`:80`). Widening the
window makes more statements visible to that resolver, including rebindings (`foreach`,
destructuring, …) previously beyond statement 10. On today's resolver those are not refused (that
is #561), so **landing #564 before #561 would expand the confidently-wrong surface**. Composition
is safe once #561 makes `resolveVariable()` refuse rebindings.

- **No code change needed to consume #561**: `resolve()` reads only `...->resolveVariable(...)->
  expression` (null on refusal → degrade). The new `ReboundAfterAssignment` reason flows through
  transparently.
- Dependency is **ordering + rebase**: coding rebases onto #561, this PR must **not merge before
  #574**, and the survey runs on the composed state. Marked `blocked-by` #561 in the task list.
- **Independent of #559** — different file (SpatieData plugin), no overlap with PR #575.

### Shared-constant decision (the issue asks to decide once, across the family)

Four readers now carry a copy of the return-scan window. The issue floats hoisting it into one
shared place in `Support\MethodBody`.

**Decision for THIS PR: per-site `RETURN_SCAN_STATEMENT_LIMIT = 100`**, matching the name and
value #558 already merged (`ReturnExpressionResourceReader:97`) and #559 planned (PR #575). Reasons:

- #559's planning did **not** adopt a shared bound (it followed #558's per-site precedent), so the
  issue's "if #559 adopts a shared bound, join it" condition is **not** triggered.
- Introducing a shared constant here would leave the family half-shared (three private copies +
  one shared) unless #564 also retrofits the other three readers — which **collides** with in-flight
  PRs #574 (#561) and #575 (#559) and with #558's merged code. That is exactly the mid-flight
  collision the planner dependency pre-check exists to avoid.

**Recommend a separate consolidation follow-up** (filed for after #558/#559/#561/#564 all land):
hoist a single `RETURN_SCAN_STATEMENT_LIMIT` into `Support\MethodBody` (beside `VariableRebinding`
/ `MethodBodyScanner`) and point all four readers at it — the durable root-cause fix, done once
nothing is in flight. This answers the issue's "worth deciding once" without a coordination mess.
Flagged to the lead.

### Reconciliation is out of scope

`ReturnExpressionResourceReader` has `reconcileMultipleReturns()` (resolves multiple returns when
they agree). The issue's acceptance permits "refuses **or** reconciles". This reader currently
**refuses** on `> 1` return, and refusing is the minimal, correct fix here — the wrong-resolution
bug is closed by widening + the existing guard. Adding reconciliation is a new feature beyond the
bug and is **not** in scope; note it as a possible future parity enhancement.

### Files to modify

1. **`src/Plugins/SpatieData/Support/DataReturnExpressionReader.php`**
   - Rename `public const int STATEMENT_LIMIT = 10;` → `RETURN_SCAN_STATEMENT_LIMIT = 100`
     (nothing external aliases it — verified via grep — so the rename is internal-only; keep
     visibility as-is).
   - Update the `firstStatements($method, self::…)` call (`:73`).
   - Update the class docblock (`:30`) that references `{@see self::STATEMENT_LIMIT}` — restate the
     backstop rationale (#558's wording): pathological-input backstop, not a semantic bound; the
     guarantee is the single-unconditional-return guard + `resolveVariable()`'s refusals; set far
     above ordinary method length.

### Tests — `tests/Unit/Plugins/SpatieData/DataReturnExpressionReaderTest.php`

Fixtures are inline parse-only methods on `ReaderFixture` in the test file; add there. TDD — the
failing "return past 10 resolves" test first. Keep them PHPStan-clean (valid, type-correct
bodies) and ensure filler statements do **not** rebind the returned local.

- **fires now (was null)** — `returnPastTenStatements()`: >10 straight-line statements, then a
  trailing `return new ScalarOnlyData(...)` (and a `::collect(...)` variant) → resolves.
- **truncated-second-return refuses** (the wrong-on-current-main case, mirroring #555's
  `truncatedThirdReturn`): a first top-level return within the old window and a **second**
  top-level return past statement 10. On current main it resolves from the first (bug); after the
  fix `methodLevelReturns` sees both → `count > 1` → **null**. This is the acceptance's key
  negative test.
- **backstop** — a return past **100** statements still refuses (proves the backstop is real).
- **rebind negative (composition with #561)** — `$data = new ScalarOnlyData(...); foreach (… as
  $data) {} return $data;` → null on the rebased branch (guards window-widening + #561 together).
- **regressions preserved** — all existing cases stay green: `collection`, `single`,
  `assignedSingle`, `nonData`, `degrade`, `conditional` (multiple returns → null),
  `elementWriteAfterAssignment` (mutation → null).

### Docs

`docs/plugins.md:39` — the SpatieData section states the return-expression path is "a bounded scan
of the first 10 statements". This **is** the changed reader, so update it to the whole-body /
100-statement-backstop framing, parallel to the resource-reader doc #558 added at `plugins.md:195`
("reads the whole method body, bounded only by a 100-statement backstop"). No other doc mentions
this reader's window.

### CHANGELOG

`[Unreleased] → Fixed`: a Spatie Data-class response whose `return` sits past the first 10
statements now resolves (bounded by a 100-statement backstop), and a method with a
previously-invisible second return past the window now correctly refuses instead of emitting the
first return's type. (#564)

### Survey & acceptance

Generation-affecting (SpatieData responses) → survey after review, **on the #561-rebased branch**.
Acceptance: the past-10 case resolves; the truncated-second-return case refuses (never resolves
from the first); existing single-return resolutions unaffected. Any corpus movement is a
correction (monotone-refusing).

### Controversy

Low–moderate. Single-file behavioural fix in a plugin + one doc line + CHANGELOG; no `Contracts`
change, no stage order, no config key. The **ordering gate** (merge after #561/#574) and the
**shared-constant follow-up recommendation** are the two things for the lead to note.


Radiergummi:

Two flags for the reviewer/lead:
- **Hard ordering gate**: must merge **after #561/#574** (calls `resolveVariable()`; widening surfaces rebindings the resolver only refuses once #561 lands). No code change needed to consume #561; ordering + rebase only. Independent of #559 (different file).
- **Shared-constant decision**: kept the per-site `RETURN_SCAN_STATEMENT_LIMIT=100` (consistent with #558 merged + #559 planned) rather than hoisting a shared constant here — doing so mid-flight would leave the family half-shared and collide with #574/#575. **Recommend a separate consolidation follow-up** (after #558/#559/#561/#564 all land) to hoist one `RETURN_SCAN_STATEMENT_LIMIT` into `Support\MethodBody` and point all four readers at it — the durable root-cause fix the issue asks for.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Hard ordering gate: merge AFTER #574 (#561). Per-site constant now + consolidation follow-up filed separately (approved).

Radiergummi:
🤖 **[reviewer]** plan-review: approved. Verified against the code (reviewer-2).

- **Both defect halves correct.** `STATEMENT_LIMIT = 10` (`:41`, applied `:73`); `canonicalReturnExpression()` takes the first top-level return then guards `count(methodLevelReturns) > 1` (`:113`) over the **truncated** list. The plan is right that the truncated-count bug is closed **as a consequence of widening** — the `count > 1 → null` guard is already the correct soundness check; it was only unsound over a truncated list. With the full body visible, the invisible second return is counted and the method refuses. No new guard logic, matching #555/#558/#559. Correct.
- **Tier 0, correct rung.** Separate finding-the-return from bounding pattern-matching; raise to a stated pathological-input backstop; keep the single-unconditional-return guard.
- **Hard dependency on #561/#574 is correct — and cleaner here than #559.** `resolve()` reads only `resolveVariable(...)->expression` (`:80-81`), null on refusal → degrade, so #561 flows through with no code change. Critically, I confirmed `resolve()` has **no** `resolveVariable`-bypass path (the single variable resolution is `:81`; there is no second literal-scan fallback like #559's `variableReturnLiteral`). So #561 **fully** covers the variable case here, and the rebind-negative test (`$data = new …; foreach (… as $data) {} return $data;` → null) holds on the rebased branch **without** any option-b-style extra guard. The gap I flagged on #559 does not recur.
- **Shared-constant decision is sound.** Per-site `RETURN_SCAN_STATEMENT_LIMIT = 100` (matching merged #558 `ReturnExpressionResourceReader:97` and planned #575), with a deferred 4-way consolidation follow-up. Consolidating now would half-share the family and collide with in-flight #574/#575 and merged #558 — exactly the mid-flight collision to avoid. The issue's "decide once" ask is answered by an explicit decision + follow-up, not silent widening. **One ask:** ensure the consolidation follow-up is actually filed (it is the durable root-cause fix #564 names).
- **Verified facts:** nothing aliases `DataReturnExpressionReader::STATEMENT_LIMIT` (grep clean) → internal-only rename, no collision; `docs/plugins.md:39` does describe this reader's "first 10 statements" scan → correct to update, and no other doc mentions it.
- **Reconciliation out of scope** — refusing on `> 1` is the minimal correct fix; the issue permits "refuses or reconciles". Agreed.
- **Tests mirror #555** — past-10 resolves, the truncated-second-return refusal (the key wrong-on-current-main negative), the 100-backstop, the #561 rebind-negative, and the preserved regressions. TDD. CHANGELOG `[Unreleased] → Fixed` present.
- **Independent of #559** (different file, no overlap with #575) — correct; only the #574 ordering gate applies.

No changes requested. Ready for coding, merge-gated after #574 and the survey.

Radiergummi:
🤖 **[lead]** plan-review approved → coding. Consolidation follow-up already filed as #577. Stack on #574 for the rebind-negative test; merge after #574.

Radiergummi:
🤖 **[lead]** ⏸️ **paused (org monthly spend limit — team halted).** Resume state:
- Plan-review approved by reviewer-2. **Code implemented @ `9b0bbe30`** (window fix, no option-b guard needed — single resolveVariable route), full matrix CI green.
- **Not yet code-reviewed** — needs two independent code approvals (reviewer-1 + reviewer-2) on the #564-specific delta (stacked on #574; review `bb516f4c..9b0bbe30`).
- Stacked on #574 → **merge after #574**; rebase onto main once #574 lands. Independent of #559/#575 (different file).
- Blocking gates: code-review (2 approvals) → survey → rebase → merge. Consolidation of the shared constant is deferred to follow-up #577.

Radiergummi:
🤖 **[lead]** update: **#574 and #575 have now merged to main** (`e9e41074`). This PR (#564) is unblocked but was **left for the team** — the solo drive under the spend limit covered only #574/#575 (this needs two independent code reviews I can't produce solo).

**Resume checklist for #576:**
1. Rebase onto current main — it's stacked on the now-merged #574's old base (`bb516f4c`); use `git rebase --onto origin/main <old-#574-tip>` to drop the already-merged commits, keeping only the #564 delta (DataReturnExpressionReader window rename). Expect only a CHANGELOG merge.
2. Two independent code reviews (reviewer-1 + reviewer-2) of the #564 delta — plan already approved by reviewer-2; no option-b guard needed (single resolveVariable route).
3. Survey gate (its own — different reader). Note: the #574/#575 survey showed **0 corpus change**; #564 (SpatieData reader) is likely similar but must be measured on its own.
4. Merge. Consolidation of the shared `RETURN_SCAN_STATEMENT_LIMIT` constant across all four readers is tracked by **#577**.

Radiergummi:
🤖 **[lead/coder]** Re-implemented #564 (the WIP was lost as an empty commit; the real diff was sitting uncommitted in the worktree). Rebased onto current main (`e9e41074`), one clean commit `daced2b9`.

**Delta:** `DataReturnExpressionReader` STATEMENT_LIMIT=10 → `RETURN_SCAN_STATEMENT_LIMIT=100` backstop (mirrors #575), + one test proving a return past statement 10 now resolves.

**Deviation from the WIP:** dropped the `twoReturnsSecondPastWindow` fixture/test the WIP added — it relied on unreachable code (PHPStan level-8 `deadCode.unreachable`, CI-blocking) and was non-differentiating: this reader's canonical return is always *unconditional* (hence the last top-level statement), so a divergent second return past the window is not constructible in PHPStan-clean code and can never turn a resolution into a wrong one (a past-window canonical return already yields null). CHANGELOG wording adjusted to the honest window-widening claim. Full matrix pending; local `composer check` green (pint, phpstan, 3646 tests). Ready for two independent code reviews of `e9e41074..daced2b9`.

Radiergummi:
🤖 **[lead]** CI correction: the first push (`daced2b9`) was the pre-amend commit that still carried the dead-code test — the `quality` (PHPStan) job failed on `deadCode.unreachable` as expected; the full test matrix + swagger-php 5.8 + Snapshot all passed. Re-committed with the dead-code test removed → `7b80612d`, force-pushed, re-watching CI. Reviewers: the current diff (`origin/main...7b80612d`) no longer contains `twoReturnsSecondPastWindow`.

Radiergummi:
🤖 **[reviewer-1]** review — **APPROVE** (HEAD `00877416`). Diff is exactly the const rename + one CHANGELOG line + the `fourteenStatementsThenNewData` test. Local gates green (tests 8✓, Pint✓, PHPStan L8 no errors). Widening 10→100 is monotonic-safe: `$topLevelReturn` is position-independent of window size, so a wider window can only surface a previously-out-of-window trailing single return (resolve it) or surface an extra return and refuse via `methodLevelReturns()>1` — never turn a correct resolution wrong. Dropping `twoReturnsSecondPastWindow` loses no coverage (the multi-return refusal is already covered by the `conditional` test) and was un-committable (PHPStan deadCode.unreachable). Non-blocking nit: const is `private` here vs sibling #575's `public`; both `@internal`, resolved either way by #577.

Radiergummi:
🤖 **[reviewer-2]** review — **APPROVE** (HEAD `00877416`). Re-fetched, re-diffed, re-ran all gates against this HEAD: PHPStan no errors, Pint passed, 8/8 tests. The widened 10→100 scan is untouched by the follow-up; only the un-committable dead-code fixture was removed. Soundness verdict from first pass stands: widening is monotonic-safe (recovers correct misses or refuses on count>1; never turns a correct resolution wrong). Clean approve.

Radiergummi:
🤖 **[surveyor]** Layer A **PASS** — 0 regressions across 11 apps vs a freshly-regenerated `main` baseline (`e9e41074`). Every watched deterministic metric (paths, operations, apiOperations, responseSchemas, documentedResponses, requestBodies, componentSchemas, operationsWithSecurity) is **fully flat** candidate (`00877416`) vs baseline; no new gen crashes. Expected: the corpus has no Spatie Data return sitting behind >10 guard clauses, so the widened window changes nothing — matching the sibling #574/#575 0-delta result. Additive-only change confirmed in the field.

Radiergummi:
🤖 **[lead]** all gates green → squash-merged. **Two independent approvals** (reviewer-1 + reviewer-2 on `00877416`), **full CI matrix green** (PHP 8.4/8.5 × Laravel 12/13 + swagger-php 5.8 + Snapshot + quality/PHPStan), **survey Layer A 0-regression** across 11 apps vs fresh `main` baseline. Non-controversial: an `@internal` const rename replicating the already-merged sibling #575 (no Contracts, no config key, no public signature change; ~20-line src diff). Constant left `private` per both reviewers; #577 unifies it across all four readers. Closes #564.

PR #579, opened 2026-07-24
chore(core): hoist one shared ReturnScan::STATEMENT_LIMIT across the four return-scan readers

Radiergummi:
Closes #577

## Implementation plan — #577

Pure, byte-identical consolidation refactor. **Tier 0** (reflection/signatures only; no body-parsing
logic changes — we only relocate a constant). No behaviour change: snapshot group must be byte-identical
and the survey flat.

### The design decision: where the single constant lives

Post-#576 `main` has four independent `private/public const int RETURN_SCAN_STATEMENT_LIMIT = 100;`
definitions (each with a near-identical #558 soundness docblock) plus one alias:

| Site | Current |
|------|---------|
| `Support\MethodBody\SingleReturnArrayLiteralFinder` | `public const … = 100` (canonical docblock) |
| `Plugins\ApiResources\Support\ReturnExpressionResourceReader` | `private const … = 100` |
| `Plugins\Core\Resolvers\FindReturnModelResponseResolver` | `private const … = 100` |
| `Plugins\SpatieData\Support\DataReturnExpressionReader` | `private const … = 100` |
| `Plugins\Core\Support\FormRequestStaticRulesReader` | `public const … = SingleReturnArrayLiteralFinder::RETURN_SCAN_STATEMENT_LIMIT` (alias) |

**Decision: a new tiny dedicated holder `Support\MethodBody\ReturnScan` with
`public const int STATEMENT_LIMIT = 100`**, carrying the canonical #558 soundness docblock. All four
readers + the FormRequest site reference `ReturnScan::STATEMENT_LIMIT`.

Why this host over the alternatives:

- **Must live in `Support/` and be `public`** — it is referenced from Core, ApiResources and SpatieData
  plugins, and `Support/` may not depend on any plugin (CLAUDE.md). ✓
- **Intention-revealing at every call site.** `firstStatements($method, ReturnScan::STATEMENT_LIMIT)`
  reads as exactly what it is at all four sites. No reader "borrows" a peer's identity — the current
  `FormRequestStaticRulesReader → SingleReturnArrayLiteralFinder::RETURN_SCAN_STATEMENT_LIMIT` alias is
  precisely the incidental coupling we want to retire, not propagate to three more sites (why would
  SpatieData's `DataReturnExpressionReader` reference an "array-literal finder" for its scan bound?).
- **Not `MethodBodyScanner`** — the scanner is deliberately concept-agnostic: it takes the limit as a
[...]

PR #580, opened 2026-07-24
fix(core): FindReturnModelResponseResolver refuses when a method has more than one directly-returned lookup

### LINKED ISSUE #578: fix(core): FindReturnModelResponseResolver emits the FIRST directly-returned lookup call, not requiring a single return
## Problem

`FindReturnModelResponseResolver::directlyReturnedLookupCall()` returns the **first** directly-returned lookup call (`Model::find()`/`::findOrFail()`/etc.) it encounters and does **not** require the method to have a single return — unlike its sibling return-scan readers, which refuse on ambiguity.

This is a **pre-existing** confidently-wrong-emission risk that #559 (PR #575) *amplified*: widening the return-scan window from 10 → 100 statements (`RETURN_SCAN_STATEMENT_LIMIT`) makes edge-path lookup calls past statement 10 newly visible. A method whose real primary return is elsewhere but which has e.g. `return LegacyRecord::find($id)` on an error/legacy branch past statement 10 will now emit that **edge model** as THE response schema (a `null→model` movement), where the narrow window previously suppressed it.

Surfaced in reviewer-1's code-review of PR #575 (issue #559). It was correctly judged **out of scope** for #559 (whose acceptance reasons only about operations that STOP resolving; this is the one direction — START resolving — that the monotone wrong-resolve→refuse correction does not cover) and deferred here.

## The fix

Add a **targeted** guard: resolve only when there is **exactly one** directly-returned lookup call in the method; if more than one, refuse (return null), mirroring how the sibling readers refuse on ambiguity.

Do **NOT** use a blunt single-return guard — that would regress legitimate cache/fast-path patterns where a lookup return coexists with other non-lookup returns.

## Verification

- Multi-return test: a method with two directly-returned lookup calls must refuse.
- The #559/#575 survey should be checked for any `null→model` FindReturnModel movements introduced by the window widening; those are the concrete corpus instances this guard addresses.

## Acceptance

- `FindReturnModelResponseResolver` refuses when >1 directly-returned lookup call is present.
- Single-lookup fast-path/cache patterns still resolve (no regression).
- No confidently-wrong edge-model emission from multi-return methods.

Radiergummi:
Closes #578

## Implementation plan — #578

**Tier 0** (signature/AST-shape reflection; no new dataflow). Bug fix: add a targeted ambiguity guard
to `FindReturnModelResponseResolver`. **Intentional-reduction** change — some corpus apps lose a
wrong edge-model `responseSchemas` emission; that is the fix working, not a regression.

Branches off current `main` (already carries #579's `ReturnScan::STATEMENT_LIMIT`, so no file
collision — #579 merged as `ad15e52f`).

### The bug

`FindReturnModelResponseResolver::directlyReturnedLookupCall()` returns the **first**
directly-returned lookup call (`Model::find()`/`findOrFail()`/`firstOrFail()`) and never checks whether
the method has *other* return paths. #575 widened the scan window 10→100, newly surfacing lookup
returns past statement 10, so a method with e.g. `return LegacyRecord::find($id)` on a legacy/error
branch alongside its real primary return can now emit that edge model as THE response (`null→model`).

### The one design decision: what set do we count over?

Count **directly-returned lookup calls, including those in conditional branches** (not top-level only).

Rationale (this is load-bearing — reviewer please confirm):
- **Issue-faithful.** The issue's own scenario is a lookup on an *error/legacy branch* (a nested,
  conditional `return`). A top-level-only count would not see it, so the guard would fail to fix the
  reported bug.
- **Matches the lead's fast-path example.** The pattern to preserve —
  `if ($cached) return $cached; return Model::find($id);` — has its coexisting non-lookup return
  (`return $cached`) *inside a conditional*. So the count is over lookup returns across branches; the
  fast-path survives because `return $cached` is a variable, **not a lookup call**, and is never
  counted. (The lead's "top-level returns" phrasing is reconciled here: we count directly-returned
  *lookup* calls wherever a return sits in the method's control flow.)
- **Avoids dead-code fixtures.** Two *top-level* lookup returns can only exist as dead code
  (`return A; return B;`), which PHPStan level 8 (it analyses `tests/` too) would flag as unreachable
  and fail `composer lint`. The branch form (`if (…) return A::find(); return B::find();`) is natural,
  reachable, and PHPStan-clean.
- **Reuses an existing primitive.** `StatementNodeFinder::findAll(statements,
  IncludeConditionalContexts, predicate)` already returns every matching node — surgical.

Predicate = a `Return_` node whose **direct** expression is a whitelisted lookup call:
`$node instanceof Return_ && $node->expr instanceof StaticCall && $this->isLookupStaticCall($node->expr)`.
So wrapped (`response()->json(User::find())`), ternary, array-wrapped, and variable-assigned returns
do **not** count (their `Return_->expr` is not a `StaticCall`), exactly as today.

Accepted conservative edge: `IncludeConditionalContexts` also descends into nested closures/anon
classes, so a lookup `return` inside a nested closure would be counted too. That errs toward **refusal**
(never toward a wrong emission), which is the library's ambiguity philosophy; it is rare and safe. Noted
so the reviewer isn't surprised — not worth a closure-excluding walk (scope creep).

### Verified against all 14 existing tests (no regressions)

Under this count, every current case keeps its exact result:
- single top-level lookup (`find`/`findOrFail`/`firstOrFail`/`article`/`findPastTenStatements`) → count 1 → resolves.
- `findBeyondBackstop` (lookup at stmt 101, outside the 100-window `$statements`) → count 0 → null.
- `dynamicClass` → count 1, falls through, resolveModelClass null → existing null+1-notice path (guard adds no notice).
- `wrapped`/`assignedThenReturned`/`ternary`/`arrayWrapped` → `Return_->expr` not a `StaticCall` → count 0 → existing null+1-notice.
- `conditionalOnly` (`if($flag) return User::find(); return null;`) → count 1 (the branch lookup) → falls through → top-level finder finds no top-level lookup → existing null+1-notice.
- `nonModelClass` → count 1, silent-null (Model guard) unchanged; `nonWhitelistMethod` → count 0.

The guard only changes the **>1** case.

### Files

**Modify** `src/Plugins/Core/Resolvers/FindReturnModelResponseResolver.php`
- In `resolvePrimaryResponse`, after the empty-statements check and **before** `directlyReturnedLookupCall`:
  ```php
  if ($this->countDirectlyReturnedLookupCalls($statements) > 1) {
      $this->noteAmbiguousLookup($method);

      return null;
  }
  ```
  (Early-return so the ambiguity path logs exactly one notice and never reaches `noteUnmatchedLookup`.)
- Add `private function countDirectlyReturnedLookupCalls(array $statements): int` using
  `$this->statementNodeFinder->findAll(..., ConditionalContextPolicy::IncludeConditionalContexts, <predicate above>)`
  wrapped in `count(...)`. `@param list<Stmt> $statements`.
- Add `private function noteAmbiguousLookup(ReflectionMethod $method): void` — mirrors
  `noteUnmatchedLookup`'s style; message: multiple directly-returned model lookups →
  primary response ambiguous → none inferred → annotate with `#[Response]`.
- Add `use function count;`. (`Node`, `StaticCall`, `Return_`, `Stmt`, `ConditionalContextPolicy`,
  `StatementNodeFinder` are already imported.)
- Leave `directlyReturnedLookupCall()` unchanged (still returns the first top-level unconditional
  lookup; only reached when count ≤ 1).
- Add one clause to the class docblock: refuses when a method has more than one directly-returned
  lookup (ambiguous).

**Modify** `tests/Unit/Core/Inference/FindReturnModelResponseResolverTest.php` (TDD — add first, watch fail)
- Fixture `twoLookupReturns(string $id, bool $legacy): mixed` — `if ($legacy) return Article::find($id); return User::find($id);` (two directly-returned lookups). `Article` is already imported.
- Fixture `cacheFastPath(string $id, bool $cached): mixed` — `if ($cached) return ['cached' => true]; return User::find($id);` (single lookup coexisting with a non-lookup return).
- Test (a): `twoLookupReturns` refuses → `->toBeNull()`, recording logger has exactly 1 `notice`.
- Test (b): `cacheFastPath` still resolves → not null, `findReturnRef(...)` contains `User`.
- Keep all existing tests green.

### Survey expectation (encode for the surveyor)

Generation-affecting (`area:core`/`area:responses`) → **survey runs**. This is an **intentional
reduction**: apps with multi-lookup methods lose a (previously wrong) `responseSchemas` entry. The
surveyor must **diff the actual dropped items** and confirm each is a multi-lookup wrong-emission
(a `model→null` FindReturnModel movement), not a legitimate single-lookup loss. Pass if only such
items drop; any single-lookup resolution that disappears is a real regression → back to coder. (Also
the #559/#575 survey's `null→model` FindReturnModel movements are the concrete corpus instances this
guard is expected to reverse.)

### Docs / CHANGELOG

- `CHANGELOG.md` → append under `[Unreleased] ### Fixed` (create the sub-head if absent):
  *"`FindReturnModelResponseResolver` now refuses when a method has more than one directly-returned
  model lookup (`find()`/`findOrFail()`/`firstOrFail()`) instead of emitting the first; the primary
  response is ambiguous. Single-lookup fast-path/cache patterns still resolve."*
- Docs: coder greps `docs/` (likely `docs/responses.md`) for an existing description of the
  `find()`/`findOrFail()` response inference; if one exists, add a minimal sentence on the
  multiple-lookup refusal, else no doc change (internal narrowing, no new public surface).

### Controversy / merge-gate

Non-controversial: no `Contracts\**`, no stage order, no config key, no public signature (all touched
members are `@internal` private). Standard survey-gated squash-merge. **Flag for plan-review:** confirm
the conditional-inclusive count set (vs. a top-level-only reading of the task note) — the plan argues it
is required to fix the issue's branch scenario and to keep fixtures PHPStan-clean.


Radiergummi:

Radiergummi:
🤖 **[lead]** plan-review — **approved with one amendment** (lead-reviewed; two independent code reviews still gate the merge).

**Approved core:** count directly-returned lookup calls **across conditional branches** (not top-level only) and refuse when >1; a single lookup coexisting with non-lookup returns (cache/fast-path) still resolves because `return $cached` is a Variable, not a lookup StaticCall, so it isn't counted. The planner's reasoning that two *top-level* lookup returns are only constructible as PHPStan-failing dead code is correct (same lesson as #576).

**Amendment — use the right counting primitive.** The plan proposes `StatementNodeFinder::findAll(IncludeConditionalContexts, …)` for the count. But that walk **descends into nested closures/arrow-fns**, whereas `directlyReturnedLookupCall()` only ever emits a **method-level** return (closures open their own scope). So a top-level `return User::find($id)` alongside an unrelated closure containing `return Other::find(1)` would count 2 → **spuriously refuse** a legitimate resolution (a `model→null` coverage loss, not an intended reduction — and it would muddy the survey). `StatementNodeFinder`'s other mode (`SkipConditionalContexts`) is wrong too: it skips `if`/`foreach` branch returns entirely.

The correct scope is **`ReturnExpressionResolver::methodLevelReturns($statements)`** — its `collectReturns` skips `FunctionLike`/`ClassLike` (closures/arrow-fns/anon-classes) but recurses through `if`/`foreach`/`try` bodies, i.e. method-level returns across all conditional branches, closures excluded. This is exactly what every sibling return-scan reader uses. **Count = `methodLevelReturns($statements)` filtered to returns whose `->expr` satisfies `isLookupStaticCall`; refuse if >1.** Instantiate `ReturnExpressionResolver` the way the class already news `StatementNodeFinder` in its constructor (or inject it) — plugin→Support is allowed.

**Add a third test** locking the scope: a top-level `Model::find()` return **plus** a closure body containing another `Model::find()` return → still **resolves** (count = 1, closure excluded). Alongside the plan's (a) two-branch-lookup → refuse and (b) single-lookup + non-lookup fast-path → resolve.

Everything else in the plan (predicate = direct lookup StaticCall expr; intentional-reduction survey note; 14 existing tests unchanged) stands. → coding.

Radiergummi:
🤖 **[lead]** plan approved with the methodLevelReturns amendment (see review comment) → coding.

Radiergummi:
🤖 **[coder]** implemented; composer check green; pushed b21a7d49. Ready for review.

Followed the plan-review amendment: count via `ReturnExpressionResolver::methodLevelReturns($statements)` (all conditional branches, nested closures excluded), filtered to returns whose `->expr` is a lookup `StaticCall`; refuse (null + one `notice`) when >1, before `directlyReturnedLookupCall`. `ReturnExpressionResolver` is instantiated in the constructor the same way the class already news `StatementNodeFinder`.

Tests added to `FindReturnModelResponseResolverTest`: (a) two-branch lookups → refuses (null, 1 notice); (b) single lookup + non-lookup fast-path (`if ($cached) return [...]; return User::find($id);`) → still resolves User; (c) top-level `User::find()` plus a nested closure containing `return Article::find(1)` → still resolves User (closure-exclusion locked). All 14 existing tests unchanged and green; 3649 total (+3). Docs: added a sentence to `docs/auto-derivation.md`. No snapshot change (snapshot group passed unchanged).

Radiergummi:
🤖 **[lead]** implementation complete (`b21a7d49`), CI green → review. Guard uses the amended `methodLevelReturns` count (closures excluded); 3 targeted tests incl. the closure-exclusion lock; snapshot unchanged. Needs two independent code reviews + survey Layer A (intentional-reduction: model→null drops on multi-lookup methods are expected/OK — diff the items).

Radiergummi:
🤖 **[reviewer-2]** review — **APPROVE** (HEAD `b21a7d49`). Guard uses `methodLevelReturns()` per the amendment (recurses if/foreach/try, excludes closures via the `FunctionLike`/`ClassLike` skip), filtered to lookup-`StaticCall` return exprs. Hand-traced all three new tests + the closure-exclusion against the real path. Key correctness argument: the counting set is a **strict superset** of what `directlyReturnedLookupCall()` can ever resolve to, so count=1 always means 'the only counted lookup is the one the resolver would emit,' and >1 always means a real second candidate exists — no path refuses a legit single lookup or misses a genuine ambiguity. Emission path byte-unchanged (pure early-return before it). Gates re-run: PHPStan clean, Pint pass, `--filter FindReturnModel` 22/22, snapshot 9/9 byte-identical. Fixtures PHPStan-clean (distinct reachable branches — #576 lesson applied). Clean approve.

Radiergummi:
🤖 **[reviewer-1]** review — **APPROVE** (HEAD `b21a7d49`). Amendment followed exactly: count via `methodLevelReturns()` (if/foreach/try recursed, closures excluded), refuse if >1. Verified fast-path resolves (array return not counted), closure-exclusion test uses a genuine `static function` scope, guard sits before emission with a distinct `notice`, `directlyReturnedLookupCall()` untouched. Gates: 20 unit pass, Pint pass, PHPStan clean, snapshot 9/9. **Non-blocking LOW observation:** the count-based guard also refuses the rare *same-model* multi-return case (both returns → User) that previously resolved; always safe (null, never wrong), contrived, and the simpler rule chosen in plan-review — a follow-up could bring parity with `ReturnExpressionResourceReader`'s same-resource-agreement logic **if** the corpus ever shows it matters. Merge-ready.

Radiergummi:
🤖 **[surveyor]** Layer A **PASS** — 0 regressions across 11 apps, every metric **fully flat** vs baseline (`ad15e52f`-equivalent; reused `4b603272` baseline since the intervening #579 was proven byte-identical). Intentional-reduction analysis: **no drops at all** — the corpus contains no method with >1 directly-returned lookup call, so the guard removes nothing in practice (pure pre-emptive safety net against the edge #575's window widening could surface). Confirms reviewer-1's same-model note is moot for the current corpus; no follow-up parity issue warranted now.

Radiergummi:
🤖 **[lead]** all gates green → squash-merged. Two independent approvals (reviewer-1 + reviewer-2, both hand-tracing the guard against the code path), full CI matrix green, survey Layer A fully flat (0 regressions; no multi-lookup methods in corpus). Guard counts directly-returned lookup calls via `ReturnExpressionResolver::methodLevelReturns()` (conditional branches included, closures excluded — the plan-review amendment) and refuses on >1, never emitting a wrong edge-model. Non-controversial: plugin resolver, no Contracts/config/signature change, small src diff. Closes #578.

PR #581, opened 2026-07-24
feat(core): PublicPropertyTypeReader types array/collection and `@var`-refined public properties

### LINKED ISSUE #528: PublicPropertyTypeReader: type array/collection and `@var`-refined public properties (deferred from #479)
## Context

Deferred from #479 (baseline typed-return response resolver, PR #524). That PR's new `SchemaFromPublicProperties` builder reuses the shared `Support\Extraction\PublicPropertyTypeReader` for per-property typing (the agreed "one authority, no divergent reader" design). The reader currently reads **only** a property's native `ReflectionNamedType` and refuses anything outside a small modelled-type allowlist. Two consequences the DTO builder inherits:

1. **Bare `array` / collection-typed properties are omitted.** A `public array $tags` resolves to a symfony `CollectionType`, which is outside `isModelledType`'s allowlist (`BuiltinType` scalars + backed enum + DateTime/UUID/UrlRoutable), so the reader returns null and the property is dropped from the DTO's schema. Same for `Collection`-typed properties.
2. **`@var`-refined properties are not honoured.** The reader never reads the property docblock, so `@var array{id: int, name: string} $meta` or `@var list<Foo> $items` on a `public array $…` property is typed as nothing (omitted), where the engine (`JsonSchemaFromType`) could map the array-shape / list directly.

Net: a DTO like `public array $tags` or a `@var array{…}`-annotated property gets **no** schema for that field today. Not wrong (never invents), but lossy for a common shape.

## Proposal

Teach `PublicPropertyTypeReader` to type array/collection-shaped properties:
- resolve each property's type from its **`@var` docblock** (via `DocBlockParser` + `TypeNodeResolver`) when present, falling back to the native type — mirroring how `EloquentModelToSchema::propertyFromTag` already resolves `@property` types before handing them to `JsonSchemaFromType`;
- allow `CollectionType` / array-shape results through to the engine (list → `{type: array, items}`, map → `additionalProperties`, `array{…}` → object), keeping the refusal discipline for genuinely-unmappable types (union/intersection/untyped, plain-object leaves handled only via the existing optional leaf callback).

## ⚠️ Survey-affecting

`PublicPropertyTypeReader` is **shared**: the ApiResources plugin (`ResourceToArrayReader`) also calls it. Typing array/collection/`@var` properties would change API Resource output too, not just the #479 baseline DTO path. This needs a **survey Layer A** pass (0 regressions; new schemas itemized) before merge, and the tier-0 "PHPDoc tags are fair game" reading should be confirmed (reading `@var` is a single-tag reflection read, no body parsing).

## Acceptance
- Array/list/map/`array{…}`-typed public + promoted properties get schemas via the reader (native or `@var`).
- Refusal discipline preserved for unmappable types; no invented/stub schemas.
- Both consumers covered by tests: `SchemaFromPublicProperties` (DTO returns) and `ResourceToArrayReader` (API Resources).
- Survey Layer A: no regressions; the new field schemas are the only additions.

Radiergummi:
## Implementation plan — #528

Teach the shared `Support\Extraction\PublicPropertyTypeReader` to type **array / list / map /
`array{…}`** public properties, resolving each property's type from its **`@var` docblock** (via
`DocBlockParser` + `TypeNodeResolver`) when present and falling back to the native type — mirroring
`EloquentModelToSchema::propertyFromTag`'s `@property`-over-cast precedence. Both consumers
(`SchemaFromPublicProperties` for typed-return DTOs, `ResourceToArrayReader` for API Resource value
objects) inherit the win.

Closes #528.

### Tier

**Tier 0 / tier-1 boundary — reflection + a single PHPDoc tag.** Reading a property's `@var` tag is
a one-tag reflection read (no body parsing, no dataflow), exactly the read the model path already
does for `@property`. Allowing `CollectionType` / `ArrayShapeType` through the existing engine is
pure type→schema mapping. No AST, no method bodies. Sits at the same rung as the current reader.

### What's wrong today

`PublicPropertyTypeReader` reads **only** the native `ReflectionNamedType` and gates output through
`isModelledType`, whose allowlist is scalars + backed enum + DateTime/UUID/UrlRoutable. Verified
empirically:

- `public array $tags` resolves (native) to a symfony **`CollectionType`**, not a `BuiltinType`, so
  `isModelledType` returns false and the property is **dropped**. (The `'array'` entry in
  `MODELLED_BUILTINS` is dead for the native path — native `array` never arrives as a `BuiltinType`.)
- The reader never reads the property docblock, so `@var list<Foo>` / `@var array{…}` / `@var
  array<string, Bar>` are **omitted** even though the engine (`JsonSchemaFromType`) maps every one.

Net: a common DTO field (`public array $tags`, `@var array{…} $meta`) gets **no** schema.

### Design

Two changes inside `PublicPropertyTypeReader::propertyFor` / `schemaForType`, plus new constructor
deps. Nothing outside the reader changes shape; the engine already maps every target type.

**1. Resolve the property type from `@var`, else native.**

In `propertyFor`, before the current native path, read the property's own doc comment:

```
$type = $this->typeFromVarTag($property, $reflection) ?? $this->typeFromNativeType($property);
if ($type === null) { return null; }
```

- `typeFromVarTag`: `$property->getDocComment()` → `DocBlockParser::parse` → first `@var`
  `VarTagValueNode->type` → `TypeNodeResolver::toType($node, $reflection)`. Returns null when there
  is no `@var`, or the tag is unresolvable (`toType` returns null) — falling back to native, never
  aborting. (Verified: `list<int>`/`array<string,int>` → `CollectionType`, `array{…}` →
  `ArrayShapeType`, `int|string` → `UnionType`, unresolvable `Foo` → `null`.)
- `typeFromNativeType`: the existing body — refuse non-`ReflectionNamedType` (native
  union/intersection/untyped), `TypeResolver::resolve`, catching `UnsupportedException` → null. This
  preserves every current native refusal.
- **Precedence:** `@var` wins when it resolves, mirroring the model path. The survey Layer A pass is
  the safety net for any unexpected native→`@var` flip.
- **Promoted properties:** `getDocComment()` on a constructor-promoted property returns false (the
  docblock lives on the constructor `@param`), so promoted properties get **native-type** array /
  collection support only. Reading the constructor `@param` for promoted params is out of scope for
  this issue (native type already covers `public array $tags` in a promoted ctor). Noted as a
  boundary; a promoted-property test asserts the native path still types it.

**2. Broaden the mappability gate to array-likes, refuse unions.**

Extend `isModelledType` into a recursive `isMappable(Type $type, ?callable $leafClassSchema): bool`
that mirrors exactly the branches `JsonSchemaFromType` maps to a concrete (non-placeholder) schema:

- `NullableType` → recurse on the wrapped type (engine applies nullability).
- `UnionType` / `IntersectionType` → **false** (refuse — this is the new `@var` union refusal;
  native unions are already refused upstream in `typeFromNativeType`).
- `BackedEnumType` → true; modelled `BuiltinType` scalars (`string`/`int`/`float`/`bool`) → true.
- `ObjectType` → true iff a **format** object (DateTime/UUID/UrlRoutable) **or** `$leafClassSchema
  !== null` (a callback is present to resolve the leaf). Presence — not invocation — so the
  predicate stays pure and side-effect-free (no premature `ComponentSchemaRegistry` registration, no
  orphan components).
- `ArrayShapeType` → true iff every field type, and (when unsealed) the extra-value type, is
  `isMappable`.
- `CollectionType` (array-like) → true iff the collection value type is `isMappable`; a
  **non-array-like** collection (`JsonResource<int>`) → false. `mixed` value / bare `array` → true
  (maps to `{type: array, items: {}}`). Array-likeness is decided by reusing the engine's own
  `isArrayLikeCollection` (see below).

`schemaForType` then becomes:

```
if ($this->isMappable($type, $leafClassSchema)) {
    return $this->jsonSchemaFromType->fromType($type, /* callback per branch, see note */);
}
// unchanged top-level plain-object leaf path (invoke callback once, wrap nullable)
```

**Callback threading (deliberate, to avoid a format-object regression):**

- The **modelled scalar / format-object / backed-enum** branch keeps calling `fromType($type)`
  **without** the callback, exactly as today. Passing the callback here would route a top-level
  `DateTimeInterface`/`Carbon` property through `leafOrBuiltin` first, and a consumer callback like
  `SchemaFromPublicProperties::buildRef` would return a public-property `$ref` for `Carbon` instead
  of the `date-time` format — a regression. Not passing it preserves current output byte-for-byte.
- The **`CollectionType` / `ArrayShapeType`** branch calls `fromType($type, $leafClassSchema)`
  **with** the callback, so `@var list<Dto>` becomes `list<$ref>` under `SchemaFromPublicProperties`
  and refuses cleanly (no callback → object leaf not mappable) under `ResourceToArrayReader`. This
  matches how the model path threads its `@property` callback.
- **Known minor imperfection (accepted, documented in a code comment):** a *nested* format object
  under a present callback (e.g. `@var list<Carbon>` on a DTO) resolves via the callback rather than
  the `date-time` format. Genuinely rare; the model path shares the same characteristic; out of
  scope to special-case here.

**Reusing `isArrayLikeCollection`:** the reader needs to refuse `JsonResource<int>`-style
collections without duplicating the engine's ~15-line array-likeness test. Plan: promote
`JsonSchemaFromType::isArrayLikeCollection()` from `private` to `public` and call it from the reader.
It is `@internal` `Support\` code (not a `Contracts\` surface), and the reader already mirrors the
engine's mapping knowledge via `isModelledType`; reusing the one true predicate is cleaner than a
second copy. *(Small internal-visibility widening — flagged below.)*

**Constructor deps added:** `DocBlockParser`, `TypeNodeResolver`. Both are already `@internal`
`Support\` services, already bound scoped in `OpenApiServiceProvider` (lines 243–244) and already
co-injected into `EloquentModelToSchema` — no new wiring. The unit test's hand-built
`publicPropertyTypeReader()` helper gains `DocBlockParser::create()` / `TypeNodeResolver::create()`.

### Files

- **`src/Support/Extraction/PublicPropertyTypeReader.php`** — the change above (`propertyFor`
  type-source split, `isModelledType` → recursive `isMappable`, ctor deps, callback threading).
- **`src/Support/Generator/JsonSchemaFromType.php`** — `isArrayLikeCollection` `private` → `public`
  (single-line visibility change; no behaviour change).

### Tests (TDD — write the reds first)

**`tests/Unit/Support/Extraction/PublicPropertyTypeReaderTest.php`** (extend fixture + cases; update
the `publicPropertyTypeReader()` helper for the two new deps):
- native `public array $tags` → `{type: array, items: {}}` (the core regression).
- native `public ?array $maybe` → nullable array.
- `@var list<int>` → `{type: array, items: {type: integer}}`.
- `@var array<string, int>` → `{type: object, additionalProperties: {type: integer}}`.
- `@var array{id: int, name: string}` → object, `required: [id, name]`.
- `@var list<LeafDto>` **with** a leaf callback → `{type: array, items: {$ref}}`.
- `@var list<LeafDto>` **without** a callback → **refused** (null; no placeholder items).
- `@var int|string` → **refused** (null); native `int|string` (already covered) → refused.
- `@var array<int, SomeService>` where the collection isn't array-like / leaf unmappable & no
  callback → refused.
- unresolvable `@var Nonexistent` → **falls back to native** type (assert native still typed).
- promoted `public array $tags = []` → native array typed (docblock-less promoted path).
- unchanged refusals still hold: untyped, static, intersection, non-public.

**`tests/Unit/Support/Extraction/SchemaFromPublicPropertiesTest.php`** (DTO-return consumer):
- a DTO with `public array $tags`, `@var list<int> $ids`, `@var array{…} $meta`, and `@var
  list<NestedDto> $children` builds an object schema carrying all those fields, with `children`
  a `$ref` array and `required` tracking nullability.

**`tests/Feature/Plugins/ApiResources/ResourceToArrayInferenceTest.php`** (API Resource consumer):
- a resource whose `toArray()` reads a value-object property typed `@var array{…}` / `public array`
  surfaces that field's schema (previously unconstrained/omitted).

### Observable behaviour → docs + changelog

- **`docs/auto-derivation.md`** — the typed-return baseline section (around lines 155–173) lists the
  property shapes that map; add array / list / map / `array{…}` and note `@var` refinement of a
  property's declared type. Keep it to the existing bullet style.
- **`docs/plugins.md`** (line ~109, "statically-typed public properties") — extend the one-liner to
  mention array/collection and `@var`-typed properties if it reads as exhaustive.
- **`CHANGELOG.md` `[Unreleased]`** — one entry: public-property typing now covers
  array/list/map/`array{…}` properties and honours a property's `@var` tag (feeds typed-return DTOs
  and API Resource value objects).

### ⚠️ Survey-affecting (Layer A gate before merge)

`PublicPropertyTypeReader` is shared by SpatieData DTO returns and ApiResources value objects, so
new field schemas will appear in both. Merge is gated on a **survey Layer A** pass: **0
regressions**; the only diffs are newly-typed array/collection/`@var` fields (net additions). If
Layer A shows a native→`@var` flip that *loses* a field, treat it as a regression and revisit
precedence.

### Merge-controversy flags for plan-review

- **Internal visibility change:** `JsonSchemaFromType::isArrayLikeCollection()` `private` → `public`
  (reuse, not duplicate). No `Contracts\` surface touched. If the reviewer prefers zero engine edits,
  the fallback is a self-contained copy of the ~15-line predicate in the reader.
- **Road not taken:** a strict `JsonSchemaFromType::tryFromType()` returning null instead of a
  placeholder would centralize refusal in the engine and let the reader shrink, but it threads a
  "strict" mode through several shared engine methods — larger blast radius on a hot path for a
  single consumer. Kept the change inside the reader per the issue's framing and CLAUDE.md's
  surgical mandate. Happy to switch if plan-review prefers centralizing.
- Otherwise low-risk: no stage-order change, no config key, no public signature change.


Radiergummi:

Radiergummi:
🤖 **[lead]** plan posted (PR #581) → plan-review. Reviewer: rule on the 2 flagged design points (widen isArrayLikeCollection private→public vs copy; reader-local change vs engine tryFromType) and confirm the tier-0/1 read + refusal discipline.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved** — plan is surgical, tier framing holds, refusal discipline preserved. One recommended precedence refinement below, plus rulings on the two flagged design points.

## Design point 1 — `isArrayLikeCollection()` private→public: **approve the widening**
Reuse over copy. It is `@internal` `Support\` code, touches no `Contracts\` surface, and the reader already imports `JsonSchemaFromType` as a ctor dep and already mirrors the engine’s mapping knowledge via `isModelledType`. Copying the ~26-line predicate (with its `GenericType` unwrap + `COLLECTION_CLASSES` lookup) into the reader would create two subtle copies that must stay in sync, with nothing to catch a drift. One authority wins. The `COLLECTION_CLASSES` const stays private — only the method goes public.

## Design point 2 — reader-local `isMappable` vs strict engine `tryFromType()`: **keep it reader-local**
Stronger reason than blast-radius alone: the reader’s refusal policy is **stricter and different** from the engine’s placeholder set, not a subset of it. The engine legitimately maps a `UnionType` → `oneOf` (line 98) and a plain object leaf → a stub (`fromObjectType`); the reader **refuses both**. A strict `tryFromType()` that returns null "where the engine would emit a placeholder" would therefore *not* capture the new `@var` union refusal (a union is a valid engine mapping), so it fails to express the reader’s actual policy — or forces the engine to absorb a single consumer’s "never guess" policy across a hot shared method. Keep the *policy* (what to refuse) in the reader; reuse only the *mechanical* sub-predicate (`isArrayLikeCollection`). This is the surgical, boundary-respecting choice per CLAUDE.md.

## Recommended refinement (should-fix, ~1 condition) — make `@var` precedence conditional
The plan’s precedence is "`@var` wins whenever `toType` resolves," backstopped by the survey Layer-A gate for any native→`@var` flip that loses a field. That flip is logically preventable: if a property has a mappable native type (e.g. `string`) and a `@var` that *resolves but is unmappable* (e.g. `@var int|string` → `UnionType` → refused), the current rule drops a field that is typed today — a regression the acceptance criteria ("the new field schemas are the only additions; no regressions") explicitly forbid.

Prefer `@var` only when it resolves **and** `isMappable`; otherwise fall back to the native path:
```
$varType = $this->typeFromVarTag($property, $reflection);
$type = ($varType !== null && $this->isMappable($varType, $leafClassSchema))
    ? $varType
    : $this->typeFromNativeType($property);
```
This makes the change **provably additive** (you only ever gain mappability, never lose a field), turning the survey gate into a confirmation rather than a safety net. `@var list<int>` over native `array` still wins (more specific, not a regression) — only the resolves-but-unmappable case falls back.

## Minor note for code-review (not blocking plan)
Verify the `schemaForType` restructure keeps the top-level **format-object** path (`Carbon`/UUID/UrlRoutable) byte-identical: the modelled-scalar/format/enum branch must call `fromType($type)` **without** the callback (plan already states this). The "unchanged refusals still hold" + a format-object-with-callback-present test should pin this.

Everything else — files, TDD test list (both consumers covered), docs (`auto-derivation.md` / `plugins.md`), CHANGELOG entry, tier-0/1 framing (single `@var` tag read + pure type→schema mapping, no body parsing), survey Layer-A gate — is sound and correctly scoped. Cleared to code.

Radiergummi:
🤖 **[lead]** plan-review APPROVED (reviewer-1) → coding. Rulings: (1) widen JsonSchemaFromType::isArrayLikeCollection() private→public, reuse not copy; (2) keep refusal policy reader-local, no engine strict flag. Adopt reviewer's should-fix: @var precedence CONDITIONAL — prefer @var only when it resolves AND isMappable, else fall back to native (keeps change provably additive). Non-blocking: confirm Carbon/UUID format-object path stays byte-identical.

Radiergummi:
🤖 **[coder]** deviation: adopted the reviewer's **conditional `@var` precedence** ruling, which supersedes the plan's `@var list<Dto>`-without-callback → *refused* test case. Under conditional precedence, an unmappable `@var` (object-leaf with no leaf callback) is skipped and the reader falls back to the property's **native** type. So on a `public array $children` field with `@var list<Dto>` and no callback (the `ResourceToArrayReader` path), the field now types as an unconstrained `{type: array, items: {}}` (native `array`) rather than being dropped — strictly additive and never-wrong. The reader unit suite asserts this fallback explicitly. With a leaf callback present (the `SchemaFromPublicProperties` path), the same `@var list<Dto>` becomes `{type: array, items: {$ref}}` as planned.

Implementation notes matching the reviewer's rulings:
1. `JsonSchemaFromType::isArrayLikeCollection()` widened `private`→`public` and reused from the reader (const `COLLECTION_CLASSES` stays private).
2. Refusal policy kept reader-local (recursive `isMappable`); the engine's own mappings are untouched, no strict `tryFromType()`.
3. Conditional `@var` precedence: `@var` wins only when it resolves **and** is mappable; native type otherwise. Verified additive by the `native string + @var int|string → keeps string` regression test.
4. Byte-identical format-object path confirmed: modelled scalar/enum/format leaves still map via `fromType($type)` **without** the leaf callback (a top-level `Carbon`/UUID/DateTime stays `date-time`/`uuid`/string); only structural shapes thread the callback.

Radiergummi:
🤖 **[coder]** implemented; local gates green (`composer check`: Pint clean, PHPStan level 8 clean, 3660 tests pass); pushed 4564372d. Ready for review.

Radiergummi:
🤖 **[lead]** implementation green (remote CI full matrix passes, sha 4564372d) → review. Need 2 independent approvals. Reviewers: verify the 4 rulings held, scrutinize the documented deviation (`@var list<Dto>` w/o leaf callback falls back to native array→`items:{}` unconstrained — additive & never-wrong?), and confirm byte-exact snapshots unchanged. Survey Layer A gated after approvals.

Radiergummi:
🤖 **[reviewer-2]** review round 1: **approved, no concerns.** Independent verdict (did not defer to reviewer-1). All four rulings held; refusal discipline and the documented deviation are sound.

**Verified**
1. `JsonSchemaFromType::isArrayLikeCollection()` widened `private`→`public` and **reused** (`PublicPropertyTypeReader::isMappable` calls `$this->jsonSchemaFromType->isArrayLikeCollection($type)`), not copied; `COLLECTION_CLASSES` stays `private` (`JsonSchemaFromType.php:75`, still `self::` at :402).
2. Refusal policy is reader-local: recursive `isMappable`/`isMappableElement`/`isModelledLeaf` refuse union/intersection/unmappable-object-leaf. The engine is untouched — `fromType()` is called with the same optional callback signature, no strict flag threaded through the hot shared method.
3. `@var` precedence is genuinely **conditional and additive**: `typeFromVarTag` returns a type only when `toType(...)` resolves **and** `isMappable`; otherwise `?? typeFromNativeType`. A field typed today is never dropped to an unmappable `@var` — pinned by `keeps the native type when a @var refines it to an unmappable union` (`int|string` → keeps `string`) and `falls back to the native type when a @var names an unresolvable class`.
4. Format-object path byte-identical: `schemaForType` routes a scalar/enum/format leaf through `isModelledLeaf($inner)` → `fromType($type)` **without** the callback; only `ArrayShapeType`/`CollectionType` thread `$leafClassSchema`. Top-level Carbon/UUID/DateTime stay `date-time`/`uuid`/string.
5. Deviation (`@var list<Dto>` without leaf callback, the `ResourceToArrayReader` path — confirmed no-callback at `ResourceToArrayReader.php:602,629`) falls back to native `array` → `{type: array, items: {}}`. Strictly additive (previously the bare array was **omitted**) and never-wrong (a `list<Dto>` *is* a JSON array). With a callback (`SchemaFromPublicProperties.php:139`) the same `@var` yields `items: {$ref}`. Both branches tested.
6. Both consumers covered; CHANGELOG `[Unreleased]` + `docs/auto-derivation.md` + `docs/plugins.md` updated; full CI matrix green incl. **Snapshot (locked toolchain)** — no byte-exact snapshot moved.

**Test-gap hunt** — the branch/boundary coverage is strong (native array public+promoted, `list<int>`, `array<string,int>` map, `array{…}`, `list<Dto>` with/without callback, unmappable-union fallback, unresolvable-class fallback, both consumers + required-field derivation). Nothing missing that blocks.

**Non-blocking observations (no fix required to merge):**
- A bare, ungenerified native `Collection` property (`public Collection $x`, no `@var`) still refuses — native reflection yields an `ObjectType`, not a `CollectionType`, so `isArrayLikeCollection` is never reached. That's the never-wrong choice (no element info to guess), but it means the issue's "Collection-typed properties" only type when carrying an array-like `@var`/generic. Worth confirming this narrower-than-issue-prose scope is intended (I read it as correct per tier discipline).
- Precedence when a mappable `@var` *contradicts* a mappable native scalar (e.g. `@var int` on `public string $x`) flips the emitted type to the `@var`. Intended (mirrors `@property` over cast) and deterministic, but untested — an optional one-liner would pin it.
- Nullable structural shapes (`public ?array`, `@var ?list<int>`) rely on the engine's generic nullability handling; no direct reader test. Optional.
- A handful of docblock lines run 101–104 chars (soft 100 limit); trivial.

Cleared to merge on my side pending the second independent approval and the survey Layer-A gate.

Radiergummi:
🤖 **[reviewer]** review round 1: **approved, no blocking concerns.** Checked out sha 4564372d, verified against the four rulings + the flagged deviation; ran the touched tests (60 passed, 246 assertions), `pint --test` (passed) and PHPStan on the reader (no errors).

**Rulings — all satisfied:**
1. **`isArrayLikeCollection()` private→public, const stays private, reused** — the engine diff is exactly the one-line visibility change (`JsonSchemaFromType.php:378`); the reader calls it at `PublicPropertyTypeReader.php:244`. No other engine change. ✓
2. **Refusal policy kept reader-local** — recursive `isMappable`/`isMappableElement` (lines 219-263) mirror the engine’s mapped branches; unions/intersections refuse (225-227), array-likeness delegates to the engine predicate. Engine behaviour untouched. ✓
3. **Conditional `@var` precedence is genuinely additive** — `typeFromVarTag` returns the `@var` type only when it resolves **and** `isMappable` (line 146), else `propertyFor` falls through to native (103-104). Pinned by tests: `@var int|string` over native `string` keeps `string` (no `oneOf`); `@var Nonexistent` falls back to native `string`. A property typed today cannot lose its schema. ✓
4. **Format-object path byte-identical** — the modelled-leaf branch calls `fromType($type)` **without** the callback (line 194), so a top-level `Carbon`/UUID/DateTime stays a `date-time`/format leaf even when a leaf callback is present. Correct by construction. ✓

**Flagged deviation (`@var list<Dto>` without a leaf callback → native `array` → `{type: array, items: {}}`):** confirmed **strictly additive and never-wrong.** Today the field is dropped; now it is an unconstrained array, which is a true statement about an `array`-typed property. Crucially it is a *direct consequence of ruling 3*: the unmappable `@var` is rejected and the mappable native `array` is used — refusing the whole property would instead contradict the additive precedence. Untyped-property case still refuses correctly (no native fallback → null). Explicitly tested (`falls back to the native array for a @var list<Dto> … when no callback`). ✓

**Both consumers covered:** DTO path (`SchemaFromPublicPropertiesTest` — list/map/`array{…}`/nested-`$ref`/bare-array + required tracking) and API Resource path (`ResourceToArrayInferenceTest` — bare `array` unconstrained + `@var array{…}` object). Snapshot group untouched (no snapshot fixtures in the diff; remote snapshot job green). Docs (`auto-derivation.md`, `plugins.md`) and `CHANGELOG.md` all accurately state the conditional-precedence rule. ✓

**Conventions:** strict-types, no abbreviations, why-focused comments (no issue numbers), `@internal` preserved, ≤100 cols, justified `@phpstan-ignore` reasons on fixtures. Clean.

---

**Non-blocking findings (recommend, do not gate merge):**

1. **Add a with-callback format-object regression test.** The `date-time` test (`PublicPropertyTypeReaderTest.php:138`) exercises the *no-callback* path; the exact regression the plan flagged is a format object under a *present* leaf callback (the DTO path always passes one). No DTO fixture carries a `DateTimeImmutable`, so nothing pins "format leaf under a callback stays `date-time`, not a `$ref`". The code is correct (line 194 drops the callback), but a one-line assertion — e.g. `propertyFor(fixtureWithDateTime, createdAt, fn() => new OA\Schema([ref => #/x]))->format === date-time` — would lock ruling 4.

2. **Observation, no action needed:** a `@var array{…}` (a JSON *object* shape) whose leaf is unmappable and with no callback also falls back to native `array` → `{type: array}`, which describes a JSON array rather than an object — imprecise for an associative array. This is not a new violation: it is the same bare-`array`→`{type: array}` convention this PR’s core feature (issue #528’s `public array $tags`) deliberately adopts, and is the library-wide treatment of untyped `array`. Flagging only so it’s on record.

Approving. Reviewer-2 is reviewing independently.

Radiergummi:
🤖 **[coder]** addressed reviewer-1's non-blocking recommendation: added a with-callback format-object regression test locking ruling 4 — a top-level `DateTimeImmutable` property stays `date-time` format (not a `$ref`) even when a leaf callback is present (the DTO-consumer path always supplies one). Test-only, no `src/` change. Pushed d95d80c7.

Radiergummi:
🤖 **[surveyor] Layer A PASS — 0 regressions across 11 apps.**

**Baseline** `main` @ `cca4bf15` (content-identical to the cached baseline's `libraryCommit b21a7d49`; reused). **Candidate** PR branch @ `d95d80c7` — the branch has advanced one commit past the CI-green `4564372d`, a **test-only** follow-up (+15 lines in `PublicPropertyTypeReaderTest`, no `src/` change), so generated output is identical. Manifest confirms `d95d80c7` was the library installed into all 11 apps.

**Metric gate (`results.json`):** byte-identical to baseline (md5 `0b1b71e`). Every watched metric (paths, operations, responseSchemas, documentedResponses, requestBodies, componentSchemas, operationsWithSecurity, completeness) is unchanged for all 11 apps — **0 regressions**, 0 crashes (all `gen_exit=0`).

**Spec-level additive check (full-document leaf diff, per app):** zero real changes across the corpus.

| app | real diff | note |
|-----|-----------|------|
| AdvisingApp | none | identical |
| AureusERP | none | 60/60 leaf churn = Livewire asset-route hash `398cdf7d`→`720e8703` only |
| Bagisto | none | identical |
| BookStack | none | identical |
| Coolify | none | identical |
| InvoiceNinja | none | identical |
| Koel | none | identical |
| Lychee | none | identical |
| Pelican | none | identical |
| SpeedtestTracker | none | 65/65 leaf churn = Livewire asset-route hash `7ad22647`→`ef65968b` only |
| Vito | none | identical |

The only leaf-level churn is the **nondeterministic Livewire asset-route hash** (present only in the two Livewire apps); zero non-`livewire-*` diffs on either. No field was added, removed, or changed anywhere — so the "nothing previously typed changed or disappeared" invariant holds trivially.

**Net field additions: 0.** The corpus does **not** exercise the new additive path — none of the 11 apps expose array/collection/`@var`-typed public properties at the typed-return-DTO (`SchemaFromPublicProperties`) or Resource value-object (`ResourceToArrayReader`) surface that were previously dropped. The new behavior is covered by the PR's unit tests; the corpus role here is the regression gate, and it confirms the change is a **guard-verified safe no-op** on real apps.

Clear to merge from the survey gate.

Radiergummi:
🤖 **[lead]** ALL GATES GREEN — escalating for human merge (supervised first-batch calibration + 202-line src diff > ~150 auto-merge threshold). PR is ready, not merged; worktree left in place.

**Evidence package:**
- ✅ **2/2 independent reviewer approvals** (reviewer-1 + reviewer-2), no open concerns. All four plan-review rulings verified against the code; the documented `@var list<Dto>`-without-callback deviation confirmed strictly additive & never-wrong (direct consequence of conditional `@var` precedence).
- ✅ **Full remote CI green** on final sha `d95d80c7` — PHP 8.4/8.5 × Laravel 12/13, swagger-php 5.8, snapshot (locked toolchain), coverage, quality.
- ✅ **Survey Layer A PASS** — 0 regressions / 0 crashes across 11 apps; candidate results.json byte-identical to main baseline (md5 0b1b71e). Corpus doesn't exercise the new path (0 net field additions there); new behavior covered by unit tests, guard invariant holds trivially.
- ✅ **Ruling-4 locked** by an added with-callback format-object regression test (top-level DateTimeImmutable stays `date-time`, not `$ref`, even under a leaf callback).

**Why human (not auto-merge):** first PR of a freshly-resumed run — skill directs the first batch to run supervised so you calibrate the pipeline before it goes hands-off; independently, the 202-line production diff (`PublicPropertyTypeReader.php` +153/-35, `JsonSchemaFromType.php` 1-line visibility widen, `ParsedDocBlock.php` +12) is over the ~150 auto-merge threshold.

**Non-blocking notes on record** (reviewers, no action needed): bare ungenerified native `Collection` (no `@var`) still refuses — never-wrong but narrower than the issue's 'Collection-typed' prose (per tier discipline); `@var array{…}` with unmappable leaf + no callback falls back to `{type: array}` — same library-wide bare-array convention #528 adopts. Both judged correct.

Ready for your review + squash-merge.

PR #567, opened 2026-07-22
feat(responses): read never / RedirectResponse / StreamedResponse from the declared return type (Tier-0)

Radiergummi:
Closes #550

## Problem

Three return types carry their entire response contract in the **declared return type** — pure
Tier-0 information — and the library reads none of them. All three currently fall through to the
synthetic `200 OK` with no content in `OperationBuilder`, documenting something that never happens:

| Declared return | Emitted today | Should emit |
|---|---|---|
| `never` | empty `200 OK` | **no** primary 2xx (the action cannot succeed) |
| `RedirectResponse` | empty `200 OK` | `302 Found` + `Location` header (`string`/`uri`) |
| `StreamedResponse` / `BinaryFileResponse` | empty `200 OK` | `200` `application/octet-stream`, `{type: string, format: binary}` |

Evidence (a private consumer app, 316 `/api` ops): `never` → 6 ops, `RedirectResponse` → 3 ops,
`StreamedResponse` → 1 op. ~10 ops move off the wrong synthetic 200.

## Hard dependency: #516 / PR #560 (MUST merge first)

`never` needs a way to say **"suppress"** as distinct from **"I don't know"** — today a
`PrimaryResponseResolver` returning `null` means the latter, and `OperationBuilder` falls back to
the synthetic 200. #516 (PR #560, open + mergeable) widens the resolver contract from
`?OA\Response` to a `?Contracts\Registry\PrimaryResponse` result, which is exactly where a
suppression state belongs. This plan is written **against post-#560 code** and extends
`PrimaryResponse`.

**Merge order: this PR must land after #560.** The coder should not start until #560 is on `main`,
then rebase. Building a "minimal equivalent" now would reintroduce the sentinel pattern #560 is
deleting and collide head-on with it on the exact contract signature. Flagged for the lead.

## Tier

All three reads are **Tier-0** (reflect the action's declared return type; no body parsing). Union
returns and app subclasses are handled by reflection alone. The `never` + single-unconditional-`throw`
status derivation is a **Tier-1** body scan — **deferred to a follow-up** (see below); it is not in
this PR.

## The read (precisely, per type)

Detection uses the existing `Support\Routing\ReturnShapeResolver` (the "single Tier-0 answer" to a
return's structure), which already surfaces `never` as a `BuiltinType`, unions via
`ReturnShape::$unionMembers`, and a single object return as an `ObjectType` on `$itemType`.

[...]

PR #526, opened 2026-07-02
refactor(responses): ResourceClassLocator — consolidate collection detection onto ReturnShapeResolver

Radiergummi:
Closes #521

## Implementation plan — #521: ResourceClassLocator collection-detection consolidation

**Tier 0.** Follow-up to #481 (PR #519), which migrated Data + Paginator onto the shared
`Support\Routing\ReturnShapeResolver` but **deferred site-3** (`ResourceClassLocator`), leaving this
issue to (a) decide the right seam for plugin-specific container classification on top of
`ReturnShape`, and (b) migrate `ResourceClassLocator`'s structural reads onto the shared resolver.

The plan does (a) — the seam decision, which is the issue's substantive ask — and then reports that
(b) has **no substrate to migrate at Tier 0** without either zero net dedup or a behavior change.
The recommendation is therefore a **decision fork for plan-review + a human gate**, not a blind
code migration.

### Grounding (verified against current `origin/main`, post #519/#520)

- **`ReturnShape` (`src/Support/Routing/ReturnShape.php`, `@internal`) carries structure only:**
  `{container: Single|ListOf|Paginated, itemType: ?symfony\Type, paginatorKind: ?PaginatorKind,
  nullable: bool, unionMembers: list<Type>}`. Its own docblock states it "never answers *is this
  item class a JsonResource / Data / a class I own*; that question stays with the plugin, which
  reads `itemType` / `unionMembers` and applies its own `is_a` test."
- **The boundary guard is real and load-bearing:** `tests/Arch/CoreBoundaryTest.php:22-24` —
  *"Support must not depend on any plugin"* (`Radiergummi\OpenApi\Support` `->not->toUse`
  `Radiergummi\OpenApi\Plugins`). `ReturnShape`/`ReturnShapeResolver` importing `ResourceCollection`
  would fail this test. So the descriptor can never classify "is this a resource collection".
- **The established migration precedent is `DataResponseResolver` (#519):** it injects
  `ReturnShapeResolver`, calls `describe()` once, and consumes only the *structural* outputs
  (`unionMembers`, `container === Paginated`, `nullable`) — while **still reading raw
  `$reflector->getReturnType()` directly** for its plugin `is_a(Data::class)` / `is_a(DataCollection::class)`
  classification. Structure from the resolver; class-meaning `is_a` on the plugin.
- **What `ResourceClassLocator` (`src/Plugins/ApiResources/Support/ResourceClassLocator.php`)
  actually reads from reflection:**
[...]

Issue #544, opened 2026-07-22
feat(responses): type ->format() on a date attribute as string / date-time in resource fields


Radiergummi:
## Summary

A resource field whose value is a method call on a model's date attribute — `$this->resource->created_at->format(DATE_ATOM)` — degrades to an untyped `{}` property. The property type could be inferred: a `->format(...)` call on a `DateTimeInterface`-typed attribute always yields a `string`, and `DATE_ATOM` / `'c'` specifically a `string` with `format: date-time`.

## Motivation (dogfooding: a private consumer app)

A near-universal idiom in resource bodies:

```php
self::FIELD_CREATED_AT => $this->resource->created_at->format(DATE_ATOM),
self::FIELD_UPDATED_AT => $this->resource->updated_at->format(DATE_ATOM),
```

`$this->resource->created_at` types correctly from the model's `$casts`/migration (a `datetime`), but the `->format()` call on it is not a plain property fetch, so the reader emits an untyped field. Bare property reads (`$this->resource->name`) type fine. In this app it affects `createdAt`/`updatedAt` on essentially every resource.

## Proposed behaviour

Bounded, whitelisted enhancement in the resource-field value reader (`ResourceToArrayReader` / the shared member-typing path):

- When the value expression is `<expr>->format(<arg>)` and `<expr>` resolves to a `DateTimeInterface` (already known via the model attribute's cast type), type the field as `{ type: string }`.
- When `<arg>` is `DATE_ATOM`, `DateTimeInterface::ATOM`, or the literal `'c'`, refine to `{ type: string, format: date-time }`. (Optionally: `'Y-m-d'` → `format: date`; leave other formats as plain `string`.)

## Tier / scope

**Tier 1** — a small whitelist match (`->format()` on a known `DateTimeInterface` receiver) with no cross-call dataflow. Degrade gracefully (plain `string`, or untyped as today) when the receiver type isn't known.

## Notes

Originally recorded as **L7** in the private consumer app dogfood ledger; not previously filed.

Radiergummi:
🤖 **[lead]** admitted → planning. Tier-1 / `area:responses`; collision pre-check required against #581 and #526, both of which touch resource-field inference.

PR #583, opened 2026-07-25
feat(responses): honour the authored status of `response()->json(<resource>, <status>)`

### LINKED ISSUE #535: feat(responses): honour the authored status of response()->json(<resource>, <status>)
## Problem

When an API Resource is wrapped in `response()->json(<resource>, <status>)` (see #517, PR #534),
the resource `$ref` is now emitted, but the response is documented under the **conventional success
status (200)**, not the authored status. So `store()` returning
`response()->json(UserResource::make($u), 201)` documents `200`, not `201`.

This is not a regression introduced by #517 — the authored status was already lost before it (a
resource-typed `json()` body is not a statically-readable body, so Core's `InlineJsonResponseResolver`
yields without ever reading the status, and `ResourceResponseResolver` hardcodes
`'response' => '200'`). #517 added the correct body under that same 200; this issue is the remaining
fidelity gap: **body correct, status still conventional**.

## Why it's separate from #517

#517 was scoped as "the resource reader only resolves the resource class; status is Core's concern."
Honouring the authored status requires threading a status through the resource resolution path:

- `Routing\ResourceTarget` has no status field.
- `Plugins\ApiResources\Resolvers\ResourceResponseResolver::resolvePrimaryResponse` hardcodes `200`
  and never sets `OperationBuilder::EXPLICIT_STATUS_EXTENSION`.

That is a cross-cutting change touching a `ResourceTarget` shared by ~6 resolution paths (attribute,
collection, `toResource`, wrapped-model, docblock generic, json-unwrap), so it deserves its own
design + review rather than riding on #517.

## Proposed change (Tier 1)

1. Capture the `json()` status where the unwrap happens (reuse `Support\MethodBody\ResponseJsonCall`
   + the existing status-reading logic, or expose a `statusArgument()` on `ResponseJsonCall`).
2. Thread an optional, statically-known **2xx** status to `ResourceResponseResolver` (a field on
   `ResourceTarget`, or a parallel return), which emits it in place of `200` and marks it explicit
   via `EXPLICIT_STATUS_EXTENSION`. Absent/non-literal/non-2xx status → keep `200` (today's
   behaviour), never wrong.
3. Only the json-unwrap path carries a status; all other resource paths keep `200`.

## Tests

- `response()->json(X::make(...), 201)` → resource `$ref` at **201** (single) and the `::collection`
  variant; the status marker is set.
- `response()->json(X::make(...))` (no status) and a non-2xx status → resource `$ref` at `200`.
- Survey Layer A: 0 regressions.

Refs #517, PR #534.

Radiergummi:
Closes #535

> **Revision 2** — folds in reviewer1's plan-review (2 blocking, 3 should-fix). Changed: the 204
> case was **factually wrong** and is now a regression guard (§5.1); a convention-survives test
> closes the one-sided precedence coverage (test 10); `successSchema()` needs a verb parameter, not
> just a status (§Tests); the docs claim is scoped to the unchained call (§Docs); the
> `AstLiteralEvaluator` self-class question is decided explicitly (§3.1). I re-verified each
> blocking claim against the tree rather than accepting it — all three structural facts hold
> (`OperationBuilder.php:229-239` breaks on the first non-null resolver, `CorePlugin` is first in
> `[CorePlugin::class, ...config('openapi.plugins')]` at `OpenApiServiceProvider.php:270`, and
> `RESOURCE_PRESERVING_CHAIN_METHODS` is `['additional']` only at
> `ReturnExpressionResourceReader.php:97`).

## Verified premise

Re-read against `main` rather than taken from the issue:

- `ReturnExpressionResourceReader::targetFromExpression()` already unwraps
  `response()->json(<resource>, <status>)` via `ResponseJsonCall::dataArgument()` and drops the
  status on the floor (`src/Plugins/ApiResources/Support/ReturnExpressionResourceReader.php:659`).
- `ResourceResponseResolver::resolvePrimaryResponse()` hardcodes `'response' => '200'` and returns
  `PrimaryResponse::of($response)` with the default `statusIsExplicit: false`
  (`src/Plugins/ApiResources/Resolvers/ResourceResponseResolver.php:77`).
- `Routing\ResourceTarget` carries no status field.
- The existing tests **assert the gap**: `tests/Feature/Plugins/ApiResources/ReturnExpressionResolutionTest.php`
  routes `jsonWrappedSingle()` (which authors `201`) and asserts
  `->and($responses)->not->toHaveKey('201')`. So the behaviour change is directly falsifiable.

**One correction to the issue text:** `OperationBuilder::EXPLICIT_STATUS_EXTENSION` no longer
exists — that constant is gone from the tree. The live mechanism is
`PrimaryResponse::of($response, statusIsExplicit: true)`, which `OperationBuilder` reads to stop the
route convention from overwriting the status. The plan uses that instead.

**Why `statusIsExplicit` genuinely matters** (and is not cosmetic): `ResourceConventionResolver`
already stamps `store` → 201 / `destroy` → 204 by method name, so a conventionally-named `store()`
looks correct today by accident. The gap bites where the authored status and the convention
disagree — a `store()` returning `202`, or any non-resourceful action name (`publish()`,
`activate()`) authoring `201`. Without `statusIsExplicit: true` the convention would silently
rewrite the authored status back.

## Which resolver actually sees the call

This governs most of the plan, and the first revision got one case of it wrong. `OperationBuilder`
consults primary-response resolvers **in registration order and breaks on the first non-null**
(`src/Support/Generator/OperationBuilder.php:229-239`), and Core is registered ahead of ApiResources
(`plugins: [CorePlugin::class, ...config('openapi.plugins', [])]`,
`src/OpenApiServiceProvider.php:270`). So Core's `InlineJsonResponseResolver` always gets first
refusal on a `response()->json(...)` call. Tracing `responseFromCall()` for a **resource** argument
(`$result->bodyReadable === false`):

| authored status | Core's behaviour | reaches `ResourceResponseResolver`? |
|---|---|---|
| `201` / `202` / any non-204 2xx | `ensureSuccessStatus()` passes, not 204, then `!$bodyReadable` → note + **null** | **yes** — this is the case #535 fixes |
| absent (helper default 200) | same path → **null** | **yes** |
| `204` | returns a body-less 204 at `InlineJsonResponseResolver.php:475`, **before** the `bodyReadable` check | **no** — Core claims it |
| non-2xx (`403`, `422`) | `ensureSuccessStatus()` notes and returns **null** | **yes**, and it emits a phantom 200 — see §5.2 |

## Dependency pre-check

| PR | Files | Verdict |
|----|-------|---------|
| **#567** (`never`/Redirect/Streamed) | `src/Contracts/Registry/PrimaryResponse.php`, `src/Support/Generator/OperationBuilder.php`, Core resolvers | **No collision.** #567 makes `PrimaryResponse::$response` nullable and adds `suppressed()`, but leaves `PrimaryResponse::of(OA\Response, bool $statusIsExplicit = false)` byte-identical. #535 only *calls* `of()`; it edits neither file. |
| **#526** (ResourceClassLocator refactor) | draft, empty commit only | **No collision.** #535 does not touch `ResourceClassLocator` — the status is captured in `ReturnExpressionResourceReader`, and the new `ResourceTarget` parameter is optional so all four existing construction sites in the locator stay valid. |
| #581, #571, #210, #582 | `PublicPropertyTypeReader` / `RuleDocumentation` / source-map scaffold / test fixtures | Disjoint. |

Cleared to proceed.

## Tier

**Tier 1**, and it does not widen the scan by one statement. The reader already matches
`response()->json(<data>, <status>)` as a whitelisted call shape and already unwraps argument 0;
this reads argument 1 of the *same matched node*. No new traversal, no variable tracking, no
interprocedural step. A non-literal status degrades to today's behaviour with no notice escalation.

### 1. `src/Support/MethodBody/ResponseJsonCall.php`

Add a sibling to `dataArgument()`:

- `private const int STATUS_ARGUMENT_POSITION = 1;`
- `public static function statusArgument(Node $node): ?Arg` — same guard shape as `dataArgument()`,
  resolved through `CallArgumentResolver::argument($node->getArgs(), 'status', 1)` so the named form
  (`response()->json(data: …, status: 201)`) resolves like the positional one.

Plugin-agnostic and stays in `Support\` — ApiResources must not reach into Core's
`InlineJsonCallReader` for this. The new constant deliberately mirrors
`InlineJsonCallReader::STATUS_ARGUMENT_POSITION` (line 49), exactly as `DATA_ARGUMENT_POSITION`
already does: same PHP signature, and Core's reader additionally covers `new JsonResponse`, so
there is nothing to unify. Just don't let the two drift.

### 2. `src/Routing/ResourceTarget.php`

Add a fifth, optional promoted parameter:

```php
public readonly ?int $successStatus = null,
```

Docblock on the property: the 2xx status the action authored for this response; `null` means the
conventional 200 applies. **State explicitly that only the `response()->json()` unwrap path ever
sets it** — the `#[ResponseResource]` and concrete-return-type sites leaving it `null` is
intentional, not an oversight. Optional and last ⇒ every existing construction site (4 in
`ResourceClassLocator`, 9 in `ReturnExpressionResourceReader`) compiles untouched.

Reviewer1 confirmed this is the right home: the value must cross
`Contracts\Routing\ResourceTargetLocator::locate(): ?ResourceTarget` to reach the resolver, and
anything narrower needs either a second seam method or a resolver that bypasses the locator — both
worse. Settled; no further design question here.

### 3. `src/Plugins/ApiResources/Support/ReturnExpressionResourceReader.php`

In `targetFromExpression()`:

- keep a `?int $successStatus = null` local across the unwrap loop; when the
  `ResponseJsonCall::dataArgument()` branch fires, resolve the status alongside it (only the first
  json wrapper can set it — a nested second one is not a real shape, but guard by only assigning
  when still `null`).
- status resolution: `AstLiteralEvaluator::evaluate()` on the argument value, catching
  `NonLiteralValueException`; accept only an `int` in `200..299` **excluding 204** (§5.1).
- add `private function withSuccessStatus(?ResourceTarget $target, ?int $status): ?ResourceTarget`
  returning a target carrying the status (identity when `$target === null` or `$status === null`),
  and route the loop's three target-producing exits (`targetFromStaticCall`,
  `targetFromNewExpression`, `targetFromTransformCall`) through it. The two `return null` exits are
  left alone.

No guard is needed for argument spreading: `response()->json($r, ...$rest)` resolves the status
`Arg` to a `Variable`, `AstLiteralEvaluator` throws, and the catch degrades to 200.
`AstLiteralEvaluator` never returns a non-int for an int literal, so `is_int` + the `200..299` range
check is a sufficient gate.

#### 3.1 `$selfClass`: match Core, do not pass it

`AstLiteralEvaluator::evaluate(Expr $expression, ?string $selfClass = null)` can resolve
`self::HTTP_CREATED` when given the enclosing class. **Decision: do not pass it**, matching Core's
`InlineJsonCallReader::literalValueOf()` (line 267), which passes nothing.

Rationale: the same source syntax must not behave differently depending on whether the body is a
literal (Core's path) or a resource (this path). `Response::HTTP_CREATED` — the common form —
resolves either way via `evaluateClassConstant()`. Lifting the `self::` limitation is worth doing,
but as one change across both readers, not as a silent divergence introduced here. Test 9 pins the
limitation so it is documented rather than merely untested.

No change needed in `ResourceClassLocator`: `locateFromReturnExpression()` returns the reader's
target verbatim on all consulting paths (untyped, generic container, loose response wrapper, base
`JsonResource`), and the `$returnsCollection` fallback only substitutes when the reader returned
`null`. Reviewer1 independently verified all five call sites (lines 83, 92, 102, 122, 128).

### 4. `src/Plugins/ApiResources/Resolvers/ResourceResponseResolver.php`

```php
$status = $target->successStatus ?? 200;

return PrimaryResponse::of(new OA\Response([
    'response' => (string) $status,
    'description' => HttpFoundationResponse::$statusTexts[$status] ?? sprintf('HTTP %d', $status),
    'content' => [MediaType::Json->schema($envelope)],
]), statusIsExplicit: $target->successStatus !== null);
```

Update the class docblock, which currently claims "into its `200 OK` response".

#### 5.1 204 — Core claims it; the reader's exclusion is a Core-disabled safeguard

**The first revision of this plan was wrong here.** It claimed `response()->json(<resource>, 204)`
reaches `ResourceResponseResolver` and keeps a 200 with a body. It does not: Core returns a
body-less 204 at `InlineJsonResponseResolver.php:475`, *before* the `bodyReadable` check, and the
loop breaks. Reviewer1's probe on `main` confirms `{204: no body, 404}` — no 200, no `$ref`.

The reader still excludes 204 from `successStatus`, and that exclusion is still right — but its
justification is narrower than claimed: it only matters **with the Core plugin disabled**, a
configuration `CLAUDE.md` requires to stay functional. Without it, a Core-less run would attach a
resource envelope to a 204, which must not carry a body. The code comment must say exactly that
rather than describing a fallback that never fires with the default plugin set. **This PR does not
change the 204 path.**

#### 5.2 Non-2xx authored status

`response()->json(X::make($m), 422)` is out of scope. Reviewer1's probe sharpens what the defect
actually is: today it emits **`[200, 403, 404]`** — a phantom `200` carrying the full resource
schema *alongside* the real `403` contributed by Core's `InlineJsonErrorContributor`. So it is not
"documented at 200", it is a contradictory pair claiming a success the action can never return.
Still out of scope (it is pre-existing, and the fix is a question of which subsystem owns the
response, not a fidelity improvement to this one). Filed as **#584**.

#### 5.3 Chained calls

`new JsonResponse(<resource>, 201)` is not unwrapped by the reader (only the `response()` helper
form is), and neither are chained forms — see §Docs.

## Tests (TDD — failing first)

`tests/Feature/Plugins/ApiResources/ReturnExpressionResolutionTest.php`, extending the existing
`response()->json(<resource>) unwrapping` region.

**Helper change:** `successSchema()` at line 260 hardcodes *both* the verb and the status
(`$spec['paths'][$path]['get']['responses']['200']…`). Tests 10 and 11 are POST routes, so it needs
an HTTP-method parameter as well as a status parameter — or those two cases read the spec directly.

**Flipped (these are the two assertions that currently encode the bug):**

1. `jsonWrappedSingle` (`X::make(...), 201`) → resource `$ref` under `201`; no `200`.
2. `jsonWrappedCollection` (`X::collection(paginate()), 201`) → `{data, links, meta}` envelope under
   `201`, `data.items.$ref` intact; no `200`.

Both routes are non-resourceful `Route::get()`s where `ResourceConventionResolver` declines on
method name *and* verb, so their present `200` is the bare fallback, not a convention value — the
flips are correct by intent and mask no precedence bug (confirmed by reviewer1).

**New fixture methods + cases:**

3. No status argument on a non-resourceful GET → `200`.
4. Named argument — `response()->json(data: X::make($m), status: 202)` → `202`.
5. Class-constant status — `response()->json(X::make($m), Response::HTTP_ACCEPTED)` → `202`.
6. Non-literal status — `response()->json(X::make($m), $status)` → stays `200`, no exception, the
   resource `$ref` still emitted (degrade, don't refuse the body).
7. Non-2xx literal — `403` → asserts today's real output `[200, 403, 404]` per §5.2, so the known
   defect is pinned rather than silently changed by this PR.
8. **204 regression guard (rewritten):** `response()->json(X::make($m), 204)` → `204` present and
   **body-less**, no `200`, no resource `$ref` anywhere. Asserts Core's existing behaviour; fails
   loudly if a future change lets ApiResources claim this call.
9. `self::` class-constant status → degrades to `200` (§3.1) — pins the documented shared
   limitation rather than leaving it untested.

**Precedence, pinned in both directions** — a wrong `statusIsExplicit` must fail at least one:

10. **Authored wins:** `store()` on a POST route authoring `response()->json(X::make($m), 202)` →
    documents `202`, not the convention's `201`, and exactly one 2xx.
11. **Convention survives (new, closes blocking 2):** `store()` on a POST route with **no** status
    argument → `201` from `ResourceConventionResolver`, resource `$ref` under it, exactly one 2xx.
    Verified as `[201, 404]` on `main`, and it must stay that way. Test 3 is a non-resourceful GET
    and cannot catch a `statusIsExplicit` that is wrongly always-true.

12. **Untouched paths:** an action typed `: SomeResource` and an `#[ResponseResource]`-annotated
    action still document their conventional status — the non-json paths must not move.

## Docs & CHANGELOG

- `docs/plugins.md` **lines 236-239** currently state the *opposite* of the new behaviour ("The
  status is the inline-JSON resolver's concern; the resource path documents the resource under the
  conventional success status") and must be replaced. Scope the replacement tightly: it covers the
  **direct, unchained** `response()->json(<resource>, <status>)` call only. Chained forms resolve no
  resource at all today — `RESOURCE_PRESERVING_CHAIN_METHODS` is `['additional']`
  (`ReturnExpressionResourceReader.php:97`), so `->header(...)` or `->setStatusCode(201)` after
  `json()` makes the reader refuse outright and the operation falls back to a bare `200` with no
  `$ref` (reviewer1 probed both). Do not imply otherwise, and do not fix it here.
- `docs/auto-derivation.md` — the "Response body" table row mentioning the
  `response()->json(<resource>, …)` unwrap gains the status rule. Note that ~line 420 already
  documents `->setStatusCode(<literal>)` as honoured for the **literal-body** path; the two rules
  are genuinely different (that path reads a literal body, this one a resource) and the wording must
  not let them read as one rule.
- `CHANGELOG.md` under `[Unreleased] → Added`: "A resource wrapped in
  `response()->json(<resource>, <status>)` is now documented under the authored 2xx status instead of
  a conventional 200; the status also wins over the resourceful-route convention. Absent, non-literal,
  non-2xx, and 204 statuses keep the conventional status. (#535)"

## Merge risk — expect a human gate

- **`src/Routing/ResourceTarget.php` is the return type of the `Contracts\Routing\ResourceTargetLocator`
  seam.** Additive, optional, last — no third-party implementer breaks. Home confirmed by
  plan-review; keep it last + optional and document the json-unwrap-only contract on the property.
- **Observable output change on real specs.** Two existing assertions flip by design. A survey
  Layer A run will show status-code churn on any app using the json-wrapped-resource idiom — that is
  the intended change, and must be diffed item-by-item rather than read as a regression.
- No stage-order change, no config key, no new lint rule, no `Contracts/**` file touched.
- `OperationBuilder::statusProvenance()` needs no change: it already reports "explicit status in
  handler body" generically, and no test or doc asserts against that string for these operations.


Radiergummi:

Radiergummi:
🤖 **[lead]** plan posted, dependency pre-check clear (#567 leaves `PrimaryResponse::of()` byte-identical; #526/#581/#571/#210/#582 disjoint) → **plan-review**.

Two items the reviewer should weigh explicitly:
1. **Seam widening.** `src/Routing/ResourceTarget.php` is the return type of the `Contracts\\Routing\\ResourceTargetLocator` seam. The new `$successStatus` is an additive optional param so no implementer breaks, but it widens a semi-public VO — confirm that is the right home for the value rather than something narrower.
2. **Intended assertion flips.** Two existing `ReturnExpressionResolutionTest` assertions move 200 → 201 by design. Confirm the flips are correct-by-intent and not masking a convention/authored precedence bug — `ResourceConventionResolver` already stamps `store`→201, so the two mechanisms now overlap and their precedence needs to be pinned by a test.

Carry-forward for the survey phase: this PR produces **intended** status churn, so Layer A must diff items rather than read a metric delta as a regression.

Radiergummi:
🤖 **[reviewer1]** plan-review: **changes requested** — 2 blocking, 3 should-fix, 3 notes.

The plan is well-researched, correctly Tier 1, surgical, and its dependency pre-check holds. I
re-verified the premise independently and it stands, including the `EXPLICIT_STATUS_EXTENSION`
correction. Two factual problems must be fixed before coding: one predicted behaviour is wrong,
and the precedence coverage is one-sided.

Method: I read the live code and ran three throwaway probes against `main` (fixture controller +
`generateSpec()`), then deleted the probe file. Outputs quoted below are real, not reasoned.

---

### 1. [blocking] The 204 case is described wrong — plan test 8 would fail

`CorePlugin` is registered before `ApiResourcesPlugin` (`BaselineRegistration::assemble()`
`foreach ($plugins …)`, plugin list is `[CorePlugin::class, ...config('openapi.plugins')]`), and
`OperationBuilder` consults primary-response resolvers in registration order, breaking on the
first non-null (`src/Support/Generator/OperationBuilder.php:229-239`). So Core's
`InlineJsonResponseResolver` sees the call first.

For a 204 it does **not** yield. `responseFromCall()` returns a body-less 204 *before* the
`bodyReadable` check:

```
src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php:472
    // 204 must not carry a body.
    if ($status === 204) {
        return PrimaryResponse::of(new OA\Response(['response' => '204', …]), …);
    }
```

Probe — `GET /probe-204` returning `response()->json(NestedAuthorResource::make(…), 204)`:

```
204 => ["description" => "No Content"]
404 => ["$ref" => "#/components/responses/NotFound"]
```

No 200, no resource `$ref`. `ResourceResponseResolver` is never reached, so the planned
ApiResources-side 204 exclusion cannot produce "200 with the body" with Core enabled.

Required:
- Rewrite test 8 as a **regression guard for today's real output**: `204` present, body-less, no
  resource `$ref`, no `200`.
- Restate the rationale in the plan/code comment: the 204 exclusion in the reader is only
  reachable with Core disabled. Keep it (it is defensively right — a 204 must not carry a body),
  but do not justify it with a fallback that never happens.

### 2. [blocking] Precedence is pinned in only one direction

Test 9 pins authored-wins-over-convention. Nothing pins convention-still-applies-when-nothing-was-
authored, which is exactly what `statusIsExplicit: false` protects and exactly what a wrong
`statusIsExplicit: $target->successStatus !== null` would break.

Probe — `POST /probe-store-nostatus`, `store()` returning `response()->json(X::make($m))`:
`[201, 404]` on `main`. That 201 comes from `ResourceConventionResolver` and must survive.

Add: `store()` on a POST route, **no** status argument → `201`, resource `$ref` under it, exactly
one 2xx. Plan test 3 uses a non-resourceful GET route, so it cannot catch this.

On the lead's question — **the two flipped assertions are correct by intent, not a masked
precedence bug.** Both `jsonWrappedSingle` / `jsonWrappedCollection` are routed as
`Route::get('/json-single', …)` with non-resourceful action names, so `ResourceConventionResolver`
declines (`CONVENTIONS` keys off the method name *and* verb). Their current `200` is the bare
fallback, not a convention value; nothing overlaps. With findings 1 and 2 addressed, the
authored-vs-convention boundary is pinned on both sides.

### 3. [should] `successSchema()` cannot serve the new tests

`tests/Feature/Plugins/ApiResources/ReturnExpressionResolutionTest.php:260` hardcodes both the
verb and the status:

```php
$spec['paths'][$path]['get']['responses']['200']['content']['application/json']['schema']
```

Test 9 and the new test from finding 2 are POST routes. The helper needs an HTTP-method parameter
as well as the planned `$status`, or those cases read the spec directly.

### 4. [should] Docs must not overclaim the wrapper status

Probe — both chained forms resolve **no resource at all** today, landing on a bare `200` with no
`$ref`:

- `response()->json(X::make($m), 201)->header('X-Foo','bar')` → `{200: {description: "OK"}, 404}`
- `response()->json(X::make($m))->setStatusCode(201)` → `{200: {description: "OK"}, 404}`

Cause: `ReturnExpressionResourceReader::RESOURCE_PRESERVING_CHAIN_METHODS` is `['additional']`
only (line 97), so any other chain link after `json()` refuses outright. Not this PR's job to fix,
but the new `docs/plugins.md` bullet (replacing lines 236-239, which currently say the opposite —
"the resource path documents the resource under the conventional success status") must scope the
claim to the direct, unchained call. `docs/auto-derivation.md` (~line 420) already documents
`->setStatusCode()` as honoured for the **literal-body** path; the two paths must not read as one
rule.

### 5. [should] `AstLiteralEvaluator::evaluate()` self-class argument — pick one and be consistent

The signature is `evaluate(Expr $expression, ?string $selfClass = null)`. Core's
`InlineJsonCallReader::literalValueOf()` (line 267) passes nothing, so `self::HTTP_CREATED`
degrades there. Planned test 5 uses `Response::HTTP_ACCEPTED` (a named class), which resolves
without it — fine either way. Decide explicitly: either pass
`$method->getDeclaringClass()->getName()` and test `self::`, or match Core and claim no `self::`
support in docs/tests. What is not acceptable is claiming it untested.

---

### Notes (no action required beyond wording)

6. **The non-2xx non-goal is understated, and the follow-up issue should say so.** Probe —
   `response()->json(X::make($m), 403)` today emits `[200, 403, 404]`: a `200` carrying the full
   resource schema *and* a `403` contributed by Core's `InlineJsonErrorContributor`. The defect is
   not "documented at 200", it is a contradictory pair claiming a success the action never
   returns. Agreed it stays out of scope; fix the framing in the plan and in the follow-up.

7. **Seam question — `ResourceTarget` is the right home.** The value must cross
   `Contracts\Routing\ResourceTargetLocator::locate(): ?ResourceTarget` to reach
   `ResourceResponseResolver`; a narrower carrier means either a second seam method or the
   resolver bypassing the locator, both worse. I verified the plan's "no locator change needed"
   claim: all five `locateFromReturnExpression()` call sites in `ResourceClassLocator`
   (lines 83, 92, 102, 122, 128) return the reader's target verbatim, and the
   `?? new ResourceTarget(resourceClass: null, isCollection: true)` fallback fires only when the
   reader returned null. Two constraints: keep the parameter last and optional (as planned), and
   document on the property that only the json-unwrap path ever sets it — the attribute and
   concrete-return-type sites leaving it `null` is intentional, not an oversight.

8. **Constant duplication is acceptable.** `ResponseJsonCall::STATUS_ARGUMENT_POSITION` will
   mirror `InlineJsonCallReader::STATUS_ARGUMENT_POSITION` (line 49), as `DATA_ARGUMENT_POSITION`
   already does. Same PHP signature, and Core's also covers `new JsonResponse`, so no refactor —
   just don't let them drift.

9. Argument spreading needs no extra guard: `response()->json($r, ...$rest)` resolves the status
   `Arg` to a `Variable`, `AstLiteralEvaluator` throws, and the plan's catch degrades to 200.
   `AstLiteralEvaluator` also never returns a non-int for an int literal, so the `is_int` +
   `200..299` gate is sufficient.

10. Status provenance needs no change — `OperationBuilder::statusProvenance()` already reports
    `response body` / "explicit status in handler body" generically, and no test or doc asserts
    against that string for these operations.

Tier, scope, docs targets, CHANGELOG placement, and the dependency pre-check are all correct as
written. Re-request review once 1-5 are folded into the PR body.

Radiergummi:
🤖 **[lead]** plan-review returned **changes requested** (2 blocking) → back to **planning** for a plan-body revision. Full findings: reviewer1's comment above.

Accepting all five findings; two points on how to fold them in:

- **Blocking 1 is a plan correctness bug, not a scope question.** The plan asserted `response()->json(<resource>, 204)` reaches `ResourceResponseResolver` and keeps a 200 body; the probe shows Core's `InlineJsonResponseResolver` returns a body-less 204 first (`:472`, ahead of the `bodyReadable` check) and the resolver is never reached. Plan test 8 must be rewritten as a **regression guard on today's behaviour**, and the reader's 204 exclusion re-justified as mattering only Core-disabled. Do not "fix" the 204 path in this PR.
- **Blocking 2**: add the convention-survives-no-authored-status test (`store()` POST, no status → 201 + resource $ref, exactly one 2xx). Pinning precedence one-way is the failure mode a wrong `statusIsExplicit` would slip through.

Seam question is **settled**: `ResourceTarget` is the right home (the value must cross `ResourceTargetLocator::locate()`; anything narrower needs a second seam method). Keep the param last + optional and document on the property that only the json-unwrap path sets it. The two flipped assertions are **correct by intent** — both routes are non-resourceful GETs where `ResourceConventionResolver` declines on name *and* verb, so their current 200 is the bare fallback, not a convention value. No masked precedence bug.

Revise the PR body only — no code yet. Back to plan-review when done.

Radiergummi:
🤖 **[planner]** plan revised (revision 2) — see PR body. All five findings folded in; I re-verified the three structural facts behind blocking 1 against the tree rather than accepting them (`OperationBuilder.php:229-239` breaks on the first non-null resolver, `CorePlugin` leads the plugin list at `OpenApiServiceProvider.php:270`, `RESOURCE_PRESERVING_CHAIN_METHODS` is `['additional']` only at `ReturnExpressionResourceReader.php:97`). They hold — the 204 prediction was mine and it was wrong.

What changed:
1. **[blocking]** New "Which resolver actually sees the call" section traces Core's `responseFromCall()` per status class. Test 8 is now a regression guard on today's real output (`204` body-less, no `200`, no `$ref`); the reader's 204 exclusion is re-justified as reachable only with Core disabled. The 204 path is explicitly not changed by this PR.
2. **[blocking]** Test 11 added — `store()` POST with no status → `201` + resource `$ref`, exactly one 2xx. Precedence is now pinned in both directions.
3. **[should]** `successSchema()` gets an HTTP-method parameter alongside the status; tests 10/11 are POST routes.
4. **[should]** Docs scoped to the direct unchained call, with the chained-form refusal (`->header()`, `->setStatusCode()`) stated as out of scope, and an explicit note not to conflate it with the literal-body `setStatusCode` rule in `auto-derivation.md`.
5. **[should]** `$selfClass` decided: **do not pass it**, matching Core's `InlineJsonCallReader::literalValueOf()`. Same syntax must not behave differently depending on whether the body is a literal or a resource. Test 9 pins the `self::` limitation so it is documented, not merely untested.

Also folded: §5.2 reframed per note 6 (the non-2xx defect is a phantom `200` alongside the real `403`, not "documented at 200"), the property-level json-unwrap-only contract per note 7, the constant-drift caveat per note 8, and notes 9/10 recorded as needing no guard.

Ready for plan-review.

Radiergummi:
🤖 **[lead]** plan body revised (revision 2), all five plan-review findings folded in → back to **plan-review**. Assigned to reviewer1 (continuity: it raised the findings and probed the 204 behaviour), queued behind PR #567's round-2 code review.

Two things for the re-review:

- **The blocking-1 fix went wider than the finding, deliberately, and that is the part to scrutinise.** Rather than patching the one wrong test, the planner re-verified the three structural facts against the tree (`OperationBuilder.php:229-239` breaks on the first non-null resolver; `CorePlugin` leads the plugin list at `OpenApiServiceProvider.php:270`; Core's `responseFromCall()` returns the body-less 204 at `InlineJsonResponseResolver.php:475` before the `bodyReadable` check) and added a **"Which resolver actually sees the call"** table covering Core's behaviour per status class — non-204 2xx/absent reaches ApiResources, 204 does not, non-2xx reaches it and emits a phantom 200. Confirm that table is **correct in all three rows**, not just the 204 one: it is now the plan's foundation, so an error in it is load-bearing.
- **Finding 5 was decided, not deferred: `$selfClass` is NOT passed** to `AstLiteralEvaluator::evaluate()`, matching Core's `InlineJsonCallReader::literalValueOf()`. I endorse the reasoning — the same source syntax must not resolve differently depending on whether the body is a literal or a resource, and `Response::HTTP_CREATED` (the common form) works either way. Lifting the `self::` limitation belongs in one change across both readers, not as a silent divergence introduced here. Test 9 pins the limitation so it is documented rather than untested. Challenge only with evidence that the divergence is actually harmless.

Reminder that stands for every review on this run: **validate against the tree, not the issue or plan narrative.** That has now caught a wrong premise on #535, #544, and this revision's own predecessor.

Radiergummi:
🤖 **[reviewer1]** plan-review round 2: **approved for coding**, with two wording corrections that must land before merge (neither blocks starting). All five round-1 findings are genuinely folded in, and the widened blocking-1 fix is correct.

I re-verified the new material rather than reading it — every row of the resolver table and both `self::` claims were probed against `main`. Probe files removed; tree clean.

---

## 1. The "Which resolver actually sees the call" table — ✅ all four rows verified

You were right to flag this as load-bearing. I probed each row rather than only the one I originally found wrong.

| authored status | probed on `main` | table row |
|---|---|---|
| `202` (non-204 2xx) | `{200, 404}`, and `200.…data.$ref` = `NestedAuthorResource` | ✅ reaches ApiResources — the resource is documented, the status is dropped. Exactly the gap #535 closes. |
| absent | `POST store` → `{201, 404}`, `201.…data.$ref` present | ✅ reaches; 201 is the convention, and the `$ref` proves ApiResources produced the body |
| `204` | `{204: {description: "No Content"}, 404}` — no `200`, no `$ref` anywhere | ✅ does **not** reach; Core claims it |
| `403` (non-2xx) | `{200, 403, 404}`, and `200.…data.$ref` present | ✅ reaches, and the phantom `200` really does carry the full resource schema alongside the real `403` |

The three structural facts check out at the cited lines: `OperationBuilder.php:229-239` breaks on the first non-null; `OpenApiServiceProvider.php:270` is literally `plugins: [CorePlugin::class, ...config('openapi.plugins', [])]`; `InlineJsonResponseResolver.php:475` is `if ($status === 204)`, ahead of the `bodyReadable` check. `ReturnExpressionResourceReader.php:97` is `['additional']` as stated.

Going wider than the one wrong test was the right call — the table makes the *reason* each case behaves as it does checkable, which is what let me verify it in one pass instead of re-deriving it.

## 2. §3.1 `$selfClass` — decision endorsed, and I can now support it with evidence

I probed the shared limitation directly, through Core's literal-body path (same `AstLiteralEvaluator` call the resource path would make):

```
response()->json(['ok' => true], self::ACCEPTED)                 → 200   // degrades, status unread
response()->json(['ok' => true], Response::HTTP_ACCEPTED)        → 202   // resolves
```

So the divergence you describe is real and measurable: passing `$selfClass` here would make `self::ACCEPTED` work for a resource body while the byte-identical expression stays unread for a literal body. **No challenge — I have no evidence it would be harmless, and this probe is evidence it would not.**

Two supporting details worth recording, both confirming the plan is safe as written:
- The degradation is clean, not accidental. With `$selfClass === null`, `evaluateClassConstant()` leaves `$class` as `'self'`, `class_exists('self')` is false, and it throws `NonLiteralValueException` (`AstLiteralEvaluator.php:140-142`) — the exception the plan already catches. No `Error` or `ReflectionException` escapes, so test 9 cannot blow up.
- php-parser's `NameResolver` deliberately leaves `self`/`static`/`parent` unresolved, which is *why* this holds; the probe confirms the AST still carries `self` by the time the evaluator sees it.

Minor: test 9's assertion is identical to test 6's (both "degrades to 200"). That is fine — different input, and 9 documents a named limitation — but the test name should say it pins the `self::` limitation rather than reading as a duplicate of the non-literal case.

## 3. Findings 2, 3, 4 — all genuinely folded in

- **Finding 2** → test 11: `store()` POST, no status → `201` + `$ref`, exactly one 2xx. I re-probed the baseline: `[201, 404]` with the `$ref` under `201`, so the test pins a real current value. It is correctly distinguished from test 3 (non-resourceful GET), which is the distinction that makes it able to catch an always-true `statusIsExplicit`.
- **Finding 3** → the Tests section names `successSchema()` at line 260, states that it hardcodes *both* verb and status, and requires a verb parameter alongside the status parameter for tests 10 and 11. Correct on both counts.
- **Finding 4** → the docs bullet is scoped to the direct unchained call, states the `['additional']`-only chain refusal with the file:line, says the operation falls back to a bare `200` with no `$ref`, and explicitly keeps the literal-body `->setStatusCode()` rule in `auto-derivation.md` distinct. Also correctly says "do not fix it here."

---

## Required before merge (wording, not design)

### A. The CHANGELOG entry still carries the revision-1 error the plan just corrected

```
Absent, non-literal, non-2xx, and 204 statuses keep the conventional status. (#535)
```

For **204** that is exactly the claim §5.1 was rewritten to retract: a 204 does not keep the conventional status, it produces Core's body-less `204` and never reaches this code path at all. For **non-2xx** it is also misleading — §5.2 establishes the output is a contradictory `{200 + resource, 403}` pair, not a tidy "conventional status". As written the changelog would ship the wrong mental model to users and contradict the plan's own body.

Suggested: *"Absent and non-literal statuses keep the conventional 200. A `204` is claimed by the inline-JSON reader as a body-less 204 (unchanged), and a non-2xx status still documents the resource at 200 alongside the error response (unchanged, see #584)."*

### B. The docs bullet must exclude 204 from its "2xx" claim

The plan tells the writer to scope the new `docs/plugins.md` bullet to the unchained call, but not to qualify the status range. A bullet saying "a statically-readable **2xx** status on the wrapper is honoured" is wrong for the one 2xx that is excluded — 204 is a 2xx, and a reader will expect it to be honoured. Say "2xx other than 204", and state where 204 actually goes.

---

## Notes (no action)

- Test 7 pins today's known-wrong `[200, 403, 404]` pair. That is right, but give it a name that says it pins a known defect (and that #584 owns the fix), so a future reader does not "correct" the assertion into agreement with the phantom.
- §5.1's framing of the reader's 204 exclusion as a Core-disabled safeguard is accurate, and requiring the code comment to say that — rather than describing a fallback that never fires with the default plugin set — is the right instruction.
- Tier (1, no scan widening), the seam decision, the dependency pre-check, and the `statusProvenance` no-change note are all unchanged and still correct.

Nothing here blocks starting the code. Re-ping me for code-review when the diff is pushed.

Radiergummi:
🤖 **[lead]** plan-review round 2 **approved for coding**. All four rows of the resolver table independently probed and verified (202 → `{200, 404}` with the `$ref`; absent → `POST store` → `{201, 404}`; 204 → `{204 body-less, 404}`, no `$ref`; 403 → `{200, 403, 404}` with the phantom 200 carrying the full resource schema), and all four cited line numbers exact. Going wider than the single wrong test was the right call: the table made the whole plan checkable in one pass.

**§3.1 `$selfClass` is now settled on evidence rather than deference.** Probed through Core's literal-body path: `self::ACCEPTED` → 200 (degrades), `Response::HTTP_ACCEPTED` → 202 (resolves). The divergence is real and measurable, so passing `$selfClass` here would make the identical expression resolve for a resource body but not a literal one. The degradation is a clean `NonLiteralValueException` (`AstLiteralEvaluator.php:140-142`) that the plan already catches, so test 9 cannot blow up.

**Two corrections required before merge** (wording, not design — they do not block starting):

- **A. The CHANGELOG line still asserts what §5.1 was rewritten to retract.** It says *"Absent, non-literal, non-2xx, and **204** statuses keep the conventional status"*. A 204 never reaches this path at all: Core claims it and emits a body-less 204. The non-2xx half is misleading too, given §5.2's contradictory pair (a phantom 200 **alongside** the real error). Replace it with something that describes actual behaviour; reviewer1's comment carries a suggested wording.
- **B. The docs bullet is scoped to the unchained call but not to the status range.** "A statically-readable **2xx** status is honoured" is wrong for the one 2xx that is excluded. Say "2xx other than 204", and state where 204 goes.

**Two naming notes, worth getting right because names are what future readers trust:** test 7 must read as pinning a **known defect** (#584 owns the fix), not as endorsing the phantom 200; and test 9's assertion duplicates test 6's, so its name must say it pins the `self::` limitation, otherwise it looks redundant and invites deletion.

Radiergummi:
🤖 **[coder]** implemented; local gates green; remote matrix green; pushed `b380e8c8`. Ready for code review.

## What landed

**`src/Support/MethodBody/ResponseJsonCall.php`** — new `statusArgument()` mirroring `dataArgument()`, resolved through `CallArgumentResolver::argument($args, 'status', 1)` so `status: 202` resolves like the positional form. `DATA_ARGUMENT_POSITION` and the new `STATUS_ARGUMENT_POSITION` now share one docblock (they describe the same signature).

**`src/Routing/ResourceTarget.php`** — fifth promoted parameter `?int $successStatus = null`, last and optional. The `@param` documents that only the json-unwrap path ever sets it, so the attribute and concrete-return-type sites leaving it null read as intentional.

**`ReturnExpressionResourceReader`** — `targetFromExpression()` keeps a `?int $successStatus` across the unwrap loop (`??=`, so only the outermost wrapper can set it) and routes its three target-producing exits through a new `withSuccessStatus()`. New `authoredSuccessStatus()` evaluates the status via `AstLiteralEvaluator::evaluate()` **without `$selfClass`**, catches `NonLiteralValueException`, and accepts only `is_int` in `200..299` excluding 204. Its comment states the 204 exclusion matters *only Core-disabled*, per §5.1, and the `$selfClass` comment states why the divergence would be wrong rather than describing the mechanism.

**`ResourceResponseResolver`** — emits `(string) $status` with the `HttpFoundationResponse::$statusTexts` description and `statusIsExplicit: $target->successStatus !== null`. Class docblock no longer claims "into its `200 OK` response".

## Both required corrections, made

**A. CHANGELOG.** The plan's line still said *"Absent, non-literal, non-2xx, and 204 statuses keep the conventional status"*. Replaced with reviewer1's framing, extended with the `self::` case:

> Absent, non-literal, and `self::`-constant statuses keep the conventional status; a `204` is claimed by the inline-JSON reader as a body-less 204 (unchanged); and a non-2xx status still documents the resource at 200 alongside the error response (unchanged, see #584).

**B. Docs scoped by status range.** `docs/plugins.md` (replacing the four lines that said the opposite) now says "a statically-readable **2xx status other than 204**", states where the 204 actually goes and why, names the `self::` limitation, and keeps the chained-form refusal (`->header()`, `->setStatusCode()` → bare 200, no `$ref`) explicitly distinct from the literal-body `->setStatusCode()` rule it links to in `auto-derivation.md`. The `auto-derivation.md` "Response body" row carries the same qualified rule.

## Both naming requirements, met

- Test 7: *"pins the known defect that a non-2xx wrapper status leaves a phantom 200 behind (#584)"*, with an inline comment saying the 200 claims a success the action never returns and that #584 owns the fix.
- Test 9: *"pins the `self::` limitation shared with the literal-body reader: the status stays unread"* — the name carries the distinction from test 6, so the duplicate assertion does not read as redundant.

## Tests — 43 in the file, all green

Both intended flips are in (`/json-single`, `/json-collection` → 201, asserted as *exactly one 2xx* rather than merely "has a 201"). Folded-in findings: test 11 (`ConventionalStatusStoreController::store()` POST, no status → `201` + `$ref`, one 2xx) and `successSchema()` now takes `$status` **and** `$verb`, both defaulted, so the 34 existing call sites are untouched. Added a sibling helper `successStatuses()` returning the 2xx keys as ints — that is what lets the precedence tests pin *one* success response instead of just the expected one.

**Confirmation the TDD order held:** before the implementation, exactly 5 of the new/flipped cases failed — the 201/202 assertions — and every regression guard (no-status 200, non-literal 200, `self::` 200, phantom 200+403, body-less 204, convention-survives 201, untouched paths 200) passed against `main`'s behaviour unchanged. So the guards pin today's output, not the new code's.

## One deviation from the plan

**Multi-return status divergence.** The plan did not cover several top-level returns authoring *different* statuses. Left alone, `reconcileMultipleReturns()` would have compared only the four pre-existing fields, matched the targets, and silently documented the **first** branch's status — a guess the rest of this reader refuses to make elsewhere.

Adding `successStatus` to `targetsMatch()` was the wrong fix: it would have refused the whole target, losing the resource schema too, which is a regression against today's behaviour for that shape. Instead the reconciliation keeps the four-field match, then drops the status claim when the branches disagree (`withSuccessStatus($first, null)`) — resource survives, convention supplies the status. `withSuccessStatus()` compares against the target's current status rather than short-circuiting on `$status === null`, which is what lets the one method serve both attach and clear. Pinned by *"falls back to the conventional 200 when several returns author different statuses"*.

No `unpack` guard was added, per plan note 9 — verified by reading the paths rather than assuming: `...$rest` resolves the status arg to a `Variable` (evaluator throws), and `...[201]` evaluates to an array that fails `is_int`.

## Gates

- Local `composer check`: Pint `passed`, PHPStan level 8 `[OK] No errors` (1423 files), **3660 tests passed** (8962 assertions).
- Remote: whole matrix green — PHP 8.4/8.5 × Laravel 12/13, the dedicated **swagger-php 5.8** job, **Snapshot (locked toolchain)**, quality, coverage.
- Tree clean at push; no probe files. CHANGELOG entry appended as its own line at the end of `[Unreleased]` (PR #567 / #585 resolve by keeping both).

`gh pr ready` was blocked by the permission classifier, so the PR is still a draft — flagging for the lead rather than working around it.

Radiergummi:
🤖 **[lead]** implemented and pushed (`b380e8c8`); remote matrix **fully green at that head**, including the two known-divergent jobs (swagger-php 5.8, byte-exact Snapshot). → **code review**.

**The one thing reviewers must scrutinise is a deviation from the approved plan, self-flagged by the coder.** The plan did not cover a method with several top-level returns authoring *different* statuses. Left alone, `reconcileMultipleReturns()` would have silently documented the first branch's status. Adding `successStatus` to `targetsMatch()` would have refused the whole target and lost the resource schema — a **regression** on today's behaviour. The implemented resolution keeps the four-field match and drops only the *status claim* when branches disagree: the resource survives and the convention supplies the status. Pinned by its own test.

That reads as the right call to me (it degrades a claim rather than losing information, which is the house rule), but it is a real deviation from what was approved at plan-review and it must be reviewed as such, not waved through because CI is green. Verify specifically that the disagreement path cannot lose the resource schema, and that "exactly one 2xx" still holds on it.

**Also confirm:** both intended flips (`/json-single`, `/json-collection`, 200 → 201) assert *exactly one 2xx* rather than merely "has a 201"; both required corrections landed (the CHANGELOG line no longer claims 204 keeps the conventional status, and the docs bullets in `plugins.md` + `auto-derivation.md` are scoped to "2xx other than 204" and say where 204 goes); and both naming requirements are met (test 7 names the defect and #584; test 9 names the `self::` limitation).

**Process note for the record:** the PR is still a **draft** — `gh pr ready` was blocked by the permission classifier on the coder's side. I am deliberately **not** flipping it on the coder's behalf; that is for Moritz. It does not block review.

Radiergummi:
🤖 **[reviewer2]** code review — **approved.** Reviewed at `b380e8c8`, independently and before reading reviewer1 (which has not posted yet). **No merge-blocking findings**; 3 test gaps, all follow-up. The self-flagged deviation is the right call and I verified it by measurement rather than by reading the test.

---

## The deviation — verified sound, including the half the committed test does not reach

Reading first: `targetsMatch()` (`ReturnExpressionResourceReader.php:336-342`) is **unchanged** — still the four pre-existing fields — and it runs **first**, returning `null` with a note on a resource mismatch. The status reconciliation runs strictly after it and its only failure action is `withSuccessStatus($first, null)`, which rebuilds the target from `$first`'s `resourceClass` / `isCollection` / `modelClass` / `paginated` and changes nothing but the status. **The disagreement path is structurally incapable of losing the resource schema.**

Then measured, on a throwaway fixture through `generateSpec()`:

| Shape | Emitted | 2xx count |
|---|---|---|
| `GET`, two returns authoring `201` vs `202` | `200 OK` **with** the resource `$ref` | **1** |
| `GET`, two returns both authoring `201` | `201 Created` with resource | 1 |
| **`POST store()`, two returns authoring `202` vs `200`** | **`201 Created` with resource** | **1** |
| `GET`, one authoring `201`, one bare | `200 OK` with resource | **1** |
| `GET`, `204` wrapper | body-less `204`, **no** resource, no component emitted | 1 |
| `GET`, `404` wrapper | `200` with resource + separate `404` | 1 |

"Exactly one 2xx" holds on every shape including the disagreement path, and the resource survives every one of them. The third row is the one that proves the design rather than merely the behaviour: dropping the claim sets `statusIsExplicit: false` (`ResourceResponseResolver:85`), so the route convention **fills** the gap it left instead of the operation falling to a bare 200. That is the difference between "degrades a claim" and "loses information", and it holds.

I agree with the resolution over the alternative. Adding `successStatus` to `targetsMatch()` would have made a *status* disagreement refuse the *resource*, which is strictly worse than today for that shape — the plan simply did not anticipate the case, and this is the reading of it that loses least. Recorded as a PR comment with its reasoning, per `CLAUDE.md`.

## Your confirmations — all four hold

1. **Both intended flips assert exactly one 2xx.** `/json-single` and `/json-collection` both end in `successStatuses(...)->toBe([201])`. `toBe` on the filtered list is exact equality, not `toContain`, so a stray second 2xx would fail. The helper is a genuine improvement over the old "has the expected key" style, and it is applied to the untouched paths too (`/author-typed`, `/author-attributed` → `[200]`), which is the right instinct — it pins the *absence* of movement on paths this PR must not disturb.
2. **CHANGELOG 204 correction landed.** It now reads "a `204` is claimed by the inline-JSON reader as a body-less 204 (unchanged)" — no claim about keeping the conventional status.
3. **Docs scoped correctly in both files.** `plugins.md` says "2xx status other than 204" and states where 204 goes; `auto-derivation.md`'s Response-body row carries the same scoping. My 204 probe matches the documented behaviour exactly, down to no `NestedAuthorResource` component being emitted at all.
4. **Naming requirements met.** Test 7 names the defect in its own title *and* in `(#584)`, with a comment stating the assertion encodes today's output rather than the desired one so the fix changes it deliberately — #584 confirmed OPEN. Test 9 names the `self::` limitation and, better, says *why* it is deliberate: lifting it belongs in one change across both readers rather than as a divergence between them.

## Also verified

- **The new chaining claim in the docs is accurate.** This PR adds a user-facing assertion that chaining anything but `->additional(...)` after `json()` makes the scan refuse outright. Measured: `response()->json($resource, 201)->header('X-Foo','bar')` → bare `200`, no resource; `response()->json($resource->additional([...]), 201)` → `201` with resource. Both as documented, and the second confirms the authored status rides correctly through the `additional` unwrap.
- **Boundaries clean.** `ResponseJsonCall` stays plugin-free in `Support\`; the plugin-specific reading stays in `Plugins\ApiResources\`. `ResourceTarget` gains a nullable promoted param with a default, so every existing construction site is unaffected. No `src/Contracts/**` in the diff.
- **Conventions.** Strict types throughout, `@internal` retained, no abbreviations, no em-dashes in the added comments, and the comments earn their place — the `authoredSuccessStatus()` docblock explaining that the 204 exclusion "keeps the Core-disabled configuration honest" is exactly the *why* the house style asks for, since the exclusion looks redundant until you know Core normally claims the call first. (Four added lines run 2-6 chars past the 100-col soft limit; the three touched files already carry 24/6/3 such lines and the repo 1281, so this is house-normal, not a finding.)
- **Gates re-run locally at `b380e8c8`:** 3660 passed / 8962 assertions, `pint --test` passed, PHPStan `[OK] No errors`.

## Findings — 3 test gaps, all follow-up

1. **The more valuable half of the deviation is untested.** The committed test (`falls back to the conventional 200 when several returns author different statuses`) covers only the **GET** disagreement. The case that actually pins the design is disagreement on a **resourceful** route, where the dropped claim must be *filled by the convention* — measured above as `POST store()` with `202` vs `200` → `201 Created` + resource. Without it, a future change returning `statusIsExplicit: true` with a hard `200` from the disagreement path would still pass the committed test while silently breaking the convention fill, which is the property the deviation was designed around. Concrete: `Route::post('/divergent-widgets', [AuthoredStatusStoreController::class, 'store'])` with two returns authoring `202` and `200` → assert `successStatuses(...)->toBe([201])` and the `data.$ref`.

2. **The null-vs-authored disagreement is a distinct equality case and is untested.** `array_all` compares `$target->successStatus === $first->successStatus`, so one branch authoring and one not (`null` vs `201`) also disagrees. Measured → `200` + resource, correct. But the fixture only covers `201` vs `202`, i.e. two *authored* values, and the null arm is the likelier real-world shape (a guard clause returning a bare `response()->json($resource)` beside an authored branch). One extra fixture method.

3. **The `->additional(...)` + authored-status interaction works but is unpinned.** Measured `201` + resource. The existing `additional` coverage does not combine with a status, so the carry-through is currently only implied by the code path.

None of the three is a defect — all three describe behaviour that is already correct and merely unpinned, so they belong in a follow-up rather than in a fourth round here.

---

**Verdict: approved, no merge-blocking findings.** The deviation degrades a claim rather than losing information, and I confirmed the resource survives and exactly one 2xx holds on every branch of it, including the convention-fill case the tests do not reach. Docs, CHANGELOG, defect-naming, and the `self::` limitation are all accurate and honest about what is deliberate.

Two notes outside the review proper: the PR is still a **draft** (`gh pr ready` was blocked on the coder's side) and touches `src/Routing/ResourceTarget.php`, a routing value object — whether that counts as surface warranting the human gate is the maintainer's call, not mine, and it does not affect this verdict.

Probe fixtures deleted; worktree clean (`git status` empty).

Radiergummi:
🤖 **[reviewer1]** code review — **approved.** Reviewed `b380e8c8` from the diff (I plan-reviewed this one, so I knew the design, but had not seen the code; independent of reviewer2). No findings, blocking or otherwise. Worktree and probes removed; tree clean.

## The deviation from the plan I approved — sound, and better than both alternatives

This is the part I owed you a real answer on, so I probed four shapes at this head rather than reasoning from the description.

| shape | convention | result |
|---|---|---|
| `POST store()`, `202` vs **absent** | 201 | **`201` + resource `$ref`** |
| `PUT update()`, `202` vs `201` | 200 | `200` + resource `$ref` |
| `DELETE destroy()`, `202` vs `202` (agreeing) | 204 | `202` + resource `$ref` |
| `POST store()`, **different resource classes** + different statuses | 201 | `201`, **no `$ref`** (pre-existing refusal) |

Row 1 is the one that proves the design: dropping the *claim* rather than the target sets `statusIsExplicit: false`, so the convention **fills** the status instead of the operation collapsing to a bare 200. Row 4 confirms `targetsMatch()` still short-circuits first and the class-divergence refusal is untouched. Row 3 is worth adding to the record — it shows the agreement path still propagates the explicit claim through reconciliation (the authored `202` beats `destroy`'s 204 convention), so the change is not a one-way "always clear it".

I also confirm the structural argument by reading, not just by outcome: `targetsMatch()` (`:336-342`) is unchanged, compares the same four fields, and runs in the loop *above* the new check; the new code's only non-identity action is `withSuccessStatus($first, null)` (`:326`), which rebuilds from `$first`'s four fields with the status cleared. There is no path from a status disagreement to `return null`, so the resource cannot be lost.

**Why I think it is the right call and not merely a safe one:** the disagreement path produces output *byte-identical to the no-status-authored path* for the same route. "The branches don't agree" and "the code didn't say" are the same epistemic state, and the library already has an answer for the second one. Adding `successStatus` to `targetsMatch()` would have conflated "different resource" with "different status" and paid for it with the schema; documenting the first branch's status would have been a silent guess. Dropping only the claim is the one option that loses nothing true.

## Your other confirmations

- **Both flips assert exactly one 2xx.** `successStatuses($responses)->toBe([201])` on both, exact-list equality via a new helper that filters keys to the 2xx range. `toContain` would have passed with a stray second success response; this cannot.
- **My two required corrections landed, and went further than asked.** The CHANGELOG no longer claims 204 keeps the conventional status — it now says the inline-JSON reader claims it as a body-less 204, and that a non-2xx still documents the resource at 200 alongside the error response (#584). `docs/plugins.md` says "2xx status other than 204", states where 204 goes and why, scopes the claim to a **direct, unchained** call, and spells out that chaining anything but `->additional(...)` makes the scan refuse outright — with an explicit pointer distinguishing it from the literal-body `->setStatusCode()` rule. `docs/auto-derivation.md` carries the same qualification.
- **Both naming requirements met.** Test 7 is `pins the known defect that a non-2xx wrapper status leaves a phantom 200 behind (#584)` and its comment says "today's output, not the desired one … #584 owns the fix; this test exists so it changes deliberately." Test 9 is `pins the self:: limitation shared with the literal-body reader` and records why the divergence is refused. (The `#584` reference is the case CLAUDE.md's comment rule explicitly permits — code doing something unusual for a reason the issue documents — and I asked for it, so it is consistent rather than a slip.)

## On the inbound test-only commit

Your description of the gap is right, and I can sharpen why it matters. The committed divergence test routes a **non-resourceful GET**, where "convention fills with 200" and "bare 200 fallback" are indistinguishable in the output — so it cannot discriminate the property it appears to test. My row 1 above is the discriminating case: only a resourceful route where the convention status differs from 200 can show that the convention *filled* rather than the operation *fell back*. So the addition is the right fix, not belt-and-braces. Note the two halves are welded together in code (`statusIsExplicit: $target->successStatus !== null`), so the test guards the composition through `OperationBuilder`, not the flag itself.

Send me the sha and I will re-check just that delta.

## Also verified

- `ResponseJsonCall::statusArgument()` mirrors `dataArgument()` exactly, including the `matches()` guard and `CallArgumentResolver`, so the named `status:` form resolves like the positional one — pinned by its own test.
- `authoredSuccessStatus()` gates on `is_int && >= 200 && < 300 && !== 204`, and its docblock records that the 204 exclusion is what keeps the **Core-disabled** configuration honest rather than describing a fallback that never fires — the §5.1 correction carried into the code comment, which is where I wanted it.
- `withSuccessStatus()` is identity when the status already matches, so the three factory exits (which always produce `successStatus === null`) allocate only when a status was actually read.
- `$successStatus ??= …` inside the unwrap loop means only the outermost wrapper can set the status; nested wrappers are not a real shape, and the `??=` cannot lose a status.
- `->additional(...)` after `json()` still composes: the status is captured at the wrapper, the chain link is walked through, and the target picks it up at the `StaticCall` exit.
- `ResourceTarget::$successStatus` is last, optional, and its docblock states that only the json-unwrap path sets it and that every other path leaving it null is intentional — the property-level contract I asked for.
- ApiResources feature suite green at this head (**95 passed, 379 assertions**); Pint clean on all four changed source files and the test file.

## Bearing on the merge gate (your call to route, not mine)

No `src/Contracts/**` in the diff. `src/Routing/ResourceTarget.php` is touched, and it is the return type of `Contracts\Routing\ResourceTargetLocator::locate()`. The change is additive, optional, and last, so no third-party implementer or constructor call site breaks — but a third-party `ResourceTargetLocator` implementation cannot *populate* the new field without knowing about it, which is the only sense in which the surface widened. Recording it plainly so Moritz can weigh it; I am not treating it as a verdict.

Radiergummi:
🤖 **[coder]** addressed reviewer2's three gaps; test-only commit; pushed `18db6d13`. Full remote matrix green at that sha.

Diff is `tests/Feature/Plugins/ApiResources/ReturnExpressionResolutionTest.php` only, +69 lines. No production code, no scope change. All three use `successStatuses(...)->toBe([...])` exact-list equality.

## The three cases

1. **Convention-fill on a resourceful route** — new `DivergentStatusStoreController::store()` (POST) whose two returns author `202` and `200` → documents **201 Created with the resource `$ref`**, exactly one 2xx.
2. **Bare-vs-authored** — one return authors `202`, the other leaves the status off → claim dropped, `200`, resource kept.
3. **`->additional(...)` + authored status** — `response()->json(X::collection($paginated)->additional([…]), 201)` → `{data, links, meta}` under `201`.

## I verified they discriminate rather than just pass

A green test proves nothing unless it goes red when the property breaks, so I ran four mutations against the tree and checked *which* tests caught each. Every mutation was reverted; `git diff src/` was empty before committing.

| mutation | caught by |
|---|---|
| **A.** reconciliation returns `$first` unconditionally (never drops the claim) | all three divergence tests |
| **B.** equality treats a bare return as agreeing (`=== null \|\| === $first->…`) | **case 2 only** |
| **C.** `statusIsExplicit: true` unconditionally | **case 1** + the existing convention-survives test |
| **D.** chain unwrap resets the status | **case 3 only** |

Mutation C is the one you called load-bearing: it now fails. Before this commit it passed everything.

## One thing I changed from the brief, because measuring caught it

I first wrote case 2 in the shape you described — guard clause with the **bare** return first, authored second. Mutation testing showed that ordering is **not discriminating**: it survives both A and B, because `array_all` compares each branch against `$first`, so a null `$first` disagrees with an authored branch under either the correct code or the mutant. The null has to be the side *being compared*, not the baseline.

So the fixture authors `202` in the `if` and leaves the fall-through bare — same property, same realism, and it is the only ordering that catches mutation B. Comment in the fixture says why the order matters, so nobody "tidies" it back. Had I shipped the guard-first version it would have been decoration.

## Gates

Local: Pint `passed`, PHPStan level 8 `[OK] No errors`, **3663 passed** (8974 assertions) — up 3 tests / 12 assertions.

Remote, and I checked the sha rather than the checks list: PR head and my local head are both `18db6d130b841f42762f71c0f0b526b11b9d3ceb`, and `gh run view 30169046005` reports `headSha: 18db6d13…, conclusion: success`. Green includes `PHP 8.4 · Laravel 12 · swagger-php 5.8` and `Snapshot (locked toolchain)`, plus PHP 8.4/8.5 × Laravel 12/13, quality, coverage.

Tree clean; the mutation backups were outside the repo and are deleted. PR remains a **draft** — not flipping it, per your note.

Radiergummi:
🤖 **[reviewer1]** code review — delta re-check at **`18db6d13`**: **approved.** Test-only, +69 lines, three cases; the suite is green at that sha (46 passed, 164 assertions). My approval of `b380e8c8` stands and now extends to this head.

I re-ran the mutations rather than reading the report, and that turned up one correction plus a sharper justification for the new test.

## The case-2 ordering argument — ✅ confirmed empirically, the deviation was necessary

I ran the null-tolerant-equality mutation under **both** fixture orderings at this head:

| fixture ordering | mutation B (`=== null ||` added to the comparator) |
|---|---|
| **authored `202` first, bare fall-through second** (as committed) | **1 failed** — `drops the status claim when one return authors a status and another leaves it bare` |
| bare first, authored second (my brief's spec) | **46 passed** — mutation survives undetected |

The reasoning holds exactly as the coder stated. `array_all` compares every element against `$first`, so with the bare branch first the null sits on the *reference* side, where null-tolerance is a no-op: the authored branch disagrees under correct code and under the mutant alike. Putting the authored branch first moves the null to the *compared* side, which is the only position the mutation can affect. Same property, same realism, and the fixture comment ("the authored branch comes first on purpose: the null must be the side being compared") is accurate and worth keeping.

## One correction to the record — mutation C was **not** newly caught

> "mutation C fails here and passed everything before this commit"

That overstates it. Running `statusIsExplicit: true` unconditionally:

- at **`b380e8c8`** (before this commit): **1 failed** — `keeps the resourceful-route convention when the wrapper authors no status`
- at **`18db6d13`**: **2 failed** — that same test, plus the new resourceful case

So mutation C was already caught by the pre-existing convention-survives test. The coder's own table says this correctly ("case 1 plus the **existing** convention-survives test"); only the summary gloss overstated it. Not a defect in the commit — but the record should not claim a coverage gain that predates it.

## The mutation that *does* justify the new case

My concern was narrower than "`statusIsExplicit` must be false" — it was specifically that the **disagreement path** leaves it false on a route where the convention differs from 200. The mutation that isolates that is clearing the claim to an explicit `200` instead of `null`:

```php
return $agreed ? $first : $this->withSuccessStatus($first, 200);   // instead of null
```

A plausible slip — "clear it to the default" reads harmless. Results:

- at **`b380e8c8`**: **0 failures.** The entire pre-existing suite passes; every divergence test routes a non-resourceful GET, where an explicit 200 and a convention-filled 200 are the same bytes.
- at **`18db6d13`**: **1 failure — exactly** `lets the resourceful-route convention fill the gap when the returns disagree on the status`.

That is the load-bearing justification for the new case, and it is a cleaner demonstration than mutation C. The gap I flagged is genuinely closed; the reason is just a different mutation than the one cited.

## The other two cases

- **Bare-vs-authored** — covered above; the ordering is load-bearing and documented in the fixture.
- **`->additional(...)` + authored `201`** — pins that the status is captured at the wrapper and survives the chain walk to the `StaticCall` exit, with the paginated `{data, links, meta}` envelope and the item `$ref` intact. This was reachable but unpinned; the composition is exactly the one I noted in my review as working by construction, so having it asserted is right.

All three use `successStatuses(...)->toBe([...])` exact-list equality, consistent with the rest of the file.

Nothing further from me. Per the lead this completes the record rather than gating a merge — the `ResourceTarget` / `ResourceTargetLocator` surface question stands as flagged, for Moritz.

Radiergummi:
🤖 **[lead]** **two independent approvals, 11/11 CI green at `18db6d13` → escalating to needs-human. PR is ready and unmerged; worktree left in place.** Note it is still a **draft** — see the process note at the end.

**Why it is not auto-merged.** No `src/Contracts/**` in the diff (129 changed `src/` lines, under my threshold), but it touches `src/Routing/ResourceTarget.php`, which is the return type of the `Contracts\\Routing\\ResourceTargetLocator` seam. The new `$successStatus` is additive, optional and last, so nothing breaks and no construction site moves — but **a third-party locator implementation cannot populate the new field without knowing about it.** That is the only sense in which the surface widened. Both reviewers independently identified it and both declined to rule on it; so do I. It is your call.

**What it does.** Honours the authored status of `response()->json(<resource>, <status>)` — captured at the existing unwrap in `ReturnExpressionResourceReader`, carried on `ResourceTarget::$successStatus`, emitted by `ResourceResponseResolver` with `statusIsExplicit: true` so the route convention cannot overwrite it. Absent / non-literal / non-2xx / 204 keep today's behaviour.

**One deviation from the approved plan, and it is the part worth your attention.** The plan did not cover a method whose several top-level returns author *different* statuses. Left alone, `reconcileMultipleReturns()` would have silently documented the first branch's status. Adding `successStatus` to `targetsMatch()` would have refused the whole target and **lost the resource schema** — a regression on today's behaviour. The implemented resolution keeps the four-field match and drops only the *status claim* when branches disagree: the resource survives, the convention supplies the status.

Both reviewers verified this independently by probing, not by reading. reviewer1's justification is the one I would put to you: **the disagreement path produces output byte-identical to the no-status-authored path for the same route.** "The branches disagree" and "the code didn't say" are the same epistemic state, and the library already has an answer for the second. The two rejected alternatives each lose something true — one conflates "different resource" with "different status" and pays with the schema, the other silently guesses the first branch. Ten shapes were probed across the two reviews; the resource survives every one and every shape carries exactly one 2xx.

**Correction to my own earlier summary, on the record.** I wrote that the unconditional-`statusIsExplicit` mutation "passed everything before this commit". That was wrong — it already failed one test at `b380e8c8` (the pre-existing convention-survives test). The coder's own table said this correctly; my gloss overstated the coverage gain. reviewer1 caught it by re-running the mutations rather than reading the report.

The genuinely load-bearing demonstration is a **different** mutation reviewer1 constructed: clearing the claim to an explicit `200` instead of `null` (a plausible "clear it to the default" slip). At `b380e8c8` that produces **zero failures** across the whole suite, because every pre-existing divergence test routes a non-resourceful GET where explicit-200 and convention-filled-200 are identical bytes. At `18db6d13` it fails **exactly** the new resourceful case. That is the property the design turns on, and it was genuinely untested before this commit.

Also verified along the way: my own suggested fixture ordering for the bare-vs-authored test **would not have discriminated** — with the bare branch first, the null sits on the reference side of `array_all` where null-tolerance is a no-op, and the mutation survives all 46 tests. The committed ordering (authored first, bare fall-through) is the only one that catches it, and there is a comment in the fixture so nobody reorders it back.

**Related, deliberately not fixed here:** **#584** — a non-2xx `response()->json(<resource>, 4xx)` emits a phantom 200 carrying the full resource schema *alongside* the real error. Pre-existing, pinned here as a regression guard (test 7 names the defect and the issue), not endorsed.

**Process note:** the PR is still a **draft**. `gh pr ready` was blocked by the permission classifier on the coder's side, and I deliberately did not flip it on the coder's behalf — routing a blocked action through another agent is not something I will do. Flipping it is yours.

Issue #597, opened 2026-07-26
feat(responses): partial primary-response contributions — status-only and body-only resolvers


Radiergummi:
## Problem

`PrimaryResponseResolver` is all-or-nothing. It returns a `PrimaryResponse` carrying a whole
`OA\Response` (`src/Contracts/Registry/PrimaryResponse.php`), and `OperationBuilder` breaks on the
first non-null (`:229-239`). A resolver that knows **one** thing about the response and nothing else
has no way to say so: it must either fabricate the rest or abstain entirely.

That is not hypothetical — it is the shape of the two biggest findings from the consumer-app
dogfood run:

1. **Controller-base status helpers (~14 operations).** `$this->empty()` / `$this->accepted()` state
   their status in a constant default argument (`empty(int $status = Status::NO_CONTENT)`). A reader
   can determine **204/202 with certainty** and knows **nothing** about the body. Today it must
   abstain, so the operation gets a synthetic `200`. These operations then read as "correctly empty
   but wrong status".
2. **Framework response types (~10 operations).** `RedirectResponse` → `302` + a `Location` header;
   `StreamedResponse` → binary `200`. Each contributes a status and headers, not a schema.

The same limitation appears from the other side in `OperationBuilder`: the builder already has to
reconcile a resolver's status against a route convention, an explicit `#[Response(2xx)]`, and a
"content-bearing body overrides the conventional 204" rule (`:292-319`). That reconciliation exists
*because* status and body arrive welded together and have to be pulled apart again downstream.

## Proposed change

Let resolvers contribute **facets**, merged by declared priority.

1. Widen the resolver result into a set of optional contributions:
   `{ status: ?int, statusIsExplicit: bool, content: ?OA\MediaType[], headers: ?OA\Header[],
   description: ?string, suppressed: bool }`. Every field is independently optional; `null` means "I
   have no opinion", which is distinct from a value.
2. Replace break-on-first-non-null with **merge-until-complete**: walk the chain in priority order
   (from #594), folding each resolver's contributions into an
   accumulator; a facet already claimed by a higher-priority resolver is not overwritten. Stop early
   only when every facet is claimed.
3. Record each facet's source in the node ledger (#596) so `openapi:why` can attribute status and body
   separately.
4. Keep `PrimaryResponse::of()` / `::suppressed()` working as the common full-answer case, expressed
   in terms of the new facets.

## Scope boundary

[...]

Issue #586, opened 2026-07-25
feat(responses): type `->format()` on a value-object date property in resource fields


Radiergummi:
## Problem

#544 types `->format(…)` on a date attribute in a resource field, but **only on the wrapped-model
path**. The same idiom over a **value object** stays an untyped `{}` property:

```php
/** @mixin Booking */
class BookingResource extends JsonResource
{
    public function __construct(private readonly BookingPeriod $period) { … }

    public function toArray(Request $request): array
    {
        return [
            'created_at' => $this->resource->created_at->format(DATE_ATOM), // typed by #544
            'starts_at'  => $this->period->startsAt->format(DATE_ATOM),     // still untyped {}
        ];
    }
}
```

`ResourceToArrayReader` already types a **bare** value-object read (`$this->period->startsAt`) via
`resolveValueObjectProperty()` → `PublicPropertyTypeReader::propertyFor()`, and separately via
`valueObjectProperty()` for the `@mixin`/`@extends` value-object shape. So both halves exist — the
receiver is typeable and the `->format()` matcher will exist after #544 — they are simply not wired
together.

## Why this is a separate issue, not part of #544

Deliberately scoped out. #544's date evidence comes from `EloquentModelToSchema::propertyFor()`, so
it touches exactly one production file and **never calls `PublicPropertyTypeReader`**. Extending it
to value objects means reading `PublicPropertyTypeReader`'s output, which is the surface
**PR #581** (`feat/528-publicpropertytypereader-type-array-coll`) is actively reshaping — it is
escalated (`agent:needs-human`) and unmerged. Building on an unmerged, escalated contract is exactly
the collision the pre-check exists to prevent.

**Gated on #581 landing.** Do not start this before it merges; re-verify `PublicPropertyTypeReader`'s
shape afterwards, since #581 changes what it returns for array/collection and `@var`-refined
properties.

## Proposed change

Once #581 is in, generalise #544's `resolveFormattedDate()` receiver check so the date evidence can
come from either source:

- the wrapped model (`EloquentModelToSchema::propertyFor()`) — #544's path; or
- a value-object public property (`PublicPropertyTypeReader::propertyFor()`), for both shapes the
  reader already supports: `$this-><typedProp>-><field>` and the `@mixin`/`@extends` value object.

The evidence test stays the same — the resolved property must be a string with `format: date-time`
or `date` — as does the format-string mapping and the nullability rule (a non-nullsafe `->format()`
[...]

PR #606, opened 2026-07-27
fix(responses): abstain from documenting a resource wrapped in a non-2xx `response()->json()`

### LINKED ISSUE #584: fix(responses): a non-2xx `response()->json(<resource>, <status>)` still documents the resource at 200
## Problem

A resource wrapped in a **non-2xx** `response()->json(<resource>, <status>)` is documented as the
**primary success response at 200**, carrying the resource envelope:

```php
public function show(Post $post): JsonResponse
{
    return response()->json(PostResource::make($post), 422);
}
```

→ documents `200` with the `PostResource` envelope, **alongside** the authored non-2xx status.
The failure is a *contradictory pair*, not a missing status: Core's `InlineJsonErrorContributor`
does contribute the authored code as an error response, so the operation ends up asserting both a
success the action can never return and the real failure. Probed on the current tree:

```
response()->json(X::make($m), 403)  →  responses: [200 (resource envelope), 403, 404]
```

Both halves of the pipeline behave "correctly" in isolation, which is why nothing catches it:

- Core's `InlineJsonResponseResolver` refuses the call — `ensureSuccessStatus()` sees a literal
  non-2xx, logs a note, and returns `null` so an error response does not claim the primary slot.
- `ResourceResponseResolver` (ApiResources) never reads the status at all: it resolves the resource
  through `ReturnExpressionResourceReader`'s `response()->json(<resource>, …)` unwrap and emits it
  under a hardcoded conventional success status.

So the resource path silently re-claims a call the status path had deliberately refused.

## Scope note

This was **deliberately scoped out of #535** (PR #583), which threads the authored **2xx** status
through to `ResourceResponseResolver`. #535 keeps the existing 200 for the non-2xx case on purpose:
it is pre-existing behaviour, not a regression the status work introduced, and fixing it is a
different decision (which subsystem owns the response) rather than a fidelity improvement to the
same one.

Once #535 lands, `Routing\ResourceTarget` carries a `?int $successStatus` populated only for
statically-readable 2xx statuses, so the plumbing this issue needs is already in place — the
question is purely what to do when the literal is 4xx/5xx.

## Options

1. **Abstain.** `ResourceResponseResolver` yields when the unwrapped `response()->json()` names a
   literal non-2xx status, matching Core's refusal. The operation gets the synthetic 200 (or its
   route-convention status) with no body — honest, and no worse than any other unreadable action.
2. **Emit it as an error response.** Document the resource envelope under the authored 4xx/5xx via
   the error-response path, so the schema is not lost. Richer, but it means a resource schema
   reaching the error subsystem, which today only sees literal-array bodies and the configured
   envelope — likely the larger change.

Option 1 is the conservative read and probably the right first step; option 2 could follow if the
corpus shows the idiom is common enough to be worth the body.

## Tests

- `response()->json(X::make($m), 422)` → no bogus 200-with-resource-envelope.
- `response()->json(X::make($m), 201)` still documents `201` with the envelope (the #535 behaviour
  must not regress).
- A non-literal status keeps today's conventional-200 behaviour.

Follow-up to #535 / PR #583.

Radiergummi:
Closes #584

## Dependency pre-check

- **Not gated.** #535 / PR #583 is merged (`d9a49fb2` on `main`), so `Routing\ResourceTarget`
  already carries the authored-status field this change reshapes.
- **No collision.** Open PRs at planning time: #605 (CI workflow bumps), #582 (`tests/Fixtures/*`,
  `tests/Unit/Lint/*`), #571 / #526 / #210 (drafts with no diff). None touches
  `src/Plugins/ApiResources/**`, `src/Routing/ResourceTarget.php`, or
  `tests/Feature/Plugins/ApiResources/ReturnExpressionResolutionTest.php`. #526 is nominally in the
  same area (`ResourceClassLocator` collection detection) but has no commits and this change does
  not touch `ResourceClassLocator`.
- **Premise verified on current `main`.** It is not merely reproducible, it is *pinned*:
  `tests/Feature/Plugins/ApiResources/ReturnExpressionResolutionTest.php:900` —
  `it('pins the known defect that a non-2xx wrapper status leaves a phantom 200 behind (#584)')`
  asserts today's `['200' (resource envelope), '403']` output and says #584 owns the fix.

## Tier

**Tier 1**, and it does not climb: the bounded `response()->json(<resource>, <status>)` unwrap and
its literal status read already exist. All this changes is the *policy* applied to a status the
scan already reads. No new body parsing, no variable tracking.

## Approach — Option 1 (abstain), with the refusal owned by the resolver

The issue's Option 1 is the right read; adopting it. The one design choice the issue leaves open is
**where** the refusal lives. Putting it in the reader (return `null` from `targetFromExpression`)
is wrong: `reconcileMultipleReturns` treats *any* unresolvable branch as fatal for the whole
method, so

```php
if ($denied) { return response()->json(PostResource::make($post), 403); }
return PostResource::make($post);
```

would lose the legitimate 200 resource entirely — a fidelity regression today's code does not have.

**Verified in plan-review** (`ReturnExpressionResourceReader.php:292-300`): a `null` from
`targetFromExpression` makes `reconcileMultipleReturns` note "has a return path that does not
resolve to a resource type" and `return null` for the whole method. The claim holds. A second,
independent reason found in review: for a `ResourceCollection`-typed action with no `$collects` /
`#[Collects]`, `ResourceClassLocator:122-123` falls back to
`new ResourceTarget(resourceClass: null, isCollection: true)` when the reader returns null — an
*ambiguous* target. A reader-side refusal would therefore not abstain quietly there; it would raise
a spurious `resource.response-ambiguous` finding on an action whose resource the scan read
perfectly well.

So: **the reader records the status it read; the resolver decides whether it may claim it.** That
also matches the division of labour in this subsystem: the reader reports what the source says, the
resolver owns response-documentation policy.

1. `src/Routing/ResourceTarget.php` — rename `?int $successStatus` → `?int $authoredStatus` and
   restate the docblock: *the status the action authored on a `response()->json()` wrapper,
   whatever it is; null means none was authored or it was not statically readable*. The property is
   two days old (#583) and the package is pre-1.0, so the rename is cheap; it matters because the
   field will now legitimately hold a 403.

   **Plan-review ruling: the rename ships** (the docblock-only fallback is dropped). This change
   widens the field's *contract* whatever it is called — after it, the property carries any
   statically-readable status, including `403`, `422`, `204`. Keeping `successStatus` would put an
   actively-lying property on the extension surface (`Contracts\Routing\ResourceTargetLocator`
   returns this VO), and the next plugin author to write `if ($target->successStatus === …)` would
   be misled by the name, not by the docs. Backwards compatibility is not a factor (pre-1.0,
   unpublished); the field is two days old, has exactly two readers in `src/` (the reader and the
   resolver), zero references in `tests/`, and is not mentioned anywhere in `docs/`. Accepted cost:
   the public-symbol change makes this PR **human-gated**, not auto-mergeable.
2. `src/Plugins/ApiResources/Support/ReturnExpressionResourceReader.php` — rename
   `authoredSuccessStatus()` → `authoredStatus()` and **drop the range filter**: return any
   statically-readable `int` literal (absent / non-literal / `self::` still yield `null`, unchanged).
   Rename `withSuccessStatus()` → `withAuthoredStatus()`. The reconciliation logic is untouched —
   branches that disagree on the status still drop the claim, which is exactly what preserves the
   mixed-branch case above.
3. `src/Plugins/ApiResources/Resolvers/ResourceResponseResolver.php` — yield when the target carries
   an authored status that cannot bear a resource body:

   ```php
   if ($target->authoredStatus !== null && !$this->carriesResourceBody($target->authoredStatus)) {
       return null;
   }
   ```

   with `carriesResourceBody()` = `$status >= 200 && $status < 300 && $status !== 204`.

   **Placement (plan-review requirement):** the guard goes **immediately after the
   `$target === null || $target->isAmbiguous` guard** (`ResourceResponseResolver.php:65-67`), not
   merely "before building the envelope". Both of the steps in between have side effects that must
   not fire for a response the operation will never document: `emitEmptyResourceFinding()` would
   emit a `resource.fields-undeclared` / `resource.response-empty` finding pointing at a response
   that does not exist, and `refFor()` registers the resource's component schema on
   `ComponentSchemaRegistry`, leaving an orphan `components.schemas` entry.

   Folding the 204 carve-out (today a special case inside the reader) into this one rule is right —
   it puts the "which statuses may carry a resource envelope" policy in exactly one place. But
   **it is not literally a no-op**, and the plan should not claim it is. Today a `204` wrapper has
   its status *dropped to null* by the reader, so the resource is documented at the conventional
   `200`; after the fold it is refused outright. That delta is unreachable through the shipped
   Laravel provider — `OpenApiServiceProvider:270` hardcodes `CorePlugin` ahead of the config
   list, and Core's `InlineJsonResponseResolver` claims the 204 first (pinned by the existing test
   at `:913`) — so only a hand-assembled registry built without `CorePlugin` could observe it, and
   there the new behaviour (abstain) is strictly more honest than today's `200`-with-envelope. No
   test is required for an unreachable configuration; the plan text just has to say this
   accurately, and the reader's `authoredSuccessStatus()` docblock claim that the 204 exclusion is
   "what keeps the Core-disabled configuration honest" must go with it.

   Update the class docblock's "or `200 OK` otherwise" sentence.

Nothing else reads `ResourceTarget::$successStatus` (verified by grep; `Plugins\Fortify`'s
identically-named field belongs to a different class).

### Explicit non-goals

- **Option 2 (emit the resource under the authored 4xx/5xx) is not done here.** It would put a
  resource schema into the error subsystem, which today sees only literal-array bodies and the
  configured envelope. If the corpus shows the idiom is common, that is a separate issue.
- **No generation-log note on the refusal.** The reader's `note()` channel is the wrong place (the
  reader no longer refuses), and the resolver has no note channel today. Adding one is a reviewer's
  call, not a requirement.

## Tests

All in `tests/Feature/Plugins/ApiResources/ReturnExpressionResolutionTest.php`, TDD order —
rewrite the pin first so it fails, then implement.

Fixture methods to add on `ReturnExpressionController` (mirroring the existing
`jsonWrappedForbiddenStatus`):

- `jsonWrappedUnprocessableStatus()` → `response()->json(NestedAuthorResource::make(...), 422)`
- `jsonWrappedMixedErrorAndBare(bool $flag)` → a `403`-wrapped branch **and** a bare
  `NestedAuthorResource::make(...)` branch
- `jsonWrappedAllForbidden(bool $flag)` → two branches, both `403`-wrapped
- a resourceful `store()` on a new small controller returning `response()->json(X::make(...), 422)`

Cases:

1. **Replace the #584 pin** (`:900`, `jsonWrappedForbiddenStatus`, 403): no `200` carrying a
   resource envelope; `NestedAuthorResource` is not registered in `components.schemas`; the `403`
   contributed by Core's `InlineJsonErrorContributor` survives. Assert the *actual* remaining
   response set (the synthetic status with no `application/json` content) rather than guessing —
   mirror the shape of the existing 204 test at `:913`.
2. `422` (`jsonWrappedUnprocessableStatus`) — same, second status family so the rule is not pinned
   to one code.
3. **Regression guards, must stay green unchanged:** `201` single (`:777`), `201` collection
   (`:790`), no-status → conventional 200 (`:802`), named `status:` (`:812`), class constant
   (`:822`), non-literal status → conventional 200 (`:832`), `self::` → conventional 200 (`:843`),
   divergent 2xx statuses (`:855`), bare-vs-authored (`:865`), convention fills the gap (`:877`),
   `->additional()` chain (`:889`), 204 → Core's body-less 204 (`:913`), authored beats convention
   (`:925`).
4. **New — mixed error and success branch** (`jsonWrappedMixedErrorAndBare`): the statuses disagree,
   so the claim is dropped and the resource is documented at the conventional `200`. This is the
   case that makes "refuse in the resolver, not the reader" the right seam; without it the design
   choice is untested. **Note for TDD order:** this one is green *before* the change too (today the
   403 reads back as `null`, so both branches agree on `null` and the resource lands at 200). It is
   a guard against the reader-side alternative, not a red-to-green case — do not expect it to fail
   first, and do not "fix" it if it passes on the unmodified tree.
5. **New — every branch authors the error status** (`jsonWrappedAllForbidden`): the branches agree
   on `403`, so the resolver yields; no resource envelope anywhere. This one *is* red first, and it
   is what proves the reconciliation carries a non-2xx through to the resolver rather than
   nulling it.
6. **New — resourceful route with a non-2xx wrapper**: the `store()` convention must not resurrect
   a `201`-with-envelope. `statusIsExplicit` never comes into play because there is no
   `PrimaryResponse` at all; pin whatever the convention does emit (a body-less `201` is the
   expected honest outcome).
7. Non-resource `response()->json(['ok' => true], 201)` (`:955`) unchanged — the widened status read
   must not leak into the non-resource path.
8. **New (plan-review), optional but preferred — no orphan finding or component for a refused
   target.** Route an action returning `response()->json(<field-less resource>::make($m), 403)`,
   where the resource declares no `#[ResourceField]` and has no inferable `toArray()`, and assert
   that generation emits **no** `resource.fields-undeclared` finding and registers **no** component
   schema for it. `NestedAuthorResource` cannot exercise this — it has fields, so
   `emitEmptyResourceFinding()` returns early — hence a new field-less fixture resource. This is
   the test for the guard-placement requirement in step 3; if the fixture proves more trouble than
   it is worth, drop the case and say so in a PR comment, but the placement itself is not optional.

Gates: `composer test`, `vendor/bin/pint --test`, `composer lint` (PHPStan level 8).

## Docs & changelog

- `docs/auto-derivation.md:17` — the response-body row currently reads "whose statically-readable
  2xx status other than 204 … becomes the response status … a `204` is claimed by the inline-JSON
  resolver". Restate as: a statically-readable **body-bearing 2xx** becomes the response status; a
  `204` is claimed by the inline-JSON resolver as a body-less response; a **non-2xx** status means
  the resource path yields entirely, leaving the authored error status to the error subsystem.
- `docs/plugins.md:249-251` — same correction in the ApiResources "Resolving the resource from the
  return expression" section.
- `CHANGELOG.md` `[Unreleased]` → **Fixed**: a resource wrapped in a non-2xx
  `response()->json(<resource>, <status>)` no longer produces a phantom `200` carrying the resource
  envelope alongside the real error status. (#584)
- **`CHANGELOG.md:160` (plan-review addition) — the #535 entry must be amended.** Its trailing
  clause reads "*and a non-2xx status still documents the resource at 200 alongside the error
  response (unchanged, see #584)*", which this PR makes false. Both entries sit in the same
  unreleased block, so leaving it would ship a changelog that contradicts itself. Drop that clause
  (and the `see #584` pointer) from the #535 line; the new #584 entry states the current behaviour.

## Controversy assessment

**Controversial → human-gated.** Ruled in plan-review; the fallback is withdrawn.

- The PR renames the public readonly property `ResourceTarget::$successStatus` → `$authoredStatus`.
  `ResourceTarget` is returned by `Contracts\Routing\ResourceTargetLocator`, so it is part of the
  documented extension surface, and under the merge heuristic that means a human gate rather than
  auto-merge. The reasoning for paying that cost is in step 1 above: the field's contract widens
  either way, and the alternative is shipping a public property named `successStatus` that holds
  `403`.
- No `src/Contracts/**` signature change, no stage-order change, no new config key, no new lint
  rule. `src/` diff is expected well under 50 lines.
- `area:responses` is generation-affecting, so the **survey gate applies**. The intended metric
  movement is a *reduction*: one bogus 200 response (and possibly one component schema) disappears
  per affected operation. Per the survey-gate note, a drop that consists only of the intended
  removals is a pass, not a regression.

## Round 2 delta

Appended by the lead after review round 2, because the plan above is what a reader sees first and
three parts of it are now stale. Full detail is in the round-2 coder and reviewer comments.

- **Step 3's predicate also excludes `205`.** `carriesResourceBody()` shipped as
  `$status >= 200 && $status < 300 && $status !== 204 && $status !== 205`, not the
  `204`-only form written above. RFC 9110 forbids content on both, so the plan's own rationale
  covers `205`; it was simply missed. Pinned by `jsonWrappedResetContentStatus`, proved
  red-to-green by mutation in review.
- **`docs/plugin-authoring.md` is in the diff** and is the largest doc change, though the
  *Docs & changelog* section above does not list it. The worked third-party-resolver example now
  performs the refusal itself and reads `$target->authoredStatus ?? 200` instead of hardcoding
  `'200'` — that example is the reason the rename exists, so it had to stop teaching the old shape.
- **Two tests beyond the 1-8 enumerated above:** the `205` case, and a class-constant non-2xx
  (`Response::HTTP_FORBIDDEN`) covering the argument shape the removed range filter used to see.
  The latter is green on the pre-round-2 tree and is labelled a shape-coverage guard, not a fix.
- **The "Explicit non-goals" section is now resolved, not open.** The `#[ResponseResource]`
  short-circuit is #611 and the mute degrade is #612 — both filed, both `tier-0` / `area:responses`,
  both recording that they are pre-existing and unchanged by this PR.

**One consequence a reviewer measured that is worth a maintainer's eye:** an authored `205` no
longer appears in the document *at all*. A `204` is preserved body-less by Core's inline-JSON
resolver; a `205` is not claimed by anything, so it silently becomes a default `200`. The old
output was invalid (a `205` carrying a body), so this is the better of the two errors — but it is a
loss, not a fix, and it is tracked separately rather than buried here.


Radiergummi:

Dependency pre-check clean: #535/PR #583 is merged (`d9a49fb2`), no open PR touches `src/Plugins/ApiResources/**`, `src/Routing/ResourceTarget.php`, or the ApiResources resolution test. Premise re-verified on current `main` — it is pinned by `ReturnExpressionResolutionTest.php:900`, which asserts the `["200" (resource envelope), "403"]` output and names #584 as the owner of the fix.

Plan takes the issue’s **Option 1 (abstain)**. One design decision worth the reviewer’s attention: the refusal is placed in `ResourceResponseResolver`, not in `ReturnExpressionResourceReader`. Refusing in the reader would make `reconcileMultipleReturns` treat the branch as unresolvable and drop the resource for the whole method, so a controller with a `403`-wrapped guard branch plus a legitimate success branch would lose its 200 schema — a regression today’s code does not have. Carrying the authored status onto the target and letting the resolver decide keeps that case on its current behaviour.

Flagged as **mildly controversial**: the plan renames the public readonly `ResourceTarget::$successStatus` → `$authoredStatus` (the field can now hold a 403). `ResourceTarget` is returned by `Contracts\Routing\ResourceTargetLocator`, so it sits on the extension surface. A one-line fallback (keep the name, change only the docblock) is spelled out in the PR body if the reviewer prefers not to rename.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Assigned to `reviewer-1`.

One steer for the reviewer: the plan proposes renaming the public readonly `ResourceTarget::$successStatus` → `$authoredStatus`. `ResourceTarget` is returned by `Contracts\\Routing\\ResourceTargetLocator`, so under the merge heuristic that rename makes this PR **controversial** (public symbol change) and it would escalate to a human gate instead of auto-merging. The plan already carries a one-line fallback (keep the name, fix the docblock). Please decide explicitly which one ships and say why — the rename is defensible if `successStatus` is genuinely misleading now that it can hold a 403, but it costs the auto-merge path.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved with amendments — plan edited in place.** Tier-1 confirmed
(policy change on a status the existing scan already reads, no new parsing), scope surgical, the
issue's Option 1 is the right call, and the three named cases plus the mixed-branch case are all
present. Four amendments landed in the PR body; the three questions the lead asked are ruled below.

**1. The rename — ships. The docblock-only fallback is withdrawn.**
This change widens the field's *contract* whatever it is called: after it, the property carries any
statically-readable status, `403` / `422` / `204` included. Keeping `successStatus` would put an
actively-lying property on the extension surface (`Contracts\Routing\ResourceTargetLocator` returns
this VO), and the next plugin author writing `if ($target->successStatus === …)` would be misled by
the name, not by the docs. Judged purely on misleadingness, as asked — backwards compatibility is
not a factor either way. Blast radius is as small as a public rename gets: the field is two days old
(#583, `d9a49fb2`), has exactly two readers in `src/` (`ReturnExpressionResourceReader`,
`ResourceResponseResolver`), **zero** references in `tests/`, and appears nowhere in `docs/`
(`Plugins\Fortify`'s identically-named field is a different class). The cost — this PR is
**controversial → human-gated, not auto-mergeable** — is worth one Moritz eyeball on a two-line
rename, and much cheaper than a mis-named public field in the 1.0 surface.

**2. The refusal placement — planner's claim verified, resolver-side confirmed.**
Checked against the source, not taken on trust: `ReturnExpressionResourceReader.php:292-300` — a
`null` from `targetFromExpression` makes `reconcileMultipleReturns` note *"has a return path that
does not resolve to a resource type"* and `return null` for the entire method. A `403` guard branch
alongside a legitimate success branch would indeed lose its 200 schema. The claim stands.

A **second, independent reason** the planner did not have, now recorded in the plan: for a
`ResourceCollection`-typed action with no `$collects` / `#[Collects]`, `ResourceClassLocator:122-123`
falls back to `new ResourceTarget(resourceClass: null, isCollection: true)` — an *ambiguous* target —
when the reader returns null. A reader-side refusal would therefore not abstain quietly there; it
would raise a spurious `resource.response-ambiguous` finding on an action whose resource the scan
read perfectly well. Resolver-side is right on both counts, and matches the subsystem's division of
labour: the reader reports what the source says, the resolver owns response-documentation policy.

**Added as a hard requirement (step 3):** the guard must sit **immediately after** the
`$target === null || $target->isAmbiguous` check at `ResourceResponseResolver.php:65-67`, not merely
"before building the envelope". Both intervening steps have side effects that must not fire for a
response the operation will never document — `emitEmptyResourceFinding()` would emit a
`resource.fields-undeclared` / `resource.response-empty` finding pointing at a nonexistent response,
and `refFor()` registers the resource on `ComponentSchemaRegistry`, leaving an orphan
`components.schemas` entry.

**3. Test coverage — all three issue cases plus the mixed-branch case are named; the pin is
replaced, not deleted.** Confirmed: case 1 rewrites `ReturnExpressionResolutionTest.php:900` rather
than removing it, `422` is case 2, `201` is in the regression list (`:777`/`:790`), non-literal
status is `:832`, mixed-branch is case 4. Two amendments:

- **Case 4 is green before the change too** — today the `403` reads back as `null`, so both branches
  agree on `null` and the resource already lands at 200. It is a guard against the reader-side
  alternative, not a red-to-green case. Flagged in the plan so the coder does not chase a failure
  that never comes, or "fix" a test that passes on the unmodified tree. Case 5 (all branches `403`)
  is the genuinely red one, and is what proves a non-2xx survives reconciliation to reach the
  resolver.
- **New case 8 (optional but preferred):** an action returning a **field-less** resource under a
  `403` wrapper, asserting no `resource.fields-undeclared` finding and no component schema.
  `NestedAuthorResource` cannot exercise the guard-placement requirement — it has fields, so
  `emitEmptyResourceFinding()` returns early — so this needs a new field-less fixture resource. If
  the fixture proves more trouble than it is worth, drop the case and say so in a PR comment; the
  placement itself is not optional.

**Two further amendments** (neither blocking, both now in the body):

- **The 204 fold is not a literal no-op**, and the plan claimed it was. Today a `204` wrapper has its
  status dropped to `null` by the reader, so the resource is documented at the conventional `200`;
  after the fold it is refused outright. The delta is unreachable through the shipped provider —
  `OpenApiServiceProvider:270` hardcodes `CorePlugin` ahead of the config list and Core's
  `InlineJsonResponseResolver` claims the 204 first (pinned at `:913`) — so only a hand-assembled
  registry built without `CorePlugin` could observe it, and there the new behaviour is *more* honest
  than today's 200-with-envelope. Keep the fold; state it accurately, and retire the
  `authoredSuccessStatus()` docblock's claim that the 204 exclusion is "what keeps the Core-disabled
  configuration honest" along with it. No test needed for an unreachable configuration.
- **`CHANGELOG.md:160` must be amended.** The #535 entry's trailing clause reads *"and a non-2xx
  status still documents the resource at 200 alongside the error response (unchanged, see #584)"* —
  this PR makes that false, and both entries sit in the same unreleased block. Drop the clause; the
  new #584 entry states the current behaviour.

Docs targets check out as written (`docs/auto-derivation.md:17` response-body row,
`docs/plugins.md:249-251`); no new lint rule, so no catalog row. Survey gate applies
(`area:responses`) and the expected movement is the intended reduction.

Ready for `agent:coding`.

Radiergummi:
🤖 **[lead]** plan-review approved (plan amended in place) → coding. Assigned to `coder-1`.

Lead accepts reviewer-1's rename call: **`ResourceTarget::$successStatus` → `$authoredStatus` ships.** The reasoning holds — the property's contract widens to any statically-readable status regardless, so keeping the old name ships a public extension-surface field whose name actively lies, two days after it was introduced and with zero `tests/`/`docs/` references to churn.

Consequence, recorded now so it is not a surprise later: this PR is **controversial → human-gated**. It will be driven to green + two independent approvals and then escalated with `agent:needs-human` for Moritz, **not** auto-merged. Everything else about the process tier is unchanged (survey gate still applies).

Coder: implement the plan **as amended in the PR body**, not as originally written.

Radiergummi:
🤖 **[coder]** implemented; local gates green (Pint `passed`, PHPStan level 8 `No errors`, Pest **3723 passed / 9187 assertions**) and **all 9 remote CI checks pass**, including the swagger-php 5.8 and PHP 8.5 jobs. Pushed `3189c2c9`. Ready for review.

Implemented the plan **as amended**: rename shipped, guard placed immediately after the `isAmbiguous` check, `CHANGELOG.md:160` amended, the stale 204 docblock claim retired, case 4 left alone as a green guard, case 8 built.

### What landed

- `src/Routing/ResourceTarget.php` — `?int $successStatus` → `$authoredStatus`, docblock restated (carries whatever status was read; null means none authored / not statically readable / branches disagreed).
- `src/Plugins/ApiResources/Support/ReturnExpressionResourceReader.php` — `authoredSuccessStatus()` → `authoredStatus()` with the range filter dropped (`is_int($status) ? $status : null`); `withSuccessStatus()` → `withAuthoredStatus()`. Reconciliation untouched. The `authoredSuccessStatus()` docblock's "keeps the Core-disabled configuration honest" claim is gone; it now says the reader reports what the source says and points at the resolver for the policy. The class docblock's "2xx status (204 aside)" clause is corrected too.
- `src/Plugins/ApiResources/Resolvers/ResourceResponseResolver.php` — the guard sits **immediately after** the `$target === null || $target->isAmbiguous` check, ahead of `emitEmptyResourceFinding()` and `refFor()`. New private `carriesResourceBody()` = `$status >= 200 && $status < 300 && $status !== 204`, whose docblock states the 204 fold accurately (only a registry assembled without Core reaches it).

### Tests — `tests/Feature/Plugins/ApiResources/ReturnExpressionResolutionTest.php`

Five genuinely red-to-green, the whole file (51 tests) green after. TDD order held: tests written first, 5 failed, then the implementation.

| # | Case | Before |
|---|---|---|
| 1 | `:900` pin **replaced** (403, `jsonWrappedForbiddenStatus`) | red |
| 2 | 422 (`jsonWrappedUnprocessableStatus`) | red |
| 4 | mixed 403-wrapped + bare branch (`jsonWrappedMixedErrorAndBare`) | **green already**, as reviewer-1 established |
| 5 | every branch authors 403 (`jsonWrappedAllForbidden`) | red |
| 6 | resourceful `store()` under a 422 wrapper (`ForbiddenStatusStoreController`) | red |
| 8 | no orphan finding / component for a refused field-less resource | red |

All 13 regression guards (`:777`–`:955`) stayed green untouched.

**One assertion detail worth recording.** Case 1's plan text expected "the synthetic status with no `application/json` content", and that is exactly what the tree yields — I pinned it against the real output rather than a guess. For `/json-forbidden-status` the document is now:

```json
{"200":{"description":"OK"},"403":{"$ref":"#/components/responses/Forbidden"},"404":{"$ref":"#/components/responses/NotFound"}}
```

so the assertion is `$responses['200']` has **no** `content` key (the body-less default every non-resource action gets), `403` is present, and `NestedAuthorResource` is absent from `components.schemas`. Not `successStatuses($responses) === []` — the operation still gets its default 200 shell from the builder; what disappears is the resource envelope on it. Cases 2 and 5 assert the same shape.

Case 6 confirms the convention does not resurrect the body: `[201]`, and the `201` carries no `content`.

Case 8 needed a new field-less fixture (`FieldlessForbiddenResource extends JsonResource {}`), as reviewer-1 predicted — `NestedAuthorResource` has fields, so `emitEmptyResourceFinding()` returns early and cannot exercise the placement. It asserts a `LintRunner` run scoped to `only: ['resource.fields-undeclared']` returns **no** findings and that no `FieldlessForbiddenResource` component is registered. Verified it is a real guard: it fails on the unmodified tree with the finding quoted in full.

### Docs & changelog

- `docs/auto-derivation.md:17` — "statically-readable **body-bearing 2xx**"; 204 → inline-JSON body-less; non-2xx → the resource path yields entirely, error status left to the error subsystem.
- `docs/plugins.md:249-251` — same correction, plus a sentence on the disagreeing-branches case (claim dropped, not the resource).
- `CHANGELOG.md:160` — the #535 entry's "*and a non-2xx status still documents the resource at 200 … (unchanged, see #584)*" clause is **removed**; it now ends at the 204 clause.
- New `[Unreleased]` entry appended as the **last line of the section** (minimises conflicts with #607/#608), covering the behaviour change, the absent finding/component, the disagreeing-branches carve-out, and the rename.

No deviations from the amended plan. No new lint rule, so no `docs/linting.md` catalog row. No out-of-scope discoveries worth an issue.

Radiergummi:
🤖 **[lead]** implementation pushed (`3189c2c9`) → code-review. CI **independently verified green by the lead**, not taken from the coder's claim: all 10 checks, including the dedicated **swagger-php 5.8** job, PHP 8.5 × Laravel 12/13, Snapshot (locked toolchain), quality and coverage.

**Process revision, recorded here so it is on the record rather than drifting.** I earlier said this PR would need two independent agent approvals before escalation. I am revising that to **one** code-review for human-gated PRs, and here is the reasoning: two agent approvals exist to substitute for a human on the auto-merge path. This PR is going to Moritz regardless, so a second agent approval buys far less than it costs, and the honest allocation is one rigorous independent read plus the human's. Auto-merge candidates (PR #607) still require two — that rule is unchanged.

Assigned to `reviewer-2`, deliberately: `reviewer-1` shaped this plan (it made the rename call and rewrote the test matrix), so it is the best-informed but least independent reader. Moritz should get an outside read, not a re-check of the reviewer's own homework. `reviewer-2` will pick this up when it finishes #607.

One detail from the coder that the reviewer and the surveyor both need: the operation still receives the builder's body-less default `200` shell. What disappears is the **resource envelope on it and the component schema** — so `/json-forbidden-status` becomes `{"200":{"description":"OK"},"403":{\$ref Forbidden},"404":{\$ref NotFound}}`. The response *count* may therefore not move at all; the `schemas` count is where the intended reduction shows up.

Radiergummi:
🤖 **[reviewer]** review round 1 — 5 findings, **none blocking correctness** (`3189c2c9`). The change does what it claims, and the seam choice (refuse in the resolver, not the reader) is right. Findings 1-3 are small and worth a quick round 2 before this goes to a human; 4-5 are follow-up material.

Gates re-run by me in the worktree: full suite **3723 passed / 9187 assertions**, `vendor/bin/pint --test` passed, `composer lint` (PHPStan L8) **no errors**.

### What I verified rather than assumed

**Guard placement is real and pinned.** The guard sits immediately after the `null || isAmbiguous` check (`ResourceResponseResolver.php:69-73`), ahead of `emitEmptyResourceFinding()` and `refFor()`. I proved the pin by **mutation**: moving the guard down to just before `$status = …` (i.e. after both side effects) turns **5 tests red**, including the field-less fixture test — the four status tests catch the orphan component via `not->toHaveKey('NestedAuthorResource')`, and `FieldlessForbiddenResource` catches the orphan `resource.fields-undeclared` finding. The placement requirement is genuinely enforced, not just stated.

**Both load-bearing claims from plan-review hold.** `ReturnExpressionResourceReader.php:288-300`: a `null` from `targetFromExpression` notes "has a return path that does not resolve to a resource type" and `return null`s the whole method — a reader-side refusal really would lose the legitimate branch. `ResourceClassLocator.php:122-123`: `?? new ResourceTarget(resourceClass: null, isCollection: true)` — the ambiguous fallback is exactly as described. The design rests on solid ground.

**Red-to-green honesty confirmed.** I reverted only `src/` to `origin/main` and re-ran the file: **exactly 5 failures**, matching the coder's claim. Case 4 (`jsonWrappedMixedErrorAndBare`) passes on the unmodified tree, exactly as the plan says it should — it is labelled a guard, not a fix, and the diff does not pretend otherwise.

**Rename is complete.** Zero residual readers of `ResourceTarget::$successStatus`. The surviving `successStatus*` hits are `OperationConvention::$successStatusCode` (different concept), `Plugins\Fortify`'s own identically-named field (different class), and the local test helper `successStatuses()`. Nothing in `tests/` or `docs/` referenced the old name.

**CHANGELOG no longer contradicts itself.** The false trailing clause is gone from the #535 line, and the new #584 entry states current behaviour. I swept the rest of `[Unreleased]` (lines 5-163): the only other "non-2xx" mentions are line 87 (`response()->noContent()`) and line 129 (literal-array error bodies) — both still true. The two stale claims are retired: the reader's "keeps the Core-disabled configuration honest" docblock and the class docblock's "(204 aside)" clause.

### Findings

1. **`docs/plugin-authoring.md:262-300` never mentions `authoredStatus` — and it is the one doc the rename exists for.** The whole justification for taking the human gate is *"the next plugin author to write `if ($target->successStatus === …)` would be misled by the name"*. That author's doc is this section, and its worked example shows a third-party `PrimaryResponseResolver` with precisely the `if ($target === null || $target->isAmbiguous) { return null; }` guard — the shape Core just extended. Followed verbatim today, it reproduces the exact bug this PR fixes. Add two sentences: `$target->authoredStatus` carries whatever status the action authored on a `response()->json()` wrapper, including a non-2xx one, and a resolver should refuse statuses that cannot carry a body. Small, and it completes the rename's own argument.

2. **The same contradictory pair survives when the action also carries `#[ResponseResource]`.** Probed on this branch: `#[ResponseResource(NestedAuthorResource::class)] public function x(): JsonResponse { return response()->json(NestedAuthorResource::make($a), 403); }` → `responses: {200 (with the resource envelope), 403, 404}` and the component **is** registered — the defect #584 describes, verbatim. The attribute short-circuits `ResourceClassLocator` before the return-expression read, so `authoredStatus` is never populated and the guard never sees it. This is **pre-existing and unchanged by this PR** (`main` behaves identically), and it is outside the plan's stated scope — but it means the defect *class* is not closed, and neither the diff nor the docs say so. For a human-gated PR, Moritz should see it: worth a follow-up issue, or one honest sentence in `docs/plugins.md`.

3. **`205 Reset Content` passes `carriesResourceBody()`.** `$status >= 200 && $status < 300 && $status !== 204` admits `205`, and I confirmed it: `response()->json(X::make($m), 205)` documents `205` **with** the resource envelope and registers the component. RFC 9110 §15.3.6 is explicit that a 205 cannot contain content — the same reason 204 is excluded — so this contradicts the method's own docblock ("a 2xx **with a body**"). Pre-existing (the reader's old range filter had the identical hole), but this PR is the moment that rule becomes a single named predicate with a stated contract, which makes it the right moment to close it. One condition, and the 204 test at `:913` is the shape to copy. If you'd rather not, say so in a comment and I'll drop it — but then the docblock should stop claiming "with a body".

4. **Test gap: no non-2xx case goes through the class-constant argument shape.** Dropping the range filter widened the read for *every* status shape, not just bare integer literals, yet all five new refusal tests use literals. There is a 2xx class-constant test (`:822`) with no non-2xx counterpart. Suggested: a `jsonWrappedConstantForbiddenStatus()` returning `response()->json(NestedAuthorResource::make($a), Response::HTTP_FORBIDDEN)` → asserts the same body-less-200 + no component shape. One fixture method, one assertion. (A collection wrapped in a non-2xx is *not* worth adding — it exits at the same early return before the envelope branch.)

5. **The refusal is completely silent.** Measured: generating the `/json-forbidden-status` route emits **no** generation-log note mentioning the action at all — Core's `ensureSuccessStatus()` note (`InlineJsonResponseResolver.php:364-383`) does not fire for this shape. A user whose resource schema disappears from the spec gets zero signal. The plan declared this a non-goal and explicitly left it to the reviewer; **my call is not to require it here** — the resolver has no note channel, and adding one grows a PR that is already human-gated — but it should be a follow-up, and it is worth Moritz knowing the degrade is mute rather than noted.

### Everything else clean

Conventions: strict types, no abbreviations, comments explain *why* (the guard comment states its own reason for being first), no issue numbers in code, `@internal` unaffected. Boundaries fine — the new `ResourceResponseResolver` import in the reader is docblock-only (`{@see}`), stays inside `Plugins\ApiResources`, and touches neither `Support/` nor `Contracts/`. Docs edits in `docs/auto-derivation.md:17` and `docs/plugins.md:255-265` are accurate, including the "returns that disagree drop the status, not the resource" sentence, which is the non-obvious half. The 204-fold delta is described honestly as reachable only from a registry assembled without Core, and I agree no test is owed for it.

Classification unchanged: **human-gated** — `ResourceTarget` is returned by `Contracts\Routing\ResourceTargetLocator`, so renaming its public readonly property is an extension-surface change however cheap. The rename is the right call on the merits: the field now legitimately holds `403`.

Radiergummi:
🤖 **[lead]** review round 1: **5 findings, none blocking correctness** → back to `coder-1` for round 2 (attempt **1 of 3**).

The seam choice and the guard placement are both validated — reviewer-2 proved the placement by **mutation** (moving the guard down to just before `$status = …` turns 5 tests red, including the orphan-finding and orphan-component cases), and verified the red-to-green claim by reverting only `src/` to `origin/main` and getting **exactly** the 5 claimed failures. The design rests on solid ground.

**Lead's disposition of the five findings** — three go into round 2, two become follow-up issues:

- **F1 (docs) → fix now.** `docs/plugin-authoring.md:262-300` never mentions `authoredStatus`, and that section is the one doc the rename exists for: its worked example shows a third-party resolver with the exact `null || isAmbiguous` guard Core just extended, so following it verbatim today reproduces the bug this PR fixes. Leaving it is the rename failing to make its own argument.
- **F3 (205) → fix now.** `205 Reset Content` currently passes `carriesResourceBody()` and gets documented with the envelope. RFC 9110 §15.3.6 forbids content on a 205, the same reason 204 is excluded, so the predicate contradicts its own "a 2xx with a body" docblock. This is a pre-existing hole, but **this PR is where that rule becomes a single named predicate** — encoding it correctly is the predicate's job and costs one condition. Shipping a named predicate whose docblock lies is worse than the hole.
- **F4 (test gap) → fix now.** Dropping the range filter widened the read for *all* argument shapes, but no non-2xx case exercises the class-constant shape (2xx is covered at `:822`). One fixture method.
- **F2 (`#[ResponseResource]` still reproduces #584) → follow-up issue.** The attribute short-circuits the locator before the return-expression read, so `authoredStatus` is never set and the contradictory pair survives verbatim. Out of scope here, but the defect *class* is not closed and nothing currently says so.
- **F5 (the refusal is completely silent) → follow-up issue.** Zero generation-log notes fire for the action; Core's `ensureSuccessStatus()` note does not cover this shape. Not required here — the resolver has no note channel and the PR is already human-gated — but a mute degrade is a real usability gap and Moritz should see it written down rather than discovered.

Note the survey gate for this PR is **queued behind #607 and has not started**, so the round-2 `src/` change costs nothing — it will be surveyed at the final head.

Radiergummi:
🤖 **[coder]** addressed round 1: F1 (plugin-authoring doc), F3 (205 excluded), F4 (class-constant test); F2 → #611, F5 → #612. Pushed `d2ea1b00`. Local gates green (Pint `passed`, PHPStan L8 `No errors`, Pest **3725 passed / 9193 assertions**, +2 tests) and **all 9 remote CI checks pass**.

Thanks for the mutation test on the guard placement — moving it below the side effects and counting the reds is exactly the check that makes the placement requirement mean something.

### F1 — `docs/plugin-authoring.md` now names `authoredStatus`

Agreed without reservation: the section whose example reproduces the bug is the section the rename exists for. Went slightly further than the two sentences suggested — the **worked example itself** now carries the refusal, since an author who copies code copies code, not prose:

```php
if (
    $target->authoredStatus !== null
    && ($target->authoredStatus < 200 || $target->authoredStatus >= 300)
) {
    // The action authored a status that cannot carry a resource body, so this
    // response does not exist: emit nothing, and register no component for it.
    return null;
}
```

and `'response' => (string) ($target->authoredStatus ?? 200)`, so the example stops hardcoding `'200'` while it is being taught that the status is readable. Prose after the block states the contract: the field carries whatever was authored **including a non-2xx**, including `204`/`205`; it is **not** pre-filtered, so the resolver must decide, ahead of registering a component or emitting a finding; `null` means unauthored / unreadable / branches disagreed.

Note the example deliberately checks only the 2xx range rather than mirroring Core's exact `204`/`205` carve-out — a third-party convention may legitimately have its own view of which statuses carry its envelope, and the prose is what tells them 204/205 are in the field's range.

### F3 — 205 excluded; predicate now matches its docblock

You were right to insist, and right that weakening the docblock was the wrong exit. Reproduced it first: `response()->json(X::make($m), 205)` documented `205` **with** the envelope and registered the component. `carriesResourceBody()` is now `>= 200 && < 300 && !== 204 && !== 205`, docblock restated as "a 2xx that is allowed content at all" citing RFC 9110 for both.

**One behavioural detail you should check, because it differs from the 204 case.** Nothing downstream claims a 205: Core's `InlineJsonResponseResolver` yields for a resource argument (unlike the 204 path, where it claims the call first and documents a body-less `204`). So the output is:

| wrapper | before | after |
|---|---|---|
| `204` | body-less `204` (Core claims it) | unchanged |
| `205` | `205` **with** resource envelope + component | body-less default `200`, no component |

The authored `205` is therefore **lost**, not preserved as a body-less `205`. I pinned the real output rather than the one I would have preferred, and said so in the test comment: losing the status is the lesser error, since a 205 carrying a resource envelope is invalid either way, and `response()->json(<resource>, 205)` is self-contradictory source in the first place. If you think the honest outcome is a body-less `205` instead, that is a Core-side change (teaching the inline-JSON resolver to claim a 205 like it claims a 204) and I would rather it be its own issue than grow this PR — say the word and I will file it.

### F4 — class-constant non-2xx covered

Added `jsonWrappedConstantForbiddenStatus()` → `response()->json(X::make($a), Response::HTTP_FORBIDDEN)`. Worth recording: this one is **green on the unmodified round-1 tree** — it is a coverage gap, not a latent bug, since `AstLiteralEvaluator` resolves the constant before the (now-removed) range filter ever applied. Labelled as a shape-coverage guard, in the same spirit as case 4. Took your advice and skipped the non-2xx collection case: same early return, no new path.

### F2 → #611, F5 → #612

Both filed rather than folded in, as directed.

- **#611** (`bug`, `area:responses`, `tier-0`) — `#[ResponseResource]` + non-2xx wrapper still yields the contradictory pair. States plainly that #584 closed the return-expression path only, quotes `ResourceClassLocator.php:66-77` as the short-circuit, and flags the real tension the fix must resolve: the attribute is the authoring escape hatch and *should* win, but it names *which resource*, not *at what status* — so the suggested shape is to consult the return expression for the **status alone** while leaving `resourceClass`/`isCollection` entirely to the attribute. Acceptance includes keeping the existing `attributeWins` fixture green.
- **#612** (`bug`, `area:responses`, `tier-0`) — the mute refusal. Grounds it in the package's own stated policy (Tier-1 inference must "degrade gracefully **and** log"), notes the reader fires a NOTICE for every shape it cannot match so this is the one silent hole in an otherwise noisy path, and lays out log-note vs. lint-finding with a recommendation to start with the note. Scoped to the whole refusal, not just non-2xx, since `204`/`205` take the same early return.

### Unchanged

Rename, guard placement, seam choice, the five round-1 tests, and the round-1 docs/CHANGELOG edits all stand. `docs/auto-derivation.md`, `docs/plugins.md`, and the CHANGELOG entry were each extended by a clause to cover `205`, so no doc now describes the predicate as 2xx-minus-204. `src/` diff for this round is one condition and one docblock.

Radiergummi:
🤖 **[surveyor]** Layer A pass — 0 regressions across 11 apps. The intended reduction has **no corpus exposure**, so it is invisible here; what the gate does prove is that the refactor around it changed nothing else.

**Provenance**
- baseline: `main` @ `d9a49fb218fa9a4184ce3d9fd823edd62a701c34`
- candidate: `3189c2c9f6e622d504706b16a1df318326fc2990` (own detached worktree at that SHA)
- `git diff d9a49fb2 3189c2c9 -- src/` touches 3 files; clean superset of the baseline.

**Removed items: none.** As briefed, I did not read the flat `responses` count as "nothing happened" — I diffed response *content* per `(METHOD path -> status)`, and then the whole document. Across all 22 specs (default + stack), with `operationId` and Livewire's per-run asset hash masked:

```
ALL SPECS STRUCTURALLY IDENTICAL
```

So: no 200 lost its resource envelope, no component schema disappeared, no path or operation dropped, no crash (`gen_exit=0`, all `booted`). `compSchemas` is unchanged on every app. Your PASS criterion — *every 200 that lost its envelope must be one wrapped in a literal non-2xx `response()->json(<resource>, <status>)`* — is satisfied vacuously: the set of such 200s is empty.

**Why the intended reduction does not show.** No app in the corpus writes `response()->json(<JsonResource>, <non-2xx>)`. Grepping all 11 apps for a non-2xx `response()->json(...)` whose data argument mentions a resource yields 3 hits, and all 3 are false positives — array literals whose *message string* contains the word "resource":

```
Coolify app/Http/Controllers/Api/DeployController.php:441   ['message' => 'No resources found.'], 404
Coolify app/Http/Controllers/Api/DeployController.php:493   ['message' => 'No resources found with this tag.'], 404
Coolify app/Http/Controllers/Api/ServicesController.php:2126 ['message' => 'Service resource not found.'], 404
```

The bogus-200 bug is real, but it is rare enough not to occur once in 3,900 corpus operations.

**What this gate is actually worth on this PR.** The abstention branch is untested by the corpus, but it is not the risky part of the change. The risk is the surrounding refactor: `successStatus` → `authoredStatus`, and moving the 2xx/204 filter out of `ReturnExpressionResourceReader` into `ResourceResponseResolver`. That rewires the path #583 shipped two commits ago, and the *positive* case is heavily exercised here — 80 non-200 2xx responses carrying content across 8 apps (InvoiceNinja 33, Bagisto 24, Pelican 8, AdvisingApp/Koel/Lychee 4 each, Vito 2, Coolify 1), on top of every 200-with-envelope in the corpus. All byte-identical. The relocation of the filter is clean on real code.

**Note for the merge order.** Both #607 and this PR produce a structurally identical corpus to `main` @ `d9a49fb2`. If both squash-merge with the trees reviewed here, the baseline stays valid and #608 can be gated against it without regenerating — worth checking `main`'s SHA against the manifest first, per the staleness rule.

Radiergummi:
🤖 **[lead]** rebased onto current `main` (`70308a89`) → **round-2 review** at `096cc2c9`. Assigned to `reviewer-1`.

The branch was 4 commits behind after #607, #605 and #609 merged. I did the rebase myself rather than routing it through the coder — it is mechanical — **but two of the three conflicts were semantic, not the usual CHANGELOG append**, so they are unreviewed content and I am flagging them for the reviewer rather than waving them through:

- **`CHANGELOG.md`** — three-way overlap. `main` had gained the #544 rewrite and the new #586 entry (both from #607); this branch had the older #544 text plus its own #584 entry and, deliberately, the #535 line **with the stale trailing clause removed** (round-1 finding). Resolution: `main`'s #544 and #586 lines verbatim, this branch's de-claused #535 line, this branch's round-2 #584 line (the one that names `204`/`205`).
- **`docs/auto-derivation.md`** — the "Response body" row was edited by *both* sides in the same cell: #607 added the value-object date wording, this branch rewrote the `response()->json()` status parenthetical. Resolution: `main`'s row with only the status parenthetical swapped for this branch's round-2 text (*"any status that may carry no content — a **non-2xx**, a `204`, or a `205` — makes the resource path yield entirely"*).
- `docs/plugins.md` auto-merged.

Gates re-run by me in the worktree after the rebase: `composer check` **exit 0** — Pint clean, PHPStan level 8 no errors, Pest **3730 passed / 9221 assertions**. Remote CI is running on `096cc2c9`; the review verdict must name that sha.

**Round-2 scope** (attempt 1 of 3 already spent on round 1): verify the three claimed fixes are really in the tree, verify #611/#612 faithfully capture F2/F5, and adversarially check my two conflict resolutions above.

**Disposition is already decided: this PR will be escalated, not auto-merged.** Round-1 review classified it human-gated and I agree — it renames the public readonly `ResourceTarget::$successStatus` to `$authoredStatus`, and `ResourceTarget` is returned by `Contracts\Routing\ResourceTargetLocator`, so this is an extension-surface change however cheap. Round 2 is therefore the last automated gate before Moritz reads it.

Radiergummi:
🤖 **[reviewer]** rebase verification at **`096cc2c9`**: **both semantic resolutions are correct.** I
checked them mechanically against `main` (`70308a89`) and against the pre-rebase tree (`d2ea1b00`)
rather than reading them for plausibility — the useful property here is provability, and both
resolutions turn out to be provable.

## 1. `CHANGELOG.md` — correct, all four entries verified individually

Against `main`, the whole file diff is **exactly two changes**: the #535 line loses its trailing
clause, and the #584 entry is appended. Nothing else moves. Per-entry, comparing the exact lines:

| Entry | Compared against | Result |
|---|---|---|
| #584 (round-2, `205` wording) | pre-rebase `d2ea1b00` | **byte-identical** |
| #535 (de-claused, round-1 work) | pre-rebase `d2ea1b00` | **byte-identical** |
| #544 (#607's rewrite) | `main` `70308a89` | **byte-identical** |
| #586 (#607's new entry) | `main` `70308a89` | **byte-identical** |

So nothing from #607 was dropped, and the branch's own two edits survived unaltered. The de-clausing
is the right half to have kept: `main`'s #535 line still carried *"and a non-2xx status still
documents the resource at 200 alongside the error response (unchanged, see #584)"*, which this PR
falsifies — taking `main`'s version there would have re-introduced the exact contradiction round 1
removed. The resolution takes `main` where `main` is ahead and the branch where the branch is ahead,
which is the only correct split.

Word-diff of the surviving change, for the record:

```
… a `204` is claimed by the inline-JSON reader as a body-less 204
  [-(unchanged); and a non-2xx status still documents the resource at 200
   alongside the error response (unchanged, see #584).-]{+(unchanged).+} (#535)
```

## 2. `docs/auto-derivation.md` — correct, and this one is provable in one command

The cell was contested by both PRs, so I diffed it in both directions.

- **vs `main`:** the only changes are inside the `response()->json()` parenthetical — `2xx` →
  `**body-bearing 2xx**`, `other than 204` removed, and the yield clause extended with
  *"…and any status that may carry no content — a **non-2xx**, a `204`, or a `205` — makes the
  resource path yield entirely…"*. #607's contributions to the same cell (`model or value object`,
  *"`->format(…)` on a date receiver typed from either source"*) do not appear in the diff at all,
  which means they are retained verbatim.
- **vs pre-rebase `d2ea1b00`:** the entire file diff is a **single** change —
  `-model, / +model or value object, `->format(…)` on a date receiver typed from either source,`.
  That is #607's addition and nothing else, which proves the branch's round-2 parenthetical survived
  byte-for-byte.

Two opposing diffs each showing only the *other* PR's contribution is as clean a demonstration of a
correct three-way merge as this gets.

## 3. Nothing else was disturbed

The five files the branch owns and `main` never touched are byte-identical pre- and post-rebase
(SHA-compared, not eyeballed): `ResourceResponseResolver.php`,
`ReturnExpressionResourceReader.php`, `ResourceTarget.php`, `ReturnExpressionResolutionTest.php`,
`docs/plugin-authoring.md`. `docs/plugins.md` was the third conflict and is a genuine
non-overlap — the branch edits the ApiResources `response()->json()` section (~`:249`), #607 edits
the value-object and `->format(…)` bullets (~`:108-128`); both sets are present in the rebased file.

So my round-1 plan-review and the subsequent code-review rounds still apply to unchanged content —
the rebase moved no reviewed code.

## Gates, re-run by me

`composer check` in the worktree at `096cc2c9`, **exit code 0**: Pint `{"tool":"pint","result":"passed"}`,
PHPStan level 8 `[OK] No errors`, Pest **3730 passed / 9221 assertions**. Worktree clean, HEAD at
`096cc2c9`. Matches the lead's numbers exactly.

**Verdict: rebase approved.** No findings.

Radiergummi:
🤖 **[reviewer]** review round 2 — **approved at `096cc2c9497d6e44e4c101c5f918a79f7662b64a`**, 2 non-blocking findings, both about *what the human reads*, not about the code.

**Sha caveat, stated plainly:** `096cc2c9` is the lead's **rebased local head**; the remote head at review time is still the pre-rebase `d2ea1b00` (`mergeStateStatus: DIRTY`). I reviewed the rebased tree in the worktree. If the force-push lands anything other than `096cc2c9`, this approval is void.

### 1. The three claimed round-1 fixes are present and correct — verified, not taken from the comment

- **F3 (205) — present, and the test is a genuine guard.** `ResourceResponseResolver.php:100-102`: `carriesResourceBody()` is `$status >= 200 && $status < 300 && $status !== 204 && $status !== 205`, and the docblock now says "a 2xx that is allowed content at all", citing RFC 9110 for both. There **is** a 205 test — `jsonWrappedResetContentStatus` → `/json-reset-content`, `ReturnExpressionResolutionTest.php:1000-1012`. I proved it is red-to-green by **mutation**: deleting `&& $status !== 205` from the predicate turns that test red with `successStatuses` returning `[205]` instead of `[200]`, 1 failed / 52 passed. Not a docblock-only fix.
- **F1 (plugin-authoring doc) — present, and it went further than asked.** `docs/plugin-authoring.md:290-298` puts the refusal **into the worked example** (not just prose), and `'response' => (string) ($target->authoredStatus ?? 200)` stops the example hardcoding `'200'` in the same breath as teaching that the status is readable. The prose at `:312-320` states the contract correctly: carries whatever was authored **including a non-2xx**, including `204`/`205`; not pre-filtered; decide ahead of registering a component or emitting a finding; `null` means unauthored / unreadable / disagreed. This is the doc the rename exists for, and it now makes the rename's own argument.
- **F4 (class-constant non-2xx) — present.** `jsonWrappedConstantForbiddenStatus()` → `Response::HTTP_FORBIDDEN`, asserted at `ReturnExpressionResolutionTest.php:986-996` with the same body-less-200 + no-component shape. The coder is right that it is green on the round-1 tree — `AstLiteralEvaluator` resolves the constant before the removed range filter ever applied — and labels it a shape-coverage guard rather than pretending it was red.

One deliberate inconsistency in F1 worth one sentence of the human's attention, **not** a finding: the example guard refuses only non-2xx, so an author copying it verbatim still emits `204`/`205` with a body. The coder documented the choice (a third-party convention may hold its own view of which statuses carry its envelope) and the prose covers it. Given F1's own premise — *authors copy code, not prose* — I'd have mirrored Core's carve-out and let the prose say "adjust to your convention", but this is a judgement call, defensible as made, and I am not re-opening it.

### 2. #611 and #612 exist and faithfully capture F2 and F5

Both open, both `bug` + `tier-0` + `area:responses`.

- **#611** states the finding accurately, including the part that matters most: *pre-existing, unchanged by #606, `main` behaves identically*. It quotes `ResourceClassLocator.php:66-77` as the short-circuit, and its scope note frames the real tension honestly (the attribute names *which resource*, not *at what status*), with the "read the status alone" shape as a suggestion rather than a decision. Acceptance keeps the `attributeWins` fixture green. No overstatement.
- **#612** likewise: grounds the mute degrade in the package's own Tier-1 "degrade gracefully **and** log" policy, notes correctly that the reader is noisy for every unmatched shape so this is the one silent hole, and scopes to the whole refusal (`204`/`205` take the same early return), not just non-2xx. Options are laid out with a recommendation rather than a mandate.

Neither issue misstates its finding. (Cosmetic only: #612 is titled `feat(…)` but labelled `bug`.)

### 3. The lead's conflict resolution — checked as adversarially as a coder's diff, and it is exact

Not "looks fine": I compared the branch's edits **word-for-word** before and after the rebase.

`git diff --word-diff d9a49fb2 d2ea1b00` (pre-rebase) and `git diff --word-diff 70308a89 HEAD` (post-rebase) produce the **identical set of word-level changes** on both conflicted files:

- `docs/auto-derivation.md` — exactly four: `2xx` → `**body-bearing 2xx**`, the deletion of `other than 204`, and the `instead),` → `instead, and any status that may carry no content — a **non-2xx**, a 204, or a 205 — makes the resource path yield entirely, …),` replacement. Everything #607 added to that row survives untouched in the merged line: `the wrapped @mixin/@extends model **or value object**` and `->format(…) on a date receiver **typed from either source**` are both present. The two edits are folded, not one overwriting the other.
- `CHANGELOG.md` — exactly two: the false trailing clause dropped from the #535 line, and the #584 entry appended. The #544 line carries **main's** post-#607 revision (`the receiver must already type as a date … (value-object receivers are covered by #586)`), not the branch's older copy, and the **#586 entry is present**. Both appear as pure context lines in the diff against `origin/main`, which is the proof nothing was clobbered. The #584 entry survived byte-intact, 205 wording included.
- `src/`, `tests/`, `docs/plugins.md`, `docs/plugin-authoring.md` — diffed pre- vs post-rebase and the patches are **byte-identical apart from blob hashes and one hunk offset** (`docs/plugins.md` `@@ -255` → `@@ -261`, from #607's earlier insertions). The rebase moved no code.

Nothing from #607, #544, or #586 was dropped. Resolution accepted.

### 4. Gates, re-run by me in the worktree at `096cc2c9`

- `vendor/bin/pint --test` → `{"tool":"pint","result":"passed"}`
- `composer lint` (PHPStan L8, 1435 files) → `[OK] No errors`
- `composer test` → **3730 passed, 9221 assertions**, 8.47s

(The delta from the coder's 3725/9193 is main's own new tests arriving via the rebase, not this branch.)

### Findings

**1. The PR body is now stale in three places, and for a human-gated PR the body is the first thing Moritz reads.** Everything below is correctly recorded in the coder's round-2 comment, but nobody reads a comment thread before the plan:

- Step 3 still specifies `carriesResourceBody()` = `$status >= 200 && $status < 300 && $status !== 204`. The shipped predicate also excludes `205`.
- The **Docs & changelog** section lists `docs/auto-derivation.md`, `docs/plugins.md`, `CHANGELOG.md` — it does not mention `docs/plugin-authoring.md`, which is in the diff and is the single largest doc change.
- The test plan enumerates cases 1-8; the two round-2 tests (205, class-constant non-2xx) are absent, as are the follow-ups #611/#612 (the "Explicit non-goals" section still reads as though the note question is unresolved).

Fix: a short **"Round 2 delta"** section appended to the body — five lines, pointing at the round-2 comment. No code change.

**2. An authored `205` now disappears from the document entirely, and nothing durable records it.** Measured (this is what the new test pins): `response()->json(<resource>, 205)` was documented as `205` *with* the resource envelope; it is now a body-less default `200` and **no `205` appears anywhere**. Unlike the `204` case, Core's `InlineJsonResponseResolver` does not claim it, so the status is lost rather than preserved body-less. I agree with the coder that this is the lesser error — the old output was invalid — but: `docs/plugins.md` says only that `205` "documents no resource", the CHANGELOG says only that the resource path yields, and **the coder explicitly offered to file the Core-side follow-up ("say the word and I will file it") and nobody answered** — there is no such issue open. Fix: file it. Worth noting for whoever does — **#597** (partial primary-response contributions: status-only resolvers) is the natural home, since "I know the status is 205 and nothing about the body" is precisely the facet-contribution shape that issue exists for. A `docs/plugins.md` clause saying the authored `205` is not preserved would also do, but the issue is the better record.

Neither finding blocks. Both are cheap, and both are the sort of thing that is much more expensive to notice after a human has already read the PR.

### Everything else

Conventions clean: strict types, no abbreviations, comments say *why* (the guard's comment states its own reason for being first), no issue numbers in code. Boundaries clean: the new `ResourceResponseResolver` import in the reader is docblock-only (`{@see}`) and stays inside `Plugins\ApiResources`; `Support/` and `Contracts/` are untouched. The reader's two stale claims ("keeps the Core-disabled configuration honest", "(204 aside)") are both retired. Line widths: the new guard line is 103 chars against a 100-char soft limit, in files where that is already the surrounding norm — not worth a round.

Test-gap sweep, so the boundary is on the record rather than merely unmentioned: a **collection** wrapped in a non-2xx and a **named** `status: 403` argument are both uncovered, and I deliberately do **not** ask for either. The guard sits ahead of the `isCollection` envelope `match`, so the collection path provably reaches the same `return null` (its 2xx twin is pinned at `:852`), and `authoredStatus()` reads a named argument identically to a positional one before any status value is inspected (2xx twin at `:874`). Neither would exercise a new branch.

**Classification unchanged: human-gated.** `ResourceTarget` is returned by `Contracts\Routing\ResourceTargetLocator`, so renaming its public readonly property is an extension-surface change. The rename remains the right call on the merits, and the CHANGELOG announces it explicitly, which is what a consumer needs.

Radiergummi:
🤖 **[reviewer]** finding 2 filed as **#614** — *fix(responses): an authored `205` on a resource `response()->json()` vanishes, where a `204` survives body-less* (`bug`, `tier-0`, `area:responses`).

Framed around the asymmetry rather than the yield: `204` and `205` are both contentless under RFC 9110 and both correctly refused a resource envelope, but `204` is **preserved body-less** while `205` **disappears** and the operation falls back to a default `200`. Both outcomes are pinned by adjacent tests in this PR (`ReturnExpressionResolutionTest.php:1073` and `:1000`).

Mechanism recorded: `ResourceResponseResolver::carriesResourceBody()` (`:100-103`) excludes both, but Core's `InlineJsonResponseResolver` (`:474-480`) carves out `$status === 204` **only**, and it returns before the body-readability check — so a resource-argument `205` falls through to `!$result->bodyReadable` and yields. `204` survives only because Core happens to claim it first.

The issue states plainly that this is a **consequence of this PR, not a pre-existing condition**: before #606 an authored `205` was documented as `205`, invalidly carrying an envelope. We traded an invalid `205` for a missing one — still the right trade, and the CHANGELOG says so, but a reader of the follow-up should not be left thinking it was always this way.

#597 (status-only contributions) is referenced as the proper mechanism — "the status is `205`, the body is unknown" is exactly the partial contribution it exists for — with a local widening of Core's carve-out to `{204, 205}` noted as a defensible interim. Filed standalone rather than folded into #597 so it stays independently closeable while #597 is `blocked`.

Nothing further from me on this PR. My approval stands at `096cc2c9`.

Radiergummi:
🤖 **[reviewer]** round-2 finding 2 filed as **#615** — *fix(responses): an authored `205` is lost entirely, while a `204` is preserved body-less* (`bug`, `tier-0`, `area:responses`).

The issue leads on the **asymmetry**, not the yield: both statuses are refused for the same RFC 9110 reason, but Core's `InlineJsonResponseResolver` claims a `204` and documents it body-less, while nothing claims a `205`, so it is silently downgraded to the builder's default `200`. Refusing the body is right in both cases; losing the status in one of them is the defect.

It states explicitly that this is a **consequence of this PR, not a pre-existing condition** — before #606 an authored `205` was documented as `205`, invalidly, with a body — and that the trade was deliberate and correct, so nobody picking it up concludes the behaviour was always this way or that #606 should be reverted. It names both candidate fixes: the narrow Core-side one (teach `InlineJsonResponseResolver` to claim a `205` as it already claims a `204`) and #597 (status-only facet contribution), referenced as the likely general mechanism but not a dependency. Acceptance requires the `205` test added here to be **updated rather than deleted**, since it currently pins the loss with a comment explaining it.

Round-2 finding 1 is closed by the lead's **Round 2 delta** section in the PR body. My approval at `096cc2c9497d6e44e4c101c5f918a79f7662b64a` stands — the force-push landed that exact sha, so nothing in this trail is stale. No further automated gate is owed; ready for the human.

Radiergummi:
🤖 **[lead]** all automated gates closed → **`agent:needs-human`**. Ready, green, and **not merged**. Requesting @Radiergummi's review.

**Why a human, not an auto-merge.** The PR renames the public readonly property `ResourceTarget::$successStatus` → `$authoredStatus`, and `ResourceTarget` is returned by `Contracts\Routing\ResourceTargetLocator` — that is the documented extension surface, so it is a public-symbol change and outside what this team merges unsupervised. Both reviewers reached that classification independently and neither wavered on it. The rename is right on the merits (the field now legitimately holds `403`), and the CHANGELOG announces it explicitly, which is what a consumer needs.

**State at `096cc2c9`** (rebased onto `main` @ `70308a89`; the earlier `d2ea1b00` is superseded):

- **CI: 11/11 green** on the whole matrix, verified by me on this sha — including `PHP 8.4 · Laravel 12 · swagger-php 5.8` and `Snapshot (locked toolchain)`.
- **Review: approved, round 2, 2 non-blocking findings — both now closed.** Not a reading-for-plausibility pass: the 205 fix was proved by mutation (deleting `&& $status !== 205` turns `jsonWrappedResetContentStatus` red), and my rebase conflict resolutions were checked word-for-word in both directions rather than by eye.
- **Follow-ups filed: #611, #612, #614.** Nothing was folded into this PR that did not belong to it.
- Gates re-run in-worktree post-rebase: Pint clean, PHPStan L8 `[OK] No errors` (1435 files), Pest **3730 passed / 9221 assertions**.

**The rebase carried two semantic conflicts, not the usual CHANGELOG append** — flagging because it is unreviewed-by-construction content that I authored. `main` had gained #607, which edited the *same* `docs/auto-derivation.md` table cell this branch rewrites, plus a three-way overlap in the CHANGELOG (#544 rewritten, #586 added, and this branch deliberately de-clausing the #535 line). Both were verified afterwards by word-diffing pre- and post-rebase against their respective bases and showing the identical change set: #607's contributions survive verbatim, this branch's survive byte-intact, and `src/`/`tests/` are byte-identical across the rebase apart from one hunk offset.

**The survey gate is NOT re-anchored to this tree, deliberately — please read this as a documented gap, not a passed gate.** Layer A passed at `3189c2c9` against `main` @ `d9a49fb2`, with 0 regressions across 11 apps and all 22 specs structurally identical; the surveyor also established the change has **zero corpus exposure** (no app in ~3,900 operations writes a non-2xx `response()->json(<resource>, …)`; the 3 grep hits are array literals whose *message* contains the word "resource"). Since then the head moved twice: the 205 exclusion, and the rebase over #607, which is itself generation-affecting. I chose not to re-run it because **any survey I run now is guaranteed stale before it can be used** — this PR cannot merge without you re-admitting it, and it will need rebasing again at that point, which would void a fresh run anyway. If you would rather see it re-anchored at merge time, say so and the team will run it against whatever `main` is then.

**One consequence worth your eye before you decide** (#614): an authored `205` no longer appears in the document at all. A `204` survives body-less only because Core's `InlineJsonResponseResolver` happens to claim it first; nothing claims a `205`, so it silently becomes a default `200`. The old output was invalid (a `205` carrying a body), so this is the better of two errors — but it is a loss this PR introduces, not a pre-existing one, and it is tracked rather than buried.

Worktree left in place at `.claude/worktrees/fix-584-fix-responses-a-non-2xx-response-json-re`. Dropping from the in-flight count.

Radiergummi:
🤖 **[lead]** trail correction: the round-2 finding was filed **twice**, as #614 and #615. **#615 is closed as a duplicate; #614 is the live issue.** My error — I sent the filing request to two agent identities after assuming the first had not received it. Nothing in #615 is missing from #614, and no other annotation on this PR is affected. Escalation state is unchanged: `agent:needs-human` at `096cc2c9`, 11/11 green, not merged.

Radiergummi:
🤖 **[lead]** `main` has moved to `ad66c7dd` (#616 merged). Status of this PR is **unchanged: escalated, ready, not merged**, awaiting @Radiergummi.

**No rebase performed, deliberately.** GitHub reports this PR still `MERGEABLE` / `CLEAN` against the new `main`, so there is nothing to fix — and rebasing would move the head and **void the sha-anchored reviewer approvals** on a PR that cannot merge until you re-admit it. Churning the sha to refresh a gate nobody can use would cost the approvals and buy nothing.

**What is owed at re-admission**, so it is on the record rather than assumed: this PR's green CI **predates** `ad66c7dd`. Before it merges it needs a rebase onto whatever `main` is at that point, a re-run of the full matrix on the rebased head, and re-confirmation of the approvals against the new sha. Expect the usual `CHANGELOG.md [Unreleased]` conflict; the resolution is to keep both entries.

The worktree is left in place.

Radiergummi:
🤖 **[lead]** **re-admitted by the maintainer.** @Radiergummi instructed "merge both 606 and 608" directly in the session. That is the human decision this PR was escalated for, so `agent:needs-human` is cleared and the merge freeze lifts.

Gates re-verified by me at this moment, not carried over from the escalation:

- `mergeable: MERGEABLE / CLEAN` against `main` at `ad66c7dd` — no rebase performed, so the sha-anchored approvals stand.
- All checks green on the current head.

Merging as-is.

Radiergummi:
🤖 **[lead]** squash-merged on the maintainer's direct instruction. Closes #584.

**Rebased before merging, deliberately** — the opposite of the call recorded earlier in this thread. That earlier reasoning ("don't churn the sha, it voids approvals a frozen PR can't use") was right *while the PR was frozen*. Once you re-admitted it the calculus inverted: the approvals were about to be spent, and the green CI behind them had been measured against `main` at `70308a89` — two merges stale, the more recent of which (#608) changed how nullable schemas are emitted, which is upstream of the resource schemas this PR touches. Merging on that green would have been merging an untested combination.

So: rebased onto `332dd300`, re-ran everything. Clean rebase, no `CHANGELOG` conflict (the append-at-end-of-section convention did its job). Local `composer check` **3866 passed / 9819 assertions**. Remote CI re-ran in full at `efa082eb` — genuine runs, not result-cache skips — all 9 checks green including swagger-php 5.8 and PHP 8.5 × Laravel 12/13.

Follow-ups from this PR's own review, all filed and now unblocked: #611 (the same contradictory pair survives when the action also carries `#[ResponseResource]` — the attribute short-circuits `ResourceClassLocator` before the return-expression read, so `authoredStatus` is never populated), #612 (the refusal is silent), #614 (an authored 205 vanishes where a 204 survives body-less).

PR #585, opened 2026-07-25
feat(responses): type `->format()` on a date attribute as string / date-time in resource fields

Radiergummi:
Closes #544

## Verified premise

Traced against `main` rather than taken from the issue.

`$this->resource->created_at->format(DATE_ATOM)` is a `MethodCall`, and every branch of
`ResourceToArrayReader::resolveValue()` declines it:

| step | why it declines |
|---|---|
| `resolveConditionalWrapper()` | `resourceMethodName()` requires the receiver to be `$this` or `$this->resource`; here the receiver is `$this->resource->created_at`, a `PropertyFetch` whose own `var` is a `PropertyFetch`, so `isResourceReceiver()` is false |
| `resolveNestedResource()` | not a `New_` / `StaticCall` |
| `resolveValueObjectProperty()` | requires a `PropertyFetch` value |
| `resolveModelProperty()` | `modelFieldName()` requires a `PropertyFetch` value |
| `SchemaDefinitionFromLiteral::fromValue()` | falls through to `AstLiteralEvaluator::evaluate()`, which throws `NonLiteralValueException` on a `MethodCall` → `[]` |

→ `unconstrained()`, an untyped `{}` property. Confirmed, and `docs/plugins.md:135` documents it
as the intended fallback today ("Anything else (method calls, ternaries, …) keeps its key with an
unconstrained schema").

**A correction to the issue's example.** The issue's headline idiom uses `DATE_ATOM`, a *global*
constant — and `AstLiteralEvaluator::evaluateConstant()` handles only `true`/`false`/`null`, throwing
for every other `ConstFetch` (`src/Support/MethodBody/AstLiteralEvaluator.php:66`). So `DATE_ATOM`
is **not** resolvable with today's primitives, while `DateTimeInterface::ATOM` is (class constants go
through `evaluateClassConstant()`, which already guards with `interface_exists`). The plan resolves
the argument through a narrow local step rather than widening the shared evaluator — see
"Resolving the format argument" below.

**The receiver typing is already free.** `created_at` / `updated_at` resolve through
`EloquentModelToSchema::timestampProperty()` to `{type: ['string','null'], format: 'date-time'}`, a
`datetime`/`immutable_datetime`/`timestamp`/`custom_datetime` cast maps to
`{type: 'string', format: 'date-time'}` and `date` to `format: 'date'`
(`EloquentModelToSchema.php:385-390`), and a `@property Carbon $x` docblock reaches
[...]

PR #607, opened 2026-07-27
feat(responses): type ->format() on a value-object date property in resource fields

### LINKED ISSUE #586: feat(responses): type `->format()` on a value-object date property in resource fields

Radiergummi:
Closes #586

## Dependency pre-check

- **Both gates are open.** The issue's hard gate — #581, "do not start this before it merges" — is
  merged as `57ef1aac`; the `PublicPropertyTypeReader::propertyFor()` contract this plan reads is
  the post-#581 one (three-argument signature with the optional `$leafClassSchema` callback; the two
  existing call sites in `ResourceToArrayReader` pass two arguments and are unaffected). The
  prerequisite #544 / PR #585 is merged as `13023dbc`.
- **No blocking collision.** Open PRs: #605 (CI workflows), #582 (`tests/Fixtures/*`,
  `tests/Unit/Lint/*`), drafts #571 / #526 / #210 (no diff), and #606 (this run's #584 plan).
  #606 touches `ResourceResponseResolver`, `ReturnExpressionResourceReader`, `ResourceTarget`, and
  the resolution test — **disjoint from `ResourceToArrayReader`**. The only shared files are
  `CHANGELOG.md` (`[Unreleased]`) and `docs/plugins.md`, and there in different sections
  (`~118-128` here vs `~249` there). Textual merge order at worst, not a contract collision.
- **Premise verified on current `main`.** `ResourceToArrayReader::resolveFormattedDate()` still
  bails at `src/Plugins/ApiResources/Support/ResourceToArrayReader.php:663` —
  `if ($fieldName === null || $modelClass === null) { return null; }` — and its docblock states the
  refusal outright ("Only the wrapped *model* answers that question here, never a value object").
  The behaviour is pinned by `ResourceDateFormatInferenceTest.php:158` and by the docblock on
  `tests/Fixtures/Resources/FormattedDateValueObjectResource.php`.

## Tier

**Tier 0/1 boundary, and it stays where #544 put it.** The AST match (`->format()` on a property
fetch) is the existing Tier-1 whitelist entry; the *evidence* is pure reflection. Nothing new is
parsed — only a second, already-existing reflection source is consulted for the receiver's type.

## Approach

`resolveFormattedDate()` currently inlines its receiver resolution: `modelFieldName($value->var)`
plus `modelToSchema->propertyFor($modelClass, …)`. Extract that into one helper that resolves a
receiver expression to its `OA\Property` using the **same precedence the field path already uses**,
then keep the date test, format mapping, nullability rule, and description carry-over unchanged.

New private helper in `ResourceToArrayReader`, e.g.
`receiverProperty(Expr $receiver, string $resourceClass, ?string $modelClass): ?OA\Property`:

1. **Shape (A)** — `$this-><typedProp>-><field>`: receiver is a `PropertyFetch` whose `var` is a
   `PropertyFetch` on `$this`. Resolve `propertyTypeClass($resourceClass, <typedProp>)` then
   `publicPropertyTypeReader->propertyFor($voClass, <field>)`. This mirrors
   `resolveValueObjectProperty()` (`:730`) exactly — reuse its receiver-matching, do not re-derive it.
2. **Shape (B) / model** — `$this-><field>` or `$this->resource-><field>`, via the existing
   `modelFieldName()`: try `modelToSchema->propertyFor($modelClass, $fieldName)` when
   `$modelClass !== null`, then fall back to `valueObjectProperty($resourceClass, $fieldName)`
   (`:771`), which is the `@mixin`/`@extends` value-object path. Identical precedence to
   `resolveModelProperty()` (`:613-616`): model first, value object second.

`resolveFormattedDate()` then becomes: match the `->format()` call as today → `receiverProperty(...)`
→ require `format ∈ {date, date-time}` → build the string property. The `$modelClass === null` bail
disappears; everything downstream (`RFC3339_FORMATS`, `DAY_FORMAT`, nullsafe handling,
`is_defined($property->description)` carry-over) is untouched and now works for either source.

### Two things this deliberately does *not* do

- **`isResourceReceiver()` is not widened.** The lead's cross-reference notes it accepts only
  `$this` / `$this->resource`, and that whatever wires the value object in "will have to widen that
  gate deliberately". It does not: shape (A) is a *separate* receiver branch keyed on a statically
  typed property of the **resource class**, which is how `resolveValueObjectProperty()` already
  works. The consequence is the desired one — a model **relation** hop
  (`$this->parent->published_at->format(…)`) stays refused, because `parent` is not a declared
  typed property on the resource and `propertyTypeClass()` returns null. Leaving
  `isResourceReceiver()` alone is what keeps that refusal intact.
- **Ordering hazard — reviewer-corrected: it is not an accident, and no test can pin the order.**
  `$this->resource-><field>` structurally also matches shape (A)'s pattern, so the plan originally
  called the model-first ordering load-bearing. It is not. `JsonResource::$resource` is declared
  `public $resource;` (untyped, `JsonResource.php:29`), and PHP **forbids** a subclass from adding a
  type to an inherited untyped property — `Type of C::$resource must be omitted to match the parent
  definition` is a fatal error, verified locally. So for any `class-string<JsonResource>`,
  `propertyTypeClass($resourceClass, 'resource')` returns null *by language rule*, not by accident,
  and shape (A) can never capture `$this->resource`. The two branches are disjoint for every input
  the caller can produce: shape (A) needs `$receiver->var` to be a `PropertyFetch`, and
  `modelFieldName()` only accepts `$this` / `$this->resource`, so the sole structural overlap is a
  resource property literally named `resource`.

  **Consequence for the coder:** keep the model branch first as cheap belt-and-braces, with a
  one-line comment saying *why* it is ordered that way, but do **not** claim a test pins it — both
  orderings produce identical output for every writable fixture, so any such test would pass either
  way and pin nothing. Case 8 below is therefore relabelled a behaviour regression pin (and it is an
  existing assertion, not new work). Do not restructure the existing `resolveValue()` ordering —
  out of scope.

## Tests

`tests/Feature/Plugins/ApiResources/ResourceDateFormatInferenceTest.php`, TDD order: flip the pinned
refusal at `:158` first (it must fail), then implement.

New fixtures:

- `tests/Fixtures/Values/BookingPeriodValue.php` — `public DateTimeImmutable $startsAt`,
  `public ?DateTimeImmutable $endsAt`, and `public MoneyValue $price` (the non-date guard).
- `tests/Fixtures/Values/MoneyValue.php` — a plain value object with a real
  `public function format(string $pattern): string`, so the guard fixture is PHPStan-clean
  (`phpstan.neon` analyses `tests/` at level 8, so a `->format()` on a non-formattable fixture type
  would fail the gate, not the assertion).
- `tests/Fixtures/Resources/FormattedDateShapeAResource.php` — a resource with a typed
  `private readonly BookingPeriodValue $period` and a `toArray()` exercising the cases below.
- A `shapeA()` action on the existing `DateFormatInferenceController` — note it is declared **inside
  the test file** (`ResourceDateFormatInferenceTest.php:21`), not under `tests/Fixtures/Controllers/`.
- Extend `tests/Fixtures/Values/IssuedDocumentValue.php` and `FormattedDateValueObjectResource` for
  the shape-(B) nullable/guard cases (additive; the existing keys stay).

Cases:

1. **Shape (A) happy path** — `$this->period->startsAt->format(DATE_ATOM)` →
   `{type: string, format: date-time}`.
2. **Shape (A), `'Y-m-d'`** → `{type: string, format: date}` — proves the mapping is shared, not
   re-implemented.
3. **Shape (B), the flipped pin** — rewrite `:158`: `$this->issuedAt->format(DATE_ATOM)` on the
   `@mixin IssuedDocumentValue` resource → `{type: string, format: date-time}`, while
   `issued_at_raw` keeps its existing bare-read schema. Update the fixture's docblock, which
   currently documents the refusal.
4. **Nullability, non-nullsafe** — `$this->period->endsAt->format(DATE_ATOM)` where `$endsAt` is
   `?DateTimeImmutable` → plain `string` (the null member dropped), key required.
5. **Nullability, nullsafe** — `$this->period->endsAt?->format(DATE_ATOM)` → `['string','null']`.
   Cases 4+5 prove the nullability rule reads the *call*, not the receiver's own nullability.
6. **Must-not-match: non-date value-object member** — `$this->period->price->format('%.2f')` →
   stays unconstrained (`[]`). Keep it, but **be accurate about what it discriminates**
   (reviewer-checked): `MoneyValue` is a plain object leaf, which `PublicPropertyTypeReader` refuses
   outright (it returns `null`), so this case fails only if the *null-property* guard is dropped —
   it never reaches the `DATE_LIKE_FORMATS` test. That test is already discriminated on the model
   path by the existing `formatted_string` key (`$this->summary->format(…)` on a `string` attribute,
   asserted `[]` at `:146`). There is no realistic value-object fixture that is both typeable by the
   reader *and* has a `format()` method, so do not invent one — record the split instead.
7. **Must-not-match: relation hop — already covered, do not add a fixture** (reviewer-checked).
   `FormattedDateResource` already carries `'formatted_relation' => $this->parent->published_at
   ->format(DATE_ATOM)`, asserted `[]` at `:149`. It is genuinely discriminating: `published_at`
   *is* a date attribute on the wrapped `DatedArticle` (the same resource types `iso_c` off it), so
   widening `isResourceReceiver()` would make this key resolve and the test would fail. With this
   change it also becomes the guard for shape (A) — `$this->parent->published_at` matches shape
   (A)'s structure, and only `propertyTypeClass(FormattedDateResource, 'parent') === null` refuses
   it. Re-run it; add nothing.
8. **Regression pin (not an ordering pin), already covered** — `atom_global`
   (`$this->resource->created_at->format(DATE_ATOM)` in `FormattedDateResource`) already asserts
   `{type: string, format: date-time}`. It proves shape (A) does not steal `$this->resource`, but it
   passes under either branch ordering, so it pins behaviour only. Add nothing.
9. **Non-literal / unrecognised format argument** on a shape-(A) receiver → plain `string`, no
   `format` (the degrade path is shared, but assert it once off the new source).
10. **Description carry-over → collapse to a single absence assertion** (reviewer-checked).
    `PublicPropertyTypeReader` contains no description handling at all, and the only descriptions
    `JsonSchemaFromType` emits are the `Unmapped …` placeholders (which the reader refuses) plus
    backed-enum case prose. A date-typed value-object property therefore never carries one. Assert
    the absence in one line and move on; do not add description synthesis, and do not build a
    docblock fixture for it.
11. **The modelless pin must stay green** — `keeps ->format() unconstrained when the resource wraps
    no model` (`:152`) is the direct pin for deleting the `$modelClass === null` bail. It should
    still pass untouched: `$this->created_at` is not shape (A) (its `var` is `$this`), and
    `valueObjectProperty()` returns null because the resource has no `@mixin`/`@extends` value
    object. If it needs editing, that is a signal the helper over-reached — stop and re-check.
12. **No regression on #544's model path** — the whole existing `region Recognised formats` and
    `region Untouched neighbours` blocks (`:57`–`:166`, `:170`–`:206`) must pass unchanged. The
    summarising generation-log note test at `:188` is expected to be **unaffected**: it filters on
    `FormattedDateResource` only, and that resource's three unconstrained keys
    (`formatted_string`, `formatted_unknown`, `formatted_relation`) all stay unconstrained under
    this change. If its expectation needs touching, treat that as a finding, not a fixup.

Gates: `composer test`, `vendor/bin/pint --test`, `composer lint`.

## Docs & changelog

- **`docs/plugins.md:127-128` is the sentence the lead flagged.** It currently ends: *"Receivers the
  model cannot type as a date, a value object's members among them, stay **unconstrained**."* That
  clause becomes false with this change. Rewrite to say the date evidence may come from the wrapped
  model **or** from a value object's statically-typed public property (both the `@mixin`/`@extends`
  shape and `$this->wrapped->field`), and that receivers neither source types as a date — including
  model relation hops — stay unconstrained. Updating this is in scope, not an afterthought.
- `docs/plugins.md:108-117` — extend the value-object bullet with the note that formatting such a
  property is now typed too, so the two bullets do not contradict each other.
- `docs/auto-derivation.md:17` — the response-body row mentions the `->format()` refinement; widen
  its wording to match.
- `CHANGELOG.md` `[Unreleased]` → **Added**: `->format(…)` on a date-typed **value object** property
  in a resource field now documents a `string` / `date-time` / `date`, the same as the wrapped-model
  path — both the `@mixin`/`@extends` value object and a `$this->wrapped->field` read. (#586)

Both bookkeeping items are named and in scope (reviewer-confirmed). PR #606 also edits
`CHANGELOG.md [Unreleased]` and `docs/plugins.md` (`~249`); a textual rebase conflict there is
expected and is resolved by **keeping both entries** — never by dropping one.

Also update the docblock on `resolveFormattedDate()` itself
(`ResourceToArrayReader.php:633-635`), which currently states the refusal outright ("Only the
wrapped *model* answers that question here, never a value object"). It becomes false with this
change and is the kind of stale comment the code review will reject.

## Controversy assessment

**Low.** No `src/Contracts/**` change, no stage-order change, no config key, no lint rule, no public
signature change — one production file (`src/Plugins/ApiResources/Support/ResourceToArrayReader.php`),
a refactor of one private method plus one new private helper, expected well under 60 `src/` lines.

Two items for the reviewer to confirm rather than gate on:

- The inference *widens* (more fields get a schema), so it can only be wrong by over-claiming. Case 6
  (`Money`-style `->format()`) and case 7 (relation hop) are the two guards that bound it; a reviewer
  should check those two are genuinely discriminating (fail when the guard is removed).
- `area:responses` → **survey gate applies**. Expected movement is an increase in typed properties
  across the corpus with no path/operation count change. A *drop* anywhere would be a real
  regression here, unlike #584.

**Reviewer verdict (plan-review):** classification **confirmed — auto-merge candidate**, measured
against the skill's four criteria: no `src/Contracts/**` and no stage-order change; no config key or
lint-rule severity; no public signature change (the new helper and the reworked
`resolveFormattedDate()` are both `private`, and `PublicPropertyTypeReader::propertyFor()` is only
*called*, with the same two arguments the two existing call sites already pass); `src/` diff one file
at roughly 40-60 changed lines against a ~150-line threshold that counts production code only.
`area:responses` still routes it through the survey gate before merge; auto-mergeable means no human
gate, not no gate.

Discriminating-guard check, answered: case 7 is genuinely discriminating and **already exists**; case
6 is discriminating but for the null-property guard rather than the date-format guard (the latter is
covered on the model path by `formatted_string`). Details in the case list above.


Radiergummi:

Dependency pre-check clean. The issue’s hard gate — #581, “do not start this before it merges” — is merged as `57ef1aac`, so the `PublicPropertyTypeReader::propertyFor()` contract the plan reads is the post-#581 one (the two existing call sites in `ResourceToArrayReader` pass two arguments and are unaffected by the new optional `$leafClassSchema` parameter). Prerequisite #544/PR #585 is merged as `13023dbc`. No open PR touches `ResourceToArrayReader`; the only files shared with the sibling plan in #606 are `CHANGELOG.md` and `docs/plugins.md`, in different sections.

Premise re-verified on current `main` and still pinned: `ResourceToArrayReader.php:663` bails on `$modelClass === null`, its docblock states the refusal outright, and `ResourceDateFormatInferenceTest.php:158` asserts it.

Carried both items from the lead’s cross-reference into the plan:

- The `docs/plugins.md:127-128` sentence (“a value object’s members among them, stay **unconstrained**”) is scoped in as a required edit, not an afterthought.
- `isResourceReceiver()` (`:353-363`) is **not** widened. Shape (A) goes in as a separate receiver branch keyed on a statically typed property of the resource class, exactly as `resolveValueObjectProperty()` already does. That is what keeps the model relation hop (`$this->parent->published_at->format(…)`) refused, and it is pinned by its own test. The `Money`-style must-not-match guard is planned as case 6, with a fixture carrying a real `format()` method so it stays PHPStan-clean (`phpstan.neon` analyses `tests/` at level 8).

One hazard recorded for the coder: `$this->resource-><field>` structurally also matches shape (A)’s pattern, and is harmless today only because `JsonResource::$resource` has no usable declared type. The new helper must keep the model branch ahead of shape (A) so correctness does not rest on that accident; there is a test for it.

Radiergummi:
🤖 **[lead]** plan posted → plan-review. Assigned to `reviewer-2`.

Lead notes for the reviewer:
- Planner reports this as **non-controversial** (one production file, ~60 `src/` lines, no `Contracts/**`, no public signature change) — i.e. an auto-merge candidate. Please confirm or refute that classification explicitly, since it decides whether a human ever sees this diff.
- The plan hinges on a claimed precedence order (model branch ahead of the new value-object shape (A) branch) because `$this->resource-><field>` structurally matches shape (A) and is harmless today only by accident. Verify that ordering is actually load-bearing in the proposed helper and that the pinning test really pins it.
- Survey expectation differs from #606: here the intended movement is an **increase** in typed properties with **no** operation-count change, so any drop is a genuine regression, not an intended removal.

Radiergummi:
🤖 **[reviewer]** plan-review: **approved with amendments** — plan edited in place. Tier and scope are right; three claims in it were wrong and are now corrected, and three planned test cases turned out to be existing assertions (net less work, not more).

**1. Merge classification — CONFIRMED, auto-merge candidate.** Checked against the four criteria rather than the summary: no `src/Contracts/**` and no registry/stage-order change; no config key and no lint-rule default severity; no public signature change (the new helper and the reworked `resolveFormattedDate()` are both `private`; `PublicPropertyTypeReader::propertyFor()` is only *called*, with the same two arguments its two existing call sites in `ResourceToArrayReader` already pass, so the post-#581 optional third parameter is untouched); `src/` diff is one file at roughly 40-60 changed lines against the ~150-line production-only threshold. `area:responses` still routes it through the survey gate — auto-mergeable means no *human* gate, not no gate.

**2. Precedence hazard — REFUTED as stated; the ordering is not load-bearing and no test can pin it.** `JsonResource::$resource` is `public $resource;` — untyped (`vendor/…/JsonResource.php:29`) — and PHP *forbids* a subclass from adding a type to an inherited untyped property: `Type of C::$resource must be omitted to match the parent definition`, a fatal error I verified locally. So `propertyTypeClass($resourceClass, 'resource')` returns null by language rule, not by accident, for every `class-string<JsonResource>`. Shape (A) can therefore never capture `$this->resource`, and the two branches are disjoint for every input the caller can produce (shape (A) requires `$receiver->var` to be a `PropertyFetch`; `modelFieldName()` accepts only `$this` / `$this->resource`, so the sole structural overlap is a resource property literally named `resource`).

Consequence: the plan's proposed **case 8 "ordering pin" would have passed under either ordering** — exactly the pin-nothing test the lead warned about. Amended to keep the model-first ordering as cheap belt-and-braces *with a one-line comment saying why*, and relabelled case 8 a behaviour regression pin. It is also **already an existing assertion** — `atom_global` in `FormattedDateResource` is literally `$this->resource->created_at->format(DATE_ATOM)`. Nothing to add.

**3. Scope discipline — refusal kept, but two guard claims needed correcting.**
- `isResourceReceiver()` stays un-widened; the plan's reasoning holds.
- **Case 7 (relation hop) already exists** and is genuinely discriminating: `FormattedDateResource` carries `'formatted_relation' => $this->parent->published_at->format(DATE_ATOM)`, asserted `[]` at `ResourceDateFormatInferenceTest.php:149`, and `published_at` *is* a date attribute on the wrapped `DatedArticle` (the same resource types `iso_c` off it) — so widening the gate would make the key resolve and the test fail. It now doubles as the shape-(A) guard, since `$this->parent->published_at` matches shape (A) structurally and only `propertyTypeClass(…, 'parent') === null` refuses it. No new fixture.
- **Case 6 (`Money` guard) discriminates a different guard than the plan claims.** `MoneyValue` is a plain object leaf, which `PublicPropertyTypeReader` refuses outright (returns `null`), so the case never reaches the `DATE_LIKE_FORMATS` test — it fails only if the *null-property* guard is dropped. The date-evidence test is already discriminated on the model path by `formatted_string` (`:146`). Keep case 6, but record the split honestly; there is no realistic value-object fixture that is both typeable by the reader *and* carries a `format()` method, so don't invent one.

**4. Bookkeeping — both named, confirmed.** `docs/plugins.md:127-128` and the `CHANGELOG.md [Unreleased]` **Added** entry are both scoped in, plus `docs/plugins.md:108-117` and `docs/auto-derivation.md:17`. Recorded in the body that the #606 overlap on `CHANGELOG.md` / `docs/plugins.md` (`~249`) is a textual rebase conflict resolved by **keeping both entries**.

**Additional amendments:**
- Added the stale docblock on `resolveFormattedDate()` (`ResourceToArrayReader.php:633-635` — *"Only the wrapped model answers that question here, never a value object"*) to scope; it becomes false with this change.
- **Case 10 (description carry-over) collapsed to a one-line absence assertion.** Verified: `PublicPropertyTypeReader` has no description handling at all, and the only descriptions `JsonSchemaFromType` emits are the `Unmapped …` placeholders (which the reader refuses) plus backed-enum case prose. A date-typed value-object property never carries one. Don't build a docblock fixture for it.
- **Added case 11: the modelless pin (`:152`) must stay green untouched** — it is the direct pin for deleting the `$modelClass === null` bail, and it should still pass on its own (`$this->created_at` is not shape (A); the resource has no wrapped value object). If it needs editing, the helper over-reached.
- Corrected the regression ranges in case 12 (`:57`–`:166`, `:170`–`:206`) and flagged that the generation-log note test at `:188` is expected **unaffected** — it filters on `FormattedDateResource`, whose three unconstrained keys all stay unconstrained. If that expectation needs touching, it is a finding, not a fixup.
- Noted `DateFormatInferenceController` lives inside the test file (`:21`), not `tests/Fixtures/Controllers/`.

Ready for coding.

Radiergummi:
🤖 **[lead]** plan-review approved (plan amended in place) → coding. Assigned to `coder-2`. Classification confirmed by reviewer-2: **auto-merge candidate**, subject to the `area:responses` survey gate.

Recorded because it changes the work, not just the paperwork: reviewer-2 **refuted** the precedence hazard the plan was built around. `JsonResource::$resource` is `public $resource;` (untyped), and PHP forbids a subclass adding a type to an inherited untyped property, so `propertyTypeClass($resourceClass, 'resource')` returns null **by language rule, not by accident** — shape (A) can never capture `$this->resource`, the branches are disjoint for every producible input, and the planned "ordering pin" would have passed under either ordering. Model-first ordering stays as belt-and-braces with a one-line why comment; case 8 is relabelled a behaviour regression pin and already exists as `atom_global`.

Net: three planned test cases (7, 8, most of 10) collapse into re-running existing assertions. Coder implements the plan **as amended**, not as originally written.

Radiergummi:
🤖 **[coder]** implemented; local gates green (`vendor/bin/pint --test` passed, PHPStan level 8 `[OK] No errors`, Pest **3722 passed / 9199 assertions**). Pushed `4293e191`. Watching remote CI, then ready for review.

**Implemented as amended**, not as originally planned. What landed:

- `ResourceToArrayReader::resolveFormattedDate()` now resolves its receiver through one new private helper, `receiverProperty(Expr, string, ?string)`, which tries the wrapped model (`modelToSchema->propertyFor`) then the `@mixin`/`@extends` value object (`valueObjectProperty()`) for a `modelFieldName()` receiver, and otherwise falls through to shape (A). The `$modelClass === null` bail is gone. Everything downstream (`RFC3339_FORMATS`, `DAY_FORMAT`, nullsafe rule, description carry-over) is untouched and now serves either source.
- Shape (A) receiver-matching was **extracted, not re-derived**: `resolveValueObjectProperty()` now delegates to a new `wrappedValueObjectProperty(Expr, string): ?OA\Property`, which both callers share. No behaviour change on the bare-read path.
- **No ordering pin test.** Model-first ordering kept as belt-and-braces with a one-line comment saying why (shape (A) needs a typed property, and the inherited `$resource` cannot be typed by a subclass).
- `isResourceReceiver()` untouched; the relation hop stays refused.
- Stale `resolveFormattedDate()` docblock ("Only the wrapped *model* answers that question here") rewritten.

**Tests** — TDD: the pin was flipped first and 4 cases failed before implementation, all green after.

| Plan case | Where |
| --- | --- |
| 1, 2, 10 (description absence) | new `types a ->format() on a typed value-object property of the resource` |
| 3 (flipped pin) | the old `:158` refusal, rewritten as `types a ->format() on a wrapped value object date property as a date-time string` |
| 4, 5 | new `reads nullability off the call, not the value-object property` |
| 6 | new `keeps ->format() on an untypeable value-object member unconstrained` |
| 9 | new `degrades a non-literal format argument on a value-object receiver to a plain string` |
| 7, 8, 11, 12 | existing assertions, re-run unchanged |

Confirming the two things you asked to be checked rather than fixed up: **the modelless pin needed no edit** (`$this->created_at` is not shape (A), and the resource has no wrapped value object), and **the generation-log note test at `:188` is unaffected** — no change to it, still green.

New fixtures: `tests/Fixtures/Values/BookingPeriodValue.php`, `tests/Fixtures/Values/MoneyValue.php`, `tests/Fixtures/Resources/FormattedDateShapeAResource.php`, plus a `shapeA()` action on the in-test-file `DateFormatInferenceController`.

Docs + changelog: `docs/plugins.md` (both the value-object bullet and the now-false `->format(…)` sentence), `docs/auto-derivation.md`, and a `CHANGELOG.md [Unreleased]` line appended at the end of the section.

Two small deviations, neither behavioural:

1. **`docs/auto-derivation.md:17` did not actually mention the `->format()` refinement** (the plan says it does). The response-body row only said `` `$this->field` typed from the wrapped `@mixin`/`@extends` model ``. I widened that clause to name the value object *and* the `->format(…)` case, which is the honest version of the intended edit.
2. **Did not extend `IssuedDocumentValue` / `FormattedDateValueObjectResource` with nullable/guard keys.** The plan's fixture paragraph asks for them, but its own case list puts every nullability and guard case on shape (A), and both shapes converge on the same downstream code immediately after `receiverProperty()` returns. Duplicating them on shape (B) would add fixtures that cannot fail independently. Shape (B) keeps its one case (the flipped pin).

3. **`price` (case 6) already passed before the implementation**, as reviewer-2 predicted — it guards the null-property gate, not the date-format gate. Recorded here so the guard is not mistaken for a date-evidence discriminator; the date-evidence guard on the shape-(A) path is unchanged from the model path's `formatted_string`.

Radiergummi:
🤖 **[lead]** implementation pushed (`4293e191`), local gates green → code-review. PR marked ready.

Assigned to **both** `reviewer-1` and `reviewer-2`, reviewing **independently**. This PR is classified auto-mergeable, so two independent approvals with no open concerns are what stands between this diff and `main` with no human ever seeing it. Review on your own merits; do not read or defer to each other's verdict.

Remote CI is still in flight — that is my gate, not yours. Start reading the diff now; I will not merge on a local-gates-green claim.

Radiergummi:
🤖 **[lead]** remote CI **fully green** on `4293e191` — whole matrix, not a local claim: PHP 8.4/8.5 × Laravel 12/13, the dedicated **swagger-php 5.8** job, snapshot (locked toolchain), quality, coverage, tooling. That clears the CI half of the merge gate.

Remaining gates for this auto-merge candidate: two independent reviewer approvals, and survey Layer A. The surveyor is already building the `main` baseline in parallel, so the gate should not be the critical path.

Reviewers: the coder documented **three deviations from the agreed plan**. Deviation from plan is a first-class review criterion, so please each rule on them explicitly rather than letting the coder's own justification stand as the record:
1. `docs/auto-derivation.md:17` did not actually mention the `->format()` refinement the plan claimed it did; the coder widened the response-body row's `$this->field` clause instead.
2. Nullable/guard keys were **not** added to the shape-(B) fixtures, on the argument that the plan's own case list puts every such case on shape (A) and the two shapes converge on identical downstream code immediately after `receiverProperty()` returns — so the duplicates could not fail independently. This is the one worth real scrutiny: if the convergence claim is wrong, a nullable value-object path just shipped untested.
3. Case 6 (`price`) passed before the implementation, as predicted — it guards the null-property gate, not the date-evidence gate. Kept with the split recorded.

Radiergummi:
🤖 **[reviewer]** review round 1: **approved, no blocking concerns** (`4293e191`).

Gates re-run by me in the coder's worktree, not taken on trust: full suite **3722 passed / 9199 assertions**, `vendor/bin/pint --test` passed, `composer lint` (PHPStan level 8, 1434 files) **no errors**.

### Against what plan-review actually required

I rewrote this plan, so I checked the diff against my amendments rather than the original:

- **No fake ordering pin re-added.** The coder did not invent a case-8-style test, and `atom_global` / `formatted_relation` are untouched. I verified the underlying claim by **mutation**: I swapped the two branches inside `receiverProperty()` and re-ran the whole `tests/Feature/Plugins/ApiResources/` directory — **117 passed, 503 assertions**, i.e. the ordering genuinely is not observable, exactly as found in plan-review. The comment at `ResourceToArrayReader.php:703-705` says so honestly ("The branches cannot overlap anyway: shape (A) needs a typed property, and the inherited `$resource` one cannot be typed by a subclass") instead of claiming a pin that does not exist. This is the right call.
- **Money guard named honestly.** `it('keeps ->format() on an untypeable value-object member unconstrained')` plus the comment *"MoneyValue is a plain object leaf the reader refuses outright, so no date evidence exists"* — attributes it to the null-property guard, not to `DATE_LIKE_FORMATS`. No misattribution reinstated.
- **Stale docblock updated** at `ResourceToArrayReader.php:633-635`; the two fixture docblocks that documented the refusal (`IssuedDocumentValue`, `FormattedDateValueObjectResource`) were flipped too, which I had not explicitly demanded and which is correct.
- **`:152` modelless pin and the `:188` generation-log note test are untouched** and green, as required. The log note's key list did not move, as predicted.
- **Description case collapsed to one line** (`->not->toHaveKey('description')` inside the shape-(A) happy path) rather than a fixture — as required. The model-path carry-over stays pinned at `:91`.
- Nothing I kept was skipped: shape (A) happy path, `'Y-m-d'` refinement, the flipped shape-(B) pin, both nullability cases, the non-literal degrade, and the guard are all present.

### Correctness / adversarial probing

The refactor is clean: `resolveValueObjectProperty()` now delegates to the extracted `wrappedValueObjectProperty()` rather than duplicating the receiver match, so there is one matcher, not two. I hunted for the input that over-claims and wrote three throwaway probes (since deleted, worktree left clean — `git status` verified):

1. `$this->resource-><voField>->format(DATE_ATOM)` on a `@mixin`-value-object resource — newly reachable now the `$modelClass === null` bail is gone → correctly `{string, date-time}` (it *is* the value object, so this is right).
2. A resource declaring a typed **Eloquent Model** property, `$this->article->published_at->format(DATE_ATOM)` → correctly `[]`. `propertyTypeClass()` succeeds here and returns a Model class; only `PublicPropertyTypeReader`'s refusal of non-public/dynamic attributes stops it.
3. Nullsafe intermediate hop `$this->period?->startsAt->format(DATE_ATOM)` → `[]`, no crash (a `NullsafePropertyFetch` var fails shape (A)'s match), consistent with the bare-read path.

Also walked by inspection: `$this->format()`, `$this->resource->format()`, and a three-hop `$this->a->b->c->format()` all refuse; a nullable value-object member keeps its `format` through the reader so cases 4/5 exercise the real nullability rule.

### Two non-blocking observations (do not bounce the PR for these)

1. **Uncovered boundary — shape (A) over a typed *Model* property.** Probe 2 above is the closest realistic over-claim vector (resources that take the model in the constructor are common), and nothing in the suite pins it. One key on the existing shape-(A) fixture would: a `private readonly DatedArticle $article` plus `'article_published_at' => $this->article->published_at->format(DATE_ATOM)` → `[]`. Worth noting that the refusal is a *boundary*, not a bug — the bare read `$this->article->published_at` is equally unconstrained today, so it is consistent — but it means shape (A) over a Model-typed property is now one line away and may deserve its own issue rather than an accidental future regression.
2. **Docs nit.** The `->format(…)` bullet still leads with `$this->field->format(…)` while shape (A) is `$this->wrapped->field->format(…)`; the bullet body states both sources explicitly, so a reader gets there, but the lead-in example is narrower than the rule. `docs/plugins.md:118` correctly cross-links the two bullets. `CHANGELOG.md` entry is in `[Unreleased]`, `docs/auto-derivation.md:17` widened, no `docs/README.md` index change needed.

Conventions clean: strict types, no abbreviations, comments say *why*, no issue numbers in code, `@phpstan-ignore method.nonObject` on the deliberately-nullable fixture call is justified inline. The two new lines over the 100-char soft limit (the `receiverProperty()` signature and its ternary) match ~20 pre-existing lines in the same file, including the identical statement at `:613` — consistent, not worth churn.

Classification from plan-review holds: one production file, ~60 changed `src/` lines, no `Contracts/**`, no config key, no public signature change → **auto-merge candidate**, pending the `area:responses` survey gate.

Radiergummi:
🤖 **[reviewer]** review round 1 — 2 findings (both small; I expect to approve on round 2).

Reviewed `4293e191` independently: read the diff, re-derived the control flow against the
unmodified file, and re-ran the gates myself in the worktree rather than taking them on report —
**3722 passed / 9199 assertions**, `pint --test` clean (1434 files), PHPStan level 8 `[OK] No errors`.

## What I verified, not assumed

- **The language-rule claim holds.** `JsonResource::$resource` is `public $resource;` (untyped,
  `JsonResource.php:29`), and PHP fatals on a subclass typing it: *"Type of C::$resource must be
  omitted to match the parent definition"* — reproduced locally. So
  `propertyTypeClass($resourceClass, 'resource')` can never resolve. No untested branch shipped.
  The shipped code is in fact **safer than that argument**: `receiverProperty()` early-returns
  inside `if ($fieldName !== null)`, so shape (A) is unreachable for `$this->resource-><field>` by
  control flow, independently of the language rule. And `wrappedValueObjectProperty()`'s guard
  requires `$value->var instanceof PropertyFetch` with `isThisVariable($value->var->var)`, so a bare
  `$this->resource->format(…)` (var is the `$this` variable) is refused there too. Triple-guarded;
  the model-first ordering plus its why-comment is the right amount of belt-and-braces, and
  reviewer-2 was right that no test could pin it.
- **`isResourceReceiver()` is untouched** (`ResourceToArrayReader.php:353`), and `formatted_relation`
  still asserts `[]`. The relation-hop refusal rests on `propertyTypeClass()` returning null for a
  relation — which is never a declared typed property — so it is structural, not incidental.
- **The three must-stay-green pins are untouched.** The modelless pin (now `:158`) and the
  generation-log note test (now `:235`, still asserting exactly `formatted_string` /
  `formatted_unknown` / `formatted_relation`) appear only as diff context; the sole rewritten test
  is the value-object refusal at old `:158`, which is the intended flip.
- **The new fallback is precedence-faithful.** `receiverProperty()` reproduces
  `resolveModelProperty()`'s `model ?? valueObjectProperty()` exactly (`:613-616`), and the fallback
  is safe on Model-mixin resources because `WrappedModelLocator::locateValueObject()` filters
  `!is_a($className, Model::class)` (`:87`). `formatted_unknown` (`$this->missing_column->format(…)`)
  is the live proof: model returns null, the value-object fallback returns null, key stays `[]`.
- Branch coverage of the new helper: model/date ✓ (`atom_global`), model/non-date ✓
  (`formatted_string`), model-miss → VO-miss ✓ (`formatted_unknown`), no model → shape (B) ✓
  (`issued_at` and the modelless pin), shape (A) untypeable receiver class ✓
  (`formatted_relation`), shape (A) null property ✓ (`price`), shape (A) date ✓ (`starts_at`),
  nullability both ways ✓, non-literal format ✓. One branch is uncovered — finding 1.

## Findings

**1. `ResourceToArrayReader.php:661-666` — the date-evidence guard is never exercised on the new
value-object source, and the plan's reason for skipping it is wrong.**

`in_array($receiverProperty->format, self::DATE_LIKE_FORMATS)` is what stops the value-object path
from typing *every* `->format()` receiver as a string. Today it is exercised only through the model
(`formatted_string`, `:146`). The `price` / `MoneyValue` case does **not** reach it — the diff's own
test comment says so ("a plain object leaf the reader refuses outright"), so it pins the
null-property guard instead. On a change whose stated failure mode is over-claiming, the guard that
bounds the new source deserves its own case.

The plan concluded no such fixture is writable ("no realistic value-object fixture that is both
typeable by the reader *and* has a `format()` method"). That is not correct:
`PublicPropertyTypeReader::isModelledLeaf()` returns `true` for a `BackedEnumType`
(`PublicPropertyTypeReader.php:272-274`), pinned by *"refers a backed enum property to its
component"* (`PublicPropertyTypeReaderTest.php:161`) — and a backed enum may declare methods. So a
backed enum with its own `format()` is typeable by the reader, callable in the fixture, and needs no
`@phpstan-ignore`.

Concrete test to add:

- `tests/Fixtures/Enums/…` — `enum CurrencyCode: string { case Eur = 'EUR'; public function
  format(string $pattern): string { return sprintf($pattern, $this->value); } }`
- `BookingPeriodValue` — add `public CurrencyCode $currency`.
- `FormattedDateShapeAResource::toArray()` — add
  `'currency' => $this->period->currency->format('%s')`.
- Assert `$properties['currency']` is `[]`.

The reader returns a non-null property for it (a component `$ref`, `format` undefined), so the case
lands *on* the `DATE_LIKE_FORMATS` check and fails if that check is ever weakened or split per
source. That is the discrimination `price` cannot provide.

**2. `CHANGELOG.md:161` — the `[Unreleased]` block now contradicts itself.**

The #544 entry still ends: *"The attribute must already type as a date through the wrapped model's
metadata; every other receiver, **including a value object's**, stays unconstrained as before."* The
new #586 entry two lines down says the opposite. "This supersedes the value-object refusal noted for
#544" helps a reader who gets that far, but the released notes still assert both, in one release.
Amend the #544 clause — drop *"including a value object's"* and the *"every other receiver"*
framing, or point it at #586 — the same treatment the stale sentences in `docs/plugins.md` and
`docs/auto-derivation.md` correctly got in this diff.

## Everything else checks out

Conventions: strict types, no abbreviations, comments say *why* (the ordering comment and the
`@phpstan-ignore method.nonObject` justification are both the right kind), no issue numbers in code,
within the width limit. Boundaries: one plugin file, nothing in `Support/` or `Contracts/`, no public
signature change — the new `receiverProperty()` and `wrappedValueObjectProperty()` are private, and
the `resolveValueObjectProperty()` extraction is behaviour-preserving. Docs: the two `plugins.md`
bullets and the `auto-derivation.md` row are all corrected and mutually consistent; I grepped `docs/`
for any other surviving statement of the refusal and found none. The stale
`resolveFormattedDate()` docblock — the thing most likely to be forgotten — was rewritten.

Sending to coder-2. Neither finding touches shipped behaviour, so round 2 should be short.

Radiergummi:
🤖 **[lead]** review round 1 split: `reviewer-2` **approved**, `reviewer-1` raised **2 findings** → back to `coder-2` for round 2 (attempt **1 of 3**).

Both findings are real and neither touches shipped behaviour, so round 2 should be short. Note that reviewer-1 independently **verified the language-rule claim** the implementation rests on and found the shipped code is in fact triple-guarded (control-flow early return in `receiverProperty()`, plus `wrappedValueObjectProperty()`'s `PropertyFetch` guard), so no untested branch shipped — reviewer-2's mutation test (swapping the branches → 117 passed) independently confirms the ordering is unobservable.

**In parallel:** I am starting the survey gate now against `4293e191` rather than waiting for round 2, because both findings are confined to a test fixture and a CHANGELOG line — neither is shipped code, so the corpus result stays valid. **If round 2 touches anything under `src/`, say so immediately** and I will re-survey; a survey against a superseded `src/` tree is worthless.

Radiergummi:
🤖 **[coder]** addressed round 2: both findings fixed. Pushed `12fdf1a5`; **remote CI green** across the matrix. Local gates: Pint clean, PHPStan level 8 `[OK] No errors`, Pest **3723 passed / 9201 assertions**.

**`src/` is untouched** — the diff is one new fixture, two fixture edits, one test, one docs line, and two CHANGELOG lines. The survey running against `4293e191` is unaffected.

### Finding 1 — date-evidence guard now pinned on the value-object source

You were right that the plan's "no writable fixture" claim was false. Implemented as you specified:

- `tests/Fixtures/Enums/CurrencyCode.php` — `enum CurrencyCode: string` with `Eur`/`Usd` and a `format(string $pattern): string`.
- `BookingPeriodValue` — `public CurrencyCode $currency`.
- `FormattedDateShapeAResource` — `'currency' => $this->period->currency->format('%s')`.
- New test `keeps ->format() on a typed but non-date value-object member unconstrained`, asserting `[]`. No `@phpstan-ignore` needed, as predicted.

**Verified discriminating by mutation, not by inspection.** I replaced the guard with `if ($receiverProperty === null) {` and re-ran the file:

| Test | Guard present | Guard removed |
| --- | --- | --- |
| `keeps ->format() on a typed but non-date value-object member unconstrained` (`currency`) | pass | **fail** |
| `keeps ->format() on a non-date model attribute unconstrained` (`formatted_string`) | pass | **fail** |
| `names only the still-unconstrained keys in the generation-log note` | pass | **fail** |
| `keeps ->format() on an untypeable value-object member unconstrained` (`price`) | pass | **pass** |

3 failed / 16 passed under the mutation. That is exactly the split you described: `currency` reaches `DATE_LIKE_FORMATS` and dies with it, `price` does not. `src/` was restored from a backup and `git diff --stat src/` is empty against the pushed head — the mutation never left the working tree.

### Finding 2 — CHANGELOG contradiction resolved

`CHANGELOG.md:161` (#544) now reads: *"The receiver must already type as a date through the wrapped model's metadata (value-object receivers are covered by #586); a receiver nothing types as a date stays unconstrained."* The false *"every other receiver, including a value object's"* clause is gone. I also dropped the now-redundant *"This supersedes the value-object refusal noted for #544"* sentence from the #586 entry, since the entry it pointed at no longer states the refusal — the two lines now simply agree.

### reviewer-2's two non-blocking observations

1. **Filed as #610**, not folded in — "ApiResources: shape (A) over a typed Eloquent Model property is unconstrained" (`spec`, `area:responses`, `tier-0`). Records that it is a consistent boundary today (the bare read is equally unconstrained), that `propertyTypeClass()` succeeds while `PublicPropertyTypeReader` refuses, that the tier-0 fix is dispatching on `is_a($class, Model::class)`, and that closing it must move the bare read and the formatted read together without widening `isResourceReceiver()`.
2. **Docs nit taken.** The `->format(…)` bullet lead-in is now `` `$this->field->format(…)` / `$this->wrapped->field->format(…)` `` so the example matches the rule the body states. One line, rewrapped.

Ready for re-review.

Radiergummi:
🤖 **[surveyor]** Layer A pass — 0 regressions across 11 apps. Caveat: the corpus shows **no lift either**, because it contains no instance of the idiom.

**Provenance**
- baseline: `main` @ `d9a49fb218fa9a4184ce3d9fd823edd62a701c34`
- candidate: `4293e19166f507b02a7b3d3fa5a42f1b9c038772` (detached worktree cut at that SHA, not the coder's live worktree)
- `git diff d9a49fb2 4293e191 -- src/` touches exactly one file, so the candidate is a clean superset of the baseline.

**Result: every watched metric flat on all 11 apps.** No path/operation-count change, no crash (`gen_exit=0`, all `booted`), no drop anywhere. Under the standard rule this is a pass.

Because a property-level type refinement cannot move any count metric, I did not stop at the metric table — I diffed the full generated documents (default + stack-enabled, 22 specs) with `operationId` and Livewire's per-run `/livewire-<hash>/` asset prefix masked:

```
ALL SPECS STRUCTURALLY IDENTICAL
```

**So the expected increase in typed properties did not materialise, and that is not a defect in the change.** The corpus simply never exercises it: across all 11 apps there is not one `->format(...)` call inside a `JsonResource::toArray()`, let alone the value-object-typed-property variant this PR adds. (The single grep hit, `Pelican/app/Filament/Admin/Resources/Webhooks/WebhookResource.php:111`, is a Filament resource with `now()->format(...)` in a closure — not a `JsonResource`, not a field read.)

**What this gate does and does not establish:** it establishes that #607 causes no collateral damage on 11 real applications — the risk in touching `ResourceToArrayReader` is regression in the *existing* shape-(A) and model-read paths, and that risk is measured and clean. It does **not** provide positive evidence of efficacy; that rests entirely on the PR's own unit tests, which are the only place the new idiom appears.

**Incidental, worth recording for future gates:** AureusERP, InvoiceNinja and SpeedtestTracker produce byte-differing specs on every run regardless of library changes — InvoiceNinja from the random `operationId` suffix, the other two additionally from Livewire's asset-route prefix, which is derived from the `APP_KEY` the bootstrap regenerates per run. Both are environment nondeterminism, not behaviour. The gate tooling now masks both, so a future surveyor should not read them as a diff.

Radiergummi:
🤖 **[lead]** round 2 pushed (`12fdf1a5`) → re-review by `reviewer-1` (the finder). CI **independently verified green** across the whole matrix, including the swagger-php 5.8 job.

**I verified the `src/`-untouched claim myself rather than accepting it**, because the running survey depends on it: `git diff 4293e191 12fdf1a5 -- src/` is **empty**. Round 2 touches exactly six files, all non-shipped: `CHANGELOG.md`, `docs/plugins.md`, the inference test, and three fixtures. **The corpus run against `4293e191` therefore stands for the final head** — no restart, no lost hour.

Worth recording: coder-2 disclosed unprompted that it *temporarily* mutated `src/` in its working tree to check the new test discriminates, then restored it before committing. That is the right way to verify a test and the right way to report it. And it did not take the discrimination on trust — mutating the guard to `if ($receiverProperty === null)` yields 3 failed / 16 passed, with `currency` and `formatted_string` failing while `price` still passes. That is exactly the split reviewer-1 predicted, and it confirms the new test pins the `DATE_LIKE_FORMATS` guard that nothing previously exercised on the value-object path.

reviewer-2's non-blocking observation 1 was filed as **#610** rather than folded in, per the self-feed rule.

Radiergummi:
🤖 **[reviewer]** review round 2: **approved, no concerns.** Both of my round-1 findings are closed,
and I verified the discrimination claim myself rather than accepting the report.

## Finding 1 — closed, and the mutation split reproduces exactly

I re-ran the mutation independently in the worktree: replaced the guard with
`if ($receiverProperty === null) { return null; }` (dropping the `DATE_LIKE_FORMATS` clause), ran the
file, then restored `src/` and confirmed the tree clean and green again.

| | with the guard | guard removed |
|---|---|---|
| `currency` (backed enum, reader-typed, non-date) | `[]` ✅ | **fails** — now typed `string` |
| `formatted_string` (model path, `summary` typed non-date) | `[]` ✅ | **fails** |
| `price` (`MoneyValue`, reader refuses outright) | `[]` ✅ | still passes |
| generation-log note | 3 keys ✅ | **fails** (`formatted_string` no longer unconstrained) |

`3 failed / 16 passed`, matching coder-2's report line for line. That is the split I described: the
new `currency` case is the *only* thing pinning the date-evidence guard on the value-object source,
and `price` demonstrably does not pin it. The fixture is clean — a backed enum with its own
`format()`, no `@phpstan-ignore`, and `CurrencyCode`'s docblock states precisely why it exists.

Given the surveyor's framing — 22 structurally identical corpus documents because **no corpus app
has a `->format(…)` inside a `JsonResource::toArray()`**, so Layer A establishes safety and not
efficacy — this test now carries more weight than the survey does. Worth saying plainly: it is
load-bearing evidence, not a box ticked.

## Finding 2 — closed; the two entries now agree, and the sweep is clean

`CHANGELOG.md:157` (#544) no longer asserts the refusal: *"The receiver must already type as a date
through the wrapped model's metadata (value-object receivers are covered by #586); a receiver
nothing types as a date stays unconstrained."* Dropping the now-redundant "supersedes" sentence from
the #586 entry was the right call — with the #544 line no longer stating a refusal, there is nothing
left to supersede, and the two entries now read as **one feature described in two scopes** rather
than one correcting the other.

Per the lead's note about this block having produced the same failure mode twice in one run, I swept
the whole `[Unreleased]` section for any other clause this change falsifies: every mention of *value
object*, *`format(`*, and *unconstrained*. The only other candidate is the #411 entry at `:57`
(value-object bare reads), which scopes itself to reading properties and makes no claim about
`->format()`. Nothing else is stale.

## Round-2 diff and gates

Six files, none shipped: CHANGELOG, `docs/plugins.md`, one test, three fixtures. `git diff` over
`src/` between `4293e191` and `12fdf1a5` is empty, so my round-1 correctness review of the
production change stands unchanged. Gates re-run by me at `12fdf1a5`: **3723 passed / 9201
assertions**, `pint --test` clean, PHPStan level 8 `[OK] No errors`.

One unprompted improvement worth noting: the `docs/plugins.md` bullet now leads with
`$this->field->format(…)` / `$this->wrapped->field->format(…)`, so the shape-(A) form appears in the
example rather than only in the prose. Small, and it closes a gap I had not flagged.

Approved.

Radiergummi:
🤖 **[lead]** all gates green + non-controversial → **squash-merged**, closes #586, worktree removed.

**Gate record** (each verified by me, not taken from a report):
- **CI:** whole matrix green at `12fdf1a5` — PHP 8.4/8.5 × Laravel 12/13, the dedicated **swagger-php 5.8** job, Snapshot (locked toolchain), quality, coverage, tooling.
- **Non-controversial, measured:** `src/` churn **85 changed lines in one file** (`ResourceToArrayReader.php`, 60+/25−), well under the ~150 threshold; **zero** files under `src/Contracts/`; **zero** config keys; no stage-order change; no public signature change (both new methods private).
- **Two independent approvals:** `reviewer-2` at round 1, `reviewer-1` at round 2 — and neither rubber-stamped. reviewer-2 proved the branch ordering unobservable by mutation (swapping the branches → 117 passed), and reviewer-1 independently reproduced the finding-1 mutation split line for line rather than accepting the coder's report.
- **Survey Layer A:** pass, 0 regressions across 11 apps, at `4293e191`; `git diff` over `src/` between that and the merged head is **empty**, so the result stands for the final head.

**One honest caveat on the record, because "Layer A pass" overstates it.** The surveyor did not stop at flat counts — it reasoned that a property-level type refinement cannot move a count metric, diffed all 22 generated documents structurally, and found them **identical**. The reason is corpus exposure, not a defect: **no app in the corpus contains a single `->format(...)` inside a `JsonResource::toArray()`**. So the survey establishes **safety, not efficacy** — it confirms this change broke nothing on 11 real apps, not that the feature works. Efficacy rests entirely on the unit tests, which makes reviewer-1's `currency` fixture (the backed enum that is the *only* thing pinning the `DATE_LIKE_FORMATS` guard on the value-object path) load-bearing evidence rather than a box ticked.

Follow-up filed rather than folded in: **#610** — shape (A) over a typed Eloquent Model property is unconstrained.

Issue #619, opened 2026-07-27
chore(autonomous-team): parallel reviewers share one worktree and one scratchpad, risking cross-contamination


Radiergummi:
Operational defect in the autonomous-team harness, observed during the 2026-07-27 run on PR #616. **The lead's fault, not the reviewers'** — I dispatched two reviewers to the same worktree path for a concurrent independent review, and nothing in the skill told me not to.

Reported by `reviewer-2`, which caught it because it checksum-guarded its own experiments.

## What happened

Two reviewers were asked to review PR #616 **independently and concurrently**, so the two approvals required for an unsupervised merge would be two genuine reads. Both were pointed at:

- the same worktree, `.claude/worktrees/feat-610-shape-a-typed-model-property`
- the same scratchpad directory

Observed consequences, all inside one review window:

- a **staged revert of `src/Plugins/ApiResources/Support/ResourceToArrayReader.php` appeared in the index and then vanished** — one reviewer's mutation experiment, visible to the other as a real working-tree state
- three test files acquired uncommitted edits mid-gate-run
- one reviewer's `Write` of its review draft was **refused because the other's draft already existed at the same path**, exposing part of the other's in-progress conclusions before its own were finalised

## Why it matters

Two things, and the second is worse than the first.

1. **Correctness of the review.** Reviewers on this project routinely verify by *mutation* — edit `src/`, run the suite, count reds, revert. Two agents doing that simultaneously in one tree can each draw conclusions from the other's half-applied experiment. In this instance no finding was contaminated (mutations were checksum-guarded and the decisive one re-run under guards, and the reviewer quoted CI rather than its local suite numbers precisely because of this), but that was the reviewer's own discipline compensating for a harness defect.

2. **Independence of the review.** The whole point of two approvals is two reads, not one read and a ratification. A shared scratchpad leaks one reviewer's conclusions to the other. The reviewer disclosed the collision rather than letting the agreement look cleaner than it was, which is the right behaviour — but the harness should not put it in that position.

3. **A live data-loss risk.** A `git commit` from either agent during the window when a whole-file revert was staged would have landed a commit reverting the entire feature, on a branch whose CI had just gone green.

## Fix

[...]

PR #608, opened 2026-07-27
refactor(core): one owner for schema nullability — retire the FieldDescriptor keyword shuffle

Radiergummi:
Closes #591

> ⚠️ **Expect a human gate.** `area:core` + `area:requests`, and the `src/` churn is roughly
> 250 changed lines (a ~150-line deletion plus a ~60-line owner). Byte-identical spec output is an
> acceptance criterion and the crux of the risk, but see the qualifier under *Snapshot table*:
> the snapshots exercise the widen path, not the new split. See *Controversy assessment* at the bottom.

## Dependency pre-check

- **Not gated.** The issue states "None. Can start immediately", and that holds: no open PR touches
  `src/Support/Extraction/FieldDescriptor.php`, `src/Support/Generator/NullableSchema.php`,
  `src/Support/Generator/JsonSchemaFromType.php`, or `src/functions.php`.
- **One soft overlap, not blocking.** Draft #571 (`feat(requests): let RuleDocumentation express an
  object property allowlist (additionalProperties/propertyNames)`) is `area:requests` and would land
  in `FieldDescriptor`'s neighbourhood. It carries a **single empty commit and no diff**, last
  touched 2026-07-24, so there is nothing to collide with; whichever resumes second rebases.
  Other open PRs — #605 (CI workflows), #582 (test fixtures), #606 / #607 (this run's response-side
  plans), #526 / #210 (empty drafts) — are disjoint.
- **Premise verified on current `main`**, with one correction to the issue's diagnosis (below).
- Note: this issue carries the `spec` label. Per the standing decision recorded for this repo,
  `spec` marks OpenAPI-spec work, not planning-only, and is not an exclusion.

## Correction to the issue's diagnosis

The problem is real and reproduces exactly, but the *third* site is misidentified, and there is a
concrete latent bug the issue does not name. Both change what the fix has to cover, so recording
them rather than planning the issue's wording verbatim.

1. **`JsonSchemaFromType` is not an independent expression of nullability.** Its only nullability
   line is `src/Support/Generator/JsonSchemaFromType.php:95`, which already delegates:
   `NullableSchema::wrap($this->fromType($type->getWrappedType(), …))`. Its `UnionType` branch
   (`:98`) never sees a bare `null` member in practice — `symfony/type-info` normalises those into
   `NullableType`, which is matched first (the class docblock at `:56` calls that ordering
   load-bearing), and `fromBuiltinType()` has no `TypeIdentifier::NULL` arm at all, so a bare null
[...]

PR #621, opened 2026-07-28
fix(lint): response.success-empty-body must not fire on a scalar 2xx body

### LINKED ISSUE #587: response.success-empty-body false-positives on a scalar 2xx body (binary download, plain string return)
## Problem

`response.success-empty-body` reports "a 2xx response declares no body schema" for a 2xx that
*does* declare a body, whenever that body is a **scalar** schema. The rule decides "has a body" by
looking for `fields` or `schemaRef` on the response node, and a scalar schema has neither: no
properties to become fields, no `$ref` to a component.

Two reproductions:

- A plain `: string` controller return → `200` with
  `{content: {application/json: {schema: {type: string}}}}` → finding fires.
- A `: StreamedResponse` / `: BinaryFileResponse` return → `200` with
  `{content: {application/octet-stream: {schema: {type: string, format: binary}}}}` → finding fires.

## Pre-existing, not caused by #550

Confirmed pre-existing: the plain `: string` reproduction was probed on untouched code (no
`feat/550-…` changes present) and fires there too. PR #567 only makes it easier to hit, because the
new `StreamedResponse` / `BinaryFileResponse` read emits a scalar binary body where the generator
previously emitted an empty synthetic `200`. Filed separately at the lead's instruction rather than
folded into that PR.

## Expected

A 2xx whose content declares *any* schema is not an empty body. The check needs a third signal
beside `fields` / `schemaRef`: a media-type schema that is present but scalar (a `type` with no
properties), which is a legitimate body, not a missing one.

Worth deciding at the same time whether `application/octet-stream` should be exempt outright, since
a binary download can never have a field-shaped body.

## Notes

- Rule: `src/Lint/Rules/ResponseSuccessEmptyBody.php` (severity Underspecified).
- The sibling signal is right there: `ResponseNode` already carries the raw `OA\Response`, so the
  media-type schema is reachable without changing the tree.

Radiergummi:
Closes #587

## Scope (fast-path — `area:lint` only, no generation change, no survey)

`response.success-empty-body` decides "has a body" from `fields` / `schemaRef` on the
`ResponseNode`. A scalar media-type schema has neither, so a legitimate `: string` or
`: StreamedResponse` / `: BinaryFileResponse` 2xx body is reported as missing.

**Fix:** add a third signal — a 2xx whose content declares *any* schema is not an empty body.
`ResponseNode` already carries the raw `OA\Response`, so the media-type schema is reachable
without touching the tree builder.

## Lead's scope decision

The issue asks whether `application/octet-stream` should additionally be exempted outright.
**No — do not special-case it.** The general rule ("any declared schema counts as a body")
already covers the binary download, and a media-type allowlist is a second, redundant
mechanism that would drift. Keep one signal.

## Out of scope

- Any change to what the generator emits (this is a lint-rule read only).
- Reworking `ResponseNode` / `SpecTreeBuilder`.

## Tests

- 2xx with a scalar `type: string` body → no finding.
- 2xx with `application/octet-stream` `{type: string, format: binary}` → no finding.
- 2xx with genuinely no `content` → finding still fires (the rule must keep its purpose).
- 2xx with content but an empty / absent schema → finding still fires.
- Existing object-bodied cases (`fields` / `schemaRef`) unchanged.

## Bookkeeping

`CHANGELOG.md [Unreleased]` entry. No new rule, so no `docs/linting.md` catalog row; check
whether the rule's catalog description needs a wording touch.


Radiergummi:
🤖 **[lead]** fast-path (`area:lint` only, no generation change → no survey): skipping planning/plan-review. Scope + test list is the PR body; the octet-stream question is decided there. → coding.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed `3da3a72`. Ready for review.

**What changed** — `ResponseSuccessEmptyBody` gains a third "has a body" signal after the
existing `fields` / `schemaRef` checks: `declaresBodySchema($response->raw)` walks the raw
response's media types and accepts any `OA\Schema` that declares a shape. Shape is a small named
constant, `SHAPE_KEYWORDS = ['type', 'allOf', 'anyOf', 'oneOf']`.

**Deviation from the plan (scope, narrower):** the plan said "any declared schema counts". I
started with a broad keyword list (`ref`, `format`, `enum`, `items`, `properties`,
`additionalProperties`, …) and trimmed it, because most of those entries are unreachable at that
point in the rule: a `$ref` media-type schema already sets `schemaRef`, and a schema with
`properties` already yields `fields` — both short-circuit before the new check ever runs. What
remains is what the generator can actually emit here with neither fields nor a ref: a scalar or
container `type`, and a composition (`oneOf` from a union return). No behaviour is lost relative
to the plan; the dead entries are.

Per the lead's decision, **no** `application/octet-stream` exemption — the binary download passes
on the general signal (`{type: string, format: binary}` declares a `type`).

**Tests** (all five bullets from the plan, plus two):
- unit: scalar `type: string` body → no finding; `application/octet-stream`
  `{type: string, format: binary}` → no finding; `oneOf`/`anyOf`/`allOf` composition → no finding
  (dataset).
- unit, rule keeps its purpose: media type with **no** schema → fires; media-type schema
  declaring **nothing** (`OA\Schema` with everything `UNDEFINED`) → fires; response with **no
  `content`** at all → fires; a 204 carrying a scalar body → still skipped (the bodiless-code
  exemption is not weakened by the new signal).
- integration (`LintRunnerTest`, real generator + `LintRunner`, new
  `tests/Fixtures/Lint/ScalarResponseController.php`): a routed `: string` action and a routed
  `: StreamedResponse` action emit no `response.success-empty-body`. Both were verified to
  **fail before** the fix, so this pins the end-to-end reproduction from the issue, not just the
  unit shape.
- existing `fields` / `schemaRef` cases unchanged.

**Bookkeeping:** `CHANGELOG.md [Unreleased]` entry appended at end of section. No
`docs/linting.md` change — no new rule, and the catalog row's description ("declares no body
schema") stays accurate, since a scalar body *does* declare one.

**Note for the reviewer:** `Generator::UNDEFINED` is a non-empty string, so `is_defined()` alone
is not a truthiness check. The `??` in `declaresShape()` folds a null/unset property to
`UNDEFINED` before `is_defined()` sees it, which is why there is no separate null guard —
PHPStan confirmed the null branch is unreachable there.

**CI:** green on `3da3a72` — all 5 matrix jobs incl. the swagger-php 5.8 job, snapshot, coverage,
tooling, and quality (Pint + PHPStan level 8).

Radiergummi:
🤖 **[coder]** rebased onto `ecf5af8c` (#608 + #606); pushed `7d8ad9b4`. CI green on that sha.

`CHANGELOG.md [Unreleased]` was the only conflict; resolved by keeping both entries.

## The #608 interaction: real, reachable, and already covered — but I proved it rather than assuming

`NullableSchema` has two output shapes, and **both reach the top level of a response body**:

- **widen** — a nullable scalar becomes `{type: ['string','null']}`. A defined `type`, so the new
  predicate accepts it.
- **split** — anything else becomes `oneOf: [<inner>, {type: 'null'}]`, with the constraint
  keywords (including `ref`) **moved onto the inner branch**. The outer schema is left with no
  `type` and no `ref`.

The split case is the one the lead flagged, and it is worse than "no `type`": because the `$ref`
moves inward, `SchemaAccessor::extractRef()` on the outer schema returns null, so `schemaRef` is
null on the node too — the response looks empty to *every* pre-existing signal.

Probed against the generator on this branch:

| return type | emitted 2xx body |
|---|---|
| `?string` | `{type: [string, null]}` |
| `?ScalarOnlyData` | `oneOf: [{$ref: '…/ScalarOnlyData'}, {type: 'null'}]` |

Then I swapped the **pre-fix** rule back in and re-ran: both emit
`response.success-empty-body`. So this is not a hypothetical — the false positive did survive in
a new costume, and `oneOf` in `SHAPE_KEYWORDS` is what kills it. No speculative branch was added;
the existing predicate already covered it, and there is now a test proving it.

**Added:** unit cases for both shapes, and both as end-to-end `LintRunnerTest` cases. I re-ran the
whole file with the pre-fix rule in place: **all 11 positive cases fail**, including all four
integration cases — none of them is vacuous.

## Two incidental things worth flagging to the reviewer

1. **A component finding is not route-scoped.** Adding the `?ScalarOnlyData` route to
   `beforeEach` broke the *pre-existing* `--no-suppress` test: the Data return contributes a
   component schema, `schema.description-missing` fires on it, and that finding is document-wide,
   so it leaked past the `lint-runner/suppressed*` glob and turned an expected exit 0 into a 1.
   Fixed by registering the four scalar routes inside their own test instead. Not a product bug
   (a component genuinely has no route), but worth knowing before someone adds a component-typed
   fixture to a shared `beforeEach`.
2. **`return null;` is not a nullable return** as far as PHPStan is concerned — it narrows the
   declared type away and fails `return.unusedType` at level 8. The fixture's nullable payloads
   come from promoted constructor properties so the declared nullability survives analysis.

**CI on `7d8ad9b4`:** all 5 matrix jobs incl. swagger-php 5.8, Snapshot, Coverage, Tooling,
quality (Pint + PHPStan level 8). Local: 3881 tests pass, PHPStan `[OK] No errors`.

Radiergummi:
🤖 **[lead]** rebased onto `ecf5af8c` and re-greened at `7d8ad9b4` (verified by me: `merge-base --is-ancestor origin/main <branch>` true, 11/11 checks on that sha) → code review. Assigned to `reviewer-2`.

Radiergummi:
🤖 **[reviewer]** review round 1: **approved at `7d8ad9b4bd54f6f8d4a4330822976014992300d3`, no concerns.**

Independent read (I did not write this diff). Everything below was exercised, not accepted from the
PR text. Work done in **my own detached worktree** at the reviewed sha with its own `vendor/` —
`coder-1`'s `fix-587-…` worktree was not touched (#619).

## The three load-bearing claims

**1. The recorded deviation — the reasoning holds for every reachable case, but the stated rationale
is slightly stronger than what's true.**

- **`properties` is genuinely unreachable.** `SpecTreeBuilder:697` calls `buildFields`, whose
  `collectComposedProperties` returns the schema's properties, so `fields` is non-empty and the rule
  short-circuits at `ResponseSuccessEmptyBody:62` before the new check. Belt and braces: the
  generator never emits `properties` without a sibling `type` — I walked all ten
  `examples/*/openapi.yaml` for a `properties`-bearing schema with no `type` and no `$ref`, and
  found **0**. Sound.
- **`ref` is unreachable only for *local component-schema* refs.** `SpecTreeBuilder:692` sets
  `schemaRef` via `SchemaAccessor::extractRef` → `ComponentReference::name`, which returns `null` for
  anything that is not `#/components/schemas/…`. So a media-type schema whose only keyword is a
  **non-component** `$ref` (`#/components/responses/Foo`) or an **external** one
  (`other.yaml#/User`) leaves `schemaRef` null, yields no fields, reaches `declaresBodySchema` with
  no shape keyword, and still fires. Probed both directly: `findings=1` each.

  This is **pre-existing** (the pre-fix rule fired on them too) and the generator never emits such a
  ref — it only arrives via harvested `@OA` annotations or `openapi.overrides`. So: no regression, no
  change required. I'm recording it only so the rationale "a `$ref` media-type schema already sets
  `schemaRef`" isn't inherited unqualified by the next maintainer.

**2. The #608 interaction — confirmed independently, and it is real rather than lucky-but-untested.**

Generated the four fixture routes and dumped the emitted 2xx bodies:

| route | emitted `200` body |
|---|---|
| `text` | `{type: string}` |
| `download` | `application/octet-stream` `{type: string, format: binary}` |
| `nullableText` | `{type: [string, null]}` |
| `nullableData` | `oneOf: [{$ref: '#/components/schemas/ScalarOnlyData'}, {type: null}]` |

Then instrumented the rule to print the node-level signals through the **real** `SpecTreeBuilder`:

```
SIGNALS /lint-runner/scalar-text          200 -> fields=0 schemaRef=NULL
SIGNALS /lint-runner/scalar-download      200 -> fields=0 schemaRef=NULL
SIGNALS /lint-runner/scalar-nullable-text 200 -> fields=0 schemaRef=NULL
SIGNALS /lint-runner/scalar-nullable-data 200 -> fields=0 schemaRef=NULL
```

`nullable-data` included. The structural reason checks out: `extractRef` on the outer schema returns
null because `split()` moved the `ref` inward, and `buildFields` unwraps the composition
(`SpecTreeBuilder:497`) to that `$ref` branch, whose `collectComposedProperties` dereferences refs
only inside `allOf` — a bare `$ref` branch contributes no properties, so `fields` stays `[]`. The
response really was invisible to every pre-existing signal, and `oneOf` in `SHAPE_KEYWORDS` is the
only thing that catches it. "Covered by luck" is the honest framing of how it got covered, but it
*is* covered, and now pinned by a test.

**3. Vacuity — the claim is exact.** Restored the pre-fix behaviour (deleted the
`declaresBodySchema` early return) and re-ran both files: **7 unit + 4 integration = 11 failures**,
matching the coder's count precisely. The four `still flags …` negative cases stayed green under the
mutant, as they must — they pin the rule's purpose, which the mutant doesn't touch.

## Extra mutation: is every keyword load-bearing?

Ablated `SHAPE_KEYWORDS` one entry at a time:

| dropped | unit failures | integration failures |
|---|---|---|
| `type` | 3 | 3 |
| `oneOf` | 2 | 1 |
| `allOf` | 1 | 0 |
| `anyOf` | 1 | 0 |

No dead entry — the trim was complete but not over-eager. `allOf`/`anyOf` are unit-only because the
generator emits `oneOf` for nullability and `type` for scalars; since this rule also reads authored,
harvested and `overrides`-injected specs, accepting the standard composition keywords is correct
rather than speculative.

## Negative direction

Probed the predicate directly rather than trusting the tests. Still fires, correctly, on: no
`content` at all (the `Generator::UNDEFINED` sentinel), `content: []`, a media type with no schema, a
media-type schema declaring nothing, and a media type whose `schema` is the UNDEFINED sentinel. The
`is_array($content)` guard is exactly the right shape here — `UNDEFINED` is a non-empty *string*, so
a truthiness or `!== null` test would have sailed straight past it. `declaresShape` folds unset
properties to `UNDEFINED` through `??` before `is_defined()` sees them, which is consistent with
`SchemaAccessor:82` and `SpecTreeBuilder:676`, as is the `@phpstan-ignore nullCoalesce.property`.
The dynamic `$schema->{$keyword}` access matches the existing idiom at
`SchemaAccessor::compositionBranches:152`.

## Gates, re-run in my own worktree

3881 tests passed (9834 assertions), PHPStan level 8 `[OK] No errors`, Pint `passed`. No new line in
the rule exceeds 100 chars (the two over-long lines, `:31` and `:85`, are pre-existing).

## Non-blocking notes

1. **`is_defined($schema->{$keyword} ?? UNDEFINED)` accepts an empty array.** `oneOf: []` and
   `type: []` both count as "declares a shape" (verified: `findings=0` for each). Unreachable from
   the generator — `NullableSchema` always emits two branches — and the failure direction is
   under-reporting a level-2 hint, not corrupting a spec. `&& $value !== []` would close it. Take or
   leave; it is the mirror of the vacuous-truth trap, so worth naming even though it can't bite here.
2. **One suggested test, because it pins a real design decision that nothing currently pins.**
   `declaresBodySchema` is an **any**-media-type quantifier — right for this rule, since the question
   is "does the response advertise *a* body" (deliberately unlike the *every*-quantifier that suits
   the `Location` predicate in #623, where the question is different). Concrete case: a 200 whose
   content is `[MediaType('application/json', no schema), MediaType('application/octet-stream',
   {type: string, format: binary})]` → expect no finding. I confirmed it passes today; it is just
   untested, so the quantifier could be silently flipped to `every` without a red test. (Also working
   but untested: a **keyed** `content` array — the by-value loop handles list and map alike.)
3. **Docs judgement is right, and I checked rather than assumed.** The catalog row
   (`docs/linting.md:655`) is the rule's *only* mention anywhere under `docs/`, its text comes from
   `$description` (unchanged), and the wording stays accurate — a scalar body *does* declare a body
   schema, so not reporting it is what the row already promises. No new rule, config key or public
   method. Nothing here needs a docs sub-phase.
4. **CHANGELOG placement looks wrong but isn't.** The entry landed at line 166, inside
   `### Documentation` rather than `### Fixed`. That section has been the de-facto append target for
   many merges — lines 150–165 already hold ~10 `Fixed:` / `Changed (internal):` / feature entries,
   including #584's from the merge immediately preceding this branch. The coder followed the file's
   actual practice and carried an inline `Fixed:` prefix, which is the sensible adaptation.
   Re-sectioning `[Unreleased]` is a separate chore, not this PR's job.
5. The three new global test helpers (`rawSuccessResponse` / `rawMediaType` / `rawSchema`) are
   generic names in the global namespace; I checked all three across `tests/` and there is no
   collision today (`tests/Feature/RawSchemaAttributeTest.php:19` defines only the near-miss
   `rawSchemaComponentFor`). Fine as-is.

## Observation, explicitly out of scope

The blind spot this PR routes around is **not** specific to one rule: a `?SomeData` 2xx body reaches
the tree with `schemaRef = null` *and* `fields = []`, so every rule keying off those signals
mis-reads a nullable Data response. Three others read them —
`OperationResponseTypeAbstract`, `ResponseExampleMissing`, `ResponseStatusUnconventional`. Fixing
that at the tree level (having `SpecTreeBuilder` recover the ref from a nullable composition's single
non-null branch, which `SchemaAccessor::classifyComposition` already isolates) would be the general
cure. Correctly out of scope here — the PR is explicit about not reworking `ResponseNode` /
`SpecTreeBuilder` — but worth an issue.

## Verdict

Approved at `7d8ad9b4bd54f6f8d4a4330822976014992300d3`. The fix is minimal and correct, the deviation
is justified where it matters, the #608 case is a genuine catch rather than a retrofitted story, and
no test in the file is vacuous. As an auto-merge candidate this still needs the second independent
approval.

Radiergummi:
🤖 **[lead]** approval 1 of 2 in from `reviewer-2` at `7d8ad9b4` (no blocking findings). Sending back for **one non-blocking addition** before the second approval — round 2 of 3.

Taking one of the five non-blocking notes and declining the rest. Reasoning on the record:

- **Taking:** the missing pin on the **any**-media-type quantifier. Nothing currently asserts that *any* media type declaring a shape suffices; a future change to *every* would leave the suite green. That is precisely the false-coverage standard I have enforced against three other agents this run, so it applies to this PR too.
- **Declining** the `&& $value !== []` guard against `oneOf: []` / `type: []`: unreachable from the generator, so it is speculative hardening.
- **Declining** the CHANGELOG re-sectioning (`### Documentation` vs `### Fixed`): real, but it follows ~10 merges of established drift including #584 immediately prior. A separate chore, not this PR's job.
- Docs judgement and the `ref`-unreachability qualification are accepted as recorded, no action.

Radiergummi:
🤖 **[coder]** addressed round 2: added the **any**-media-type quantifier pin; pushed `c990412c`.

Rebased onto `321d42cc` first — see the note at the end, the ancestor check disagreed with the
"no rebase needed" call.

## The one test

Exactly `reviewer-2`'s case, schema-less media type **first** so ordering can't mask a narrowed
quantifier:

```php
rawSuccessResponse(
    rawMediaType('application/json'),
    rawMediaType('application/octet-stream', rawSchema(['type' => 'string', 'format' => 'binary'])),
)
// → no finding
```

**Verified it pins what it claims to.** A passing test proves nothing here on its own — the whole
point is that it must die if someone flips `declaresBodySchema()` from *any* to *every*. So I
mutated the loop to an every-quantifier (return false on the first media type without a shape;
return true only if at least one media type was seen) and re-ran both files:

```
Tests: 1 failed, 42 passed
⨯ it does not flag a 200 when only one of several media types declares a schema
```

**One failure, and it is this test.** That is the result I wanted for two reasons: it confirms the
test actually discriminates the quantifier, and the *absence* of collateral failures independently
confirms `reviewer-2`'s finding that nothing else covered it — the mutation was previously silent
across the whole file. Rule file restored bit-for-bit afterwards (`git diff --stat` on it: empty).

Scope: this one test only. The four declined items stay declined — no `!== []` guard, no CHANGELOG
re-sectioning, no docs change, no `ref`-rationale edit. I've noted the `ref` qualification
(non-component and external refs still fire, pre-existing, not generator-emitted) so the claim
isn't inherited unqualified, but changed nothing.

## Rebase note

`main` had moved to `321d42cc`, and `git merge-base --is-ancestor origin/main origin/<branch>`
returned **false** at `7d8ad9b4` — so on the instrument we agreed to use, the branch was stale.
The lead's reasoning was sound (the delta is `.claude/`-only: `SKILL.md`, `bin/open-pr`,
`bin/start-issue` — nothing under `src/`, `tests/`, or `composer.json`, so no CI-relevant change),
but I rebased rather than assert green on a branch the check calls stale. Clean rebase, no
conflict. `--is-ancestor` is now **true** at `c990412c`.

**CI green on `c990412c`:** all 5 matrix jobs incl. swagger-php 5.8, Snapshot, Coverage, Tooling,
quality. Local: 3882 tests, PHPStan `[OK] No errors`.

Radiergummi:
🤖 **[reviewer]** review round 2: **approved at `c990412c2f9b2fbf73b94c02aff28cbb2a1304f2`, no concerns.**

Delta re-read, not a fresh full review — my round-1 pass at `7d8ad9b4` stands on its merits and this
re-states the approval at the new sha. Scope checked: the new test is what it claims, the rebase
dragged nothing along, and the four declined notes are absent. Own detached worktree at the reviewed
sha, own `vendor/`, removed afterwards.

## The rule file did not change — by blob identity, not by inspection

```
7d8ad9b4:src/Lint/Rules/ResponseSuccessEmptyBody.php  9ca903887937c27d39a1c275f019aa3330488024
c990412c:src/Lint/Rules/ResponseSuccessEmptyBody.php  9ca903887937c27d39a1c275f019aa3330488024
```

Identical object. That settles the declined items more firmly than a grep could: no `&& $value !==
[]` guard could have been added to a file whose blob hash is unchanged. Likewise `CHANGELOG.md` and
`docs/` are untouched between the two shas — no re-sectioning, no docs edit — and the `ref`
qualification remains a PR-comment note with no code change behind it.

Delta excluding `.claude/`: **one file, `tests/Unit/Lint/Rules/ResponseSuccessEmptyBodyTest.php`,
+20/−0.** Nothing else.

## The rebase dragged nothing along

`git diff origin/main c990412c` lists five files, all this PR's own, and **zero** under `.claude/` —
so the `.claude/` entries that show up when diffing against my old approved tree are simply #622
arriving via `main`, not something this branch picked up. Merge-base is `321d42cc`, which is current
`origin/main`; `--is-ancestor` true; two commits (seed + implementation).

For the record, I think rebasing rather than asserting green on a branch the agreed instrument calls
stale was the right call, even though the lead's reasoning about the `.claude/`-only delta was sound.
Cheaper to re-run than to argue about which staleness signal to trust.

## The new test — independently mutated, same verdict

It is my case, with the schema-less media type placed **first** so ordering cannot mask a narrowing.

I did not take the coder's mutation result on report. I rewrote `declaresBodySchema()` to a faithful
every-quantifier (return `false` on the first media type without a shape; return `true` only if at
least one was seen) and ran it across **both** affected files, 43 tests:

```
Tests: 1 failed, 42 passed
⨯ it does not flag a 200 when only one of several media types declares a schema
```

Exactly one failure, and it is this test. Rule file restored bit-for-bit afterwards.

That does the double duty the lead identified, and I'll state the second half plainly because it is
the more interesting one: the **absence** of collateral failures is independent confirmation of my
round-1 claim that nothing pinned the quantifier. The same mutation was previously silent across the
entire file — a reviewer's "this is untested" is itself a testable assertion, and this is the
instrument that tests it. Good practice; worth repeating when a review note claims a coverage gap.

## Conventions on the 20 added lines

Zero em-dashes. Three lines exceed the 100-char soft limit, but all three are the file's own
established idiom — the sha I already approved carries **32** lines over 100, and the
`OperationNodeFactory::makeOperation(pathUri: …)` one-liner appears **15** times. CLAUDE.md asks for
matching the surrounding style, so wrapping only the new occurrences would be the inconsistent
choice. Not a finding.

## Gates, re-run in my own worktree at `c990412c`

**3882 tests passed** (9835 assertions) — exactly one more than the 3881 I measured at `7d8ad9b4`,
which is the arithmetic you want when the delta is a single added test. PHPStan level 8 `[OK] No
errors`. Pint `passed`.

## Verdict

Approved at `c990412c2f9b2fbf73b94c02aff28cbb2a1304f2`. This supersedes my approval at `7d8ad9b4`,
which the head move voided.

Radiergummi:
🤖 **[reviewer]** review round 1: **approved at `c990412c2f9b2fbf73b94c02aff28cbb2a1304f2`**, no concerns.

Full independent pass, not a delta — I had not seen this PR before. I did not re-run what
`reviewer-2` and `coder-1` already established; I went at the two directions neither had mutated, and
both hold. Worked in my own detached worktree, restored to the PR sha afterwards.

---

## The negative direction holds, by two independent methods

**Through the real pipeline.** I put five genuinely-empty 2xx shapes through the actual `LintRunner`
and generator: a `void` return, a `JsonResponse` with a variable payload (the `ResponseEmptyController`
shape), an untyped return, a plain `Response`, and an `array` return with no shape. Every one emits
**no `content` at all**, and the rule still fires on every one. That is the substantive result: the
generator never produces the "content present but shapeless" shape for an empty body, so the new
early return cannot swallow a real empty response. The shapes the unit tests hand-build (media type
with no schema, schema declaring nothing) are reachable through authored `#[Response]` attributes
rather than inference, which is why building them by hand is the right call.

**By mutation.** Reading tests tells you they exist; only a mutant tells you they bite.

- `declaresBodySchema() → return true` (always-suppress) kills **exactly 4** tests — the three
  `still flags …` cases plus `emits a finding for a 200 response with no body schema` — and
  **nothing else across 1035**. The absence of collateral failures is itself the evidence that those
  four are precisely the negative-direction guards, rather than incidental passengers.
- The `Generator::UNDEFINED` truthiness trap you flagged is pinned. Substituting the naive
  `(bool) ($schema->{$keyword} ?? null)` for `is_defined(…)` — the exact confusion available here,
  since `UNDEFINED` is a non-empty string and therefore truthy — kills **exactly 1** test:
  `still flags a 200 whose media-type schema declares nothing`. One test, but the right one, and it
  would fail loudly the moment someone "simplified" that line.

Both mutations were confirmed applied via `git diff --quiet` before the suite was trusted. Worth
saying because the first attempt *didn't* apply: a `perl -0pi` substitution silently missed, and
without the check I would have read the resulting green as a real result.

## The `ref` / `properties` trim is right, and `reviewer-2`'s qualification is exact

Confirmed at source rather than accepted:

- **`properties` is genuinely unreachable.** `SpecTreeBuilder::buildFields()` builds fields from
  `properties` irrespective of any `type`, so a properties-only schema is already caught by the
  pre-existing `fields !== []` guard before the new code runs.
- **`ref` is unreachable only for local component-*schema* refs**, exactly as qualified.
  `SchemaAccessor::extractRef()` defers to `ComponentReference::name()`, which returns null unless
  the pointer matches `^#/components/schemas/`. So `#/components/responses/Foo` and an external
  `other.yaml#/User` leave `schemaRef` null, produce no fields, and declare no shape keyword — and
  the rule fires. That behaviour is **identical before and after this PR**; it is pre-existing, not
  introduced.

One refinement to the record: the reachable surface is slightly wider than hand-authored annotations.
`OperationBuilder::resolveRefSchema()` delegates to registered `RefSchemaResolver` plugins, which
return an arbitrary string, so a third-party resolver emitting a non-`schemas` pointer reaches it
too. Still narrow, still pre-existing, and still out of scope for #587 — adding `ref` to
`SHAPE_KEYWORDS` would *change* behaviour for a case this issue never covered. The trim is correct.

## One thing both passes missed — observation, no change requested

**No command-level test discriminates this rule's over-suppression direction.** Under the
always-suppress mutant, all 32 `LintCommandTest` tests pass. I traced why rather than leaving it as a
curiosity:

- `--no-suppress disables directives` asserts exit 1, but `SuppressedResponseEmptyController` under
  `--no-suppress` surfaces **two** findings — `response.no-error` *and*
  `response.success-empty-body` — so exit 1 survives without this rule contributing.
- `config disabled_rules …` asserts exit 0, which is insensitive to over-suppression by construction.

So the inline comments at `LintCommandTest:78` and `:124`, both asserting this rule is the only
finding at that level, overstate. (The unsuppressed `ResponseEmptyController` genuinely does emit
exactly one finding, so *its* comment is accurate — only the suppressed fixture's is not.)

This is pre-existing, lives in a file this PR does not touch, and changes nothing about the verdict:
the four unit tests are a real and sufficient defence, as M1 proved. I raise it only because the
command-level belt-and-braces one might assume is there, is not.

## The four declined items are absent — checked, not assumed

- **No `&& $value !== []` guard.** The only `!== []` in the rule is the pre-existing
  `$response->fields !== []` at `:62`.
- **No CHANGELOG re-sectioning.** A single added line, sitting under the existing `### Documentation`
  heading (`:125`); no headings moved or added.
- **No `docs/` change.** And it is right rather than merely deferred: the catalog row at
  `docs/linting.md:655` is the rule's `description` property verbatim, and that string describes
  *when the rule fires*, not which internal signals it consults — so it stays accurate under this
  change. Nothing stale was left behind.
- **No edit to the `ref` rationale.** The only `ref`-mentioning line in the `src/` diff is new
  explanatory text for the new guard.

## Gates

`composer check` green on a clean tree at the reviewed sha: **3882 passed**, 9835 assertions,
PHPStan `[OK] No errors`.

Nothing further from me. Approved.

Radiergummi:
🤖 **[lead]** all gates green → squash-merged. Closes #587.

**Gate, itemised:** two independent approvals both anchored to `c990412c` (`reviewer-2` re-approved after its `7d8ad9b4` approval was voided by the round-2 push; `reviewer-1` gave a fresh full pass). CI verified `headSha`-anchored via `gh run list`, not `gh pr checks` — see the note below. `src/` diff **55 lines in one file**, no `Contracts/`, no config key, no stage order, no public signature. Survey **N/A** (`area:lint`, cannot move a generated spec).

**What the two reviews established beyond the diff.** The false positive genuinely survived #608 in a new costume: a `?ScalarOnlyData` 2xx splits to `oneOf: [{$ref}, {type: 'null'}]`, and because `split()` moves `ref` onto the inner branch, `extractRef()` returns null on the outer and `schemaRef` is null too — so the response was invisible to *every* pre-existing signal, not merely to a `type` check. `oneOf` in `SHAPE_KEYWORDS` is what catches it.

Verified by mutation rather than reading, at four independent points: removing the early return fails exactly 11 tests; ablating each `SHAPE_KEYWORDS` entry individually kills tests for all four, so none is dead; an every-quantifier mutant kills exactly the new pin and nothing else across the file (that silence being what confirms the quantifier was previously untested); and `declaresBodySchema() → return true` kills exactly 4 of 1035. The `Generator::UNDEFINED` truthiness trap is pinned too — the naive `(bool) ($schema->{$keyword} ?? null)` kills exactly one test, and it is the right one.

The negative direction was confirmed through the real pipeline as well as by hand: five genuinely-empty 2xx shapes (`void`, variable-payload `JsonResponse`, untyped return, plain `Response`, shapeless `array`) emit no `content` at all and still fire, so the generator never produces a content-present-but-shapeless body the early return could swallow.

**Two limitations recorded rather than papered over**, both pre-existing and out of #587's scope:

1. The `ref` trim is unreachable only for **local component-schema** refs (`extractRef()` → `ComponentReference::name()` requires `^#/components/schemas/`). A media-type schema whose sole keyword is `#/components/responses/Foo`, an external pointer, or a string from a registered `RefSchemaResolver` plugin still fires. Adding `ref` would change behaviour for a case the issue never covered.
2. No **command-level** test discriminates this rule's over-suppression direction: all 32 `LintCommandTest` tests pass under an always-suppress mutant, because `SuppressedResponseEmptyController` under `--no-suppress` surfaces two findings so exit 1 survives without this rule, and the `disabled_rules` case asserts exit 0 and is insensitive by construction. The comments at `LintCommandTest:78` and `:124` claiming this rule is the sole finding overstate. The four unit tests are sufficient; noted because the belt-and-braces one might assume is there, is not.

**Method note worth carrying:** `gh pr checks --watch` returned the *previous* sha's green results with identical run IDs while new runs were queued. Trust only `gh run list --json headSha,status,conclusion` with a run carrying the current head.

PR #622, opened 2026-07-28
chore(autonomous-team): sign the seed commit and name the PR head explicitly

### LINKED ISSUE #617: chore(autonomous-team): `start-issue` pushes an unsigned seed commit and `open-pr` cannot create from an unchecked-out branch
Two bugs in `.claude/skills/autonomous-team/bin/`, both hit during the 2026-07-27 run while planning #610. Each cost a failed command and a manual workaround; both will recur on every run until fixed.

Reported by `planner-1`; filed by the lead. **I have not verified the proposed fixes myself** — they are the reporter's diagnosis, recorded here so whoever picks this up starts from the symptom, not from trust.

## 1. `start-issue` builds the seed commit with `git commit-tree`, which ignores `commit.gpgsign`

The repository ruleset requires verified signatures, so the push is rejected:

```
GH013: Repository rule violations found
```

`git commit-tree` does not honour the `commit.gpgsign` config the way `git commit` does — it needs the flag passed explicitly. Proposed fix: `git commit-tree -S`.

Workaround used: recreate the seed commit signed, by hand.

## 2. `open-pr` calls `gh pr create` with no `--head`

`gh pr create` infers the head from the *checked-out* branch. But `start-issue` deliberately checks nothing out — that is its design, so the primary checkout stays free for the lead while coders work in worktrees. The two scripts therefore contradict each other, and running `open-pr` from the shared checkout aborts:

```
you must first push the current branch to a remote, or use the --head flag
```

Proposed fix: pass `--head "$branch"` (and `--base main` for good measure).

Workaround used: replicate the script's logic by hand — labels minus `spec`/`agent:*`, `Closes #<n>`, the footer, and the assignee. PR #616 is fully conventional despite not being script-created, but the next planner will hit the same wall.

## Why this is worth fixing rather than working around

These are the two scripts that open **every** piece of work the team does, and the failure mode is not a clean error at the top — it is a rejected push partway through, leaving a local branch whose state the next step assumes is pushed. That is the same class of failure as the "coder reported ready with the work uncommitted" incident the skill already guards against in its verification step.

## Note for whoever fixes it

Do **not** patch these mid-run. The scripts are in use by live agents; changing them under a running team is how you get a half-migrated harness. Fix on a branch, land it between runs.

Radiergummi:
Closes #617

## What broke

Both bugs reported in #617 hit **this run**, in the first two commands: `start-issue` for #587 was
rejected by the ruleset (`GH013: Commits must have verified signatures`), and `open-pr` aborted
with `you must first push the current branch to a remote, or use the --head flag`. Together they
made it impossible to open any PR at all, which is why this was fixed before anything else rather
than left for between runs as the issue advises. No agents were running at the time — the fix
landed in the gap the issue asks for.

Both fixes are the reporter's proposed ones, now **verified end-to-end**: they are what opened
PR #621 (#587) and this PR.

## 1. `start-issue`: sign the seed commit

`git commit-tree` does not honour `commit.gpgsign` the way `git commit` does. Pass `-S`.

## 2. `open-pr`: name the head explicitly

`gh pr create` infers the head from the checked-out branch, but `start-issue` checks nothing out
on purpose — that is what keeps the primary checkout free while coders work in worktrees. So the
head has to be named.

Rather than add a required argument (which would break every existing call site and every role
doc that documents the two-argument form), the branch is **derived from the issue number**: the
unique pushed `<type>/<issue>-*` ref. `PR_HEAD` overrides it, and both the zero-match and
multi-match cases fail with an actionable message instead of falling through to the old
inferred-head behaviour. `--base main` is pinned at the same time.

Deliberately avoids `mapfile` — the scripts run under `#!/usr/bin/env bash`, which on macOS may
resolve to bash 3.2.

## Verification

- `bash -n` on both scripts.
- `start-issue fix 587 …` → signed seed commit, push accepted.
- `open-pr 587 …` from a checkout sitting on an unrelated branch → PR #621 created, labels
  mirrored, `Closes #587` and footer present.
- `start-issue chore 617 …` + `open-pr 617 …` → this PR.

## Not included

No `CHANGELOG.md` entry: this is agent harness tooling under `.claude/`, not library behaviour,
and the changelog documents the published package.

### DIFF
--- a/.claude/skills/autonomous-team/SKILL.md
+++ b/.claude/skills/autonomous-team/SKILL.md
@@ -143,7 +143,7 @@ single-active-label / closing-keyword / footer invariants.
 | `set-phase <num> <phase> [comment]` | lead | move phase: enforce one `agent:*` label + post annotation; on a PR, also clears the closed issue's stale label (issue→PR migration) |
 | `finish-pr <pr> [comment]` | lead | un-draft + squash-merge + delete branch + remove worktree + comment; cleanup is best-effort and warns about leftovers |
 | `start-issue <type> <N> [slug]` | planner | branch off fresh `origin/main`, empty commit, push; echoes branch. Checks nothing out — the primary checkout stays free |
-| `open-pr <N> <title> [body-file]` | planner | open PR with labels/`Closes`/footer/assignee (`PR_DRAFT=1` for draft) |
+| `open-pr <N> <title> [body-file]` | planner | open PR with labels/`Closes`/footer/assignee (`PR_DRAFT=1` for draft; head defaults to the unique pushed `<type>/<N>-*` branch, override with `PR_HEAD`) |
 | `worktree-add <branch>` | coder / docs-writer | add/reuse a worktree under `.claude/worktrees/`; echoes path |
 | `sync-branch` | coder / docs-writer | from a worktree: fetch + rebase `origin/main` + `composer check` |
 
--- a/.claude/skills/autonomous-team/bin/open-pr
+++ b/.claude/skills/autonomous-team/bin/open-pr
@@ -9,6 +9,8 @@
 # Env:
 #   PR_ASSIGNEE   maintainer to assign/request review (default: Radiergummi)
 #   PR_DRAFT=1    open as a draft instead of ready-for-review
+#   PR_HEAD       branch to open the PR from; defaults to the unique pushed
+#                 `<type>/<issue>-*` branch, which is what `start-issue` creates
 #
 # Echoes the created PR URL on success.
 set -euo pipefail
@@ -23,22 +25,28 @@ assignee="${PR_ASSIGNEE:-Radiergummi}"
 # issue number unless the caller pins it.
 head="${PR_HEAD:-}"
 if [ -z "$head" ]; then
-  candidates=$(
+  # Prune first: a remote-tracking ref for a branch deleted after its merge would
+  # otherwise look like a perfectly good single match, and fail at GitHub instead.
+  git fetch --prune origin >/dev/null 2>&1 || true
+
+  refs=$(
     git for-each-ref --format='%(refname:short)' \
       'refs/remotes/origin/feat/*' 'refs/remotes/origin/fix/*' \
-      'refs/remotes/origin/chore/*' 'refs/remotes/origin/docs/*' \
-      | sed 's|^origin/||' \
-      | grep -E "^(feat|fix|chore|docs)/${issue}-" || true
+      'refs/remotes/origin/chore/*' 'refs/remotes/origin/docs/*'
   )
+  # The trailing '-' keeps '587' off 'fix/5871-…'; sed strips the remote prefix.
+  candidates=$(printf '%s\n' "$refs" | sed 's|^origin/||' \
+    | grep -E "^(feat|fix|chore|docs)/${issue}-" || true)
   count=$(printf '%s' "$candidates" | grep -c . || true)
   if [ "$count" -eq 1 ]; then
     head="$candidates"
   elif [ "$count" -eq 0 ]; then
-    echo "error: no pushed branch matches '<type>/${issue}-*'; run bin/start-issue first, or set PR_HEAD" >&2
+    echo "error: no pushed branch matches '<type>/${issue}-*';" >&2
+    echo "       run bin/start-issue first, or set PR_HEAD" >&2
     exit 2
   else
     echo "error: several branches match '<type>/${issue}-*'; set PR_HEAD to pick one:" >&2
-    printf '  %s\n' $candidates >&2
+    printf '%s\n' "$candidates" | sed 's/^/  /' >&2
     exit 2
   fi
 fi
--- a/.claude/skills/autonomous-team/bin/open-pr
+++ b/.claude/skills/autonomous-team/bin/open-pr
@@ -18,6 +18,31 @@ title="${2:?usage: open-pr <issue-number> <title> [body-file]}"
 body_file="${3:--}"
 assignee="${PR_ASSIGNEE:-Radiergummi}"
 
+# `gh pr create` infers the head from the checked-out branch, but `start-issue` checks
+# nothing out on purpose, so the head has to be named explicitly. Derive it from the
+# issue number unless the caller pins it.
+head="${PR_HEAD:-}"
+if [ -z "$head" ]; then
+  candidates=$(
+    git for-each-ref --format='%(refname:short)' \
+      'refs/remotes/origin/feat/*' 'refs/remotes/origin/fix/*' \
+      'refs/remotes/origin/chore/*' 'refs/remotes/origin/docs/*' \
+      | sed 's|^origin/||' \
+      | grep -E "^(feat|fix|chore|docs)/${issue}-" || true
+  )
+  count=$(printf '%s' "$candidates" | grep -c . || true)
+  if [ "$count" -eq 1 ]; then
+    head="$candidates"
+  elif [ "$count" -eq 0 ]; then
+    echo "error: no pushed branch matches '<type>/${issue}-*'; run bin/start-issue first, or set PR_HEAD" >&2
+    exit 2
+  else
+    echo "error: several branches match '<type>/${issue}-*'; set PR_HEAD to pick one:" >&2
+    printf '  %s\n' $candidates >&2
+    exit 2
+  fi
+fi
+
 # Read the bespoke body (file or stdin).
 if [ "$body_file" = "-" ]; then
   body=$(cat)
@@ -50,6 +75,8 @@ draft_args=()
 [ "${PR_DRAFT:-0}" = "1" ] && draft_args+=(--draft)
 
 gh pr create \
+  --base main \
+  --head "$head" \
   --title "$title" \
   --body "$body" \
   --assignee "$assignee" \
--- a/.claude/skills/autonomous-team/bin/start-issue
+++ b/.claude/skills/autonomous-team/bin/start-issue
@@ -51,7 +51,9 @@ fi
 git fetch origin main >/dev/null
 
 base=$(git rev-parse origin/main)
-commit=$(git commit-tree "${base}^{tree}" -p "$base" \
+# `-S` is not optional: the repository ruleset rejects unsigned commits, and unlike
+# `git commit`, `commit-tree` ignores `commit.gpgsign`.
+commit=$(git commit-tree -S "${base}^{tree}" -p "$base" \
   -m "${type}: ${title} (#${issue})" \
   -m "Co-Authored-By: ${CO_AUTHOR:-Claude <noreply@anthropic.com>}")
 


Radiergummi:
🤖 **[lead]** fast-path, lead-implemented: the two harness bugs in #617 blocked every `start-issue`/`open-pr` call this run, so this was fixed before any agent was spawned (the "between runs" window the issue asks for). Mechanical, tooling-only, no `src/` diff. → review.

Radiergummi:
🤖 **[reviewer]** review round 1 — 3 findings (none block correctness), reviewed at `81c2ab76e4ce65b491c664dafc1c0683e558fc10`.

I treated the head-derivation as new, unproven logic and exercised it rather than reading it. Every
one of the specific doubts raised on handoff came back clean; the evidence is below so it doesn't
have to be re-derived. Test copies were verified byte-identical to the reviewed sha before running.

## Verified, not assumed

**`grep -c` under `set -euo pipefail`.** Sound in all three shapes. Empty input: grep prints `0`,
exits 1, `|| true` absorbs it, `count=0` — no `set -e` abort. Single unterminated line (which is
what `$(...)` always yields, trailing newlines stripped): counted as `1`. Two lines: `2`. Run under
`/bin/bash` 3.2.57, not the homebrew 5.x.

**Failure modes, run against the real script.** Zero-match (`open-pr 999999`) exits 2 with the
actionable message and never reaches `gh` — so a bad derivation can't half-create anything.
Multi-match, fabricated in an isolated temp repo with `refs/remotes/origin/{feat/700-alpha,
fix/700-beta}`, exits 2 and lists both candidates. `PR_HEAD` bypasses derivation entirely — with
the same ambiguous refs present it sailed past the block and failed later at the `gh` stage, which
is the proof it skipped derivation rather than resolved it.

**Prefix-collision.** The trailing `-` in `^(feat|fix|chore|docs)/${issue}-` is load-bearing and
works: against the live ref set, `587` selects only `fix/587-lint-scalar-2xx-body` despite
`fix/584-…` sitting next to it, and neither `61-` nor `6170-` cross-matches `617-`.

**`-S` on `commit-tree`.** The diagnosis is correct and the flag is the right shape. Built a
throwaway repo with `commit.gpgsign=true`: `commit-tree` without `-S` produces a commit with **no
`gpgsig` header**; with `-S` the header appears. I ran that under `gpg.format=ssh` while this
machine is openpgp — bare `-S` resolves through `user.signingkey` in either format, so `-S"$key"`
would be strictly worse (it would hardcode one key; note this machine already carries two different
`user.signingkey` values across global/local config, and bare `-S` correctly honours the winner).
It also **fails loudly**: with a bogus signing key, `commit=$(git commit-tree -S …)` aborts the
script at exit 1 before `git branch`/`git push` — verified no branch was created. That's the right
side of the "rejected push leaves a local branch the next step assumes is pushed" failure the issue
complains about.

**Portability.** `bash -n` passes on both scripts under 3.2.57. Nothing newer than 3.2 was added:
no `mapfile`, no `${var,,}`, no associative arrays, no `&>>`. The pre-existing
`"${label_args[@]+"${label_args[@]}"}"` idiom is untouched (`open-pr:99-100`) and still needed
under `nounset`.

**`--base main`.** Not a break. `gh pr create` already defaulted to the repo's default branch (it
does *not* infer the parent branch), so this is a no-op that documents intent; and nothing in the
skill opens a PR against a non-`main` base — `SKILL.md` and the role docs describe only
branch-from-`main`, and stacking is explicitly discouraged.

**Scope.** Two files, +30/-1, nothing extraneous. Comments state *why*, carry no issue numbers and
no em-dashes. Omitting the `CHANGELOG` entry is right and consistent with precedent — `972d1165`
touched six `.claude/` files including `bin/start-issue` and added no entry.

## Findings

1. **`open-pr:9-11` — `PR_HEAD` is documented nowhere.** The `# Env:` block lists `PR_ASSIGNEE` and
   `PR_DRAFT` but not `PR_HEAD`, and `SKILL.md:146` still describes the command as
   "`open-pr <N> <title> [body-file]` … (`PR_DRAFT=1` for draft)". Both new error paths
   (`open-pr:37`, `open-pr:40`) instruct the operator to set a variable that appears in no
   documentation, so the only way to learn the knob exists is to read the script — which is the
   exact friction this PR exists to remove. Add the line to the `# Env:` block and mention
   `PR_HEAD` in the `SKILL.md` row.

2. **`open-pr:26-32` — the derivation reads a possibly-stale local view of the remote.** Unlike
   `start-issue`, which does `git fetch origin main` before it acts, `open-pr` never fetches or
   prunes; it trusts `refs/remotes/origin/*` as-is. This checkout currently holds 10 remote-tracking
   type-branches, several for long-merged work (`chore/521-*`, `feat/208-*`, `feat/417-*`), so the
   staleness is real, not theoretical. Two shapes follow: a branch deleted on the remote but still
   tracked locally becomes a *phantom single match* and fails at GitHub with a much less legible
   message than the script's own; and a re-attempted issue with an abandoned first branch becomes a
   false multi-match. A `git fetch --prune origin >/dev/null 2>&1 || true` before the block makes
   the view authoritative for one round-trip on a script that already calls `gh` twice, and pruning
   remote-tracking refs is safe for concurrent worktrees (they hold local branches).

3. **Nits, batched — take or leave.** `open-pr:37` is 109 chars against the repo's 100-char soft
   limit. `open-pr:41`'s `printf '  %s\n' $candidates` is unquoted, so it gets pathname expansion
   as well as the intended word-splitting — unreachable while slugs stay `[a-z0-9-]`, but `IFS`-only
   splitting would be the honest spelling. `open-pr:31`'s `|| true` covers the whole pipeline, so a
   genuine `git for-each-ref` failure is reported as "no pushed branch matches" rather than as a git
   error; only the `grep` needs the guard. And `PR_HEAD` now has no `PR_BASE` counterpart, which is
   fine under a linear-history policy but is worth being a deliberate choice rather than an
   asymmetry.

## One latent hazard, explicitly not a finding

`--head` decouples the created PR from local state. Before this change, running `open-pr` from a
checkout sitting on the branch with unpushed commits made `gh` refuse; now the PR is created from
whatever the remote ref holds, silently omitting them. The documented flow never exercises this —
the planner runs `start-issue` then `open-pr` against the seed commit, with nothing local — and it
is not a regression from a working state, since that invocation previously failed outright. I'm
recording it because it is the same "assumed pushed, wasn't" class the issue itself calls out, so
if `open-pr` ever moves to the coder's worktree it should first warn when `$head` is the current
branch and `HEAD != origin/$head`.

## Verdict

The two fixes are correct, minimal, and now independently verified rather than taken on the
author's word. Finding 1 is a two-line documentation fix I'd want before merge; finding 2 is a
one-line hardening I'd recommend. Neither changes the logic, so re-approval at the new sha is
immediate.

Radiergummi:
🤖 **[lead]** addressed review round 1 at `3410db12` (also rebased onto `ecf5af8c` — both frozen PRs merged since this branch opened).

- **F1 (want before merge)** — `PR_HEAD` now documented in `open-pr`'s `# Env:` block and in the `SKILL.md` helper table. You were right that it was self-defeating: the error paths named a variable discoverable only by reading the script this PR exists to stop people reading.
- **F2 (recommend)** — `git fetch --prune origin >/dev/null 2>&1 || true` before deriving. Verified the absorption matters: in an isolated repo with no `origin` the fetch fails and derivation still completes, so a remote-less checkout is not bricked by the prune.
- **F3 (nits)** — all three. The ref listing is split out of the pipeline so `|| true` no longer masks a genuine `for-each-ref` failure as "no branch matches"; `$candidates` is quoted (indenting now via `sed 's/^/  /'`, since quoting it breaks the per-line `printf`); the 109-char line is wrapped.

Re-verified the three derivation paths after the change, since I touched the control flow you had already cleared:

| path | result |
|---|---|
| live issue 587, one branch | `head=fix/587-lint-scalar-2xx-body` |
| issue 99999, no branch | exit 2, actionable message, no `gh` call |
| fabricated `fix/777-alpha` + `feat/777-beta` | exit 2, both listed and correctly indented |

`/bin/bash -n` (3.2.57, not homebrew) clean.

**Your latent hazard is recorded, not fixed** — `--head` decouples the PR from local state, so unpushed commits are silently omitted where `gh` used to refuse outright. Agreed it is not a regression and the documented planner flow never reaches it, so I am not adding a guard on speculation. It goes in the run summary; if `open-pr` ever moves into the coder's worktree it becomes real and wants a `git rev-list origin/$head..$head` check first.

Radiergummi:
🤖 **[reviewer]** review round 2: **approved at `3410db12191e1a54d82ba8a398356b893396e4b4`, no concerns.**

Second independent read. I took `reviewer-1`'s five cleared items on its stated method and spent the
budget on the four things that are genuinely new since round 1 — all four came back clean, and I
exercised them rather than reading them. Work done in an isolated temp repo with fabricated refs and
`git`/`gh` shims; the script copies were `shasum`-verified byte-identical to `3410db12` before
running, and nothing touched any agent's worktree.

## 1. The control-flow split — `set -e` now aborts on a real `for-each-ref` failure

This is the one that mattered most, since the `|| true` used to swallow it. Shimmed `git` to fail
only on `for-each-ref` and delegate everything else:

```
$ SHIM_FAIL_FOREACHREF=1 open-pr 700 t /dev/null
fatal: simulated for-each-ref failure
exit=128
```

Git's own message, git's own exit code, no "no pushed branch matches", and the `gh` shim was never
reached — so a genuine ref-listing failure can no longer half-create anything or masquerade as a
missing branch. `refs=$(…)` is a top-level simple assignment, so `set -e` applies to the command
substitution (it would not inside a `local`, but it isn't).

Nothing regressed in the split. The zero-ref case still degrades correctly: `for-each-ref` succeeds
with empty output → `printf`/`sed` yields one empty line → `grep -E` exits 1 → `|| true` → `count=0`
→ the actionable exit 2. Re-ran all three derivation paths against the shipped script:

| path | result |
|---|---|
| single match (`701`) | `--base main --head chore/701-solo` |
| zero match (`999999`) | exit 2, actionable message, `gh` never called |
| multi match (`700`) | exit 2, both candidates listed, correctly indented |
| `PR_HEAD` set with ambiguous refs present | `--head pinned/branch`, derivation skipped entirely |

Worth noting the last row also means `PR_HEAD` short-circuits the new fetch, so pinning costs no
network round-trip.

## 2. The prune — safe under concurrency, and failing open is the right call

**Safe.** Built a repo with a worktree on `fix/700-work` holding an *unpushed* commit, deleted the
branch on the remote, then pruned from the primary checkout. After the prune:

```
HEAD:                fix/700-work @ 2b72510   (the unpushed WIP, intact)
local branch:        YES
remote-tracking ref: PRUNED
status:              ## fix/700-work...origin/fix/700-work [gone]
worktree list:       both entries intact
```

Prune only ever deletes `refs/remotes/*`; a worktree holds a *local* branch, so its HEAD, index and
unpushed work are untouched. Recovery is automatic in both directions that matter: `git push -u`
from the worktree recreates the tracking ref, and `sync-branch`'s rebase is onto `origin/main`,
which is never a prune target because `main` is never deleted. `worktree-add` also works off the
local branch. No data loss, no corruption — the `[gone]` line is cosmetic.

**Failing open is correct**, and for a stronger reason than "it degrades gracefully":

- *Lock contention.* A hard failure here would make `open-pr` fail at random whenever another agent
  happens to be fetching the same repo — strictly worse for a multi-agent harness than running on a
  slightly stale view, which is exactly the pre-fix behaviour. (Six concurrent `fetch --prune`
  processes on one repo: all exit 0, no contention observed, but the guard is right regardless.)
- *A real outage cannot be masked past the next line.* If the network or auth is down and a stale
  ref still matches, `head` is set and the script's very next actions are `gh issue view` (`:78`)
  and `gh pr create` (`:85`) — both fail loudly with `gh`'s own diagnostic. If nothing matches, it
  exits 2 before mutating anything, and "run `bin/start-issue` first" is in fact the correct advice
  in the only flow that reaches it (a successful `start-issue` leaves the tracking ref behind via
  `push -u`, so the normal path always has a match). There is no arrangement where swallowing the
  fetch error produces a *wrong* PR.

## 3. The indent — the pathological case is impossible at the git level, not just the slug level

`git check-ref-format` settles it independently of `start-issue`'s slugifier:

```
fix/617-a b   REJECTED     fix/617-a?b   REJECTED     fix/617-a\b   REJECTED
fix/617-a*b   REJECTED     fix/617-a[b   REJECTED     fix/617-a&b   ALLOWED
```

Whitespace and every glob metacharacter are refused by git itself (`git branch 'fix/617-a b'` →
`fatal: … is not a valid branch name`), so no such ref can exist to be listed — the quoting fix is
correct but the hazard it closes was already unreachable. `&` is the one sed-adjacent character git
permits, and it is inert here: `&` is only special in a sed *replacement*, and the replacement is a
literal two spaces. Verified `fix/617-a&b` passes through unmangled.

Byte-checked the multi-match output (`sed -n l`): exactly two spaces, one line per candidate, no
stray blank line. The single-candidate case never reaches the `sed` at all — `count -eq 1` assigns
`head` and returns — and the primitive handles it correctly anyway. `start-issue`'s slug is
`[a-z0-9-]` by construction (`start-issue:36`), so even the hand-passed `[slug]` argument can't
smuggle anything in.

## 4. The rebase — clean, nothing dragged along

`git merge-base ecf5af8c 3410db12` is exactly `ecf5af8c`; three commits on top; diff is exactly
`SKILL.md` + `bin/open-pr` + `bin/start-issue`, +39/−2. No `src/`, no test, no doc outside
`.claude/`. Both merges crossed cleanly with no interaction, as expected.

## Beyond the brief

- **`shellcheck` is clean on both scripts at `3410db12`** (exit 0). More usefully, running it on the
  round-1 sha `81c2ab76` reproduces **SC2086 at exactly old line 41** — so F3's quoting nit was a
  real defect, and the fix genuinely resolves it rather than restating it. Line widths: no line over
  100 chars in either file, so F3's 109-char line is resolved too.
- **F1 is fully addressed** at the PR head: `PR_HEAD` is in `open-pr`'s `# Env:` block (`:12-13`)
  and in `SKILL.md:146`. `roles/planner.md:34` needs no change — it uses the two-argument form,
  which still works.
- **Probed the raw `${issue}` interpolation into the `grep -E`.** Unvalidated input reaches a
  regex, so I tried to make it mis-select: `7|701`, `70[01]`, `.*`, `70.` all broaden the match and
  are caught by the `count -eq 1` guard into the multi-match exit 2; `#701` narrows to zero-match.
  Every metacharacter form lands on the safe side, and the trailing `-` holds (`70` does not match
  `700-`/`701-`). Not worth a validation guard.
- **Fix #1 proves itself against the real ruleset.** This PR's own seed commit `f6aaf29b` was
  produced by the patched `start-issue`; GitHub's API reports `{"verified": true, "reason":
  "valid"}` for it and for the head. That is stronger evidence than the throwaway-repo test.
- **`-S` is needed nowhere else** — `start-issue:56` is the only plumbing commit creation in
  `bin/`; everything else goes through `git commit`, which honours `commit.gpgsign`.

## Two observations, recorded not required

1. The type list `feat|fix|chore|docs` now lives in three places: `start-issue`'s validating `case`
   (`:25`), `open-pr`'s `for-each-ref` patterns (`:34-35`), and `open-pr`'s `grep -E` alternation
   (`:39`) — where the type group is actually redundant, since `for-each-ref` already restricted to
   those four. Adding a fifth type to `start-issue` alone would silently zero-match. Divergence
   fails legibly (exit 2 with `PR_HEAD` as the escape hatch), so this is a note for whoever adds a
   type, not a change to make now.
2. Pre-existing, not this PR: `start-issue:57` builds the subject as `"${type}: ${title}"`, and
   issue titles already carry a conventional-commit prefix — hence this branch's seed commit reading
   `chore: chore(autonomous-team): …`. Squash-merge discards it, so it only ever shows in
   pre-merge `git log`. Worth a one-line follow-up if it ever annoys anyone.

## Verdict

Approved at `3410db12191e1a54d82ba8a398356b893396e4b4`. Both fixes are correct and minimal, the
round-1 remediation is real rather than cosmetic (shellcheck confirms), and the four newly-introduced
behaviours hold up under adversarial exercise. Lead authorship made no difference to the read.

Radiergummi:
🤖 **[reviewer]** review round 2: **approved at `3410db12191e1a54d82ba8a398356b893396e4b4`**, no concerns.

This supersedes my round-1 approval at `81c2ab76`, which is void — the head moved. Rebased onto
`ecf5af8c`; all three findings are addressed. The head-derivation block was restructured rather than
patched, so I re-ran the paths against the new script instead of carrying the old result forward.

**F1 — `PR_HEAD` documented.** Both places: the `# Env:` block in `open-pr` and the `SKILL.md:146`
row, and the wording matches what the code actually does ("the unique pushed `<type>/<N>-*` branch").

**F2 — the prune genuinely fixes the phantom.** I A/B'd this rather than reading it, because the
zero-match error alone doesn't prove the prune did anything (a pattern that never matched would look
identical). In a throwaway origin/clone pair: the script's exact `for-each-ref` patterns **do** see
`origin/fix/700-alpha`, deleting that branch on the remote leaves the tracking ref behind as a
phantom, and the new script prunes it and reports the clean, actionable zero-match — where a phantom
single match would previously have gone to `gh pr create --head <gone>` and failed at GitHub with a
much worse message. `origin/chore/701-only`, still live on the remote, survived the prune, so it
isn't over-pruning.

**F3 — all three nits.** Longest line is now 98 chars (the message is split across two `echo`s).
The multi-match list is `printf '%s\n' "$candidates" | sed 's/^/  /'` — quoted, so no word-splitting
or pathname expansion. And splitting `refs=$(git for-each-ref …)` out of the `|| true` pipeline does
what it was meant to: run from a non-git directory, the script now dies with git's own
`fatal: not a git repository` at **exit 128**, instead of swallowing it and reporting the misleading
"no pushed branch matches". `shellcheck -s bash` is clean on both scripts.

**Re-verified on the new code, not assumed:**

- Single match reaches `gh`; multi-match exits 2 listing both candidates indented; `PR_HEAD` still
  bypasses derivation entirely (proven by it sailing past two ambiguous refs).
- **Offline degrade is preserved** — the new `git fetch --prune` is the one thing that could have
  made this script fail where it used to work. With an unreachable remote the fetch fails, `|| true`
  absorbs it, and the derivation runs on the stale view and reaches `gh`. Graceful, as intended.
- Pruning is safe against concurrent worktrees: it only removes `refs/remotes/origin/*`, and a
  worktree holds a *local* branch ref, which is untouched.
- `bash -n` passes on both under bash 3.2.57; nothing newer than 3.2 was introduced by the rewrite.
- The `start-issue` hunk is **byte-identical** to `81c2ab76`, so round 1's `-S` evidence carries over
  unchanged: `commit-tree` demonstrably ignores `commit.gpgsign`, bare `-S` resolves via
  `user.signingkey` under either `gpg.format`, and a broken key aborts before `git branch`/`git push`.

The latent `--head`/unpushed-commits hazard I recorded in round 1 is unchanged and still not a
finding — the planner flow doesn't exercise it.

Radiergummi:
🤖 **[lead]** all gates green → squash-merged. Closes #617.

**Merge gate, itemised** (this PR is lead-authored, so the record should show the gate was applied to it exactly as to anyone else's):

- **Two independent approvals, both anchored to `3410db12`.** `reviewer-2` approved the current head outright; `reviewer-1` re-approved after its round-1 findings were fixed, explicitly voiding its earlier `81c2ab76` approval. I did **not** count the round-1 approval — the head had moved, which voids it under this run's own rule, and the fact that a re-read was offered as "immediate" is not the same as it having happened.
- **CI green.** The 5-9s job times are the workflow's result-cache short-circuiting, which is correct here: the marker key hashes `src/**`, `tests/**`, `config/**`, `examples/**` and the toolchain files, and this PR touches only `.claude/**`, so the key matches `main`'s and the cached result stands for a genuine full run on identical test-relevant content.
- **Non-controversial:** `src/` diff is **0 lines**. No `Contracts`, no stage order, no config key, no public signature.
- **Survey: N/A** — nothing generation-affecting.

**What the two reviews established that the diff alone does not.** Fix #1 is proven against the real ruleset rather than a throwaway repo: this PR's own seed commit `f6aaf29b` was produced by the patched `start-issue`, and GitHub's API reports `{"verified": true, "reason": "valid"}`. Bare `-S` is the right form rather than `-S"$key"`, since this machine carries two different `user.signingkey` values across global and local config. The prune was A/B'd rather than inferred — a zero-match error alone cannot distinguish "pruned the phantom" from "the pattern never matched" — and it neither over-prunes a live branch nor breaks offline use. `shellcheck` reproduces SC2086 at the round-1 sha and is clean at this one, so F3 was a real defect fixed, not reworded.

**Two observations recorded, deliberately not fixed here:**

1. The `feat|fix|chore|docs` type list now lives in three places; adding a fifth type to `start-issue` alone would silently zero-match in `open-pr`. It fails legibly and `PR_HEAD` is the escape hatch, so it is not worth a mechanism today.
2. `--head` decouples the PR from local state, so unpushed commits are silently omitted where `gh` previously refused outright. Not a regression (that invocation used to fail entirely) and the documented planner flow never reaches it — but it is the same "assumed pushed, wasn't" class this issue itself calls out. If `open-pr` ever moves into the coder's worktree it becomes real and wants a `git rev-list origin/$head..$head` guard first.

Issue #614, opened 2026-07-27
fix(responses): an authored `205` on a resource `response()->json()` vanishes, where a `204` survives body-less


Radiergummi:
## Problem

A resource wrapped in `response()->json(<resource>, 205)` loses its status entirely. The operation
documents a body-less **`200`** — there is no `205` anywhere in the document.

```php
public function reset(Post $post): JsonResponse
{
    return response()->json(PostResource::make($post), 205);
}
```

→ `responses: { 200: { description: OK } }`. The authored `205` is gone.

## The defect is the asymmetry with `204`, not the yield

Both statuses are contentless under RFC 9110, both are correctly refused a resource envelope, and
the two calls are one digit apart — but they end up in opposite places:

| authored status | resource envelope | status in the document |
|---|---|---|
| `204` | refused (correct) | **`204`, body-less — preserved** |
| `205` | refused (correct) | **absent — silently downgraded to a default `200`** |

Both outcomes are pinned by adjacent tests on the same file:

- `tests/Feature/Plugins/ApiResources/ReturnExpressionResolutionTest.php:1073` — 204 →
  `successStatuses($responses) === [204]`, no `content`.
- `…:1000` — 205 → `successStatuses($responses) === [200]`, no `content`.

Refusing the envelope is right in both cases. Emitting `200` for an action that can only return
`205` is not: it is a positive claim about a response the action never produces, and it is *less*
honest than the `204` case, which documents the real status with no body.

## Mechanism

Two resolvers, and only one of them knows about `205`.

1. `ResourceResponseResolver::carriesResourceBody()`
   (`src/Plugins/ApiResources/Resolvers/ResourceResponseResolver.php:100-103`) excludes both `204`
   and `205`, so the ApiResources path yields for either. Correct, and deliberate.
2. Core's `InlineJsonResponseResolver` (`src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php:474-480`)
   carries a carve-out for `204` **only**:

   ```php
   // 204 must not carry a body.
   if ($status === 204) {
       return PrimaryResponse::of(
           new OA\Response(['response' => '204', 'description' => 'No Content']),
           statusIsExplicit: $statusIsExplicit,
       );
   }

   if (!$result->bodyReadable) { … return null; }
   ```

   The carve-out returns *before* the body-readability check. A resource argument is not a
   statically readable body, so a `205` falls past it, hits `!$result->bodyReadable`, and returns
[...]

PR #633, opened 2026-07-28
fix(responses): read the authored wrapper status on the `#[ResponseResource]` path

Radiergummi:
Closes #611

## Premise re-verified against `main` (@ 649231ea)

- `ResourceClassLocator::locate()` still returns early for a valid `#[ResponseResource]` at
  `src/Plugins/ApiResources/Support/ResourceClassLocator.php:68-77` (the issue cited 66-77;
  the branch begins at 68 on `main`). The constructed `ResourceTarget` passes only
  `resourceClass` + `isCollection`, so `authoredStatus` stays `null`.
- The guard the fix has to reach is `ResourceResponseResolver::resolvePrimaryResponse()`
  (`src/Plugins/ApiResources/Resolvers/ResourceResponseResolver.php:71-73`), immediately after
  the `isAmbiguous` check. It is correct and unreachable on the attribute path — confirmed by
  reading, not assumed.
- `#[ResponseResource]` implements `PrimaryResponseAuthoringAttribute`, and
  `InlineJsonResponseResolver` skips any action declaring one
  (`src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php:73`). So once ApiResources yields,
  **nothing else** claims the primary response on an attributed action — the operation is left
  with its error responses and a body-less conventional status, exactly like the unattributed
  #584 path.
- #606 and #608 are merged; no unmerged contract is in the way. No open PR touches any file in
  this plan (#631 owns `NullableSchema`, #626 owns `EloquentModelToSchema`, #571 is
  request-side, #526/#210 are stale drafts with no pushed diff). `EloquentModelToSchema` and
  `NullableSchema` do **not** appear below.

## The scope call

The issue flags a genuine tension and asks for it to be settled against the documented contract.
Verified:

- `src/Attributes/ResponseResource.php` — "Names the resource **class** an operation returns …
  `collection: null` lets the resolver infer cardinality." The docblock scopes the attribute to
  class + cardinality. Silent on status.
- `docs/attributes.md:29` — "Explicit response-resource class **for the 200 response**." This is
  the one place the documented contract contradicts the suggested shape, and it does so only by
  naming `200` as the status the class attaches to.
- `docs/plugins.md:310-313` — "`#[ResponseResource]` is the escape hatch and always wins" (about
  the *resource resolution* the reader refuses).

So: the attribute's constructor and its own docblock support "it names the resource, not the
status"; one docs cell asserts `200`. **Recommendation: adopt the issue's suggested shape** —
[...]

Issue #640, opened 2026-07-28
fix(responses): an attributed `json(<resource>, <status>)` followed by `->header(…)` still documents the phantom 200


Radiergummi:
## This is a residual of #611, not a regression of #633

**Do not file this as a regression when you meet it after #633 merges.** #633 closes #611 for the
direct `response()->json(<resource>, <status>)` wrapper on an attributed action. This shape escapes it
and was equally broken before, on `main`. #633 narrows the defect; this issue is what is left.

## Symptom

An action carrying `#[ResponseResource]` whose wrapper is followed by a **status-preserving chain
link** still documents the phantom `200`:

```php
#[ResponseResource(NestedAuthorResource::class)]
public function show(): JsonResponse
{
    return response()
        ->json(NestedAuthorResource::make($author), 403)
        ->header('X-Author', 'yes');
}
```

Measured on #633's head (`3a71e890`), with `->withHeaders([…])` behaving identically:

| body | success `200` | component registered | `403` present |
|---|---|---|---|
| **attributed** `json($r, 403)->header(…)` | **carries the resource envelope** | **yes** | yes |
| **attributed** `json($r, 403)->withHeaders([…])` | **carries the resource envelope** | **yes** | yes |
| **unattributed**, byte-identical body | no content | no | yes |

That last row is what makes this a defect rather than a design choice: the same body without the
attribute documents no envelope and registers no component, because the unrecognised chain costs the
*resource*. With the attribute the resource comes from the attribute, so the refusal costs only the
*status* — and the result is #611's reproduction verbatim, a `200` claiming a success the action never
returns, next to the real `403`.

## Which chain links actually matter here

`header()`, `withHeaders()` and `cookie()` are **declared on `JsonResponse`**, and none of them can
change a status. `return response()->json($r, 403)->header(…)` is therefore ordinary code, and it is
the shape this issue is about.

The #633 plan reasoned instead about whether the status read had to unwrap
`ReturnExpressionResourceReader::RESOURCE_PRESERVING_CHAIN_METHODS` (currently `['additional']`), and
concluded it did not, because `additional()` is declared on
`Illuminate\Http\Resources\Json\JsonResource` and not on `JsonResponse`. That conclusion is right for
the practical reason that unwrapping the *resource*-preserving list buys nothing here — but **the
justification must not be stated as "such a call is impossible"**, which is how an earlier draft of
this issue and of #633's review both put it. `Illuminate\Http\JsonResponse` composes `Macroable`
[...]

PR #642, opened 2026-07-28
fix(responses): document a literal `response()->json(<body>, 205)` as a body-less 205

### LINKED ISSUE #636: fix(responses): a literal `response()->json(<body>, 205)` documents a 205 carrying content, which RFC 9110 forbids
## Problem

An action returning `response()->json(<literal>, 205)` is documented as a **`205` carrying content**,
which RFC 9110 §15.3.6 forbids: `205 Reset Content` must not have a body. The generator emits an
invalid document.

```php
public function reset(): JsonResponse
{
    return response()->json(['ok' => true], 205);
}
```

→ `responses: { 205: { content: { application/json: { schema: { … } } } } }`

## Measured, not inferred

Fresh worktree at `main` @ `649231ea`, one route per generated spec, no simulation and no decorator —
these are the generator's real outputs today:

| shape | measured output |
|---|---|
| `GET` + `response()->json(['ok' => true], 205)` | **`205`, HAS CONTENT** ← invalid |
| `DELETE` `destroy()` + `response()->json(['ok' => true], 205)` | **`205`, HAS CONTENT** ← invalid |
| `GET` + `response()->json(['ok' => true], 204)` | `204`, no content — correct |
| `GET` + `response()->json(['ok' => true], 304)` | `200`, no content |

Two things the matrix settles:

- **Not route-dependent.** It reproduces on a plain `GET` and on a resourceful `destroy()` alike, so
  it is not about the resourceful-route convention.
- **`204` is fine, `205` is not** — the same one-digit asymmetry as #614, but on a different path and
  with a different symptom.

## Mechanism

`InlineJsonResponseResolver` carves out exactly one status:

```php
// src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php:474-480
// 204 must not carry a body.
if ($status === 204) {
    return PrimaryResponse::of(
        new OA\Response(['response' => '204', 'description' => 'No Content']),
        statusIsExplicit: $statusIsExplicit,
    );
}

if (!$result->bodyReadable) { … return null; }
```

A `205` is not `204`, so it falls past the carve-out. Unlike a *resource* argument (which is not
statically readable and therefore bails at `!$result->bodyReadable`), a **literal array is perfectly
readable**, so the resolver proceeds to the bottom of the method and emits
`205` + `content` + schema. The status is also marked `statusIsExplicit`, which suppresses
`OperationBuilder`'s conventional-status pass, so nothing downstream corrects it.

The `304` row degrades for an unrelated and correct reason: `ensureSuccessStatus()`
(`:364-383`) rejects any status outside 200-299 and notes it, so the resolver yields and the route
convention supplies a body-less `200`.

## Not the same defect as #614

Both involve `205`, so they will look like duplicates. They are not:

- **#614** is about an authored `205` **losing its status** when the body is a *resource* (not
  statically readable). Symptom: a missing `205`.
- **This issue** is about an authored `205` **keeping both its status and a body** when the body is a
  *literal*. Symptom: an invalid `205`.

This one is on the **unattributed** path with a readable body, so it is not touched by #611 / PR #633
(which only routes the authored status on the `#[ResponseResource]` path) and it is not
attribute-specific.

## It invalidates one of #614's acceptance bullets

#614's test list contains:

> A non-resource `response()->json(['ok' => true], 205)` keeps whatever it does today, so the
> contentless carve-out does not swallow a readable body.

Taken literally that instructs an implementer to **pin this bug**: what it does today is emit content
on a `205`. The intent behind the bullet is right (widening the carve-out must not silently discard a
readable body); the assertion is wrong. Whoever plans #614 should pin *this* issue's expected output
instead — `205` with **no** content, and the readable body dropped with a note, since a body the spec
forbids cannot be documented anywhere.

## The underlying gap

Nothing in the pipeline owns the rule *"a contentless status must not carry content."* The set of
bodiless statuses is currently rediscovered in **three** places, each with a different membership:

| location | set | purpose |
|---|---|---|
| `src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php:474` | `{204}` | skip the body |
| `src/Plugins/ApiResources/Resolvers/ResourceResponseResolver.php:100-103` | `{204, 205}` | refuse the envelope |
| `src/Lint/Rules/ResponseSuccessEmptyBody.php:34` | `{204, 205, 304}` | exempt from the empty-body finding |

Three sets, three sizes, and the smallest one is the emitter. That is the shape that produces
one-digit gaps like this and like #614: each contentless status gets rediscovered a carve-out at a
time, per claimant.

I think this framing does hold, and that it makes **#597** (partial primary-response contributions —
status-only and body-only resolvers) a plausible home for the general fix, the same as #614's
consequence #1 argues: a resolver could contribute the `205` as a status while contributing no body,
with no per-status carve-out on any claimant. Two honest caveats, so this is not tidied into a
unification it does not deserve:

- **#597 is not required for this one.** Unlike #614, this defect has a correct local fix that needs
  no new contribution shape: Core already produces the response, so it only has to not attach content
  to a bodiless status. That is a small change at a single site.
- Even under #597, *something* still has to hold the set. The improvement is having **one** owner
  instead of three, not eliminating the knowledge.

Also worth noting there is no lint rule for the inverse of `response.success-empty-body` — a bodiless
status that *does* declare a body. If the generator can be made unable to emit one, a rule is
unnecessary; if authored `#[Response]` attributes can still declare it, a rule may be the better
guard. That is a separate call.

## Proposed change

Smallest correct fix: make Core's contentless handling status-set-driven rather than a `204` literal,
so a readable body is **not** attached to a bodiless status, and note the discarded body (the user
authored something that cannot be documented — the same reasoning as the existing degrade notes).
Whether the set is `{204, 205}` or shared with the lint rule's `{204, 205, 304}` is the implementer's
call; `304` never reaches this point today because `ensureSuccessStatus()` filters it first.

## Tier

**Tier 0.** The status is already statically read and in hand at the decision point, and "a `205`
carries no content" is a constant. No body parsing, no new source of information.

## Tests

- `response()->json(['ok' => true], 205)` on a plain route → `205` documented with **no** `content`.
- The `204` case is unchanged (regression guard on the existing carve-out).
- A `200` with the same literal body still documents its schema — the fix must not swallow readable
  bodies on statuses that may carry them.
- The discarded body emits exactly one note per action and run, naming the action and the status.
- `304` still degrades to the conventional status via `ensureSuccessStatus()`.

Related: #614 (the same `205` asymmetry, resource bodies, missing-status symptom), #597 (possible
home for the general rule).

Radiergummi:
Closes #636

## Premise verified — re-measured independently, not re-read

The issue was filed from a measured probe. I re-measured on **`main` @ `82041e0c`** (this branch =
that commit + the seed commit) in a fresh worktree, one route per generated spec, no decorator, and
extended the matrix to the shapes the issue did not cover. All rows below are **generated output**,
not reasoning:

| shape (unattributed, `JsonResponse` return) | measured today |
|---|---|
| `response()->json(['ok' => true], 205)` | **`205` + content** ← invalid |
| `new JsonResponse(['ok' => true], 205)` | **`205` + content** ← invalid (new) |
| `response()->json(['ok' => true])->setStatusCode(205)` | **`205` + content** ← invalid (new) |
| resourceful `destroy()` + literal `205` | **`205` + content** ← not route-dependent (confirms the issue) |
| `response()->json([], 205)` (empty literal) | `200`, body-less (new) |
| `response()->json(<resource>, 205)` | `200`, body-less (new) |
| `response()->json($unreadable, 205)` | `200`, body-less (new) |
| `response()->json(['ok' => true], 204)` | `204`, body-less — correct |
| `response()->json([], 204)` / `<resource>` / `$unreadable` at `204` | `204`, body-less in **all three** (new) |
| `response()->json(['ok' => true], 200)` | `200` + content — correct |
| `response()->json(['ok' => true], 304)` | `200`, body-less — correct (`ensureSuccessStatus()`) |

Line numbers hold on `82041e0c`: the carve-out is at
`src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php:474-480`, `ensureSuccessStatus()` at
`:364-383`. #632 (the only commit since `649231ea`) touched `EloquentModelToSchema` only.

**The 204/205 asymmetry is total, and that is the real finding.** `204` is claimed *before* the body
is read, so it lands body-less no matter what the body evidence is (empty literal, resource,
unreadable — all three measured `204`). `205` is claimed only in the one case where it is wrong:
when the body **is** readable. Three of the six `205` shapes lose the status; the other three keep
it with content the RFC forbids.

Precision on the symptom: the emitted document is *OAS-valid* — OpenAPI does not police HTTP
semantics — but it describes a response RFC 9110 §15.3.6 forbids. "Invalid document" in the issue
means HTTP-illegal, not schema-illegal.

## The design fork, decided by measurement

Two shapes are available. I implemented **both** as throwaway probes and measured them.

**Variant A — widen the existing pre-read gate to `{204, 205}`** (make `205` behave exactly like
`204`). Measured: *all six* `205` rows become `205` body-less. Coherent, one gate, one set.
Blast radius on the existing suite: **exactly one assertion** —
`tests/Feature/Plugins/ApiResources/ReturnExpressionResolutionTest.php:1009` flips from `[200]` to
`[205]` (132 other tests in the four adjacent suites pass).

**Variant B — drop a *readable* body on a bodiless status, after the readability gate.** Measured:
the three invalid rows become `205` body-less; the other three keep today's `200`. Blast radius:
**zero** — all 133 tests pass unchanged.

**Recommendation: B.** Not because it is smaller, but because A *is* #614:

- #614's reproduction is `return response()->json(PostResource::make($post), 205);` on an
  **unattributed** action, and the tests it cites are that same `:1000`/`:1073` pair. Variant A
  turns that row into a body-less `205`, which is #614's acceptance verbatim. Implementing it here
  is folding in #614, which this run has ruled out.
- The one assertion A must rewrite lives in the file **PR #633 is adding ~14 tests to**. Two
  in-flight PRs editing it is the collision the run is avoiding.

So the two issues split cleanly along the gate: **#636 owns "a readable body must not be attached to
a bodiless status"** (post-read), **#614 owns "a bodiless status must be emitted even when no body
was read"** (pre-read widening). #614's implementation then collapses both gates into one
set-driven gate and updates `:1009` itself, in the PR that owns that file. Merge order follows the
planner rule that the narrowing lands before the widening.

### The asymmetry B leaves must be justified truthfully or not at all

B encodes a real difference: `204` is claimed before the body read, `205` after. The **truthful**
reason — and the only one to write in a comment — is affirmative emptiness: a `204` mirrors
`response()->noContent()`, which the resolver already documents as *"an affirmative body-less
response"* (`:92`), so an unreadable body does not weaken it. A `205` whose body the scan cannot
read carries no affirmative claim of emptiness at all — the response is *unknown*, not *empty*, and
degrading to the conventional status stays the honest answer.

**Do not write any of these** (each is a factual assertion I checked and it is false):

- ~~"nothing else claims a `205`, so it is left for another resolver"~~ — nothing does claim it;
  measured, the row degrades to a conventional `200`.
- ~~"the runtime strips the body"~~ — true for `204`, **false for `205`**: Symfony's
  `Response::isEmpty()` is `[204, 304]` (`vendor/symfony/http-foundation/Response.php:1264-1267`),
  so a Laravel app really does send a body on a `205`. `docs/auto-derivation.md:509` uses that
  justification for `204`; extending the sentence to `205` verbatim would state something untrue.
- ~~the existing `note()` suffix, "no response inferred. Annotate the action with `#[Response]`"~~ —
  a response *is* inferred here (the `205`), and that advice invites authoring the very
  invalid body we just removed. See below.

If the implementer finds they need a justification beyond the affirmative-emptiness one, that is the
signal to stop and escalate rather than invent one.

### The note is load-bearing, not a nicety

Because the runtime does **not** strip a `205` body, dropping it from the document means the
document and the app genuinely disagree: the app sends a body the spec says cannot exist. Without a
note the user silently loses a documented body and is told nothing. So the note is what makes this
fix honest, and it needs its own wording — the shared `note()` helper hardcodes a suffix that is
false here.

## Tier

**Tier 0.** The status is statically read and in hand at the decision point; "a `205` carries no
content" is a constant. No new parsing, no new information source.

## Implementation

One production file: **`src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php`**. Nothing in
`src/Contracts/**`, no stage order, no config key, no public signature change → **not
controversial**. Does not touch either of PR #633's files
(`ResourceClassLocator.php`, `ReturnExpressionResourceReader.php`), nor `ResourceResponseResolver.php`,
nor the lint rule.

1. Name the set once in the class rather than adding a second status literal:
   `private const array CONTENTLESS_STATUSES = [204, 205];` — with a one-line note that only 2xx
   values can reach these gates because `ensureSuccessStatus()` runs two lines above (verifiable by
   the reader; measured for `304`).
2. Keep the existing `204` gate at `:474` where it is, reading the set membership rather than the
   literal, so behaviour there is unchanged.
3. After the `bodySchema === null` check, add the new gate: a bodiless status returns the status
   with **no** `content`, passing `statusIsExplicit: $statusIsExplicit` exactly as the `204` gate
   does, and calling the new note.
4. Add `noteDiscardedBody(ReflectionMethod $method, int $status): void` — a distinct NOTICE saying
   the status is documented, the readable body is **not**, and RFC 9110 forbids content on that
   status. It must not reuse `note()`'s suffix.

≈ **30 changed `src/` lines**, well under the 150 flag.

## Test matrix

The measured table above is the oracle. TDD: row 1 first, red on `main`.

**Feature — `tests/Feature/InlineJsonResponseTest.php`** (fixtures in
`tests/Fixtures/InlineJsonFixtureController.php`, following its existing style):

| # | shape | asserts |
|---|---|---|
| 1 | `response()->json(['ok' => true], 205)` | `205` present, **no `content` key**, description `Reset Content` — the acceptance bullet, red today |
| 2 | `new JsonResponse(['ok' => true], 205)` | same — the construction form funnels through the same site (measured) |
| 3 | `response()->json([...])->setStatusCode(205)` | same — the chained form too (measured) |
| 4 | resourceful `destroy()` + literal `205` | `205`, no content — pins route-independence, and that the conventional `204` does not reclaim it |
| 5 | `response()->json(['ok' => true], 204)` | unchanged: `204`, no content — regression guard on the untouched gate |
| 6 | `response()->json(['ok' => true], 200)` and `…, 201` | unchanged: content **present** with the literal's schema — the "must not swallow readable bodies" guard |
| 7 | `response()->json(['ok' => true], 304)` | unchanged: degrades to a body-less conventional `200` via `ensureSuccessStatus()` |
| 8 | `response()->json(<resource>, 205)`, unattributed | unchanged: body-less conventional `200`. **Pins the #614 boundary** — this row is what #614 will change, and pinning it here makes that PR's movement visible instead of silent |

**Unit — `tests/Unit/Core/Inference/InlineJsonResponseResolverTest.php`** (has `recordingLogger()`
and `sameClassHelperSerialized()` for the `content`-absence check):

| # | asserts |
|---|---|
| 9 | the resolver's `OA\Response` for a literal `205` serialises with no `content` key, `statusIsExplicit` true |
| 10 | **exactly one** record is logged, naming the action and the status, and its message does **not** claim "no response inferred" |
| 11 | the `204` literal case still logs **nothing** (the discarded-body note must not start firing on the row this PR does not change — a silent scope leak otherwise) |
| 12 | a `200` literal logs nothing and keeps its schema |

Row 11 is deliberate: extending the note to `204` would be a defensible separate change, but it is
not this issue, and today's `204` path is silent.

## Docs

- `docs/auto-derivation.md:509` — the sentence *"A literal `204` documents as `204 No Content`
  without a body schema — the runtime strips the body"*. Add `205` as a **separate** clause with its
  own (true) reason: a literal `205` documents as `205 Reset Content` without a body schema because
  RFC 9110 forbids content on it, and a body written in the call is dropped with a generation-log
  note. Do **not** extend the runtime-strips justification to `205` (measured false).
- The same section's *Known limitations* neighbourhood is where the honest consequence belongs: the
  app is still sending that body at runtime, so the note is the only place the disagreement
  surfaces.
- No `docs/linting.md` row: this PR adds no rule.

## CHANGELOG

One entry under `## [Unreleased]` → `### Fixed`: a literal `response()->json(<body>, 205)` (and the
`new JsonResponse(…, 205)` / `->setStatusCode(205)` forms) is documented as a body-less
`205 Reset Content` instead of a `205` carrying content, which RFC 9110 §15.3.6 forbids; the
unrepresentable body is dropped with one generation-log note per action and run. `204`, `200`/`201`,
and the `304` degrade are unchanged. `(#636)`

## Survey gate — applies, and here is exactly what it can and cannot prove

`area:responses`, generation-affecting, so **run it** — but with the right expectation, measured
rather than assumed. I grepped the 11-app corpus (app code, excluding `vendor/`, `node_modules/`):

- **`205` status usages: zero.** No `, 205)`, no `HTTP_RESET_CONTENT`, no `status: 205`. (Instrument
  sanity-checked against a control: the same grep finds 199 `204`/`noContent()` sites, and the only
  literal `205` anywhere in the corpus is a digit inside JSON fixture strings.)
- **The surrounding code path is exercised heavily:** 648 `response()->json(<args>, <3-digit>)`
  calls, 108 `setStatusCode(`, 1 `new JsonResponse(…, <status>)`.

Therefore: a flat/green survey **cannot speak to the defect at all** — no corpus app can produce a
`205`. What it *can* discriminate is collateral damage: the change sits inside the single hottest
response-inference method, four lines from 648 live call sites, so a green result is real evidence
that no body-bearing status lost its schema. State it that way in the gate report rather than
reporting green as confirmation. The unit + feature matrix is the only thing that verifies the fix.

(Worth noting alongside the newly-filed corpus-pristineness caveat: this claim is a *differential*
one — "no `205` inputs exist" is a property of the app sources, unaffected by leftover annotation
scratch.)

## The "one owner" question — I agree with planner-2, and it is not this PR

The set is rediscovered three times with three memberships: `{204}` in the emitter
(`InlineJsonResponseResolver`), `{204, 205}` in `ResourceResponseResolver::carriesResourceBody()`,
`{204, 205, 304}` in `ResponseSuccessEmptyBody::BODILESS_CODES`. The framing holds, and it is the
mechanism for these one-digit gaps: the smallest set is the emitter, so the emitter is where an
unhandled status becomes wrong output rather than a missed exemption.

The unifiable concept is precise: **"this status forbids content"** = `{204, 205, 304}` (plus 1xx,
unreachable here). The two 2xx-scoped callers already carry their own range checks
(`ensureSuccessStatus()`; the `>= 200 && < 300` clause), so a single predicate would serve all three
with **zero behaviour change**.

I still recommend against doing it here, for three reasons rather than tidiness:

1. Two of the three sites are adjacent to live work — `ResourceResponseResolver` is the resolver PR
   #633 is built around, and its set is exactly what #614 will touch next.
2. A zero-behaviour-change refactor earns no test of its own; folded into a bug fix it would
   dilute the diff whose correctness reviewers must check.
3. #597 (partial primary-response contributions) may relocate where the knowledge lives entirely —
   as both #636 and #614 argue. Unifying now risks unifying into the wrong home.

Even under #597, *something* holds the set; the win is one owner instead of three, not eliminating
the knowledge. Recommend filing it as a `chore` once #614 has landed and Core's two gates have
collapsed into one — at that point the emitter's set is `{204, 205}` in one place and the refactor
is genuinely mechanical.

## Follow-up worth filing (measured, not folded in)

The issue leaves open whether the inverse of `response.success-empty-body` is needed. Measured
answer: **yes, the door stays open.** An authored
`#[Response(status: 205, schema: ['type' => 'object', …])]` emits `205` **with content** today, and
this fix does not close that path — it only governs what Core infers. Combined with the runtime not
stripping a `205`, that means after this PR the generator can still be made to emit an HTTP-illegal
response, and nothing in `openapi:lint` says so (measured: linting the offending route yields only
`response.no-error`). A `response.contentless-status-has-body` rule would be the guard. Separate
call, separate issue — recorded here so the decision is made on evidence.

## Interaction with #614, preserved deliberately

#614's test list contains *"A non-resource `response()->json(['ok' => true], 205)` **keeps whatever
it does today**"*. Taken literally that pins this bug. This PR contradicts that bullet on purpose:
after it, the expected output for that shape is `205` with **no** content and the body dropped with
a note. #614's planner must read the bullet as superseded rather than as an instruction — the intent
behind it (a widened carve-out must not silently discard a readable body) is satisfied here by the
note, not by keeping the content. Row 8 of the matrix pins #614's own row so its movement is
visible.

## Gates

`composer test`, `vendor/bin/pint --test`, `composer lint` (PHPStan level 8). No new PHPStan risk:
the new gate reads an `int` already narrowed by `ensureSuccessStatus()`.


Radiergummi:

**Premise re-measured, not re-read.** Fresh worktree at `main` @ `82041e0c`, one route per generated spec. The issue reproduces, and I extended its matrix: `new JsonResponse([...], 205)` and `response()->json([...])->setStatusCode(205)` are **also** `205`-with-content (both funnel through the same site, so one fix covers three shapes). The `204` row and the `304` degrade are correct and untouched.

**The sharper finding: the 204/205 asymmetry is total.** Measured, `204` lands body-less for *every* body shape — empty literal, resource, unreadable — because its gate sits before the readability check. `205` is claimed only in the one case where it is wrong (readable body); the other three `205` shapes silently lose the status and degrade to `200`.

**The design fork was decided by measurement, not argument — I implemented both as throwaway probes.**
- **Variant A** (widen the pre-read gate to `{204, 205}`, i.e. make `205` behave exactly like `204`): all six `205` rows become body-less `205`. Blast radius: **exactly one** existing assertion, `ReturnExpressionResolutionTest.php:1009`; 132 other tests in the adjacent suites pass.
- **Variant B** (drop a *readable* body after the readability gate): the three invalid rows are fixed, **zero** test churn (133 pass).

**Recommended: B — because A *is* #614.** #614's reproduction is an *unattributed* `response()->json(PostResource::make($post), 205)` and it cites that same `:1000`/`:1073` pair, so A delivers #614's acceptance verbatim, in the test file PR #633 is actively extending. The two issues split cleanly along the gate: **#636 = a readable body must not be attached to a bodiless status** (post-read); **#614 = a bodiless status must be emitted even when no body was read** (pre-read widening, collapsing both gates into one and updating `:1009` in the PR that owns that file). Narrowing before widening.

**Two prescribed-wording traps, both checked and both false if copied:**
- `docs/auto-derivation.md:509` justifies the `204` rule with *"the runtime strips the body"*. That is **false for 205**: Symfony's `Response::isEmpty()` is `[204, 304]` (`Response.php:1264-1267`), so the app really does send the body. The `205` clause needs its own reason (RFC 9110 §15.3.6), and the plan says so explicitly.
- The shared `note()` helper hardcodes *"no response inferred. Annotate the action with `#[Response]`"* — false here (a response **is** inferred) and it invites authoring the invalid body we just removed. The plan requires a separate note method.

Because the runtime does not strip a `205`, the note is load-bearing rather than a nicety: without it the user silently loses a documented body while their app keeps sending one.

**Survey gate: applies, and I measured what it can discriminate.** Corpus (11 apps, app code only): **zero** `205` usages — no `, 205)`, no `HTTP_RESET_CONTENT`, no `status: 205` (instrument sanity-checked: the same grep finds 199 `204`/`noContent()` sites). But **648** `response()->json(<args>, <status>)` calls and **108** `setStatusCode(`. So a green **cannot speak to the defect** and must not be reported as confirming it; what it does discriminate is collateral damage four lines from 648 live call sites. The unit + feature matrix is the only verification.

**One-owner question: I agree with planner-2, and it should not ride on this PR.** The unifiable concept is precise (`"this status forbids content"` = `{204, 205, 304}`; both 2xx-scoped callers already carry their own range check, so a shared predicate is a zero-behaviour-change refactor). Against doing it here: two of the three sites are adjacent to live work (#633's resolver, #614's next edit), a zero-delta refactor earns no test and would dilute the diff under review, and #597 may relocate the knowledge entirely. Recommend a `chore` after #614 collapses Core's two gates into one.

**Follow-up worth filing (measured, not folded):** an authored `#[Response(status: 205, schema: …)]` still emits `205` with content, and lint says nothing about it (only `response.no-error` fires). So the generator can still be driven to emit an HTTP-illegal response after this fix — evidence that the inverse of `response.success-empty-body` is justified.

≈30 changed `src/` lines in one file (`InlineJsonResponseResolver.php`) — under the 150 flag, and no overlap with #633's files.

Radiergummi:
🤖 **[lead]** planning → plan-review. One fork referred up; deciding it here so plan-review inherits it.

**Decision: variant B (drop a *readable* body after the readability gate). Not A.**

The planner settled the fork properly — it implemented both as throwaway probes and measured the churn
rather than arguing from the shapes: **A** fixes all six `205` rows and breaks exactly one assertion
(`ReturnExpressionResolutionTest.php:1009`, 132 others passing); **B** fixes the three invalid rows with
**zero** test churn. Its recommendation is B and I am taking it, for a reason it identified and one it
did not lean on hard enough:

1. **A resolves #614 as a side effect.** A delivers #614's acceptance criteria verbatim — #614's
   reproduction is an unattributed `response()->json(PostResource::make($post), 205)` citing the same
   `:1000`/`:1073` pair. Absorbing another tier-0 issue into this PR would close it **without it ever
   being planned or reviewed on its own terms**, and #614's mechanism has already needed correcting
   **twice** in this run. It is the last issue in the backlog that should be resolved implicitly.
2. **A touches a file #633 is actively extending.** `ReturnExpressionResolutionTest.php` is in #633's
   diff, which is mid-rebase onto the new `main` right now. I set the disjoint-files constraint at
   admission precisely to avoid two agents in one file, and I am not going to breach it for tidiness.

The split the planner articulates is principled rather than merely convenient, which is what makes B the
right narrow fix instead of a partial one: **#636 = a readable body must not be attached to a bodiless
status** (post-read); **#614 = a bodiless status must be emitted even when no body was read** (pre-read
widening, which then collapses Core's two gates into one and updates `:1009` in the PR that owns that
file). Narrowing before widening. Note for whoever plans #614: A is one assertion plus its comment away,
and after this lands it becomes the natural shape of #614's fix.

**Endorsed without change:**

- **The one-owner refactor stays off this PR.** Agreed with both planners. The concept is precise
  (`{204, 205, 304}` = "this status forbids content") and the refactor is zero-behaviour-change, but two
  of the three sites sit next to live work, a zero-delta refactor earns no test of its own and would
  dilute the diff reviewers must actually check, and #597 may relocate the knowledge entirely. File the
  `chore` for after #614.
- **Survey gate applies, and its power was measured rather than assumed.** Corpus app code carries
  **zero** `205` usages (instrument sanity-checked against a control finding 199 `204`/`noContent()`
  sites), so a green **cannot** speak to the defect and must not be reported as confirming it. What it
  genuinely discriminates is collateral damage four lines from **648** live `response()->json(<args>,
  <status>)` calls and **108** `setStatusCode(` calls. That is real detection power, and stating the
  limit before the run is exactly what #631 failed to do.

**Two prescribed-wording traps caught before they shipped**, which is this run's newest standing rule
working as intended — a review or plan that prescribes wording is making a claim, and prescribed wording
gets probed like prescribed code:

- `docs/auto-derivation.md:509` justifies the `204` rule with "the runtime strips the body". **False for
  `205`:** Symfony's `Response::isEmpty()` is `[204, 304]` (`Response.php:1264-1267`), so a Laravel app
  really does send a `205` body. The `205` clause needs RFC 9110 §15.3.6 as its reason instead.
- The shared `note()` helper hardcodes "no response inferred. Annotate the action with `#[Response]`" —
  false here (a response *is* inferred) and it would invite authoring the very invalid body being
  removed. **Because the runtime does not strip a `205`, that note is load-bearing**: without it the user
  silently loses a documented body while the app keeps sending one.

**Also measured and worth knowing:** the `204`/`205` asymmetry is total — `204` lands body-less for
*every* body shape because its gate precedes the readability check, while `205` is claimed only in the
one case where it is wrong and silently degrades to `200` in the other three. And `new JsonResponse([...],
205)` plus `response()->json([...])->setStatusCode(205)` are the same defect; all three shapes funnel
through `responseFromCall`, so one site fixes all three.

≈30 changed `src/` lines, one file, no overlap with #633.

Radiergummi:
🤖 **[reviewer]** plan-review: **approve with 3 changes.** This is the best-founded plan I have reviewed in this run — implementing both variants as throwaway probes and measuring the churn is the right way to settle a fork. I re-derived every claim I was asked to; **four hold, one is confirmed in substance but has two wrong numbers**, and the test matrix has a gap. I also found one structural argument that makes B's central claim stronger than the plan states, and one argument that bears on #614 later.

## 1. "One site fixes all three" — confirmed, and it is deductive, not inductive

The plan rests this on three measured rows. It is stronger than that: **`:496` is the only `content`-emitting site in the entire resolver** — `grep -n "'content' =>" src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php` returns exactly one hit. Both entry points funnel into `responseFromCall()` (`:105` via `findJsonCall`, `:111` via `findJsonResponseConstruction`), and every shape's status arrives in the same `$result->status` from `callReader->read()` — the chained `->setStatusCode()` included, which is why all three measure `205 + content` today: that output is *only* reachable through `:496`. The other `PrimaryResponse::of` (`:294`, the `$this->helper()` delegation) emits status-only and cannot produce the defect.

So a gate anywhere above `:496` provably covers every shape that can exhibit this bug, present or future. Worth saying that way in the plan — it turns "three shapes measured" into "no fourth shape can exist", which is what a reviewer of the eventual diff needs.

## 2. Both prescribed-wording traps — confirmed

**(a) The Symfony claim is right, checked in the installed vendor, not from memory.** `Response::isEmpty()` is `\in_array($this->statusCode, [204, 304], true)` at `vendor/symfony/http-foundation/Response.php:1264-1267` — the exact lines cited. I also traced the *mechanism*, not just the predicate: the body is dropped at `:254-255` (`if ($this->isInformational() || $this->isEmpty()) { $this->setContent(null); … }`), gated on that same `isEmpty()`. So "the runtime strips the body" is true for `204`/`304` and **false for `205`**: a Laravel app really does put a `205` body on the wire.

Two consequences, both as the plan says. `docs/auto-derivation.md:509` — *"A literal `204` documents as `204 No Content` without a body schema — **the runtime strips the body** — as does a bare `response()->noContent()`"* — must gain a separate `205` clause citing RFC 9110 §15.3.6, never an extension of that dash-clause. And **the note is load-bearing**: the document and the running app genuinely disagree, so without it the user silently loses a documented body while the app keeps sending one.

**(b) Confirmed.** `note()` (`:343-357`) hardcodes `'; no response inferred. Annotate the action with #[Response] to document it.'` Both halves are false here — a response *is* inferred, and the advice invites re-authoring the body just removed. The plan's implementation item 4 requires a distinct `noteDiscardedBody()` that must not reuse the suffix. Requirement present and correctly scoped.

## 3. The asymmetry is real — but the matrix pins one of three degrade rows (**change 1**)

Control flow confirms it: the `204` gate sits at `:474`, **above** `if (!$result->bodyReadable)` at `:481`, so `204` is claimed before any body evidence exists; `205` reaches `:496` only when the body *is* readable. Exactly as claimed.

The gap: the plan measures **three** `205` shapes that degrade to a conventional `200` — empty literal `json([], 205)`, `json(<resource>, 205)`, and `json($unreadable, 205)` — and the test matrix pins only the resource one (row 8). Add the other two. They matter for the reason row 8 exists: these are the rows #614 will move, and pinning a third of them makes a third of that movement visible. Two fixtures.

## 4. Row 8 — confirmed

It is #614's own row (#614's reproduction is `response()->json(PostResource::make($post), 205)` on an unattributed action) and it is explicitly labelled "**Pins the #614 boundary**". Good, subject to change 1.

## 5. Survey framing — right in substance, two numbers wrong (**change 2**)

I re-ran the greps over the 11-app corpus (25 967 PHP files under `apps/*/repo`, excluding `vendor/`, `node_modules/`). My first attempt returned `0` for the *control*, which is the tell that the instrument was broken (`$(cat …)` over 26k paths), so I redid it with `xargs -0`:

| claim | plan | measured |
|---|---|---|
| `204` / `noContent()` control | 199 | **199** ✅ exact |
| `setStatusCode(` | 108 | **108** ✅ exact |
| `205` status usages | zero | **zero** ✅ — one raw match, `'rgb(255, 205, 86)'` (`apps/Pelican/repo/app/Filament/Admin/Resources/Nodes/Widgets/NodeStorageChart.php:58`), a CSS colour, not a status |
| `response()->json(<args>, <status>)` | 648 | **not reproducible**; I measure 1742 `response()->json(` occurrences and 1138 `->json(` lines carrying a second argument |
| `new JsonResponse(…, <status>)` | 1 | **5** status-bearing (264 constructions total) |

Two exact matches make me confident we are measuring the same corpus the same way, which is why the two misses are worth correcting rather than shrugging at. Both err in the *strengthening* direction — the hot path is hotter than claimed and the construction form is exercised 5× more — so neither changes the conclusion, but the plan presents them as measured facts. Either correct them or keep only the two verified figures. Minor precision: the stray `205` is an `rgb()` string, not "a digit inside JSON fixture strings".

**The framing itself is exactly right and is stated before the run**: a green cannot speak to the defect (no corpus app can produce a `205`), and its real power is collateral damage four lines from the hottest response-inference path. That is the discipline #631 lacked, and it should be the template.

## Change 3: the CHANGELOG must say the `205` fix is partial

The draft names what is unchanged (`204`, `200`/`201`, the `304` degrade) but not that **three of the six `205` shapes still degrade to a conventional `200`** — including `json([], 205)`, where a body-less `205` would arguably be more correct than today's `200`. A reader would reasonably conclude all `205` handling is now right. One clause fixes it, and it keeps #614's remaining scope visible from the changelog alone. In a plan this scrupulous about honest justifications, this is the one place the honesty thins.

## The fork: B is right, and I have a third reason the plan does not give

I did not re-open it, but the reasons invited attack and they survive: closing #614's acceptance criteria as a side effect of another issue is bad on its own terms given #614's mechanism has been corrected twice this run, and the disjoint-files constraint is exactly what stops a mid-rebase collision with #633.

There is a **substantive** third reason, which falls out of check 2(a): variant A widens the **pre-read** gate, so `json($unreadable, 205)` would document a body-less `205`. The `204` version of that is safe *because* the runtime strips the body — whatever the code passed, the wire carries none. For `205` the runtime strips nothing, so A would assert emptiness on **no evidence** about a response that may well carry a body. A is therefore not merely "#614's job" but questionable as stated, and #614 will have to confront it: its acceptance criteria ("a body-less `205` is emitted even when no body was read") asserts something the runtime does not guarantee. Worth writing into #614 now, while it is fresh — it is the same class of finding as the `docs:509` trap.

## The two deferrals: both correctly out of scope

**One-owner `{204, 205, 304}`** — agree, and the strong reasons are 1 and 3 (two of the three sites are adjacent to live work, `ResourceResponseResolver` being both #633's centre and #614's next target; and #597 may relocate the home entirely, so unifying now risks unifying into the wrong one). I put less weight on reason 2 as phrased — "earns no test" is not the issue; enlarging the diff whose correctness reviewers must check is, which is what the plan actually means. The filing trigger ("once #614 has landed and Core's two gates have collapsed") is right.

**Authored `#[Response(status: 205, schema: …)]` still emitting content** — file it, do not fold it. It is a different path (authoring, not the body scan) and, importantly, a different *fix shape*: for an authored response, silently dropping the body would override an explicit instruction, so a lint finding is more likely correct there than the drop-with-note this PR chooses. Please record that in the follow-up so nobody plans it as "do what #636 did". It also belongs to the family of #635/#638: the generator can still be driven to emit something no client can satisfy.

## Verdict

**Approve**, conditional on: (1) matrix rows for the two unpinned `205` degrade shapes; (2) the two survey numbers corrected or dropped; (3) one CHANGELOG clause naming the partial `205` coverage. Tier 0, ~30 `src/` lines, one file, disjoint from #633 (`ResourceClassLocator.php` / `ReturnExpressionResourceReader.php` untouched, and `ReturnExpressionResolutionTest.php` is not in B's blast radius — the whole point of choosing B). None of the three blocks coding; they are additions to the plan, not redirections of it.

Radiergummi:
🤖 **[lead]** plan-review → coding. Approved with **three changes**, none blocking. Variant **B** stands.

**Confirmed, and one claim upgraded from inductive to deductive.** `:496` is the **only** `content`-emitting
site in the resolver (single grep hit); both entry points funnel into `responseFromCall()` (`:105`, `:111`)
and all three statuses arrive in the same `$result->status`. So a gate above `:496` provably covers every
shape that can exhibit this — **no fourth shape can exist.** The plan should state it that way; "one site
fixes all three" reads as an enumeration when it is actually a structural guarantee.

**The Symfony trap is confirmed in the installed vendor, with the mechanism traced** rather than just the
predicate: `isEmpty()` = `[204, 304]` at `Response.php:1264-1267`, and the body is dropped at `:254-255`
gated on that same call. So a `205` body really is sent, the `204` docs clause must not be extended to it,
and `noteDiscardedBody()` is genuinely load-bearing.

**Change 1 — pin all three degrade rows, not one.** The plan measures three `205` shapes degrading to `200`
(empty literal, resource, unreadable) and pins only the resource one. Pin the other two: they are the rows
**#614** moves, and pinning a third of them makes a third of that movement visible.

**Change 2 — correct two survey figures.** The framing is right and is the template I want (limit stated
*before* the run), but two numbers are presented as measured facts and do not reproduce. The reviewer's
first grep returned `0` for the **control**, which is the tell that the instrument was broken (`$(cat …)`
over 26k paths); redone with `xargs -0`:
- `204`/`noContent` control **199** ✅ and `setStatusCode(` **108** ✅ — exact matches, so both are measuring
  the same corpus the same way, which is what makes the misses worth correcting rather than shrugging at.
- `205` usages **zero** ✅ — though the single raw match is `'rgb(255, 205, 86)'` in Pelican, a **CSS colour**,
  not the "JSON-fixture digit" the plan calls it.
- `response()->json(<args>, <status>)`: **648 not reproducible** — 1742 `response()->json(` occurrences,
  1138 `->json(` lines carrying a second argument.
- `new JsonResponse(…, <status>)`: claimed 1, actually **5** (264 constructions).
Both misses err toward *strengthening* the gate's collateral-damage power, so the conclusion is unchanged.

**Change 3 — the CHANGELOG must say the `205` fix is partial.** It names what is unchanged but not that
**three of six** `205` shapes still degrade to `200`, including `json([], 205)` where a body-less `205`
would arguably be more correct. As written a reader concludes all `205` handling is now right. This run has
already had to correct #614 twice for exactly that kind of overstatement.

**A third reason B is right, and it changes #614.** It falls out of the Symfony finding: A widens the
**pre-read** gate, so `json($unreadable, 205)` would document a **body-less `205`**. The `204` version is
safe *because* the runtime strips the body — for `205` it strips nothing, so A asserts emptiness on **no
evidence** about a response that may well carry a body. So A is not merely "#614's job": **#614's acceptance
criteria as written assert something the runtime does not guarantee.** Being written into #614 now, while
it is fresh — same class as the `docs:509` trap, and the third correction that issue has needed.

**Both deferrals stay out of scope**, with one refinement: of my three reasons for keeping the one-owner
refactor off this PR, reason 2 as phrased is weak — the real cost is **enlarging the diff reviewers must
check**, which is what it was reaching for. And for the authored-`#[Response(205)]` follow-up (**#643**),
record that its fix shape **differs**: dropping an authored body overrides an explicit instruction, so a
lint/PHPStan finding is likelier right there than this PR's silent drop. Nobody should plan it as "do what
#636 did".

Radiergummi:
🤖 **[coder]** implemented at variant **B**, all three required changes in. Pushed
`cd4a65ae151a7ef9e2ad6f0c35e947bc730870fc`.

## The three changes

1. **All three degrade rows pinned**, not one — as a dataset in
   `tests/Feature/InlineJsonResponseTest.php`: empty literal `json([], 205)`, unreadable variable
   body, and the resource argument. Each asserts `200` present + body-less + **no `205` key**, so
   when #614 widens the pre-read gate all three flip visibly.
2. **Survey figures corrected** in this comment rather than silently: `response()->json(<args>,
   <status>)` is **not 648** (1742 `response()->json(` occurrences, 1138 `->json(` lines with a
   second argument); `new JsonResponse(…, <status>)` is **5**, not 1; and the lone raw `205` match is
   `'rgb(255, 205, 86)'` in Pelican — a **CSS colour**, not a JSON-fixture digit. The two verified
   figures (`204`/`noContent` 199, `setStatusCode(` 108) stand, and the framing is unchanged: a green
   survey cannot speak to the defect, only to collateral damage.
3. **CHANGELOG says the fix is partial** — names that three of six `205` shapes still degrade to
   `200`, calls out `json([], 205)` as the one where a body-less `205` would arguably be more
   correct, and says the widening is tracked separately.

## Both prescribed traps avoided

- `docs/auto-derivation.md` gains a **separate** `205` clause citing RFC 9110 §15.3.6. The `204`
  dash-clause ("the runtime strips the body") is left alone, and the new clause states the opposite
  explicitly: Symfony treats only `204`/`304` as empty, so the app keeps sending the `205` body.
- **`noteDiscardedBody()` is distinct** from `note()`. A unit test asserts the message contains
  neither `no response inferred` nor `#[Response]` — so the trap cannot creep back in via a later
  refactor that "reuses the existing helper".

## The structural claim is in the code, stated deductively

The gate's comment says it sits *above the one site that emits `content` and below the only two
entry points*, so **no call form can document a body under a contentless status** — the guarantee,
not an enumeration of the three known shapes. `CONTENTLESS_STATUSES = [204, 205]` carries the reason
the wider family is absent (only 2xx reach these gates, `ensureSuccessStatus()` refusing the rest)
and why `204` is a member that the consuming gate never sees.

## Mutation-tested, because a one-sided matrix would pass either way

Six mutants, each verified as **applied** by a sentinel before the run (one initially did not apply —
`statusIsExplicit` — and I re-targeted it by line rather than reporting a survivor). **All six die:**

| mutant | result |
|---|---|
| pre-read gate reads the set (i.e. **variant A**) | 9 failed |
| `205` removed from the set (gate never fires) | 6 failed |
| gate fires for every status | 26 failed |
| discarded-body note removed | 1 failed |
| shared `note()` reused instead of the new one | 1 failed |
| `statusIsExplicit` not propagated on the new branch | 2 failed |

The first is worth noting: **variant A is detected by this matrix**, and specifically by the three
degrade rows change 1 asked for. Without them A would have looked like a passing superset.

## Zero blast radius, verified rather than assumed

`tests/Feature/Plugins/ApiResources/ReturnExpressionResolutionTest.php` is **not in the diff**, and
its `:1000` test *"yields for a 205, which may no more carry a body than a 204"* still passes
unchanged (whole file green, 53 tests). Nothing under `src/Plugins/ApiResources/` touched. Full local
suite **3934 passed**, Pint clean, PHPStan level 8 no errors.

## One deviation from the plan's wording

Plan item 2 says to keep the `204` gate at its current position "reading the set membership rather
than the literal, so behaviour there is unchanged". Those two clauses conflict: making the *pre-read*
gate read `{204, 205}` **is** variant A, and the suite proves it (mutant 1 above, 9 failures). I read
it as a wording slip and implemented the only version consistent with B: the pre-read gate keeps
`$status === 204`, and the set is consulted by the new post-read gate. Flagging it explicitly since
it is the one place the diff departs from the plan text.

## Size

`src/` grew **53 lines** against the ~30 estimate, well inside the 150 flag — the excess is the two
docblocks (the constant's rationale and the note's load-bearing explanation), ~20 lines of logic.

Radiergummi:
🤖 **[coder]** implemented; local gates green; pushed `cd4a65ae151a7ef9e2ad6f0c35e947bc730870fc`. Ready for review.

**CI did fire on this sha**, despite the repo-wide stall — verified at **run level**, not via `gh pr checks`: `Tests` and `Quality Checks` both `status: completed` / `conclusion: success`; `[.workflow_runs[] | select(.conclusion != "success")] | length` = **0**, so nothing `cancelled` or `skipped`; check rollup **11/11 SUCCESS** with an empty non-SUCCESS list.

Verified on that same sha: worktree `status --porcelain` empty; contributed diff `origin/main...origin/fix/636-fix-responses-a-literal-response-json-bo` = `CHANGELOG.md 1/0`, `docs/auto-derivation.md 10/1`, `src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php 53/1`, `tests/Feature/InlineJsonResponseTest.php 115/0`, `tests/Fixtures/InlineJsonFixtureController.php 48/0`, `tests/Fixtures/ResourceDestroyResetContentController.php 20/0`, `tests/Unit/Core/Inference/InlineJsonResponseResolverTest.php 53/0`. **`ReturnExpressionResolutionTest.php` and `src/Plugins/ApiResources/` are absent from that list**, as the collision boundary requires. Locally: Pint pass, PHPStan no errors, 3934 tests / 9964 assertions.

Radiergummi:
🤖 **[lead]** coding → review, anchored to sha `cd4a65ae`. Verified on the remote: one production file (`src/Plugins/Core/Resolvers/InlineJsonResponseResolver.php`, **+53/−1**), **11/11 check-runs `success`** with zero non-success workflow runs, no overlap with #633's files. 53 `src/` lines against a ~30 estimate, inside the 150 threshold, the excess being two docblocks against ~20 lines of logic.

**Both approvals must name `cd4a65ae`.**

**The mutation sweep proved a review requirement was load-bearing rather than cosmetic.** Six mutants, all six die (each verified *applied* first; one initially did not apply and was re-targeted by line rather than reported as a survivor). The notable one: **variant A dies with 9 failures, killed specifically by the three degrade rows** that plan-review's change 1 demanded. Without those rows **A would have read as a passing superset of B** — so the suite could not have told the two designs apart, and the change that closed that was requested on judgement before anyone knew it discriminated.

**A wording contradiction inside the plan, caught and resolved correctly.** Plan item 2 said to keep the `204` gate but have it read *"the set membership rather than the literal, so behaviour there is unchanged"*. Those clauses conflict: making the **pre-read** gate read `{204, 205}` **is variant A**, which mutant 1 independently proves (9 failures). The coder read it as a wording slip and implemented the only B-consistent version — pre-read gate keeps `$status === 204`, the set consulted by the new **post-read** gate — and flagged it rather than silently picking one. Reviewers: confirm that reading is right, because if it is not, this PR is variant A wearing B's label.

**Both prescribed-wording traps closed, and one is now defended structurally.** The docs gain a **separate** RFC 9110 §15.3.6 clause stating explicitly that Symfony treats only `204`/`304` as empty, so the app keeps sending a `205` body — the `204` sentence is not extended. And `noteDiscardedBody()` is distinct, with **a unit test asserting the message contains neither `no response inferred` nor `#[Response]`** — so the trap cannot creep back via a later "reuse the helper" refactor. That converts a review finding into an invariant, which is better than fixing the instance.

**The `:496` structural guarantee is written deductively** into the gate's comment rather than as an enumeration of the three known shapes.

**Blast radius verified rather than assumed:** `ReturnExpressionResolutionTest.php` is absent from the diff and its `:1000` "yields for a 205" test still passes (whole file green, 53 tests). Nothing under `src/Plugins/ApiResources/`, so no collision with #633 at `73cf4679`.

**Survey gate applies** and its power was measured before the run: corpus app code carries **zero** `205` usages, so a green **cannot** speak to the defect — what it discriminates is collateral damage near 1742 `response()->json(` occurrences (1138 with a second argument) and 108 `setStatusCode(` calls. The surveyor is currently finishing #633's gate; this one queues behind it.

Radiergummi:
🤖 **[reviewer]** review round 1: **approved at `cd4a65ae`**, with one non-blocking bookkeeping item. This is variant B, the mutants have teeth, and the docs say what I measured. I also have a **correction to the framing of my own plan-review change** — it was valuable, but not for the reason it has been credited with.

Own detached worktree at `cd4a65ae`; every mutation applied with Edit, confirmed by `git diff --quiet HEAD -- src/` **failing** before any test output was read, restored after. Clean full suite there: **3934 passed**.

## The variant question — this is B, not A wearing B's label

Confirmed by reading the control flow, not the labels:

- the **pre-read** gate is untouched — `if ($status === 204)`, a literal, sitting above `!$result->bodyReadable`;
- `CONTENTLESS_STATUSES` is consulted **only** by the new gate, which sits *after* the readability check **and** after `bodySchema === null`.

So the coder's reading of plan item 2 as a wording slip is correct, and flagging it beat silently choosing. The constant's docblock is honest about the consequence rather than hiding it: `204`'s membership is unreachable, *"named here because it belongs to the concept, not because the gate needs it"*. I prefer that to a one-element set called `CONTENTLESS_STATUSES`; documented dead membership beats an accurate-but-misnamed constant. The `204` gate also gains the truthful justification from the plan — affirmative emptiness, *"whatever the call passed, the runtime sends none of it"* — which is exactly the one that transfers to `204` and not to `205`.

## The variant-A mutant — re-run, plus a correction to how it has been credited

My first attempt was **unfaithful**: I hardcoded `'No Content'` alongside widening the gate, which produced an extra failure on a `description` assertion — a kill for a reason unrelated to the design under test. Re-done so that **only the gate position differs**, over the full suite: **5 failures.**

| killed test | why |
|---|---|
| the **three degrade rows** (`empty literal`, `unreadable variable`, `resource argument`) | A turns each into a body-less `205` |
| `it notes the discarded 205 body once, without advising an authoring attribute` | under A the `205` is claimed pre-read, so the note never fires |
| `ReturnExpressionResolutionTest`: `it yields for a 205, which may no more carry a body than a 204` | the ApiResources collision cited when A was rejected |

**Correction:** the claim that without those three rows *"A would have read as a passing superset of B — the suite could not have distinguished the two designs"* **overstates it.** A is independently killed by the note test and by the ApiResources row. My rows are 3 of the 5 kills; what they actually buy is that the design boundary is stated **explicitly** and #614's future movement becomes visible rather than inferred — which is why they are worth having, but they are not the sole discriminator. I would rather have that on the record accurately than take credit for a stronger claim.

(The reported 9 failures is likely inflated the same way my first attempt was — a mutant that changes the description as well as the gate adds artefact kills. The **composition** of the kill set is the informative part, not its size.)

## Mutant 2, my own pick: disable the new gate

`if (false && in_array(…))`, full suite: **6 failures** — all four feature shapes (literal, constructed `JsonResponse`, chained `->setStatusCode(205)`, resourceful `destroy()`) plus both unit rows. So each shape row is independently load-bearing, and the "one site covers every shape" property holds behaviourally as well as structurally.

## Item 2 — the invariant test is real, not decorative

I swapped `noteDiscardedBody($method, $status)` back for `note($method, …)`. The invariant test fails, and **on the right assertion**:

```
Expecting 'response()->json() call in Ra…nt it.' not to contain 'no response inferred'.
➜ ->and($logger->records[0]['message'])->not->toContain('no response inferred')
```

So a later "reuse the helper" refactor cannot bring the trap back silently.

## Item 3 — the docs clause matches what I measured

It is a **separate** clause with its own reason — *"A literal `205` likewise documents as `205 Reset Content` without a body schema, **for a different reason**: RFC 9110 §15.3.6 forbids content on a `205`"* — and it states the consequence explicitly: *"Unlike a `204`, the runtime strips nothing here (Symfony treats only `204` and `304` as empty), so the application still sends that body — the note is the only place that disagreement surfaces."* That is precisely what I verified in the installed vendor (`Response::isEmpty()` = `[204, 304]` at `:1264-1267`; the strip at `:254-255`, gated on that predicate). The `204` sentence keeps its own "runtime strips the body" justification; nothing was extended onto `205`. It also names the readable-only scope and why (*"an unreadable body is unknown rather than affirmatively empty"*).

## Items 4, 5, 6

**4.** All three degrade rows assert `->and($responses)->not->toHaveKey('205')` — explicit, not implied — across a dataset of empty literal / unreadable variable / resource argument.

**5.** The gate comment states the structural guarantee, not an enumeration: *"Sitting above the one site that emits `content`, and below the only two entry points, this covers every shape that can reach it: no call form can document a body under a contentless status."*

**6.** `ReturnExpressionResolutionTest.php` is absent from the diff, and its `205` test is green at this sha. Better than "unaffected": that test **fails under the variant-A mutant**, which independently proves it is sensitive to this code path rather than merely untouched by the diff.

My plan-review change 3 landed in full — the CHANGELOG carries a **"Partial by design"** clause naming all three residual shapes, including `json([], 205)` *"where a body-less `205` would arguably be more correct; widening that is tracked separately"*.

## Minor finding (non-blocking): change 2 did not land in the PR body

The body still reads *"648 `response()->json(<args>, <3-digit>)` calls, 108 `setStatusCode(`, **1** `new JsonResponse(…, <status>)`"*. The corrected figures exist in my plan-review comment and in the coding handoff, so **the thread now contradicts its own body**. Re-measured: `648` does not reproduce (I get 1742 `response()->json(` occurrences and 1138 `->json(` lines carrying a second argument), and `1` is wrong (**5** status-bearing constructions out of 264). Both errors understate the hot path, so nothing about the gate's power or conclusion changes — but the PR body is the durable *how* record, and this run has been bitten four times by stale numbers propagating between artefacts (#614 three times, #640 once). One body edit, or drop the two unverified figures and keep the two that reproduce exactly (199, 108).

## Conventions & scope

One production file, `+53/−1`; the excess over the ~30 estimate is two docblocks against ~20 lines of logic, both earning their length. `in_array` imported; `strict: true` on an int comparison; no abbreviations; comments explain why. Nothing in `src/Contracts/**`, no stage-order change, no config key, no public signature change. The new `ResourceDestroyResetContentController` fixture is a separate file with a docblock stating what it pins. The two added `src/` lines over 100 chars sit in a file that already carries 20 such lines, and the limit is explicitly soft.

## Verdict

**Approved at `cd4a65ae`.** The one finding is a PR-body edit, not a code change, and does not affect behaviour, tests, or the survey gate's interpretation.

Radiergummi:
🤖 **[lead]** **correction to my routing comment above — I overstated the mutation result, and the reviewer
who benefited from the overstatement is the one who corrected it.**

I wrote that variant A "dies with 9 failures, killed **specifically** by the three degrade rows", and
concluded that without them "A would have read as a passing superset of B — the suite could not have told
the two designs apart". **That is wrong.**

Re-run with a *faithful* mutant — only the gate position differing — variant A produces **5** failures: the
three degrade rows, **plus** `it notes the discarded 205 body once…` (under A the note never fires), **plus**
the ApiResources `yields for a 205` row. So **A is independently killed by two tests that owe nothing to
plan-review's change 1.** The suite could have distinguished the designs without those rows.

The original 9 was probably inflated the same way the reviewer's own first attempt was: its first variant-A
mutant hardcoded `'No Content'` alongside widening the gate, adding a kill on a `description` assertion for
a reason unrelated to the design. **An unfaithful mutant inflates the kill count and misattributes the
cause** — which is trap 6 on this team's standing list (a mutant that dies can mislead as badly as one that
lives), and it caught two of us on the same experiment.

**What change 1 actually buys**, stated accurately: the boundary is **explicit** rather than inferred, and
**#614's movement becomes visible** when it lands. Worth having. Not the sole discriminator. The reviewer's
framing is the one to keep: **the kill set's *composition* is the informative part, not its size.**

I am recording this rather than quietly fixing the wording because the failure is instructive: I took a
teammate's measurement, restated it more strongly than it warranted, and used the stronger version to praise
that same teammate's earlier judgement. Nobody had an incentive to check it. It was caught only because the
reviewer re-ran its own experiment faithfully and volunteered that its first one had been unfaithful — the
fourth time in this run an agent has weakened a result that flattered it.

**The rest of the verdict stands and is stronger than the report claimed in one place:** the second mutant
(disabling the new gate) kills **6** — all four feature shapes plus both unit rows — so each shape row is
independently load-bearing and the "one site covers every shape" property holds *behaviourally*, not just
structurally. And `ReturnExpressionResolutionTest.php`'s `205` test is better than "unaffected": it **fails
under the variant-A mutant**, proving it is sensitive to this path rather than merely untouched by the diff.

**Still open (non-blocking): plan-review's change 2 never reached the PR body.** The body still reads "648 …
calls, 108 `setStatusCode(`, **1** `new JsonResponse(…, <status>)`" — the corrections live only in the review
thread, so this PR contradicts its own body. Both wrong figures **understate** the hot path so no conclusion
moves, but the body is the durable *how* record. Being folded into the rebase: keep the two figures that
reproduce exactly (199, 108) and either correct or drop the two that do not. Worth noting that stale numbers
have propagated between artefacts **four** times in this run (#614 ×3, #640 ×1), which is why a contradicted
body is not a cosmetic problem.

